PR #2, opened 2015-12-27
Added an account manager

Almighty-Alpaca:
Didn't implemented the USER_UPDATE event which the websocket will recieve when the account settings have been changed.
If you have any problems with design of that part. tell me and i will change it.


DV8FromTheWorld:
PRs should be made against the development branch please :)

PR #5, opened 2015-12-27
Added an account manager

Almighty-Alpaca:
Didn't implemented the USER_UPDATE event which the websocket will recieve when the account settings have been changed.


DV8FromTheWorld:
Okay so, I am really sorry. I screwed up the git history and made this PR unmergable because the history that it is based on no longer exists D:

To fix, checkout the current development branch, and then git diff this PR with the checkout
then apply the diff on top of the checkout, commit, send to PR

Atleast that is how I would do it. Once again, I am sorry, especially since this is the second time you've had to make this PR

PR #10, opened 2016-01-02
Cleanup imports, propper lambda support

ShadowLordAlpha:
Some basic import cleanup of imports that were not being used (some of them may have also been moved around as I simply mass deleted the ones that were not in use in my IDE).

Allows Lambda functions to be used for the EventListener instead of needing to make a class that implements it.


ShadowLordAlpha:
made it push to the proper branch now -_-

PR #11, opened 2016-01-02
Cleanup imports, propper lambda support

ShadowLordAlpha:
Some basic import cleanup of imports that were not being used (some of them may have also been moved around as I simply mass deleted the ones that were not in use in my IDE).

Allows Lambda functions to be used for the EventListener instead of needing to make a class that implements it.

This is pull #10 but going to the proper branch now -_-

### DIFF
--- a/src/main/java/net/dv8tion/jda/JDA.java
+++ b/src/main/java/net/dv8tion/jda/JDA.java
@@ -19,7 +19,6 @@ import net.dv8tion.jda.entities.*;
 import net.dv8tion.jda.hooks.EventListener;
 import org.apache.http.HttpHost;
 
-import javax.security.auth.login.LoginException;
 import java.util.List;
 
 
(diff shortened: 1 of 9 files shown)


ShadowLordAlpha:
*sigh* If needed I can change it so that that other merge thing isn't there but i need to go to sleep

kantenkugel:
can you base of your branch from our dev one? I really don't wanna merge our dev->master pr back into dev

DV8FromTheWorld:
Honestly, I've been considering moving JDA down a Java version to 1.7 instead of 1.8 because the only 1.8 stuff we use is streams and a few lambdas.. It would probably be better to support more java versions considering we are an API. I'll revisit this PR after @kantenkugel and I have had a chance to discuss it.

Also yeah, you need to base the branch off of the dev branch. We don't need PRs from Dev -> Master to be merged back into Dev.

ShadowLordAlpha:
Closing this until I actually have time to properly rebase it and you guys figure out if you want 1.8 or 1.7

PR #17, opened 2016-01-26
Fixed a problem with the URLPlayer and added a volume control to the Player.

Almighty-Alpaca:
The other changes in the build.gradle are for eclipse to not use build/filtered as source folder.


DV8FromTheWorld:
Please rePR the bufferedstream fix with the updates, base it against Development since Audio is now gone.

Almighty-Alpaca:
Will do it.

PR #38, opened 2016-04-03
@me instead of user id  to get private channel

hlaaftana:


kantenkugel:
I guess i accidentally did this with the cleanup-fix before 1.4 xD Pls don't hate me :sob:

PR #39, opened 2016-04-06
move user, kick and ban

Patrickpk91:
Can you add some function to jda like:

- move user between voice channel
- kick user from server
- ban user from server


hlaaftana:
Do you know how GitHub works?

kantenkugel:
From dev to master? i think not!

kantenkugel:
PRs are not for asking questions... Ask us on our discord channel (linked in readme, or at least create a new issue - That being said: we already have that in GuildManager

PR #51, opened 2016-05-10
TextChannelImpl#deleteMessages() & TextChannel#deleteMessages()

MinnDevelopment:
please don't have another request ;-;


MinnDevelopment:
Need to rethink a few code choices

PR #52, opened 2016-05-10
Added TextChannel#deleteMessages()

Frederikam:
- Added TextChannel#deleteMessages() https://github.com/hammerandchisel/discord-api-docs/issues/40
- Added project files specific to Netbeans (harmless)
- Fixed a typo because I am a nazi


src/main/java/net/dv8tion/jda/entities/TextChannel.java
```diff
@@ -34,4 +36,14 @@
      *      The String needed to mention this Channel
      */
     String getAsMention();
+    
+    /**
+     * Bulk deletes a list of messages. Must not be more than 100 messages at a time.
+     * You must have permission to delete a message. You may always delete your own messages. In the event that you do not have permission to delete a message, the message will be skipped.
```
DinosParkour: This isn't quite correct, you need MESSAGE_MANAGE to bulk delete anyway so all messages in the list will be deleted..

src/main/java/net/dv8tion/jda/entities/impl/TextChannelImpl.java
```diff
@@ -355,6 +355,38 @@ public TextChannelImpl setPosition(int position)
...
+    @Override
+    public void deleteMessages(List<Message> messages)
+    {
+        if(messages.size() > 100)
+        {
+            throw new IllegalArgumentException("Can't delete more than 100 messages at a time, got " + messages.size());
+        }
+        else if(messages.size() == 1)
+        {
+            messages.get(0).deleteMessage();
```
DinosParkour: Check if the Bot has the permission required to delete the message (aka if it's not its own msg)?

src/main/java/net/dv8tion/jda/entities/impl/TextChannelImpl.java
```diff
@@ -355,6 +355,38 @@ public TextChannelImpl setPosition(int position)
...
+            return;
+        }
+        else if(messages.isEmpty())
+        {
+            return;
+        }
+        
+        if(PermissionUtil.checkPermission(getJDA().getSelfInfo(), Permission.MESSAGE_MANAGE, this) == false)
+        {
+            throw new PermissionException(Permission.MANAGE_PERMISSIONS, "Must have MESSAGE_MANAGE in order to bulk delete messages in this channel regardless of author.");
```
DinosParkour: Did you mean to throw
```java
new PermissionException(Permission.MESSAGE_MANAGE, ...
```

src/main/java/net/dv8tion/jda/entities/impl/TextChannelImpl.java
Frederikam: `messages.get(0).deleteMessage();` will throw an exception in such a case.

kantenkugel:
looks good!
Waiting for DV8 before merging tho

kantenkugel:
you could (if you want) properly link the deleteMessage method ( `{@link net.dv8tion.jda.entities.Message#deleteMessage}`)

Frederikam:
Thanks Kant. Went ahead and committed that.

DV8FromTheWorld:
Code looks good. Please compress the commits into a single commit considering many are for simple changes.
If possible, cleanse the first commit which is a merge and unneeded. If you can't, that's fine. Just compress and drop a message when done and I'll merge it

Frederikam:
DV8 you can just squash all the commits into one when merging this PR. This will produce a single commit for this entire PR.

DV8FromTheWorld:
Actually, I missed something: Line 376 of TextChannelImpl, please change the if statement to use `!` instead of `== false`

DV8FromTheWorld:
And add a newline at the end of the .gitignore file

DV8FromTheWorld:
Okay, I didn't know github had that functionality now. Dope. Other changes still stand.

Frederikam:
:+1:

DV8FromTheWorld:
Looks good. Question: what happens if you pass ids for messages sent in a different channel? What about if you pass it messages built with the message builder? We might want to add catches for those situations.

Otherwise, this is mergable.

Frederikam:
The endpoint doesn't care if the message belongs in another channel or if it doesn't exist. However, if you use an unsent message from a builder the message array will contain a `""` which is invalid formatting, causing a silent fail. Will fix.

Frederikam:
Unsent messages will now be ignored

kantenkugel:
You ready to merge @Frederikam ? BTW: i find it funny how you use the `lastProcessedMessage` var instead of just using `ids.get(0)` xD

Frederikam:
Ready to merge.

`lastProcessedMessage` exists because `ids.get(0)` is only the id (String) of the message. It seems dirty to convert the id to a Message object or just do a delete directly using REST.

kantenkugel:
oh right... kk

DV8FromTheWorld:
Im concerned that silently removing and returning when dealing with messages built from MessageBuilder may prove to cause hidden problems for some users not knowing why their messages aren't being deleted because that are attempting to delete a MessageBuilder built message, however, I'll leave it for now.

Frederikam:
I don't see the problem. The desired result is to not have the message exist in the given channel. If the message is never sent this goal is already fulfilled.

PR #57, opened 2016-05-20
Adding support for setting and receiving streaming status.

jagrosh:
Added ability to detect Twitch streams in User#getCurrentGame()
Added ability to set Twitch stream in AccountManager#setStreaming(Game, Url)
Added Game and GameImpl to handle future game types

### DIFF
--- a/src/main/java/net/dv8tion/jda/entities/Game.java
+++ b/src/main/java/net/dv8tion/jda/entities/Game.java
@@ -1,0 +1,78 @@
+/*
+ *     Copyright 2015-2016 Austin Keener & Michael Ritter
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @author John Grosh (jagrosh)
+ */
+package net.dv8tion.jda.entities;
+
+/**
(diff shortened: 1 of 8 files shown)


DV8FromTheWorld:
Okay so I have a few issues with this.
The syntax style/formatting doesn't match JDA. Please use whitespace.
Example: 
`if(user.getCurrentGame()==null?nextGame!=null:!user.getCurrentGame().equals(nextGame))`
vs
`if (user.getCurrentGame() == null ? nextGame != null : !user.getCurrentGame().equals(nextGame))`
This needs to be fixed across the entire commit.

The excessive amount of ternary statements is.. excessive. It makes the code really annoying to read, especially the one in EntityBuilder. Ternary is okay sometimes for simple stuff, but don't go nut

Lastly, you left the auto-generated @author statement in the License header. Please remove it :D

jagrosh:
Alright, I fixed the whitespace and excessive number of ternary statements (found a couple extraneous ones too) - please double check that it's up-to-par with the rest of JDA

After testing, I found that the Discord Client only displays a user as "Streaming" if the provided URL matches the regex `^https?:\/\/(www\.)?twitch\.tv\/.+`; as such, the "streaming" status can only be set if the provided URL matches this regex as well.

Displaying streaming is client-side, which means users can still have a "Streaming" status in presence updates, even if the client doesn't show them as streaming. These are not differentiated in the Game object, but developers can use Game#isValidTwitchUrl to determine if a user's stream is displayed client-side.

DV8FromTheWorld:
The class level Javadoc in Game is Guild's javadoc.

DV8FromTheWorld:
I think we should also change the isValidTwitchUrl to isValidStreamingUrl due to Discord planning to support other stream sources.

DV8FromTheWorld:
Lastly, make sure your braces are on newlines.

DV8FromTheWorld:
If you have an if-statement that is more than 1 line please put it in brackets. Multiline ifs that don't use brackets can be confusing to read and easily introduce bugs into the code if something changes. Specifically I see this in the AccountManager and ReadyHandler.

DV8FromTheWorld:
After those are fixed I'll merge. Everything else looks fine.

DV8FromTheWorld:
Looks good. Merged! Should be build 267.

PR #59, opened 2016-05-25
Fixed potential non-integer value for game type

jagrosh:


Frederikam:
Thanks Discord ¯\_(ツ)_/¯

DV8FromTheWorld:
Wait what? What else might it possibly give to us? I'd like to see an example explosion before accepting this.

PR #60, opened 2016-05-27
InviteUtil Failing when trying to get Guild

jagrosh:
When receiving an invite (with old InviteUtil) expecting a "code" value back only if the invite is valid, and we expect the code to be the same one we sent. However, Discord replies with {"code": 10006, "message": "Unknown Invite"} on unknown invites. The old InviteUtil just looks for a "code" field as verification of an invite, whereas it should actually check that the code is not an error-code (or in the case here, that it has an associated guild).

I did some testing, and the api never returns a guild on invalid invites, including: expired invites, invites where the channel no longer exists, invites where the guild no longer exists, and invites that are not valid. Note that we _shouldn't_ check that the code we receive is the same one we send, because we get a different one in return when sending a "human-readable" code.


DV8FromTheWorld:
Probably okay to merge this. I trust you tested it well

PR #61, opened 2016-05-29
Add default avatar

jagrosh:
Discord has default avatars for users (used when they either have not set an avatar, or have removed their avatar)


DV8FromTheWorld:
Sure, why not. I can see use-case for this.

PR #64, opened 2016-06-04
Fixed an NPE

Xetanai:
Caused when calling getContent() on a message which has been built but not sent.


DV8FromTheWorld:
The NPE only occurs when a user builds a message using MessageBuilder and then calls `getContent()` on that unsent built message. At that point, the channelId of the message isn't set because it has never been sent to a channel. Maybe in the future we should look into the best way to deal with this, however, for now, the changes proposed in the PR don't fix the problem.

PR #74, opened 2016-06-22
getStatus returning null, changed buildBlocking to use getStatus

demodude4u:


demodude4u:
The reasoning to switch to getStatus is because if the event handler is changed it may not operate in a way that works for buildBlocking()

demodude4u:
I forgot I can't use my auto formatter... and @DV8FromTheWorld pointed out that CONNECTED status needed to be moved to end of method to work correctly.

PR #84, opened 2016-07-09
Give ability to ban users by id (without them being in the guild)

MinnDevelopment:


MinnDevelopment:
It returns status 500 if you enter an id that is higher than Integer.MAX_VALUE. Should we check for that?

PR #96, opened 2016-08-04
Fixed AudioConnection and added callbacks to Players

houraineet:
AudioConnection.java's setupSendThread() was being called before the sendHandler being set, causing it to skip over the if condition on line 205.

### DIFF
--- a/src/main/java/net/dv8tion/jda/audio/AudioConnection.java
+++ b/src/main/java/net/dv8tion/jda/audio/AudioConnection.java
@@ -126,14 +126,14 @@ public class AudioConnection
 
     public void setSendingHandler(AudioSendHandler handler)
     {
-        setupSendThread();
         this.sendHandler = handler;
+        setupSendThread();
     }
 
     public void setReceivingHandler(AudioReceiveHandler handler)
     {
-        setupReceiveThread();
         this.receiveHandler = handler;
+        setupReceiveThread();
     }
 
     public void setQueueTimeout(long queueTimeout)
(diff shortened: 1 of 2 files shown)


DV8FromTheWorld:
I have no idea why you added the Runnable to the Player classes. It is not needed. Write your own checks into the class to determine when a source is done if you want, but it shouldn't be in that class, especially not that method. 

Additionally: when the handler is set doesn't matter. It is a looping thread, it doesn't matter if it misses the first call because it is called nano-seconds later.

houraineet:
before: [http://puu.sh/qq5W6.png](url) after: [http://puu.sh/qq75c.png](url)
But it's not being called again, and player.play() ends up not doing anything because of that in the AudioExample, unless there's another function that needs to be called to get the player to work that I missed. 

As for the callbacks, it may have been a bit selfish of me, for which I apologise, but I just wanted a simple way to implement a queueing system without having to extend the Player classes or performing the checks manually.
Providing a callback for the player to call after the source is done is possible in discord.py ([https://github.com/Rapptz/discord.py/blob/e977bddc1f909240b00548e837fefc058e7a832d/discord/voice_client.py](url) line 326) and I thought it would have been a reasonable addition.

Almighty-Alpaca:
That the AudioExample isn't working for you is because of your changes. The setupSend/ReceiveThread() have to be called inside the thread and on handler being changed. if the threads really have to be created is checking inside the  methods.

Almighty-Alpaca:
So after looking at the changes again, i noticed you were right with the first changes. The `setup...Thred()` have to be called after setting the fields. Otherwise the threads won't be created sometimes.

But this PR still won't be accepted due to the callback commit. For such specific needs java has the ability to sublass. 
If you want you can create a new PR with just the audio connection changes, otherwise I will change that myself.

DV8FromTheWorld:
I agree with @Almighty-Alpaca here. Indeed, somehow we've overlooked the lack of the sending thread setting up, however, a callback isn't really needed in this situation.

As a note: The Player class is being completely removed (along with FilePlayer and URLPlayer) in JDA 3.0

PR #97, opened 2016-08-10
Add a method to request additional permissions.

Xetanai:


MinnDevelopment:
Can you fix the brace on line 109 and the indentation to 4 spaces instead of tabs?
Other than that it looks ok to me

DV8FromTheWorld:
This needs to use 4 spaces instead of tabs.
Method should be an overload of `getAuthInvite`

MinnDevelopment:
Still missing indentation with spaces and new line braces.
Missing `@param` doc for JDA (Which I don't understand why it needs to be a parameter here btw)

DV8FromTheWorld:
It is a parameter to better respect the original overload. Not sure it is needed but it isn't -exactly- wrong.

It could lead to an interesting situation however where someone could provide a JDA object and a Guild object which aren't related.

DV8FromTheWorld:
Needs `@param` tag if you want to keep the JDA param.

Xetanai:
@DV8FromTheWorld Seeing as we only need the ID from the guild, not the guild itself, I'll avoid that interesting situation by changing the Guild parameter to just a String for the ID. It'll also allow for invites for guilds the bot isn't already in. Making the edits right now.

Xetanai:
Also, clarifying, I'm gonna change it a bit to be more general use rather than specifically for adding perms.

MinnDevelopment:
You're doing `String#getId()` now. Did you even run this through a compiler?

Xetanai:
No, had like 5 minutes to work on it yesterday and didn't think.

DV8FromTheWorld:
Personally.. I think this would have been better if it was a `Guild` parameter instead of a `JDA` and `String` parameter. You can retrieve a known connected JDA instance from Guild and also retrieve the Guild's id. I see no real reason to do it this way.

Xetanai:
I was thinking it could be used to get an invite for a guild it's not already in. Get the ID from some sort of oauth webapp. As for the JDA param, I was trying to keep it consistent with the other methods.

PR #102, opened 2016-08-11
Created new checkPermission methods using varargs

MinnDevelopment:
Also made old methods deprecated (together with a little note)


MinnDevelopment:
Note: Keep in mind that the old methods might still be utilized inside of JDA's internal methods.

PR #105, opened 2016-08-15
Switched guild collection in EmoteImpl to Set

MinnDevelopment:
now remove emotes from guild maps properly.

### DIFF
--- a/src/main/java/net/dv8tion/jda/entities/Emote.java
+++ b/src/main/java/net/dv8tion/jda/entities/Emote.java
@@ -43,10 +43,17 @@ public interface Emote
     boolean isAvailable();
 
     /**
-     * Returns an unmodifiable list for guilds this emote is used on.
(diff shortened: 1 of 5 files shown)


DV8FromTheWorld:
As we spoke about in the JDA Guild, I'd like to test to make sure this actually works as expected if possible.

kantenkugel:
There is another patch incoming that switches to single guild per emote. But that one needs this to merged before (bug-fix)

MinnDevelopment:
According to night, emotes can not share two guilds.

MinnDevelopment:
The current code has been on Momo for 2 hours without any unexpected behavior and all emotes returned by JDA are available to the bot.

DV8FromTheWorld:
I'm assuming that means that this should be working, yes?

MinnDevelopment:
Yes this *should* be working according to the data my bot has collected.

DV8FromTheWorld:
:+1:

PR #113, opened 2016-08-23
Added getVoice-/TextChannelById to Guild

MinnDevelopment:


src/main/java/net/dv8tion/jda/entities/impl/GuildImpl.java
```diff
@@ -325,6 +325,26 @@ public TextChannel getPublicChannel()
     }
 
     @Override
+    public TextChannel getTextChannelById(String id)
+    {
+        return getJDA().getTextChannelById(id) == null 
+            ? null 
+            : getTextChannels().contains(getJDA().getTextChannelById(id)) 
+                ? getJDA().getTextChannelById(id) 
+                : null;
```
Almighty-Alpaca: Isn't it very inefficient to lookup the channel 3 times in the global channel map ? I would store it in a local variable instead

DV8FromTheWorld: I am also confused as to why you do this. Within GuildImpl exists `textChannels` and `voiceChannels`, both of which are HashMaps which you can use a channel's id to do lookups.
These methods should not event touch the JDA object. Please refactor.

PR #123, opened 2016-09-08
Added oldNick field to both GuildMemberLeave- and BanEvent. 

MinnDevelopment:
Now cleaning nickname map on member removal as well.


MinnDevelopment:
mfw the other PR is merged already.

PR #125, opened 2016-09-15
Giving ability to check Permissions for Role in a Channel.

MinnDevelopment:
Also made Channel#checkPermission(User, Permission...) default.


DV8FromTheWorld:
This looks right to me.

PR #128, opened 2016-09-17
fixed .equals and added .hashCode() in GameImpl

jagrosh:
.equals() now correctly checks each element
.hashCode() generates hash based on game, url, and type
git diff why you being so weird

### DIFF
--- a/src/main/java/net/dv8tion/jda/entities/impl/GameImpl.java
+++ b/src/main/java/net/dv8tion/jda/entities/impl/GameImpl.java
@@ -16,13 +16,14 @@
  */
 package net.dv8tion.jda.entities.impl;
 
+import java.util.Objects;
 import net.dv8tion.jda.entities.Game;
 
 public class GameImpl implements Game
 {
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/entities/impl/GameImpl.java
```diff
@@ -1,73 +1,83 @@
...
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        if(!( obj instanceof Game ))
+            return false;
+        Game other = (Game) obj;
+        if(other.getType() != type)
+            return false;
+        if(name==null && other.getName()!=null)
```
MinnDevelopment: It looks cleaner if you do `name == null` not `name==null`. Same for the following.

MinnDevelopment: It looks cleaner to do `name == null` not `name==null` and would follow the style of JDA more accurately.

PR #132, opened 2016-09-24
Fixing some javadocs for 3.x

MinnDevelopment:


src/main/java/net/dv8tion/jda/core/entities/Member.java
```diff
@@ -62,15 +98,66 @@
...
+    /**
+     * The roles applied to this Member.
+     *
+     * @return
+     *      An immutable List of {@link net.dv8tion.jda.core.entities.Role Roles} for this Member.
+     */
     List<Role> getRoles();
 
+    /**
+     * An {@link java.awt.Color Color} this Member is displayed with.<br>
```
DV8FromTheWorld: This is incorrect grammar. Change `An` the `The`.
Probably should be a bit more extensive as to what this actually is.

"The {Color} of this Member's name in a Guild. This is determined by the color of the highest role assigned to them that does not have the default color. If all roles have default color, this returns null".
Atleast, I think it returns null. Need to check.

src/main/java/net/dv8tion/jda/core/entities/Member.java
```diff
@@ -62,15 +98,66 @@
...
+     * The Guild-Wide Permissions this Member holds.
+     *
+     * @return
+     *      An immutable List of Permissions granted to this Member.
+     */
     List<Permission> getPermissions();
+
+    /**
+     * The Permissions this Member holds in the specified {@link net.dv8tion.jda.core.entities.Channel Channel}.
+     *
```
DV8FromTheWorld: add: Permissions returned by this may be different from {#getPermissions()} due to the Channel's {PermissionOverrides}

src/main/java/net/dv8tion/jda/core/entities/Game.java
```diff
@@ -48,6 +48,11 @@
 
     /**
      * Checks if a given String is a valid Twitch url (ie, one that will display "Streaming" on the Discord client.
```
DV8FromTheWorld: Missing closing `)`

src/main/java/net/dv8tion/jda/core/entities/Member.java
```diff
@@ -24,16 +24,52 @@
 import java.time.OffsetDateTime;
 import java.util.List;
 
+/**
+ * Represents a Guild-specific User.
+ * <p>
+ * Used to get data about this user's presence within a {@link net.dv8tion.jda.core.entities.Guild Guild}.
```
DV8FromTheWorld: This is actually more than just the User's presence. It contains "all guild-specific information about a User"

src/main/java/net/dv8tion/jda/core/hooks/EventListener.java
```diff
@@ -17,9 +17,26 @@
 
 import net.dv8tion.jda.core.events.Event;
 
+/**
+ * JDA pushes {@link net.dv8tion.jda.core.events.Event Events} to the registered EventListeners.<p>
+ * Register an EventListener at either a {@link net.dv8tion.jda.core.JDA JDA} object<br>
```
DV8FromTheWorld: change "at either" -> "with either"

src/main/java/net/dv8tion/jda/core/hooks/ListenerAdapter.java
```diff
@@ -22,6 +22,27 @@
...
+ *     }
+ *
+ *    &nbsp;@Override
+ *     public void onMessageReceived(MessageReceivedEvent event)
+ *     {
+ *         System.out.printf("[%s]: %s\n", event.getAuthor().getUsername(), event.getMessage().getContent());
+ *     }
+ * }</code></pre>
+ * @see net.dv8tion.jda.core.hooks.EventListener
+ */
```
DV8FromTheWorld: Good example doc

DV8FromTheWorld:
Side note, you need to bring your branch up to date with `3.0` so that the readme commit isn't part of this PR anymore.

PR #134, opened 2016-09-27
Added support for DND and INVISIBLE

MinnDevelopment:


MinnDevelopment:
Note: bots can't use dnd but can use invisible/idle/online (idle only if 'afk' is true)

MinnDevelopment:
I'm thinking maybe it would be better to combine setStatus and setIdle in one method or at least provide an overload with both parameters. I would change it myself if I could.

PR #136, opened 2016-09-29
Add and Remove Multiple Listeners

adriantodt:
Just a way to Bulk-Add and Remove Listeners.

(I'm not good with Documentation btw)


DV8FromTheWorld:
These should not be different methods. This should just be a modification of the two methods that already exist. Remake by modifying existing methods.

PR #142, opened 2016-10-12
Added/modified embeds to match the documentation found here: https://…

jagrosh:
Added/modified some embed internals to match documentation here:
https://github.com/hammerandchisel/discord-api-docs/blob/feature/webhooks/docs/resources/Channel.md#embed-object

Alternatively, instead of merging this PR, it could be used as a template for these changes (if you don't like the names and/or structure)

EDIT: Looks like the diff is freaking out due to line endings; I'll see if I can resolve this so it's easier to read.


DV8FromTheWorld:
Yeah, once you figure out the line endings thing I'll be willing to review and probably pull this in.
This PR shows 1785 lines added and 1401 removed, and that's just not right. Once you figure it out, make a new PR so that the above commit isn't in history (thus not messing up git blame)

PR #159, opened 2016-10-30
Added createEmote

MinnDevelopment:


DV8FromTheWorld:
The RestAction return type should be an Emote, no?

MinnDevelopment:
I thought the discord API would not return the emoji object but was wrong.

PR #175, opened 2016-11-15
Added MessageChannel#editMessageById

Asourcious:


src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
```diff
@@ -242,4 +247,38 @@
...
     RestAction<List<Message>> getPinnedMessages();
+
+    default RestAction<Message> editMessageById(String id, String newContent)
+    {
+        return editMessageById(id, new MessageBuilder().appendString(newContent).build());
+    }
+
+    default RestAction<Message> editMessageById(String id, Message newContent)
+    {
+        JSONObject json = new JSONObject().put("content", newContent.getRawContent()).put("tts", newContent.isTTS());
```
DV8FromTheWorld: There need to be Args.notNull checks on both id and Message.

DV8FromTheWorld: MessageBuilder does that already. Let the MessageBuilder catch that error and filter back.

Asourcious:
Done

DV8FromTheWorld:
So uh, this kinda got borked due to the implementation of MessageImpl#toJSONObject.

If you could re-PR with the changes that'd be great. Sorry about that.

PR #183, opened 2016-11-17
Added JDA#retrieveUser

MinnDevelopment:


src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
```diff
@@ -297,6 +298,36 @@ public User getUserById(String id)
...
     @Override
+    public RestAction<User> retrieveUser(String id)
+    {
+        if (accountType != AccountType.BOT)
+            throw new AccountTypeException(AccountType.BOT);
+        Args.notNull(id, "User id");
+        // check cache
+        User user = this.getUserById(id);
+        if (user == null)
+            user = this.fakeUsers.get(id);
```
DV8FromTheWorld: Don't check the fake user cache in this situation. Unlike the live cache which is always up to date, the fake user cache may not be up to date.

src/main/java/net/dv8tion/jda/core/JDA.java
```diff
@@ -156,6 +157,8 @@ public int getShardTotal()
      */
     User getUserById(String id);
 
+    RestAction<User> retrieveUser(String id);
```
DV8FromTheWorld: This should probably be `retrieveUserById`

DV8FromTheWorld: It should also come after all of the getUser methods, but before the getGuild methods. It should not be placed between them

PR #185, opened 2016-11-20
IPermissionHolder Interface

meowingtwurtle:


meowingtwurtle:
I know that it isn't used anywhere, but I feel that this is a step towards modularity and could be used in several places.

DV8FromTheWorld:
While I agree with the Modularity of it, is there an actual use-case for this?

meowingtwurtle:
The first that would come to mind is that the two maps in the Channels for PermissionOverrides could be merged into one. Several of the canInteract/canTalk things could be updated so that either could be accepted, because sometimes someone may want to check if a Role could use it instead of just a Member.

meowingtwurtle:
It also prepares for possible future permissions systems (unlikely, but not impossible).

DV8FromTheWorld:
Not ignoring this. Will be looking into it more soon. :+1:

meowingtwurtle:
ummmm

meowingtwurtle:
Tried to fix conflicts, ended up breaking something
Will fix later

meowingtwurtle:
FINALLY!

PR #187, opened 2016-11-22
Made a few methods in MessageImpl synchronized

MinnDevelopment:
This should fix issues with concurrent calls to the modified methods.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/impl/MessageImpl.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/impl/MessageImpl.java
@@ -218,7 +218,7 @@ public class MessageImpl implements Message
     }
 
     @Override
-    public String getStrippedContent()
+    public synchronized String getStrippedContent()
     {
         if (strippedContent == null)
         {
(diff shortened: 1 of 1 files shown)


meowingtwurtle:
Wouldn't the setters also have to be synchronized for this to actually have any effect?
Because the get and set calls could still interleave.

meowingtwurtle:
A better solution would be to have a lock that is only locked and released if subcontent is null, otherwise just return subcontent without locking.

MinnDevelopment:
What setters are you talking about exactly?

meowingtwurtle:
setContent() mainly

MinnDevelopment:
`setContent` sets the raw content returned by `getRawContent`.
`getContent` however overrides  the raw content and provides a content without the discord formats like `<@86699011792191488>` for mentions and instead replaces them with `@Minn`.
That new content with stripped formats is then saved as `subContent`.
`getStrippedContent` does something similar for markdown formatting.

MinnDevelopment:
The `setContent` method is only used by internal builders and should never be used when the user accesses the Message.

meowingtwurtle:
Okay, sorry I misunderstood what the setter methods were for!

PR #189, opened 2016-11-23
Event disabling

meowingtwurtle:
I saw this on your TODO list in the Discord, so I tried to implement it.


DinosParkour:
You should follow this lib's formatting to begin with (`{` in a new line etc)

meowingtwurtle:
Sorry! Forgot about that. My IDE does the way in the commits by default :P

meowingtwurtle:
Fixed

meowingtwurtle:
Told was to be rejected by DV8 because it still requires time/resources to construct/parse event.

PR #198, opened 2016-11-29
`/channels/{channel.id}/messages/bulk_delete` is deprecated

Frederikam:
This endpoint has been deprecated. Use `/channels/{channel.id}/messages/bulk-delete` instead.

https://discordapp.com/developers/docs/resources/channel#bulk-delete-messages


MinnDevelopment:
Jake told us they won't remove it. If they did they would announce it.

PR #200, opened 2016-11-30
Update EmbedBuilder.java

jagrosh:


jagrosh:
I'll fix this properly later, plus some other requested changes and a proper permissions check

PR #201, opened 2016-12-01
Fixed javadoc errors

MinnDevelopment:
- Replaced some \<code> tags to use {@code} instead
- Fixed self-closing \<br> tags
- Removed unknown references
- Updated references to use the new paths
- Removed old @param tags


MinnDevelopment:
- [x] This does compile for both the javadoc and compile tasks
- [x] All javadoc errors are cleaned up

PR #203, opened 2016-12-05
EmbedBuilder bugfixes, changes, and permission check

jagrosh:
no more https check
can add a MessageEmbed.Field object
added isEmpty() method
removed unused imports
fixed deprecated messagebuilder method usage
added permission check for embed + empty message
made lock final because my IDE kept yelling at me

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
@@ -16,9 +16,7 @@
 package net.dv8tion.jda.core;
 
 import java.awt.Color;
-import java.time.Instant;
 import java.time.OffsetDateTime;
-import java.time.ZoneId;
 import java.time.temporal.TemporalAccessor;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -299,14 +306,26 @@ public EmbedBuilder setFooter(String text, String iconUrl)
...
     
     /**
+     * Adds a Field to the embed
+     * 
+     * @param field the field object to add
+     * @return the builder after the field has been added
+     */
+    public EmbedBuilder addField(MessageEmbed.Field field)
+    {
+        return addField(field.getName(), field.getValue(), field.isInline());
```
MinnDevelopment: What about `Args.notNull`?

DV8FromTheWorld: oh yeah, that probably should've been there.

jagrosh: It could be (I'll put that in next time I'm updating this stuff), although the end result is the same: a NPE when someone tried to make `field` null (difference in description and internal cleanliness)

DV8FromTheWorld: Na, not an npe. It results in an IllegalArgumentException.

PR #213, opened 2016-12-13
Added checks to prevent incorrect member behavior

MinnDevelopment:
Making sure we really don't remove us from the member cache and check whether the member was actually in the guild or not.

If this contains an incorrect change let me know as I'm not too sure it would help.


DV8FromTheWorld:
In the situation, I am pretty sure that the member is null because it is *actually* referencing the selfMember which was removed already in an out of order GUILD_DELETE. I'd like to have some testing done on this, but either way, im not sure that an error needs to be printed to the console in any case.

Issue #216, opened 2016-12-14
Bind to SLF4J instead of own logger implementation


meowy:
For some reason, JDA uses its own logger implementation, but considering it essentially forms a library rather than a standalone app (and I like customizing my logger format), it should use the industry-standard SLF4J API. This would add a single dependency, but let users of the library use any SLF4J logger they want and configure the everloving shmoo out of it while integrating the log output with the rest of their app's. 

If I feel like it, I'll fork this, port all the SimpleLogger uses to SLF4J, and submit a PR. Does that sound like a plan?

Frederikam:
What I do to circumvent this is to mute SimpleLogger and add an adapter event handler to log to SLF4J

DV8FromTheWorld:
As I mentioned on the closed PR, We have discussed this already internally and have designed a new logger system that we are going to use in conjunction with SFL4J. So SFL4J will be coming to JDA.

meowy:
Do you have a timeline when it'll be released? I'm interested.

DV8FromTheWorld:
Current focus is the creation and release of full-project Javadoc. 2.x had it, but in the rushed development of 3.x I didn't take the time to write them. @MinnDevelopment is working on this with me. We hope to have that done in a week or two.
We are also focusing on rereleasing Invites, spearheaded by @Almighty-Alpaca, however this is more of a feature branch and will not block loggers.
The logger rewrite is coming immediately after Minn and I have finished the Javadoc pass.

PR #217, opened 2016-12-14
Port to SLF4J (removes SimpleLogger)

meowy:
Haven't tested this change yet. This adds a dependency to slf4j-api, and requires the user to add a dependency to e.g. logback or slf4j-simple to get log messages from JDA. References issue #216.


DV8FromTheWorld:
I am closing this because we already have extensive plans internally for a logger rewrite, moving to SLF4J. Thank you for the PR, however, it doesn't address everything we want to do.

PR #220, opened 2016-12-18
Fixed an error in AccountManager#update

MinnDevelopment:
This should fix the 401 errors after updating account settings.

As this breaks the build entirely I consider this one worth merging, even for a deprecated version.


DV8FromTheWorld:
Required due to changes made to the returned `token` value from PATCH `users/@me`

PR #224, opened 2016-12-22
Multiple Embed-Related Fixes

jagrosh:
added convenience method editMessage(MessageEmbed) to Message
implemented convenience method in MessageImpl
combined setUrl into setTitle in EmbedBuilder
allow use of more TemporalAccessor by allowing a default zone
remove setVideo and setProvider
check both urls in setAuthor
no NPE for null field, and field docs have warning on limit (25)

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
@@ -16,7 +16,11 @@
 package net.dv8tion.jda.core;
 
 import java.awt.Color;
+import java.time.DateTimeException;
+import java.time.Instant;
+import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
 import java.time.temporal.TemporalAccessor;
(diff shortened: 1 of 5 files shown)


MinnDevelopment:
This will be on the waiting list until we merged JavaDocs.

Wiildanimal:
No idea if this is the right place, but:

- add removeField(String title)
- add setFieldValue(String title, String value)
- add setFieldInline(String title, boolean inline)

I made this little class because those aren't implemented:  https://hastebin.com/xitosecari.java

Almighty-Alpaca:
@Wiildanimal how would these methods behave if there are multiple fields with the same title?

Wiildanimal:
@Almighty-Alpaca In my CustomEmbedBuilder I have it so that it does it for all fields with that name, but tbh that's for you guys to decide.

MinnDevelopment:
@jagrosh this needs to be rebased to take the new JavaDoc changes into account.
In addition it must follow the JavaDoc formatting template.

jagrosh:
Replaced by #259

PR #226, opened 2016-12-23
support for gif avatars

jagrosh:


MinnDevelopment:
Shouldn't `jpg` be replaced by `webp`?

DV8FromTheWorld:
Yeah, Discord is switching to `webp`, so this patch won't work in a bit.

PR #233, opened 2017-01-11
Proofread Message.java docs

Svetroid:
Proofread the docs in Message.java and made it much more consistent. Also added some. I plan on doing much more work similar to this in the future if permitted.


MinnDevelopment:
The javadocs are a work-in-progress on the `javadoc` branch.
In addition, if we would accept javadoc PRs you would have to make it to either the `javadoc` branch or the development branch.

DV8FromTheWorld:
I'm currently working on the interfaces in `core.entities`, so those aren't necessarily needed.

PR #234, opened 2017-01-13
Added new RestAction extensions

MinnDevelopment:
Added `RoleAction` and `ChannelAction` to create either a Role or Channel with preset settings.
This is not yet finished and needs proper documentation which I will add once the role creation changes are rolled out.

~~When creating a channel via `ChannelAction`the permissions of the provided PermissionOverrides are not applied (possible discord bug?). This needs to be investigated before merging.~~ Checked.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
+++ b/src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
@@ -49,6 +49,7 @@ public class ChannelAction extends RestAction<Channel> //todo documentation
     protected final Guild guild;
     protected final boolean voice;
     protected String name;
+    protected String topic = null;
 
     // --voice only--
     protected Integer bitrate = null;
@@ -59,6 +60,7 @@ public class ChannelAction extends RestAction<Channel> //todo documentation
         super(api, route, null);
         this.guild = guild;
         this.voice = voice;
+        this.name = name;
     }
 
     public ChannelAction setName(String name)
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
```diff
@@ -0,0 +1,178 @@
...
+            throw new IllegalArgumentException("Specified Member is not in the same Guild!");
+
+        String id = member.getUser().getId();
+        PromisePermissionOverride override = new PromisePermissionOverride(USER_TYPE, id, deny, allow);
+        overrides.add(override);
+        return this;
+    }
+
+    // --voice only--
+    public ChannelAction setBitrate(Integer bitrate)
```
DV8FromTheWorld: Voice-only methods should throw exceptions on attempted use when current ChannelAction is for a TextChannel

src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
```diff
@@ -0,0 +1,178 @@
...
+    {
+        Args.notNull(name, "Channel name");
+        if (name.length() < 2 || name.length() > 100)
+            throw new IllegalArgumentException("Provided channel name must be 2 to 100 characters in length");
+
+        this.name = name;
+        return this;
+    }
+
+    public ChannelAction addPermissionOverride(Role role, Collection<Permission> allow, Collection<Permission> deny)
```
DV8FromTheWorld: Needs override to allow a `long` to be passed for `allow` and `deny` to provide raw permission values.

src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    protected final class PromisePermissionOverride implements JSONString
+    {
+        protected final String id;
+        protected final Set<Permission> deny;
+        protected final Set<Permission> allow;
+        protected final int type;
+
+        public PromisePermissionOverride(int type, String id, Collection<Permission> deny, Collection<Permission> allow)
```
DV8FromTheWorld: Due to needing to provide the `long` overrides discussed above, this should take only `long` for allow and deny, and force the addPermissionOverride in ChannelAction to convert the Collection provided to a long before passing to constructor.

Additionally, keep the order of allow-deny the same between the add override methods and the PromisePermissionOverride constructor (currently opposite of eachother)

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleAction.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public RoleAction setPermissions(Permission... permissions)
+    {
+        Args.notNull(permissions, "Permissions");
+        this.permissions = new HashSet<>();
+        Collections.addAll(this.permissions, permissions);
+        return this;
+    }
+
+    public RoleAction setPermissions(Collection<Permission> permissions)
```
DV8FromTheWorld: Provide override for `long` permissions.

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleAction.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public RoleAction setMentionable(Boolean mentionable)
+    {
+        this.mentionable = mentionable;
+        return this;
+    }
+
+    public RoleAction setColor(Color color)
+    {
+        return this.setColor(color != null ? color.getRGB() & 0xFFFFFF : null);
```
DV8FromTheWorld: Move the `& 0xFFFFFF` to `setColor(Integer)`

src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
```diff
@@ -0,0 +1,178 @@
...
+        PromisePermissionOverride override = new PromisePermissionOverride(USER_TYPE, id, deny, allow);
+        overrides.add(override);
+        return this;
+    }
+
+    // --voice only--
+    public ChannelAction setBitrate(Integer bitrate)
+    {
+        if (bitrate != null && bitrate < 8000)
+            throw new IllegalArgumentException("Bitrate must be greater than 8000.");
```
DV8FromTheWorld: There is a max bitrate, right?

src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
```diff
@@ -0,0 +1,178 @@
...
+    protected final Set<PromisePermissionOverride> overrides = new HashSet<>();
+    protected final Guild guild;
+    protected final boolean voice;
+    protected String name;
+
+    // --voice only--
+    protected Integer bitrate = null;
+    protected Integer userlimit = null;
+
+    public ChannelAction(JDA api, Route.CompiledRoute route, String name, Guild guild, boolean voice)
```
DV8FromTheWorld: `name` provided is never actually used.

With that said, if we want to keep the `.createTextChannel(String name)` method, then we should have it return `new ChannelAction(...).setName(name);`
I.e: Remove the `name` parameter from ChannelAction's constructor.

MinnDevelopment: You cannot create a channel without a name, which is why I made a forced constructor parameter.

src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
MinnDevelopment: The max bitrate is different for VIP guilds. I need a way to know whether the current guild is VIP or not before I can add a max check. The min always works.

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleAction.java
```diff
@@ -85,15 +82,20 @@ public RoleAction setColor(Integer rgb)
...
     {
-        this.permissions = permissions == null ? null : new HashSet<>(permissions);
+        this.permissions = permissions == null ? 0 : Permission.getRaw(permissions);
+        return this;
+    }
+
+    public RoleAction setPermissions(long permissions)
+    {
+        Args.notNegative(permissions, "Raw Permissions");
+        this.permissions = permissions;
```
DV8FromTheWorld: Should probably also make sure that the provided long isn't larger than `Permission.getRaw(Permission.values())`

src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
```diff
@@ -71,41 +73,75 @@ public ChannelAction setName(String name)
...
+
+    public ChannelAction addPermissionOverride(Member member, Collection<Permission> allow, Collection<Permission> deny)
+    {
+        final long allowRaw = allow != null ? Permission.getRaw(allow) : 0;
+        final long denyRaw = deny != null ? Permission.getRaw(deny) : 0;
+
+        return addPermissionOverride(member, allowRaw, denyRaw);
+    }
+
+    public ChannelAction addPermissionOverride(Role role, long allow, long deny)
```
DV8FromTheWorld: `allow` and `deny` need `Args.notNegative` checks.

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleAction.java
```diff
@@ -117,21 +117,22 @@ public RoleAction setMentionable(Boolean mentionable)
...
      * @param  rgb
-     *         The color for the new role in integer form, null to use default white/black
+     *         The color for the new role in integer form, {@code null} to use default white/black
      *
      * @return The current RoleAction, for chaining convenience
      */
     public RoleAction setColor(Integer rgb)
     {
-        this.color = rgb;
+        this.color = rgb & 0xFFFFFF;
```
DV8FromTheWorld: need to check if this is null before using an `&` on it, otherwise, NPE.

DV8FromTheWorld:
@MinnDevelopment go ahead and squash-merge this into `development` :+1:

PR #237, opened 2017-01-18
Adding ability to change position of Channels and Roles.

DV8FromTheWorld:
- [x]  Implemented support for modifying Channel positions.
- [x]  Implemented support for modifying Role positions.
- [x]  Added Javadoc related to Channels
- [x]  Added Javadoc related to Roles
- [x] Implemented permission checks
- [x] Implemented hierarchy checks for Role positions.


src/main/java/net/dv8tion/jda/core/managers/GuildController.java
```diff
@@ -1712,6 +1710,21 @@ protected void handleResponse(Response response, Request request)
         };
     }
 
+    public ChannelOrderAction modifyTextChannelPositions()
+    {
+        return new ChannelOrderAction(guild, true);
+    }
+
+    public ChannelOrderAction modifyVoiceChannelPositions()
```
MinnDevelopment: Missing generic type

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleOrderAction.java
```diff
@@ -0,0 +1,68 @@
...
+import net.dv8tion.jda.core.requests.Route;
+import org.apache.http.util.Args;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+import java.util.List;
+
+public class RoleOrderAction extends OrderAction<Role, RoleOrderAction>
+{
+    protected final Guild guild;
```
MinnDevelopment: There should be a getter for `guild`

src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelOrderAction.java
```diff
@@ -0,0 +1,61 @@
...
+import net.dv8tion.jda.core.requests.Route;
+import org.apache.http.util.Args;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+import java.util.Collection;
+
+public class ChannelOrderAction<T extends Channel> extends OrderAction<T, ChannelOrderAction>
+{
+    protected final Guild guild;
```
MinnDevelopment: There should be a getter for `guild`

src/main/java/net/dv8tion/jda/core/requests/restaction/OrderAction.java
```diff
@@ -0,0 +1,188 @@
...
+import net.dv8tion.jda.core.requests.Response;
+import net.dv8tion.jda.core.requests.RestAction;
+import net.dv8tion.jda.core.requests.Route;
+import org.apache.http.util.Args;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public abstract class OrderAction<T, M extends OrderAction> extends RestAction<Void>
```
MinnDevelopment: Would it be possible to add an abstract `reset()` method that would bring it back to the old state?

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleOrderAction.java
```diff
@@ -0,0 +1,68 @@
...
+            array.put(new JSONObject()
+                    .put("id", role.getId())
+                    .put("position", i + 1)); //plus 1 because position 0 is the @everyone position.
+        }
+
+        this.data = array;
+    }
+
+    @Override
+    protected void validateInput(Role entity)
```
MinnDevelopment: This should also check the `position` of the role to make sure we can move it in the first place.

src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelOrderAction.java
MinnDevelopment: I think we should also add `getChannelType()` that would either return `TEXT` or `VOICE` depending on the boolean value passed to the constructor. In addition to be more future-proof (channel categories) we should maybe just pass a `ChannelType` directly instead of using a boolean to determine the order type.

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleOrderAction.java
DV8FromTheWorld: Done.

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleOrderAction.java
DV8FromTheWorld: Should we check on every move or on submit with the finalizeData?

src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelOrderAction.java
DV8FromTheWorld: Added getter for `guild`.
Added getter for channel type.
Changed boolean defining whether text or voice channels to be a passed `ChannelType`.

The boolean defining ordering (ascending vs descending) is not related to Channels.

src/main/java/net/dv8tion/jda/core/managers/GuildController.java
DV8FromTheWorld: Fixed

src/main/java/net/dv8tion/jda/core/requests/restaction/OrderAction.java
DV8FromTheWorld: I suppose it would be possible.. but I don't really find it useful to be honest. Just create a new object.

MinnDevelopment: Fair enough

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleOrderAction.java
MinnDevelopment: I would say to put it in `validateInput`

src/main/java/net/dv8tion/jda/core/requests/restaction/OrderAction.java
```diff
@@ -174,18 +350,49 @@ public M swapPosition(T swapEntity)
...
     public M shuffleOrder()
     {
         Collections.shuffle(this.orderList);
         return (M) this;
     }
 
+    /**
+     * Sorts the {@link #getCurrentOrder() current order} based on
+     * the specified {@link java.util.Comparator Comparator}.
+     * <br>Using {@link java.util.ArrayList#sort(java.util.Comparator) ArrayList.sort(comparator)}
```
DV8FromTheWorld: We should reference List#sort, not ArrayList. In the future we might chose to change to a different List implementation if we so chose.

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleOrderAction.java
```diff
@@ -25,10 +25,27 @@
 
 import java.util.List;
 
+/**
+ * Implementation of {@link net.dv8tion.jda.core.requests.restaction.OrderAction OrderAction}
+ * designed to modify the order of {@link net.dv8tion.jda.core.entities.Role Roles} of the
+ * specified {@link net.dv8tion.jda.core.entities.Guild Guild}.
+ * <br>To apply the changes you must finish the {@link net.dv8tion.jda.core.requests.RestAction RestAction}
+ *
+ * <p><b>This uses descending order!</b>
```
DV8FromTheWorld: Example of descending order would probably be good to add. E.g: "Positions are descending, meaning that the highest Role in the Guild is at `n-1` in the list, which is based on Role#getPosition` or something.

DV8FromTheWorld: Example of descending order would probably be good to add. E.g: "Positions are descending, meaning that the highest Role in the Guild is at `n-1` in the list, which is based on Role#getPosition` or something.

src/main/java/net/dv8tion/jda/core/requests/restaction/OrderAction.java
DV8FromTheWorld: We should reference List#sort, not ArrayList. In the future we might chose to change to a different List implementation if we so chose.

DV8FromTheWorld:
Only thing blocking acceptance of this is permission checks and hierarchy checks

src/main/java/net/dv8tion/jda/core/managers/GuildController.java
```diff
@@ -1710,21 +1710,95 @@ protected void handleResponse(Response response, Request request)
...
     {
         return new ChannelOrderAction(guild, ChannelType.VOICE);
     }
 
+    /**
+     * Modifies the positional order of {@link net.dv8tion.jda.core.entities.Guild#getRoles() Guild.getRoles()}
+     * using a specific {@link net.dv8tion.jda.core.requests.RestAction RestAction} extension to allow moving Roles
+     * {@link net.dv8tion.jda.core.requests.restaction.OrderAction#moveUp(int) up}/{@link net.dv8tion.jda.core.requests.restaction.OrderAction#moveDown(int)down}
+     * or {@link net.dv8tion.jda.core.requests.restaction.OrderAction#moveTo(int) to} a specific position.
+     * <br>This uses <b>ascending</b> order with a 0 based index.
```
DV8FromTheWorld: This **does not** use ascending order. It uses Discord's order which is descending for Roles.
Perhaps make reference to the difference similarly to how the documentation for the constructor of RoleOrderAction does for the `useDiscordOrder` boolean parameter.

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleOrderAction.java
```diff
@@ -92,6 +95,11 @@ public Guild getGuild()
     @Override
     protected void finalizeData()
     {
+        final Member self = guild.getSelfMember();
+        final int selfPos = guild.getRoles().indexOf(self.getRoles().get(0));
```
DV8FromTheWorld: If the current user has no roles, this will IndexOutOfBounds

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleOrderAction.java
```diff
@@ -92,6 +95,11 @@ public Guild getGuild()
...
+        if (!self.hasPermission(Permission.MANAGE_ROLES))
```
DV8FromTheWorld: Check this before the self.getRoles(). If you do, then the member would **have** to have a role to actually have this permission, thuse you don't need to actually check if they have any roles before calling `self.getRoles().get(0)`

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleOrderAction.java
DV8FromTheWorld: Also, I'm not quite sure why you're doing `indexOf` here. Use `.getRoles().get(0).getPosition()`

PR #243, opened 2017-01-28
RestAction default behavior

Kaaz:
allow to override the default success/failure consumers


Sgmaniac1255:
Would it be better to do this through. Method of some sort?

PR #247, opened 2017-02-05
can send embeds and files in the same message

jagrosh:


jagrosh:
whoops, line ending issue again

PR #248, opened 2017-02-05
can send embeds and files in the same message

jagrosh:


DV8FromTheWorld:
This needs to be based on the recent changes made to `javadoc`

PR #250, opened 2017-02-07
use long identifiers rather than string

Kaaz:


MinnDevelopment:
This is already in focus on an experimental branch. `experimental/cache`

PR #252, opened 2017-02-07
Adds User#getMutualGuilds()

ArsenArsen:
Adds a utility method User#getMutualGuilds() which returns a collection of guilds that are common for the User and its JDA instance.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/JDA.java
+++ b/src/main/java/net/dv8tion/jda/core/JDA.java
@@ -210,7 +210,7 @@ public interface JDA
     User getUserById(String id);
 
     /**
-     * Gets all {@link Guild}s that contain all given users as their members.
+     * Gets all {@link net.dv8tion.jda.core.entities.Guild Guilds} that contain all given users as their members.
      *
      * @param  users
      *         The users which all the returned {@link net.dv8tion.jda.core.entities.Guild Guilds} must contain.
--- a/src/main/java/net/dv8tion/jda/core/JDA.java
+++ b/src/main/java/net/dv8tion/jda/core/JDA.java
@@ -25,6 +25,7 @@ import net.dv8tion.jda.core.requests.RestAction;
 import net.dv8tion.jda.core.requests.ratelimit.IBucket;
 import org.apache.http.HttpHost;
 
+import java.util.Collection;
 import java.util.List;
 
 /**
@@ -32,6 +33,7 @@ import java.util.List;
  */
 public interface JDA
 {
+
     /**
      * Represents the connection status of JDA and its Main WebSocket.
      */
@@ -206,6 +208,24 @@ public interface JDA
      * @return Possibly-null {@link net.dv8tion.jda.core.entities.User User} with matching id.
      */
     User getUserById(String id);
+
+    /**
+     * Gets all {@link Guild}s that contain all given users as their members.
+     *
+     * @param users The users which all the returned {@link Guild}s must contain.
+     *
+     * @return Unmodifiable list of all {@link Guild} instances which have all {@link User}s in them.
+     */
+    List<Guild> getMutualGuilds(User... users);
+
+    /**
+     * Gets all {@link Guild}s that contain all given users as their members.
+     *
+     * @param users The users which all the returned {@link Guild}s must contain.
+     *
+     * @return Unmodifiable list of all {@link Guild} instances which have all {@link User}s in them.
+     */
+    List<Guild> getMutualGuilds(Collection<User> users);
 
     /**
      * This unmodifiable returns all {@link net.dv8tion.jda.core.entities.User Users} that have the same username as the one provided.
--- a/src/main/java/net/dv8tion/jda/core/JDA.java
+++ b/src/main/java/net/dv8tion/jda/core/JDA.java
@@ -220,7 +220,7 @@ public interface JDA
     List<Guild> getMutualGuilds(User... users);
 
     /**
-     * Gets all {@link Guild}s that contain all given users as their members.
+     * Gets all {@link net.dv8tion.jda.core.entities.Guild Guilds} that contain all given users as their members.
      *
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/core/entities/impl/UserImpl.java
```diff
@@ -243,4 +255,4 @@ public String toString()
             return text;
         }
     }
-}
+}
```
Almighty-Alpaca: pls add a new line char at the end of the file

src/main/java/net/dv8tion/jda/core/entities/impl/UserImpl.java
```diff
@@ -125,6 +129,14 @@ protected void handleResponse(Response response, Request request)
     }
 
     @Override
+    public Collection<Guild> getMutualGuilds() {
+        return getJDA().getGuilds()
+                .parallelStream()
```
Almighty-Alpaca: not sure if we should really use a parallel stream here, they can be very slow

src/main/java/net/dv8tion/jda/core/JDA.java
```diff
@@ -208,6 +210,24 @@ public boolean equals(Object o)
     User getUserById(String id);
 
     /**
+     * Gets all {@link Guild}s that contain all given users as their members.
```
MinnDevelopment: This has to be a full qualified path to the `Guild` object like `{@link net.dv8tion.jda.core.entities.Guild Guild}`
Always include the path and an alias name.

src/main/java/net/dv8tion/jda/core/JDA.java
```diff
@@ -208,6 +210,24 @@ public boolean equals(Object o)
...
+     *
+     * @param users The users which all the returned {@link Guild}s must contain.
```
MinnDevelopment: The description of a parameter needs to be aligned in the next line as you can see in the javadoc template.
Same for docs below.

src/main/java/net/dv8tion/jda/core/JDA.java
```diff
@@ -208,6 +210,24 @@ public boolean equals(Object o)
...
+     *
+     * @return Unmodifiable list of all {@link Guild} instances which have all {@link User}s in them.
```
MinnDevelopment: Missing `@throws` with full qualified name and newline description
Include the plural `s` in the alias name for the entity that you are linking `Users`

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
```diff
@@ -44,10 +44,7 @@
 
 import javax.security.auth.login.LoginException;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
+import java.util.*;
```
MinnDevelopment: This wasn't needed. We only wildcard for 5+ imports from the same package usually.

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
```diff
@@ -304,6 +301,33 @@ public User getUserById(String id)
...
+        return getMutualGuilds(Arrays.asList(users));
+    }
+
+    @Override
+    public List<Guild> getMutualGuilds(Collection<User> users)
+    {
+        Args.notNull(users, "users");
+        for(User u : users)
+        {
+            if(u == null)
```
MinnDevelopment: This should use `Args.notNull`, too.

src/main/java/net/dv8tion/jda/core/entities/impl/UserImpl.java
```diff
@@ -125,6 +128,12 @@ protected void handleResponse(Response response, Request request)
     }
 
     @Override
+    public List<Guild> getMutualGuilds()
```
MinnDevelopment: There should also be an option to add other users here, imo

DV8FromTheWorld: (This was countered in chat. Decision was not to include it here as it would be redundant to `JDA#getMutalGuilds(User...)`

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
DV8FromTheWorld: Wildcard came from the use of `Arrays` which set the value to 5, thus it wildcarded.

src/main/java/net/dv8tion/jda/core/entities/User.java
```diff
@@ -111,6 +113,14 @@
     RestAction<PrivateChannel> openPrivateChannel();
 
     /**
+     * Finds and collects all {@link net.dv8tion.jda.core.entities.Guild Guild} instances that contain this {@link net.dv8tion.jda.core.entities.User User} within the current {@link net.dv8tion.jda.core.JDA JDA} instance.<br>
```
DV8FromTheWorld: Technically, `<br>` should be on the next line In this case however, there is no line that is part of this paragraph after this line so there actually shouldn't even be a `<br>` as the following `<p>` handles the paragraph separation.

src/main/java/net/dv8tion/jda/core/entities/User.java
```diff
@@ -111,6 +113,14 @@
...
+     * <p>This method is a shortcut for {@link net.dv8tion.jda.core.JDA#getMutualGuilds(User...) JDA.getMutualGuilds(User)}.</p>
```
DV8FromTheWorld: Paragraph doesn't *need* to be closed for future reference.
The start of a new paragraph *should* be separate from the previous paragraph by an extra line of space in the javadoc comment.

PR #254, opened 2017-02-10
Added CloseCode Enum

MinnDevelopment:
In addition I added checks on whether reconnecting makes sense via the close code context.


src/main/java/net/dv8tion/jda/core/events/DisconnectEvent.java
```diff
@@ -17,14 +17,17 @@
...
- * <b><u>DisconnectEvent</u></b><br>
- * Fired if our connection to the WebSocket was disrupted.<br>
- * <br>
- * Use: Reconnect manually or stop background threads that need fired events to function properly.
+ * Indicates that JDA has been disconnected from the remote server.
+ * <br>When this event is fired JDA will try to reconnect if possible
+ * unless {@link net.dv8tion.jda.core.JDABuilder#setAutoReconnect(boolean) JDABuilder.setAutoReconnect(Boolean)}
+ * has been provided {@code false}!
+ *
+ * <p>When reconnecting was successful a {@link net.dv8tion.jda.core.events.ReconnectedEvent RecconectEvent} is fired
```
DV8FromTheWorld: ReconnectedEvent, not RecconectEvent.
Also, either ReconnectedEvent or ResumedEvent could be fired after a connection is reestablished.

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -864,5 +887,67 @@ protected String handleInternally(JSONObject content)
             });
         }
     }
+
+    public enum CloseCode
```
DV8FromTheWorld: This enum should be pulled out of WebSocketClient to it's own class in `requests` and provided with javadoc.

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -864,5 +887,67 @@ protected String handleInternally(JSONObject content)
...
+
+        public boolean isReconnect()
+        {
+            return isReconnect;
+        }
+
+        @Override
+        public String toString()
+        {
+            return code + ": " + meaning;
```
DV8FromTheWorld: prepend `"CloseCode"` in some form similar to JDA's normal `toString` style.

PR #256, opened 2017-02-10
Fixed PermissionOverride cache bug

MinnDevelopment:
I'm not pushing this directly to development because I need @DV8FromTheWorld to verify this.


DV8FromTheWorld:
This needs to take VoiceChannels into consideration with a similar system.
Afterwards, it will be merged.

PR #259, opened 2017-02-13
Embed-Related Changes

jagrosh:
can use more temporal accessors
edit message with embed only
send embeds with a file
permissions check for editing a message
clear embeds from embedbuilder

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
@@ -19,7 +19,11 @@ import net.dv8tion.jda.core.entities.MessageEmbed;
 import net.dv8tion.jda.core.entities.impl.MessageEmbedImpl;
 
 import java.awt.Color;
+import java.time.DateTimeException;
+import java.time.Instant;
+import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
 import java.time.temporal.TemporalAccessor;
 import java.util.LinkedList;
 import java.util.List;
@@ -40,7 +44,7 @@ public class EmbedBuilder
     public final static int TEXT_MAX_LENGTH = 2048;
     public final static int URL_MAX_LENGTH = 2000;
     public final static String ZERO_WIDTH_SPACE = "\u200E";
-    public final static Pattern URL_PATTERN = Pattern.compile("\\s*https?:\\/\\/.+\\..{2,}\\s*", Pattern.CASE_INSENSITIVE);
+    public final static Pattern URL_PATTERN = Pattern.compile("\\s*(https?|attachment):\\/\\/.+\\..{2,}\\s*", Pattern.CASE_INSENSITIVE);
     
     private String url;
     private String title;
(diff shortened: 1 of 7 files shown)


src/main/java/net/dv8tion/jda/core/entities/Message.java
```diff
@@ -307,6 +307,35 @@
...
+     *         The edit was attempted after the Message had been deleted.</li>
+     * </ul>
+     *
+     * @param  newContent
+     *         the new content of the Message
+     *
+     * @throws java.lang.IllegalStateException
+     *         If the message attempting to be edited was not created by the currently logged in account.
+     *
+     * @return a new Message-Object for the edited message
```
MinnDevelopment: not a `Message-Object`. 
Change this to the rest action

src/main/java/net/dv8tion/jda/core/entities/Message.java
```diff
@@ -307,6 +307,35 @@
...
+     *     <br>The edit was attempted after the account lost access to the
+     *         {@link net.dv8tion.jda.core.entities.Guild Guild} or {@link net.dv8tion.jda.client.entities.Group Group}
+     *         typically due to being kicked or removed.</li>
+     *
+     *     <li>{@link net.dv8tion.jda.core.requests.ErrorResponse#MISSING_PERMISSIONS MISSING_PERMISSIONS}
+     *     <br>The edit was attempted after the account lost {@link net.dv8tion.jda.core.Permission#MESSAGE_WRITE Permission.MESSAGE_WRITE} in
+     *         the {@link net.dv8tion.jda.core.entities.TextChannel TextChannel}.</li>
+     *
+     *     <li>{@link net.dv8tion.jda.core.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
+     *         The edit was attempted after the Message had been deleted.</li>
```
MinnDevelopment: Missing `<br>` after link

src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -460,6 +498,17 @@ public EmbedBuilder addBlankField(boolean inline)
         return this;
     }
     
+    /**
+     * Clears all fields from the embed.
+     *
+     * @return the builder after the field has been added
+     */
```
MinnDevelopment: This is a bit ambiguous. Please somehow link this to __#addField__ and explain what you exactly mean with "fields"

src/main/java/net/dv8tion/jda/core/entities/Message.java
MinnDevelopment: What happens if you pass `null`

MinnDevelopment:
What about #224 , does this replace the other PR or will 224 be updated?
If this includes all changes from 224 please close it.

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
```diff
@@ -466,8 +473,15 @@ protected void handleResponse(Response response, Request request)
...
-            body.field("content", message.getRawContent());
-            body.field("tts", message.isTTS());
+            if (message.getEmbeds().isEmpty())
+            {
+                body.field("content", message.getRawContent());
+                body.field("tts", message.isTTS());
+            }
+            else
+            {
+                body.field("payload_json", ((MessageImpl) message).toJSONObject().toString());
```
DV8FromTheWorld: Question: If `payload_json` exists, is there even a reason to use the above `body.field`s? Can we just send any form of message uploaded to Discord with a file in the `payload_json` or is it limited to only messages with embeds?

src/main/java/net/dv8tion/jda/core/entities/Message.java
```diff
@@ -302,11 +302,43 @@
...
+     *         the new content of the Message
+     *
+     * @throws java.lang.IllegalStateException
+     *         If the message attempting to be edited was not created by the currently logged in account, or
+     *         if the passed-in embed is null
+     *
+     * @return {@link net.dv8tion.jda.core.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.core.entities.Message Message}
+     *     <br>The {@link net.dv8tion.jda.core.entities.Message Message} with the updated content
+     */
+    RestAction<Message> editMessage(MessageEmbed newContent);
```
DV8FromTheWorld: Because we now provide this in Message, should we also provide it in MessageChannel?

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
jagrosh: we could switch completely, but this way is a bit more efficient (no json object creation and less total data to send)

DV8FromTheWorld: If we can switch completely then we should. The reason behind that being: If anything new is ever added to Message, it will be represented in the JSONObject returned by `toJSONObject()`, and will automatically be handled here with no code change. Where as if we use `body.field` we have to modify those fields in the future.

As such, lets change all file upload situations to use `payload_json`

MinnDevelopment:
This has been merged with #260 into the new Pull-Request #265

PR #262, opened 2017-02-17
Added PermissionOverrideAction

MinnDevelopment:
Idk why we don't have this already but here it is 🌱


DV8FromTheWorld:
We need to come to an agreement about OverrideType, then this can be merged.

PR #260, opened 2017-02-14
Embedbuilder; Append to description

Kaaz:
A method to append to the description of an embed. You can close the other one

PR #265, opened 2017-02-19
Embed-Related Changes / Embedbuilder; Append to description

MinnDevelopment:
This merges #259 and #260
> __jagrosh__
> can use more temporal accessors
> edit message with embed only
> send embeds with a file
> permissions check for editing a message
> clear embeds from embedbuilder

> __Kaaz__
> A method to append to the description of an embed. 

In addition I made `description` use a `StringBuilder`

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
@@ -401,17 +401,7 @@ public interface MessageChannel extends ISnowflake
         body.field("file", data, fileName);
 
         if (message != null)
-        {
-            if (message.getEmbeds().isEmpty())
-            {
-                body.field("content", message.getRawContent());
-                body.field("tts", message.isTTS());
-            }
-            else
-            {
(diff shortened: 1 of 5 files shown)


src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
```diff
@@ -402,8 +402,15 @@ protected void handleResponse(Response response, Request request)
...
-            body.field("content", message.getRawContent());
-            body.field("tts", message.isTTS());
+            if (message.getEmbeds().isEmpty())
+            {
+                body.field("content", message.getRawContent());
+                body.field("tts", message.isTTS());
+            }
+            else
+            {
+                body.field("payload_json", ((MessageImpl) message).toJSONObject().toString());
```
jagrosh: I think it was decided to always use payload_json

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
```diff
@@ -466,8 +473,15 @@ protected void handleResponse(Response response, Request request)
...
```
jagrosh: (and again here)

PR #281, opened 2017-03-12
Fixed a minor grammatical mistake.

thvardhan:
While reading the readme file I found this small mistake, which is not good for a high level API like JDA so I thought we should fix this up.


meowingtwurtle:
@thvardhan You need a comma between "Currently" and "there", so "Currently, there are..."

MinnDevelopment:
Please remake this in a single commit and open Pull Requests only to the `development` branch.
Make a new branch, base it on `development` and make your change.

See: https://github.com/DV8FromTheWorld/JDA/wiki/5)-Contributing

PR #282, opened 2017-03-12
Fixed a minor mistake in readme file.

thvardhan:
While reading the readme file I found this small mistake, which is not good for a high level API like JDA so I thought we should fix this up.


MinnDevelopment:
This has been pushed to development just now. See e787f5c

PR #286, opened 2017-03-15
New Event: guildMemberGameUpdateEvent

DreadMoirai:
Adam told me to do it


MinnDevelopment:
There are multiple things wrong here:
- There already is a UserGameUpdateEvent you just fire a second event 
- You used 4 commits for this change which could've been one

In addition, we are currently not looking for single events. We plan to look into all events with this new release.
We want to not only give this event attention but also all other events for documentation, hierarchy and convenience.
This only covers the convenience aspect and does that only for a single event without making a change but adding one additional event.

I know you probably had good intentions to add an event rather than replacing it but you should at least deprecate the one you want to replace so it can be safely removed in a successive version later on.

Based on these aspects I will close this Pull Request and hope you understand my reasoning.

DreadMoirai:
👌

PR #287, opened 2017-03-15
Added new PaginationAction and moved OrderAction to its own package. [Rebased]

MinnDevelopment:
The `PaginationAction` is iterable!
Example:
```java
MessageReaction reaction = message.getReactions().get(0);
for (User user : reaction.getUsers())
{
    System.out.printf("%s#%s%n", user.getName(), user.getDiscriminator());
}
```
Where **MessageReaction#getUsers()** returns a `ReactionPaginationAction : PaginationAction<User, ReactionPaginationAction>`

The `PaginationAction` itself is a `RestAction` but defines a `List` of the specified entity: `RestAction<List<T>>`

The `PaginationAction.PaginationIterator` iterates over all cached entities and retrieves a new list of entities as needed.
That way you can seamlessly iterate over all entities of an endpoint!


src/main/java/net/dv8tion/jda/core/requests/restaction/pagination/PaginationAction.java
```diff
@@ -0,0 +1,307 @@
...
+        return cached.size();
+    }
+
+    /**
+     * Whether the cache of this PaginationAction is empty.
+     * <br>Logically equivalent to {@code recentSize() == 0}.
+     *
+     * @return True, if no entities have been retrieve yet.
+     */
+    public boolean isEmpty()
```
DV8FromTheWorld: What is the point of this method and `recentSize`? Can't you just `getRecent` and check the values on that?

src/main/java/net/dv8tion/jda/core/requests/restaction/pagination/PaginationAction.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    /**
+     * The currently cached entities of recent execution tasks.
+     * <br>Every {@link net.dv8tion.jda.core.requests.RestAction RestAction} success
+     * adds to this List. (Thread-Safe due to {@link java.util.concurrent.CopyOnWriteArrayList CopyOnWriteArrayList})
+     *
+     * @return Immutable {@link java.util.List List} containing all currently cached entities for this PaginationAction
+     */
+    public List<T> getRecent()
```
DV8FromTheWorld: Consider renaming this as `getRecent` doesn't really explain what this does at a glance.
Additionally, the jdoc should specify that this **Doesn't** contain all entities unless pagination has been completely iterated through.

src/main/java/net/dv8tion/jda/core/requests/restaction/pagination/PaginationAction.java
MinnDevelopment: The idea was not having to wrap the cached entities in an immutable decoration just to know whether it is empty or not

PR #288, opened 2017-03-15
Added MessageChannel#getLatestMessageId and associated methods

MinnDevelopment:
Includes:

* MessageChannel#getLatestMessageId() : `String`
* MessageChannel#hasLatestMessage() : `boolean`
* ~~MessageChannel#getLatestMessage() : `RestAction<Message>`~~

It is not possible to keep track of the latest message when it is deleted as we do not provide cached messages in JDA and thus
are unable to go back in history to look for the next most recent message.

As a substitute it is always possible to simply use a `MessageHistory` instance

### DIFF
--- a/src/main/java/net/dv8tion/jda/client/entities/impl/GroupImpl.java
+++ b/src/main/java/net/dv8tion/jda/client/entities/impl/GroupImpl.java
@@ -41,11 +41,27 @@ public class GroupImpl implements Group
     private User owner;
     private String name;
     private String iconId;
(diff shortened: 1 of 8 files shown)


DV8FromTheWorld:
I would rather not include `getLatestMessage` strictly due to the fact that `MessageHistory` exists and learning how to properly use it should be encouraged. providing `getLastestMessage` is an overlap of API and also uses two different Routes in it's implementation.

As such, this is acceptable after `getLatestMessage` is removed.

PR #289, opened 2017-03-15
Fix typo in managed emote error message

foxbot:
A high-level lib like JDA should really avoid small spelling errors like these, glad to help out!


DV8FromTheWorld:
Pls

PR #293, opened 2017-03-17
Updated server error handling in Requester to account for 5xx errors …

MinnDevelopment:
…rather than simply buttflare html responses

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/requests/Requester.java
+++ b/src/main/java/net/dv8tion/jda/core/requests/Requester.java
@@ -30,10 +30,7 @@ import net.dv8tion.jda.core.entities.impl.JDAImpl;
 import net.dv8tion.jda.core.requests.Route.CompiledRoute;
 import net.dv8tion.jda.core.requests.ratelimit.BotRateLimiter;
 import net.dv8tion.jda.core.requests.ratelimit.ClientRateLimiter;
-import net.dv8tion.jda.core.requests.ratelimit.IBucket;
 import net.dv8tion.jda.core.utils.SimpleLog;
-
-import java.util.List;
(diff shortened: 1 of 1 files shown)


Almighty-Alpaca:
wouldn't it be nice if you could set the number retry attempts yourself in `JDA`/`JDABuilder` ?

DV8FromTheWorld:
Provide a way to customize the amount of attemps via JDABuilder. Default to 3 attempts (which is what it was before, this has made it 4).

MinnDevelopment:
@DV8FromTheWorld what should the maximum amount of attempts be

DV8FromTheWorld:
I guess like 10 i suppose. Perhaps it shouldn't be customizable. Is there really a reason to attempt a request more than 3 times if we get a 5xx error? it means something is wrong with discord..

DV8FromTheWorld:
This is acceptable after the trial attempts is changed from `4` to `3`.

src/main/java/net/dv8tion/jda/core/requests/Requester.java
```diff
@@ -115,33 +114,34 @@ public Long execute(Request apiRequest)
...
-                response = request.asString();
-                attempt++;
             }
-            if (response.getBody() != null && response.getBody().startsWith("<"))
+            while (attempt < 3 && response.getStatus() >= 500);
+
+            if (response.getStatus() >= 500)
             {
-                //Epic failure due to cloudflare. Attempted 4 times.
+                //Epic failure from other end. Attempted 4 times.
```
Almighty-Alpaca: `Attempted 4 times.` -> `Attempted 3 times.`

PR #295, opened 2017-03-19
Use proper member mention format if has nickname

iiBlake:


src/main/java/net/dv8tion/jda/core/entities/impl/MemberImpl.java
```diff
@@ -23,9 +23,7 @@
 import net.dv8tion.jda.core.utils.PermissionUtil;
 import org.apache.http.util.Args;
 
-import java.awt.Color;
 import java.time.OffsetDateTime;
-import java.util.*;
```
MinnDevelopment: Why?

MinnDevelopment:
This does not compile, you removed essential imports.
I suggest you reset your commits up to upstream/master and do another commit without modifying imports. If you used intellij make sure you also check the check marks out that are in the commit dialogue to not include "optimize imports" as your settings are apparently off.

DV8FromTheWorld:
Does not compile. Modification of imports is completely unneeded.

PR #297, opened 2017-03-22
Moved to primitive longs as primary snowflake storage type

MinnDevelopment:
In this experiment I changed almost all `id` fields to use `long` over `String`.
To increase performance and memory usage we use `trove4j` which provides HashMap implementations that use backing primitive arrays instead of `Long` arrays.

Test:
- [x] Compiles with no errors
- [x] Runs with no JSONExceptions
- [x] Runs with large sharded bots (LewdBot, DabBot, etc.)
- [x] Benchmarking

### DIFF
--- a/src/main/java/net/dv8tion/jda/client/entities/impl/ApplicationImpl.java
+++ b/src/main/java/net/dv8tion/jda/client/entities/impl/ApplicationImpl.java
@@ -66,7 +66,7 @@ public class ApplicationImpl implements Application
         return new RestAction<Application.Bot>(this.api, Route.Applications.CREATE_BOT.compile(getId()), null)
         {
             @Override
-            protected void handleResponse(final Response response, final Request request)
+            protected void handleResponse(final Response response, final Request<Application.Bot> request)
             {
                 if (response.isOk())
                     request.onSuccess(ApplicationImpl.this.bot = new BotImpl(response.getObject()));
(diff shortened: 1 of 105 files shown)


MinnDevelopment:
First benchmarking checks (VisualVM) have shown a major decrease in memory usage for this experiment.
Specific data is not yet retrieved.

DV8FromTheWorld:
Would like some graphs / statistics posted here of a comparison between `experimental/cache` and `master`'s 3.0 release.

DV8FromTheWorld:
The graphs that were posted in Discord, move them to here for public comparisons.
Will review the code in the coming days.

MinnDevelopment:
Here are the comparison graphs.
image
> Taken using Java VisualVM

src/main/java/net/dv8tion/jda/client/entities/impl/ApplicationImpl.java
```diff
@@ -62,7 +63,7 @@ public ApplicationImpl(final JDA api, final JSONObject object)
         if (this.hasBot())
             return new RestAction.EmptyRestAction<>(this.bot);
 
-        return new RestAction<Application.Bot>(this.api, Route.Applications.CREATE_BOT.compile(this.id), null)
+        return new RestAction<Application.Bot>(this.api, Route.Applications.CREATE_BOT.compile(getId()), null)
```
Almighty-Alpaca: why don't you use the long for Routes? would need one string creation less per RestAction

MinnDevelopment: The compile method only accepts String but maybe I can change that

DV8FromTheWorld:
This is on hold until #287 and #302 are merged.

MinnDevelopment:
This will be updated to fix conflicts once #302 is merged

MinnDevelopment:
@DV8FromTheWorld This is now ready for review.

PR #307, opened 2017-04-10
Fixed Events

MinnDevelopment:
- [x] Make all combined message events inherit GenericMessageEvent
- [x] Make GenericMessageEvent only provide messageId and MessageChannel
- [x] Split Reaction events into specific events like GuildMessageReactionEvent etc
- [x] Add Emote events

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
@@ -51,9 +51,9 @@ public class EmoteImpl implements Emote
     private final JDA api;
     private final HashSet<Role> roles;
 
+    private final Object mngLock = new Object();
     private volatile EmoteManager manager = null;
     private volatile EmoteManagerUpdatable managerUpdatable = null;
-    private Object mngLock = new Object();
 
     private boolean managed = false;
     private String name;
@@ -222,6 +222,7 @@ public class EmoteImpl implements Emote
         return "E:" + getName() + '(' + getIdLong() + ')';
     }
 
+    @Override
     public EmoteImpl clone()
     {
         if (isFake()) return null;
--- a/src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
@@ -49,13 +49,13 @@ public class EmoteImpl implements Emote
     private final long id;
     private final Guild guild;
     private final JDA api;
+    private final HashSet<Role> roles;
 
     private volatile EmoteManager manager = null;
     private volatile EmoteManagerUpdatable managerUpdatable = null;
     private Object mngLock = new Object();
 
     private boolean managed = false;
-    private HashSet<Role> roles = null;
     private String name;
 
     public EmoteImpl(long id,  Guild guild)
@@ -71,6 +71,7 @@ public class EmoteImpl implements Emote
         this.id = id;
         this.api = api;
         this.guild = null;
+        this.roles = null;
     }
 
     @Override
@@ -219,5 +220,14 @@ public class EmoteImpl implements Emote
     public String toString()
     {
         return "E:" + getName() + '(' + getIdLong() + ')';
+    }
+
+    public EmoteImpl clone()
+    {
+        if (isFake()) return null;
+        EmoteImpl copy = new EmoteImpl(id, guild).setManaged(managed).setName(name);
+        copy.roles.addAll(roles);
+        return copy;
+
     }
 }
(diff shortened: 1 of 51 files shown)


src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
```diff
@@ -220,4 +221,13 @@ public String toString()
     {
         return "E:" + getName() + '(' + getIdLong() + ')';
     }
+
+    public EmoteImpl clone()
+    {
+        if (isFake()) return null;
+        EmoteImpl copy = new EmoteImpl(id, guild).setManaged(managed).setName(name);
+        copy.roles.addAll(roles);
```
DV8FromTheWorld: Won't this NPE?

src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
```diff
@@ -71,6 +71,7 @@ public EmoteImpl(long id,  JDA api)
         this.id = id;
         this.api = api;
         this.guild = null;
+        this.roles = null;
```
DV8FromTheWorld: Should this really be null?

src/main/java/net/dv8tion/jda/core/events/message/react/MessageReactionRemoveAllEvent.java
```diff
@@ -16,31 +16,41 @@
...
     {
-        return messageId;
+        return getChannel() instanceof PrivateChannel ? (PrivateChannel) getChannel() : null;
     }
 
-    public MessageChannel getChannel()
+    public Group getGroup()
     {
-        return channel;
+        return getChannel() instanceof Group ? (Group) getChannel() : null;
```
DV8FromTheWorld: These should be reworked to use ChannelType instead of 2 calls to `getChannel()`

DV8FromTheWorld:
Actually, I don't see *any* of the `events.emote` classes. Did you forget to commit them?

src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
MinnDevelopment: If the `guild` is null this is a fake emote, which means we don't copy it `if (isFake()) return null;`

src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
MinnDevelopment: This was only moved from line 58 into the constructor to allow a final field.
It would also be null for fake emotes without these changes.

src/main/java/net/dv8tion/jda/core/handle/MessageReactionBulkRemoveHandler.java
```diff
@@ -54,10 +60,33 @@ protected Long handleInternally(JSONObject content)
...
+                manager.handle(
+                    new PrivateMessageReactionRemoveAllEvent(
+                            api, responseNumber,
+                            messageId, (PrivateChannel) channel));
+        }
+
+        manager.handle(
+            new MessageReactionRemoveAllEvent(
+                    api, responseNumber,
+                    messageId, channel));
```
Almighty-Alpaca: I don't like that we are currently firing two events where we only need one. We could fix this by making the `[Private/Group/Guild]MessageReactionRemoveAllEvent` events extend `MessageReactionRemoveAllEvent` and change the generic events to interfaces. 
This is just an example and applies to all other events too.

MinnDevelopment: I plan to go through all events again later on, this was just to fix message related events for now

DV8FromTheWorld: No. I do not want this changed.
The idea is that there are combined events and specific events. MessageReaction* is similar to the combined events for normal messages being stuff like MessageReceivedEvent.
To make them extend would be to remove the fact that they are considered "Combined" events.

src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
DV8FromTheWorld: Forgive me, Github had collapsed the code such that it looked like this was part of the the `public EmoteImpl(long, Guild)` constructor.

src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
DV8FromTheWorld: See above reply comment in different convo

DV8FromTheWorld:
This is squash-ready @MinnDevelopment

PR #309, opened 2017-04-11
Consider Snowflake IDs as unsigned

MinnDevelopment:
- Use `Long.toUnsignedString` to ensure we only return Strings without a leading `'-' `
- Add `MiscUtil.parseSnwoflake` to have a custom `NumberFormatException` to specify we are talking about a snowflake value

Should we add documentation for this exception, too?


DV8FromTheWorld:
@MinnDevelopment add `@throws` for the NFE and then we merge.

PR #313, opened 2017-04-14
Added abstraction to ErrorResponse 'FailureResponse' to handle unknown errors

MinnDevelopment:
This is to counter code 0 server errors and unknown response failures from discord.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/requests/FailureResponse.java
+++ b/src/main/java/net/dv8tion/jda/core/requests/FailureResponse.java
@@ -1,0 +1,51 @@
+/*
+ *     Copyright 2015-2017 Austin Keener & Michael Ritter
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


src/main/java/net/dv8tion/jda/core/requests/FailureResponse.java
```diff
@@ -0,0 +1,51 @@
...
+            return meaning;
+        }
+
+        @Override
+        public int getCode()
+        {
+            return code;
+        }
+    }
+
```
DV8FromTheWorld: Add a `default` `toString` implementation override here for ResponseError if possible.
If not, add one to UnknownFailure and modify ErrorResponse's toString to include the error code as well as the meaning.

DV8FromTheWorld: Also need to consider implementation details of `equals` and `hashcode` for UnknownFailure

DV8FromTheWorld:
Closing this as we have decided to pursue a different implementation to support unknown request failures.

PR #315, opened 2017-04-15
Add getMember() to Message.

Avarel:
This seems like a logical addition to the library as many messages are from Guilds, and reducing ceremony would be desirable.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/Message.java
@@ -132,7 +132,7 @@ public interface Message extends ISnowflake, Formattable
 
     /**
      * Returns the author of this Message as a {@link net.dv8tion.jda.core.entities.Guild Guild} member.
-     * <br>This is just a shortcut to {@link #getGuild()} ()}{@link net.dv8tion.jda.core.entities.Guild#getMember(User)} () .getMember(getAuthor())}.
+     * <br>This is just a shortcut to {@link #getGuild()}{@link net.dv8tion.jda.core.entities.Guild#getMember(User)} () .getMember(getAuthor())}.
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/core/entities/Message.java
```diff
@@ -131,6 +131,17 @@
     User getAuthor();
 
     /**
+     * Returns the author of this Message as a {@link net.dv8tion.jda.core.entities.Guild Guild} member.
+     * <br>This is just a shortcut to {@link #getGuild()} ()}{@link net.dv8tion.jda.core.entities.Guild#getMember(User)} () .getMember(getAuthor())}.
```
MinnDevelopment: `{@link #getGuild()} ()}` doesn't work. please validate your javadoc.

src/main/java/net/dv8tion/jda/core/entities/Message.java
```diff
@@ -131,6 +131,17 @@
...
+     * <br><b>This is only valid if the Message was actually sent in a TextChannel.</b> This will return {@code null}
+     * if it was not sent from a TextChannel.
```
MinnDevelopment: "valid" is an interesting choice of word here

meowingtwurtle:
This should be remade into one commit.

src/main/java/net/dv8tion/jda/core/entities/Message.java
Avarel: I basically based it on the Message#getGuild() javadoc. Since I thought it had a familiar function.

Avarel:
I updated the JavaDoc for the PR.

Avarel:
👍 Done.

PR #318, opened 2017-04-18
Various fixed to user limit handling

ArsenArsen:


MinnDevelopment:
This change is not enough as `PermissionUtil` does not handle channel permissions properly since some permissions make others available.
This handling has not been fixed and that is also the reason why this __known isssue__ was not addressed yet.

This PR will not be merged until `PermissionUtil` has been fixed to properly handle channel permissions.

DV8FromTheWorld:
This has been properly handled by #324

PR #320, opened 2017-04-19
Added a ShardManager

Almighty-Alpaca:
This makes it much easier to implement sharding as you can access the ShardManager fro mevery JDA entity, e.g. `event.getJDA().asBot().getShardManager()`. As it's in interface, users can also make their own implementations for sharding for sharding across multiple servers for example. This is also the reason why the methods are allowed to throw UnsupportedOperationExceptions.


Comportment:
v gud much comment

Sanduhr32:
such Alapaca :D

Almighty-Alpaca:
this is waiting for [`experimental/okhttp`](https://github.com/DV8FromTheWorld/JDA/tree/experimental/okhttp)

MinnDevelopment:
@Almighty-Alpaca can we make use of unified cache views in shard manager? For example ShardManager#getUserCache would use `SnowflakeCacheView.all( () -> shards.stream().map(JDA::getUserCache) )` and so forth. This way you can also implement the other methods like `getUserById(long)` by using `SnowflakeCacheView#getElementById(long)` (see JDAImpl)

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,546 @@
...
+    protected boolean enableShutdownHook = true;
+    protected boolean enableVoice = true;
+    protected IEventManager eventManager = null;
+    protected Game game = null;
+    protected boolean idle = false;
+    protected final List<Object> listeners = new ArrayList<>();
+    protected int maxReconnectDelay = 900;
+    protected OnlineStatus status = OnlineStatus.ONLINE;
+    protected String token = null;
+    private SessionReconnectQueue reconnectQueue;
```
MinnDevelopment: Group these fields by type such as `boolean` or `int` to give it a better structure.
Why is reconnectQueue the only `private` field here?

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,546 @@
...
+     *          If the provided token is empty or null.
+     * @throws  RateLimitedException
+     *          If we are being Rate limited.
+     *
+     * @return A {@link net.dv8tion.jda.bot.sharding.ShardManager ShardManager} instance that has started the login process. It is unknown as
+     *         to whether or not loading has finished when this returns.
+     */
+    public ShardManager buildAsync() throws LoginException, IllegalArgumentException, RateLimitedException
+    {
+        final DefaultShardManagerImpl manager = new DefaultShardManagerImpl(this.shardsTotal, this.shards, this.listeners, this.token, this.eventManager, this.audioSendFactory, this.game, this.status, this.httpClientBuilder, this.wsFactory, this.maxReconnectDelay, this.corePoolSize, this.enableVoice, this.enableShutdownHook, this.enableBulkDeleteSplitting, this.autoReconnect, this.idle, this.reconnectQueue, this.backoff);
```
MinnDevelopment: Can we wrap this over multiple lines

src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
```diff
@@ -0,0 +1,852 @@
...
+     */
+    default void removeEventListener(final Object... listeners)
+    {
+        this.getShardCache().forEach(jda -> jda.removeEventListener(listeners));
+    }
+
+    /**
+     * Restarts all shards
+     */
+    void restart();
```
MinnDevelopment: We should give more details on how this is done, via the SessionReconnectQueue and that it will take shardTotal * 5200 ms

src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
```diff
@@ -0,0 +1,852 @@
...
+
+    /**
+     * Adds a new shard with the given id to this ShardManager and starts it.
+     *
+     * @param shardId
+     *        The id of the shard that should be started
+     */
+    void start(int shardId);
+
+}
```
MinnDevelopment: Add newline at end of file

src/main/java/net/dv8tion/jda/core/JDA.java
```diff
@@ -76,27 +76,15 @@
     /**
      * Represents the information used to create this shard.
      */
-    class ShardInfo
+    interface ShardInfo
```
MinnDevelopment: Why was this changed?

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -524,11 +526,41 @@ public JDA buildAsync() throws LoginException, IllegalArgumentException, RateLim
...
+        this.listeners.add(listener);
         JDA jda = buildAsync();
-        while(jda.getStatus() != Status.CONNECTED)
+        synchronized (listener)
         {
-            Thread.sleep(50);
+            while (!listener.isReady())
+            {
+                listener.wait();
+            }
```
MinnDevelopment: JDA internals should not be using event listeners as user implementations could hurt this system. The old loop was fine.

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
```diff
@@ -122,23 +123,89 @@ public JDAImpl(AccountType accountType, OkHttpClient.Builder httpClientBuilder,
...
         setToken(token);
         verifyToken();
-        this.shardInfo = shardInfo;
         LOG.info("Login Successful!");
 
+        Pair<String, Integer> gateway = getGatewayAndRecommendedShardCount();
+
+        if (gateway == null) 
+        { 
+            throw new RuntimeException("Could not fetch WS-Gateway!");
```
MinnDevelopment: Can we use a better exception than `RuntimeException`

src/main/java/net/dv8tion/jda/core/utils/MiscUtil.java
```diff
@@ -90,6 +89,44 @@ public static String getDateTimeString(OffsetDateTime time)
...
+     * <br>Source for formula: <a href="https://discordapp.com/developers/docs/topics/gateway#sharding">Discord Documentation</a>
+     *
+     * @param guildId
+     *             The guild id.
+     * @param shards
+     *             The amount of shards.
+     * 
+     * @return The shard id for the guild.
+     */
+    public static int getShardForGuild(long guildId, int shards)
```
MinnDevelopment: The formatting of the `@param` docs is off

src/main/java/net/dv8tion/jda/core/utils/MiscUtil.java
```diff
@@ -90,6 +89,44 @@ public static String getDateTimeString(OffsetDateTime time)
...
+     * <br>Source for formula: <a href="https://discordapp.com/developers/docs/topics/gateway#sharding">Discord Documentation</a>
+     *
+     * @param guildId
+     *             The guild id.
+     * @param shards
+     *             The amount of shards.
+     * 
+     * @return The shard id for the guild.
+     */
+    public static int getShardForGuild(String guildId, int shards)
```
MinnDevelopment: The formatting of the `@param` docs is off

src/main/java/net/dv8tion/jda/core/JDA.java
Almighty-Alpaca: because I needed a way to set the total amount of shards while logging in for auto fetching the recommended amount of shards without adding additional REST calls

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
Almighty-Alpaca: as `gateway` actually is impossible to be `null` it is pretty useless to add a custom exception for it

MinnDevelopment: It doesn't need to be a custom exception, just RuntimeException is considered a "raw exception" which is not a very high code quality

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
```diff
@@ -0,0 +1,336 @@
...
+    protected final OnlineStatus status;
+    protected final String token;
+    protected ScheduledFuture<?> worker;
+
+    public DefaultShardManagerImpl(final int shardsTotal, final Collection<Integer> shardIds, final List<Object> listeners, final String token,
+            final IEventManager eventManager, final IAudioSendFactory audioSendFactory, final Game game, final OnlineStatus status,
+            final OkHttpClient.Builder httpClientBuilder, final WebSocketFactory wsFactory, ShardedRateLimiter shardedRateLimiter,
+            final int maxReconnectDelay, final int corePoolSize, final boolean enableVoice, final boolean enableShutdownHook,
+            final boolean enableBulkDeleteSplitting, final boolean autoReconnect, final boolean idle, final SessionReconnectQueue reconnectQueue,
+            final int backoff)
```
MinnDevelopment: Maybe write documentation on these fields? I see this being a public class that can be extended.

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
```diff
@@ -0,0 +1,336 @@
...
+import okhttp3.OkHttpClient;
+
+/**
+ * JDA's default {@link net.dv8tion.jda.bot.sharding.ShardManager ShardManager} implementation.
+ * To  create new instances use the {@link net.dv8tion.jda.bot.sharding.DefaultShardManagerBuilder DefaultShardManagerBuilder}.
+ *
+ * @since  3.4
+ * @author Aljoscha Grebe
+ */
+public class DefaultShardManagerImpl implements ShardManager
```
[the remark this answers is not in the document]
MinnDevelopment: This documentation has an excessive space here at `To  create new`

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
```diff
@@ -0,0 +1,336 @@
...
+
+        this.listeners.forEach(jda::addEventListener);
+        jda.setStatus(JDA.Status.INITIALIZED); //This is already set by JDA internally, but this is to make sure the listeners catch it.
+
+        // Set the presence information before connecting to have the correct information ready when sending IDENTIFY
+        ((PresenceImpl) jda.getPresence()).setCacheGame(this.game).setCacheIdle(this.idle).setCacheStatus(this.status);
+
+        final JDAImpl.ShardInfoImpl shardInfo = new JDAImpl.ShardInfoImpl(shardId, this.shardsTotal);
+
+        final int shardTotal = jda.login(this.token, shardInfo, this.reconnectQueue);
```
MinnDevelopment: If I see this correctly this will use the GET_GATEWAY route on every shard? I think we can optimize this here.

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
```diff
@@ -0,0 +1,336 @@
...
+            // as this happens before removing the shardId if from the queue just try again after 5 seconds
+            catch (RateLimitedException e)
+            {
+                LOG.warn("Hit the login ratelimit while creating new JDA instances, maybe try raising the login backoff");
+            }
+            catch (LoginException e)
+            {
+                // this can only happen if the token has been changed
+                // in this case the ShardManager will just shutdown itself as there currently is no way of hot-swapping the token on a running JDA instance.
+                LOG.warn("The token has been invalidated and the ShardManager will shutdown!");
```
MinnDevelopment: This should include the actual exception, at least on debug level. For all catches

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
```diff
@@ -0,0 +1,336 @@
...
+                    else
+                        shardId = i;
+                }
+                while (shardId == -1 && !this.queue.isEmpty());
+
+                if (shardId == -1)
+                {
+                    this.queue.poll();
+                    return;
+                }
```
MinnDevelopment: Why do you poll here?

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
```diff
@@ -0,0 +1,336 @@
...
+                // in this case the ShardManager will just shutdown itself as there currently is no way of hot-swapping the token on a running JDA instance.
+                LOG.warn("The token has been invalidated and the ShardManager will shutdown!");
+                this.shutdown();
+            }
+            catch (final Exception e)
+            {
+                if (api != null)
+                    api.shutdown();
+            }
+        }, 5000 + this.backoff, 5000 + this.backoff, TimeUnit.MILLISECONDS); // 5s for ratelimit + backoff for safety
```
MinnDevelopment: This `+ this.backoff` wouldn't be needed if you would wait for it to send the payload (waiting on JDA.Status to change to AWAITING_LOGIN_CONFIRMATION)

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
```diff
@@ -0,0 +1,336 @@
...
+            jda.shutdown();
+    }
+
+    @Override
+    public void start(final int shardId)
+    {
+        Checks.notNegative(shardId, "shardId");
+        Checks.check(shardId < this.shardsTotal, "shardId must be lower than shardsTotal");
+        this.queue.add(shardId);
+    }
```
MinnDevelopment: Does this allow starting a shard multiple times?

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,573 @@
...
+     * @param   listeners
+     *          The listener(s) to add to the list.
+     *
+     * @return The {@link net.dv8tion.jda.bot.sharding.DefaultShardManagerBuilder DefaultShardManagerBuilder} instance. Useful for chaining.
+     *
+     * @see    net.dv8tion.jda.bot.entities.impl.DefaultShardManagerImpl#addEventListener(Object...) JDA.addEventListener(Object...)
+     */
+    public DefaultShardManagerBuilder addEventListener(final Object... listeners)
+    {
+        Collections.addAll(this.listeners, listeners);
```
MinnDevelopment: Where do you do the null checks?

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,573 @@
...
+     *          If the provided token is empty or null.
+     * @throws  RateLimitedException
+     *          If we are being Rate limited.
+     *
+     * @return A {@link net.dv8tion.jda.bot.sharding.ShardManager ShardManager} instance that has started the login process. It is unknown as
+     *         to whether or not loading has finished when this returns.
+     */
+    public ShardManager buildAsync() throws LoginException, IllegalArgumentException, RateLimitedException
+    {
+        final DefaultShardManagerImpl manager = new DefaultShardManagerImpl(this.shardsTotal, this.shards, this.listeners, this.token, this.eventManager, this.audioSendFactory, this.game, this.status, this.httpClientBuilder, this.wsFactory, this.shardedRateLimiter, this.maxReconnectDelay, this.corePoolSize, this.enableVoice, this.enableShutdownHook, this.enableBulkDeleteSplitting, this.autoReconnect, this.idle, this.reconnectQueue, this.backoff);
```
MinnDevelopment: This line length is unacceptable.

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,573 @@
...
+     * @param  backoff
+     *         The new backoff in ms to use.
+     *
+     * @return Returns the {@link net.dv8tion.jda.core.JDABuilder JDABuilder} instance. Useful for chaining.
+     */
+    public DefaultShardManagerBuilder setLoginBackoff(int backoff)
+    {
+        this.backoff = backoff;
+        return this;
+    }
```
MinnDevelopment: This won't be needed if you properly wait until it sends the payload before backing off

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,573 @@
...
+     * @param  shardTotal
+     *         The number of overall shards or {@code -1} if JDA should use the recommended amount from discord.
+     *
+     * @return The {@link net.dv8tion.jda.bot.sharding.DefaultShardManagerBuilder DefaultShardManagerBuilder} instance. Useful for chaining.
+     *
+     * @see    #setShardRange(int, int)
+     */
+    public DefaultShardManagerBuilder setShardTotal(final int shardsTotal)
+    {
+        Checks.check(shardsTotal == -1 || shardsTotal > 0, "shardsTotal must either be -1 or more than 0");
```
MinnDevelopment: ~~more~~ greater

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,573 @@
...
+     *
+     * @param  shardedRateLimiter
+     *         ShardedRateLimiter used to keep track of cross-session rate limits
+     *
+     * @return The {@link net.dv8tion.jda.bot.sharding.DefaultShardManagerBuilder DefaultShardManagerBuilder} instance. Useful for chaining.
+     */
+    public DefaultShardManagerBuilder setShardedRateLimiter(ShardedRateLimiter shardedRateLimiter)
+    {
+        this.shardedRateLimiter = shardedRateLimiter;
+
```
MinnDevelopment: Excessive newline

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,573 @@
...
+     * <br>You can also create your own EventManager (See {@link net.dv8tion.jda.core.hooks.IEventManager}).
+     *
+     * @param  manager
+     *         The new {@link net.dv8tion.jda.core.hooks.IEventManager} to use.
+     *
+     * @return The {@link net.dv8tion.jda.bot.sharding.DefaultShardManagerBuilder DefaultShardManagerBuilder} instance. Useful for chaining.
+     */
+    public DefaultShardManagerBuilder setEventManager(final IEventManager manager)
+    {
+        this.eventManager = manager;
```
MinnDevelopment: Null checks?

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,573 @@
...
+     *         The new {@link net.dv8tion.jda.core.audio.factory.IAudioSendFactory IAudioSendFactory} to be used
+     *         when creating new {@link net.dv8tion.jda.core.audio.factory.IAudioSendSystem} objects.
+     *
+     * @return The {@link net.dv8tion.jda.bot.sharding.DefaultShardManagerBuilder DefaultShardManagerBuilder} instance. Useful for chaining.
+     */
+    public DefaultShardManagerBuilder setAudioSendFactory(final IAudioSendFactory factory)
+    {
+        this.audioSendFactory = factory;
+        return this;
+    }
```
MinnDevelopment: Some users use multiple different audio send factory implementations per shard, consider adding something like a provider as well.

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,573 @@
...
+     *
+     * @return The {@link net.dv8tion.jda.bot.sharding.DefaultShardManagerBuilder DefaultShardManagerBuilder} instance. Useful for chaining.
+     *
+     * @see    net.dv8tion.jda.core.managers.Presence#setGame(Game)
+     */
+    public DefaultShardManagerBuilder setGame(final Game game)
+    {
+        this.game = game;
+        return this;
+    }
```
MinnDevelopment: This method should be grouped together with `setIdle` and `setStatus` for better navigation.

src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
```diff
@@ -0,0 +1,855 @@
...
+ * @author Aljoscha Grebe
+ */
+public interface ShardManager
+{
+    /**
+     * Adds all provided listeners to the event-listeners that will be used to handle events.
+     * This uses the {@link net.dv8tion.jda.core.hooks.InterfacedEventManager InterfacedEventListener} by default.
+     * To switch to the {@link net.dv8tion.jda.core.hooks.AnnotatedEventManager AnnotatedEventManager}, use {@link #setEventManager(IEventManager)}.
+     *
+     * Note: when using the {@link net.dv8tion.jda.core.hooks.InterfacedEventManager InterfacedEventListener} (default),
```
MinnDevelopment: Missing `<p>` tag?

src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
```diff
@@ -0,0 +1,855 @@
...
+     * @throws net.dv8tion.jda.core.exceptions.AccountTypeException
+     *         Thrown if the currently logged in account is {@link net.dv8tion.jda.core.AccountType#CLIENT}
+     *
+     * @return The {@link net.dv8tion.jda.bot.JDABot} registry for this instance of JDA.
+     */
+    default RestAction<ApplicationInfo> getApplicationInfo()
+    {
+        return this.getShardCache().stream()
+                .findAny()
+                .orElseThrow(() -> new IllegalStateException("no active shards"))
```
MinnDevelopment: This is not a documented exception

src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
```diff
@@ -0,0 +1,855 @@
...
+    }
+
+    /**
+     * The average time in milliseconds between all shards that discord took to respond to our last heartbeat.
+     * <br>This roughly represents the WebSocket ping of this session.
+     *
+     * <p><b>{@link net.dv8tion.jda.core.requests.RestAction RestAction} request times do not
+     * correlate to this value!</b>
+     *
+     * @return time in milliseconds between heartbeat and the heartbeat ack response
```
MinnDevelopment: average time*

src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
```diff
@@ -0,0 +1,855 @@
...
+     *         If the provided ID is not a valid {@code long}
+     */
+    default Category getCategoryById(final String id)
+    {
+        return this.getCategoryCache().getElementById(id);
+    }
+
+    /**
+     * {@link net.dv8tion.jda.core.utils.cache.SnowflakeCacheView SnowflakeCacheView} of
+     * all cached {@link net.dv8tion.jda.core.entities.Category Categories} visible to this ShardManager session.
```
MinnDevelopment: Its not a "session" but an instance. Sessions is what we call actual discord connections.

src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
```diff
@@ -0,0 +1,855 @@
...
+
+    /**
+     * Gets the {@link net.dv8tion.jda.core.entities.Category Category} that matches the provided id. <br>If there is no
+     * matching {@link net.dv8tion.jda.core.entities.Category Category} this returns {@code null}.
+     *
+     * @param id
+     *         The snowflake ID of the wanted Category
+     * @return Possibly-null {@link net.dv8tion.jda.core.entities.Category Category} for the provided ID.
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided ID is not a valid {@code long}
```
MinnDevelopment: The formatting on some of these docs are bad/inconsistent.

src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
```diff
@@ -0,0 +1,855 @@
...
+     * @param users
+     *        The users which all the returned {@link net.dv8tion.jda.core.entities.Guild Guilds} must contain.
+     *
+     * @return Unmodifiable list of all {@link net.dv8tion.jda.core.entities.Guild Guild} instances which have all {@link net.dv8tion.jda.core.entities.User Users} in them.
+     */
+    default List<Guild> getMutualGuilds(final Collection<User> users)
+    {
+        Checks.notNull(users, "users");
+        for (final User u : users)
+            Checks.notNull(u, "All users");
```
MinnDevelopment: You can use `Checks.noneNull(Collection)` too.

src/main/java/net/dv8tion/jda/bot/utils/cache/ShardCacheView.java
```diff
@@ -0,0 +1,49 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.dv8tion.jda.bot.utils.cache;
+
+import net.dv8tion.jda.core.JDA;
+import net.dv8tion.jda.core.utils.cache.CacheView;
+
+public interface ShardCacheView extends CacheView<JDA>
```
MinnDevelopment: Class-level docs

src/main/java/net/dv8tion/jda/bot/utils/cache/ShardCacheView.java
```diff
@@ -0,0 +1,49 @@
...
+     *
+     * @param  id
+     *         The ID of the entity
+     *
+     * @return Possibly-null entity for the specified ID
+     */
+    JDA getElementById(int id);
+
+    /**
+     * Retrieves the JDA instance represented by the provided shard ID.
```
MinnDevelopment: `Or {@code null} if none of the connected shards match the provided id.`

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -44,7 +44,7 @@
  */
 public class JDABuilder
 {
-    protected final List<Object> listeners;
+    protected final List<Object> listeners = new LinkedList<>();
```
MinnDevelopment: Why change this?

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
```diff
@@ -124,23 +126,89 @@ public JDAImpl(AccountType accountType, OkHttpClient.Builder httpClientBuilder,
...
+                    {
+                        if (response.isOk())
+                        {
+                            JSONObject object = response.getObject();
+
+                            Pair<String, Integer> pair;
+                            if (accountType == AccountType.BOT)
+                                pair = Pair.of(object.getString("url") + "?encoding=json&v=" + WebSocketClient.DISCORD_GATEWAY_VERSION, object.getInt("shards"));
+                            else
+                                pair = Pair.of(object.getString("url") + "?encoding=json&v=" + WebSocketClient.DISCORD_GATEWAY_VERSION, null);
```
MinnDevelopment: Where do we now add the `compress=zlib-stream` parameter?

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
```diff
@@ -467,7 +535,7 @@ public void shutdown()
     }
 
     @Override
-    public JDAClient asClient()
+    public JDAClientImpl asClient()
```
MinnDevelopment: Why?

MinnDevelopment:
This branch currently won't start at all due to incorrect handling of zlib compression when retrieving the gateway.

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
MinnDevelopment: Just checked, nowhere. This is breaking and unacceptable.

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
Almighty-Alpaca: This is a good question. If the code gets here the queue is empty anyway. Fixed.

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
Almighty-Alpaca: It does allow adding an Id multiple times to the queue, but if the building thread polls from the queue it checks if the shard already exists and just moves on to the next queue entry if so.

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
Almighty-Alpaca: I need the implementation of JDABot to use the ShardManager setter and thought if I change that method this should be changed too. There's also no reason to hide the real class from implementation only methods as it will just lead to more casts.

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
MinnDevelopment: Can you document that case if it isn't already documented

Almighty-Alpaca: Will do that

src/main/java/net/dv8tion/jda/bot/entities/impl/DefaultShardManagerImpl.java
Almighty-Alpaca: have to think about the best way to make this working with #520

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
```diff
@@ -0,0 +1,577 @@
...
+    @Override
+    public void restart()
+    {
+        for (final int shardId : this.shards.getMap().keys())
+        {
+            final JDA jda = this.shards.getMap().remove(shardId);
+            if (jda != null)
+            {
+                jda.shutdown();
+                this.queue.offer(shardId);
```
MinnDevelopment: Use queue.add to throw an exception when rejected

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
```diff
@@ -0,0 +1,577 @@
...
+
+        JDA api = null;
+        try
+        {
+            api = shards == null ? null : shards.getElementById(shardId);
+
+            if (api == null)
+                api = this.buildInstance(shardId);
+            else if (api.getStatus() == JDA.Status.RECONNECT_QUEUED)
+                ((JDAImpl) api).getClient().reconnect(true, true);
```
MinnDevelopment: Declare the `api` variable as `JDAImpl` and cast when initlializing

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
```diff
@@ -0,0 +1,577 @@
...
+
+            this.gatewayURL = gateway.getLeft();
+
+            if (this.shardsTotal == -1)
+            {
+                this.shardsTotal = gateway.getRight();
+
+                this.shards = new ShardCacheViewImpl(this.shardsTotal);
+
+                for (int i = 0; i < shardsTotal; i++)
```
MinnDevelopment: You have too many empty lines in this section that just stretch this code for no apparent reason

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
```diff
@@ -0,0 +1,577 @@
...
+            || jda.getStatus().ordinal() < waitUntil.ordinal()) // Wait until status is bypassed
+        {
+            if (jda.getStatus() == JDA.Status.SHUTDOWN)
+                throw new IllegalStateException("JDA was unable to finish starting up!");
+
+            try
+            {
+                Thread.sleep(50);
+            }
+            catch (InterruptedException e) { /* ignored */ }
```
MinnDevelopment: See above

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,661 @@
...
+    protected IAudioSendFactory audioSendFactory = null;
+    protected IEventManager eventManager = null;
+    protected IntFunction<Game> gameProvider = null;
+    protected OkHttpClient.Builder httpClientBuilder = null;
+    protected IntFunction<OnlineStatus> statusProvider = null;
+    protected ShardedRateLimiter shardedRateLimiter = null;
+    protected String token = null;
+    protected WebSocketFactory wsFactory = null;
+    protected boolean retryOnTimeout = true;
+    protected final List<Object> listeners = new ArrayList<>();
```
MinnDevelopment: Group these fields by type

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -0,0 +1,661 @@
...
+     *          If the provided token is invalid.
+     * @throws  IllegalArgumentException
+     *          If the provided token is empty or null.
+     * @throws  RateLimitedException
+     *          If we are being Rate limited.
+     *
+     * @return A {@link net.dv8tion.jda.bot.sharding.ShardManager ShardManager} instance that has started the login process. It is unknown as
+     *         to whether or not loading has finished when this returns.
+     */
+    public ShardManager buildAsync() throws LoginException, IllegalArgumentException, RateLimitedException
```
MinnDevelopment: This should be at the bottom

MinnDevelopment: Is there no buildBlocking? Why call it buildAsync in the first place then?

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
```diff
@@ -0,0 +1,577 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void addEventListener(final Object... listeners)
+    {
+        Collections.addAll(this.listeners, listeners);
+        ShardManager.super.addEventListener(listeners);
+    }
```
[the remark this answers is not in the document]
Almighty-Alpaca: It's funny because the methods in JDABuilder/JDAImpl don't have null checks either. Will add them there, too.

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
Almighty-Alpaca: There used to be `buildBlocking()` but it was a really bad implementation. Additionally having a method that can block up to for 30 minutes (and potentially even way more depending on amount of shard) sounds like a bad idea to me. 
I think the best idea is to just rename `buildAsync()` to just `build()`.

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
```diff
@@ -0,0 +1,587 @@
...
+
+    @Override
+    public void restart(final int shardId)
+    {
+        Checks.notNegative(shardId, "shardId");
+        Checks.check(shardId < this.shardsTotal, "shardId must be lower than shardsTotal");
+
+        final JDA jda = this.shards.getMap().remove(shardId);
+        if (jda != null)
+            jda.shutdown();
```
Kodehawa: You might want to either use shutdownNow() or give the user the option to use shutdownNow() as in large bots this can make quite the difference on the time it takes from rebooting a shard for it to be unavaliable (example when on an error the shard starts spamming "unknown..."

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
```diff
@@ -0,0 +1,642 @@
...
+        if (gameProvider != null)
+            presence.setCacheGame(this.gameProvider.apply(shardId));
+        if (idleProvider != null)
+            presence.setCacheIdle(this.idleProvider.apply(shardId));
+        if (statusProvider != null)
+            presence.setCacheStatus(this.statusProvider.apply(shardId));
+
+        if (this.gatewayURL == null)
+        {
+            Pair<String, Integer> gateway = jda.getGatewayBot().complete();
```
MinnDevelopment: Reminder to handle invalid tokens here

PR #324, opened 2017-04-20
Cleanup for PermissionUtil and fixed user_limit check.

MinnDevelopment:
Changed some structure and design for the PermissionOverrideImpl internals

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/handle/ChannelUpdateHandler.java
+++ b/src/main/java/net/dv8tion/jda/core/handle/ChannelUpdateHandler.java
@@ -60,16 +60,16 @@ public class ChannelUpdateHandler extends SocketHandler
         List<IPermissionHolder> contained = new ArrayList<>();
 
         final long channelId = content.getLong("id");
-        String name = content.getString("name");
-        int position = content.getInt("position");
+        final int position = content.getInt("position");
+        final String name = content.getString("name");
         JSONArray permOverwrites = content.getJSONArray("permission_overwrites");
         switch (type)
         {
             case TEXT:
             {
                 String topic = content.isNull("topic") ? null : content.getString("topic");
-                TextChannelImpl channel = (TextChannelImpl) api.getTextChannelMap().get(channelId);
-                if (channel == null)
+                TextChannelImpl textChannel = (TextChannelImpl) api.getTextChannelMap().get(channelId);
+                if (textChannel == null)
                 {
                     api.getEventCache().cache(EventCache.Type.CHANNEL, channelId, () -> handle(responseNumber, allContent));
                     EventCache.LOG.debug("CHANNEL_UPDATE attempted to update a TextChannel that does not exist. JSON: " + content);
@@ -77,35 +77,35 @@ public class ChannelUpdateHandler extends SocketHandler
                 }
 
                 //If any properties changed, update the values and fire the proper events.
-                if (!StringUtils.equals(channel.getName(), name))
+                final String oldName = textChannel.getName();
+                final String oldTopic = textChannel.getTopic();
+                final int oldPosition = textChannel.getPositionRaw();
+                if (!StringUtils.equals(oldName, name))
                 {
-                    String oldName = channel.getName();
-                    channel.setName(name);
+                    textChannel.setName(name);
                     api.getEventManager().handle(
                             new TextChannelUpdateNameEvent(
                                     api, responseNumber,
-                                    channel, oldName));
+                                    textChannel, oldName));
                 }
-                if (!StringUtils.equals(channel.getTopic(), topic))
+                if (!StringUtils.equals(oldTopic, topic))
                 {
-                    String oldTopic = channel.getTopic();
-                    channel.setTopic(topic);
+                    textChannel.setTopic(topic);
                     api.getEventManager().handle(
                             new TextChannelUpdateTopicEvent(
                                     api, responseNumber,
-                                    channel, oldTopic));
+                                    textChannel, oldTopic));
                 }
-                if (channel.getPositionRaw() != position)
+                if (oldPosition != position)
                 {
-                    int oldPosition = channel.getPositionRaw();
-                    channel.setRawPosition(position);
+                    textChannel.setRawPosition(position);
                     api.getEventManager().handle(
                             new TextChannelUpdatePositionEvent(
                                     api, responseNumber,
-                                    channel, oldPosition));
+                                    textChannel, oldPosition));
                 }
 
-                applyPermissions(channel, content, permOverwrites, contained, changed);
+                applyPermissions(textChannel, content, permOverwrites, contained, changed);
 
                 //If this update modified permissions in any way.
                 if (!changed.isEmpty())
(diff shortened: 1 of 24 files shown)


src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
```diff
@@ -420,79 +481,165 @@ public static long getEffectivePermission(Channel channel, Role role)
...
+     *
+     * @throws IllegalArgumentException
+     *         If any of the arguments is {@code null}
+     *         or the specified entities are not from the same {@link net.dv8tion.jda.core.entities.Guild Guild}
+     *
+     * @return Primitive (unsigned) long value with the implicit permissions of the specified member in the specified channel
+     *
+     * @since  3.1
+     */
+    public static long getImplicitPermission(Channel channel, Member member)
```
DV8FromTheWorld: Is this code not identical to `getEffectivePermission(Channel, Member)`, minus the if statements that give the set of permissions based on `ADMINISTRATOR`, `MANAGE_CHANNEL` and `MANAGE_PERMISSIONS` ?

I ask, because if it is, shouldn't `getEffectivePermission` call `getImplicitPermission` and then just have those 2 if statements modify the returned value? Looking to remove duplicate code if possible.

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
```diff
@@ -420,79 +481,165 @@ public static long getEffectivePermission(Channel channel, Role role)
...
+     *
+     * @throws IllegalArgumentException
+     *         If any of the arguments is {@code null}
+     *         or the specified entities are not from the same {@link net.dv8tion.jda.core.entities.Guild Guild}
+     *
+     * @return Primitive (unsigned) long value with the implicit permissions of the specified role in the specified channel
+     *
+     * @since  3.1
+     */
+    public static long getImplicitPermission(Channel channel, Role role)
```
DV8FromTheWorld: Doesn't look like this method takes into account the actual permissions granted to the Public role nor the provided role, it only checks the override information. Remember, overrides can have "inherit" which neither allows nor denies.

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
```diff
@@ -420,79 +481,165 @@ public static long getEffectivePermission(Channel channel, Role role)
...
+    private static boolean isApplied(long permissions, Permission permission)
+    {
+        return (permissions & permission.getRawValue()) > 0;
+    }
+
     private static long apply(long permission, long allow, long deny)
     {
-        permission = permission | allow;    //Allow all the things that the cascade of roles allowed
-        permission = permission & (~deny);  //Deny everything that the cascade of roles denied.
+        deny &= ~allow;
```
DV8FromTheWorld: Why was this line added?
Just because something isn't specifically marked as allow **does not** mean it is denied.
Overrides have concept of inherit.

src/main/java/net/dv8tion/jda/core/handle/ChannelUpdateHandler.java
```diff
@@ -115,15 +117,28 @@ protected Long handleInternally(JSONObject content)
...
+                        .filter(Objects::nonNull)
+                        .filter(user -> !containedMembers.contains(user))
+                        .forEach(member -> {
+                            changedMembers.add(member);
+                            toRemove.add(member.getUser().getIdLong());
+                        });
+
+                toRemove.forEach((id) -> {
+                    channel.getOverrideMap().remove(id);
+                    return true;
```
MinnDevelopment: This is part of trove collections, when true is returned the next iteration step starts otherwise the forEach is interrupted.

DV8FromTheWorld: Interesting. Alrighty.

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
MinnDevelopment: I'm not sure what you mean. It does take the public role into account:
```java
long permission = 0;
PermissionOverride override = channel.getPermissionOverride(role.getGuild().getPublicRole());
if (override != null)
    permission = apply(permission, override.getAllowedRaw(), override.getDeniedRaw());
if (role.equals(role.getGuild().getPublicRole()))
    return permission;
```

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
MinnDevelopment: This makes sure that no bits in `deny` are also 1 in `allow`
Example:
`1010 & ~1000` is `1010 & 0111` -> `0010`
This is done because permissions that are allowed override the ones that are denied.

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
MinnDevelopment: The reason I did not do this is because in `getEffectivePermission` it is able to return without knowing all implicit permissions.
e.g. it can return if the public role has `Permission.ADMINISTRATOR` but `getImplicitPermission` retrieves **all** permissions and then returns which would make `getEffectivePermission` less efficient.

src/main/java/net/dv8tion/jda/core/handle/ChannelUpdateHandler.java
```diff
@@ -198,15 +213,28 @@ protected Long handleInternally(JSONObject content)
...
                 //Loop through all of the json defined overrides. If we find a match, remove the User or Role from our lists.
                 //Any entries remaining in these lists after this for loop is over will be removed from the Channel's overrides.
-                List<Role> collect = channel.getRoleOverrideMap().keySet().stream().filter(role -> !containedRoles.contains(role)).collect(Collectors.toList());
-                collect.forEach(role -> {
+                TLongList toRemove = new TLongLinkedList();
+                TDecorators.wrap(channel.getOverrideMap().keySet()).stream()
+                    .map(id -> channel.getGuild().getRoleById(id))
+                    .filter(Objects::nonNull)
+                    .filter(role -> !containedRoles.contains(role))
+                .forEach(role -> {
```
MinnDevelopment: This was due to bad copy-paste

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
DV8FromTheWorld: Then you should break the code up into helper methods. We needn't have so much duplicate code

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
DV8FromTheWorld: That is only dealing with the override information. I don't see you ever taking into consideration the `role.getPermissionRaw`.

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
DV8FromTheWorld: wouldn't it be better to just reorder the apply such that
```
permission &= ~deny
permission |= allow
```
?

MinnDevelopment: That probably works too.

src/main/java/net/dv8tion/jda/core/managers/impl/AudioManagerImpl.java
```diff
@@ -103,9 +105,15 @@ public void openAudioConnection(VoiceChannel channel)
                 return;
 
             final int userLimit = channel.getUserLimit(); // userLimit is 0 if no limit is set!
-            if (!self.hasPermission(channel, Permission.MANAGE_CHANNEL) && userLimit > 0 && userLimit <= channel.getMembers().size())
-                throw new PermissionException(Permission.MANAGE_CHANNEL,
-                        "Unable to connect to VoiceChannel due to userlimit! Requires permission MANAGE_CHANNEL to bypass");
+            final long perms = PermissionUtil.getImplicitPermission(channel, self);
```
DV8FromTheWorld: Calculate the implicit perms inside of the below if statement to avoid unnecessary permission calculations.

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
```diff
@@ -259,46 +255,40 @@ public static boolean checkPermission(Guild guild, Member member, Permission...
...
 
         GuildImpl guild = (GuildImpl) channel.getGuild();
         if (!guild.equals(member.getGuild()))
             throw new IllegalArgumentException("Provided channel and member are not from the same guild!");
 
-        if (guild.getOwner().equals(member) // Admin or owner? If yes: no need to iterate
-                || guild.getPublicRole().hasPermission(Permission.ADMINISTRATOR)
-                || member.getRoles().stream().anyMatch(role -> role.hasPermission(Permission.ADMINISTRATOR)))
-            return true;
+//        if (guild.getOwner().equals(member) // Admin or owner? If yes: no need to iterate
```
DV8FromTheWorld: Above code should use `checkGuild`

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
```diff
@@ -420,79 +432,179 @@ public static long getEffectivePermission(Channel channel, Role role)
...
-        long permission = getEffectivePermission(guild, member);
+        Args.notNull(channel, "Channel");
+        Args.notNull(role, "Role");
 
-        //override with channel-specific overrides of @everyone
-        PermissionOverride override = roleOverrides.get(guild.getPublicRole());
+        final Guild guild = role.getGuild();
+        checkGuild(channel.getGuild(), guild, "Role");
+
+        long permission = role.getPermissionsRaw() & guild.getPublicRole().getPermissionsRaw();
```
DV8FromTheWorld: Shouldn't this be an `|` ?

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
```diff
@@ -420,79 +432,179 @@ public static long getEffectivePermission(Channel channel, Role role)
...
+        PermissionOverride override = channel.getPermissionOverride(guild.getPublicRole());
         if (override != null)
-        {
             permission = apply(permission, override.getAllowedRaw(), override.getDeniedRaw());
+        if (role.equals(guild.getPublicRole()))
```
DV8FromTheWorld: `role.isPublicRole()`

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
```diff
@@ -420,79 +432,179 @@ public static long getEffectivePermission(Channel channel, Role role)
...
-                }
-                else
-                {
-                    allow = po.getAllowedRaw() | allow;     //Give all the stuff allowed by this Role's allow
-
-                    deny = (po.getDeniedRaw() | deny) & (~allow);  //Deny everything that this role denies.
-                    // This also rewrites the previous role's denies if this role allowed those permissions.
-                }
+                deny.accumulateAndGet(override.getDeniedRaw(), PermissionUtil::accumulate);
+                allow.accumulateAndGet(override.getAllowedRaw(), PermissionUtil::accumulate);
```
DV8FromTheWorld: Honestly these should probably just be in-line BiConsumers

MinnDevelopment: Can you elaborate?

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
MinnDevelopment: Good catch

src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java
MinnDevelopment:
```java
getImplicitOverrides(channel, member, (a, d) ->
{
    allow.accumulateAndGet(a, PermissionUtil::accumulate);
    deny.accumulateAndGet(d, PermissionUtil::accumulate);
});
```
Do you mean something close to this?

DV8FromTheWorld: no, I meant the `PermissionUtil::accumulate`

`allow.accumulateAndGet((val1, val2) -> val1 | val2);`

PR #326, opened 2017-04-21
Fixed 5 possible memory leaks 

Almighty-Alpaca:
Many thanks to @napstr for help me with these


src/main/java/net/dv8tion/jda/core/managers/impl/AudioManagerImpl.java
```diff
@@ -65,8 +66,8 @@
 
     public AudioManagerImpl(Guild guild)
     {
-        this.guild = guild;
-        this.api = (JDAImpl) guild.getJDA();
+        this.guild = (GuildImpl) guild;
```
MinnDevelopment: Wouldn't it make sense to just make the parameter `GuildImpl` itself?

ShanerX: Correct me if I'm wrong, but doesn't this defy the purpose of Guild being an interface at all?
If it's an interface, one can write their own implementation of it, however, here only GuildImpl works.
(I'm not too sure, so please enlighten me.)

jagrosh: Guild should only ever be a GuildImpl; its an interface to prevent users from making their own Guild objects

Almighty-Alpaca: The purpose of the interfaces is that user can't see the methods and fields used by their implementations as they have to be public because they have to be accessible from other parts of the API. You shouldn't and also can't create your own implementations of these classes (well, technically you can but as JDA will never use them it doesn't make any sense).

ShanerX: Thanks for clarifying, although that makes partially sense to me: can't one simply mess with the methods and fields of GuildImpl?

MinnDevelopment: JDA doesn't provide the user with an impl. It only gives them the interface which doesn't define setters. Unless the user makes a cast to the impl class they cannot access internal methods.

PR #331, opened 2017-04-27
Create getters for several fields in EmbedBuilder

Avarel:
Just a general quality of life improvement for when using EmbedBuilder.

Should other fields within the class also get the same treatment?

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
@@ -173,6 +173,26 @@ public class EmbedBuilder
     }
 
     /**
+     * The text of the embed title, set using {@link EmbedBuilder#setTitle(String, String)}.
+     *
+     * @return Possibly-null String containing the title of the embedded resource.
+     */
+    public String getTitle()
+    {
+        return title;
+    }
+
+    /**
+     * The URL of the embed title, set using {@link EmbedBuilder#setTitle(String, String)}.
+     *
+     * @return URL of the title of the embed or {@code null} if it is not set.
+     */
+    public String getUrl()
+    {
+        return url;
+    }
+
+    /**
      * The {@link java.lang.StringBuilder StringBuilder} used to
      * build the description for the embed.
      * <br>Note: To reset the description use {@link #setDescription(CharSequence) setDescription(null)}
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -238,6 +258,18 @@ public EmbedBuilder appendDescription(CharSequence description)
...
     /**
+     * The description of the embed, can be set using {@link EmbedBuilder#setDescription(CharSequence)}
+     * or appended using {@link EmbedBuilder#appendDescription(CharSequence)}.
+     *
+     * @return Possibly-null String containing a description of the embed.
+     */
+    public String getDescription()
+    {
+        return description.toString();
+    }
```
MinnDevelopment: `Possibly-null String containing a description of the embed`
In this case it would never be null, however it should be null if the StringBuilder is empty.

src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -173,6 +173,26 @@ public EmbedBuilder setTitle(String title, String url)
...
+    {
+        return title;
+    }
+
+    /**
+     * The URL of the embed title, set using {@link EmbedBuilder#setTitle(String, String)}.
+     *
+     * @return URL of the title of the embed or {@code null} if it is not set.
+     */
+    public String getUrl()
```
MinnDevelopment: Here you should mention that the url can only be non-null if `getTitle` isn't null either.

src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -292,6 +324,16 @@ else if (temporal instanceof OffsetDateTime)
         }
         return this; 
     }
+
+    /**
+     * The timestamp of the embed, set using {@link EmbedBuilder#setTimestamp(TemporalAccessor)}.
+     *
+     * @return Possibly-null OffsetDateTime object representing the timestamp.
```
MinnDevelopment: Link to `OffsetDateTime` would be helpful.

src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -303,6 +345,16 @@ public EmbedBuilder setColor(Color color)
         this.color = color;
         return this;
     }
+
+    /**
+     * The color of the EmbedBuilder, set using {@link EmbedBuilder#setColor(Color)}.
+     *
+     * @return Possibly-null Color of the embed.
+     */
+    public Color getColor()
```
MinnDevelopment: Missing link to `java.awt.Color`

PR #334, opened 2017-04-28
added nsfw channel support

jagrosh:


DV8FromTheWorld:
Doesn't handle different letter casing.

src/main/java/net/dv8tion/jda/core/entities/impl/TextChannelImpl.java
```diff
@@ -223,6 +223,11 @@ public String getTopic()
     }
 
     @Override
+    public boolean isNSFW() {
+        return name.equals("nsfw") || name.startsWith("nsfw-");
```
ShanerX: Change to ``return name.equalsIgnoreCase("nsfw") || name.toLowerCase().startsWith("nsfw-");`` should fix it

MinnDevelopment:
Does the casing matter for text channel names?

aurieh:
It doesn't.

DV8FromTheWorld:
I really don't like this "feature", but it is now squashable.

PR #335, opened 2017-04-28
PR for editing fields in an EmbedBuilder

jagrosh:
added EmbedBuilder#getFields for an editable list of embed fields
modified related methods to not disturb EntityBuilder while still providing checks

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
@@ -484,16 +484,6 @@ public class EmbedBuilder
     {
         if (name == null && value == null)
             return this;
-        else if (name == null || value == null)
-            throw new IllegalArgumentException("Both Name and Value must be set!");
-        else if (name.length() > MessageEmbed.TITLE_MAX_LENGTH)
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -528,6 +518,16 @@ public EmbedBuilder clearFields()
         this.fields.clear();
         return this;
     }
+    
+    /**
+     * Returns a modifiable list of MessageEmbed Fields that the builder contains.
```
MinnDevelopment: Make `modifiable` bold

src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -37,8 +38,9 @@
 public class EmbedBuilder
 {
     public final static String ZERO_WIDTH_SPACE = "\u200E";
-    public final static Pattern URL_PATTERN = Pattern.compile("\\s*(https?|attachment):\\/\\/.+\\..{2,}\\s*", Pattern.CASE_INSENSITIVE);
-    
+    public final static Pattern URL_PATTERN = Pattern.compile("\\s*(https?|attachment)://.+\\..{2,}\\s*", Pattern.CASE_INSENSITIVE);
```
DV8FromTheWorld: Why was this change made?

PR #342, opened 2017-05-01
Added TextChannel#clearReactionsById

DreadMoirai:
also added TextChannel#clearReactionsById(long messageId) as a default method
copied over the docs and changed relevant parts

also removed a `@throws IllegalArgumentException` from Message#clearReactions since it doesn't take any arguments

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/TextChannel.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/TextChannel.java
@@ -193,9 +193,7 @@ public interface TextChannel extends Channel, MessageChannel, Comparable<TextCha
 
     /**
      * Attempts to remove all reactions from a message with the specified {@code messageId} in this TextChannel
-     *
      * <br>This is useful for moderator commands that wish to remove all reactions at once from a specific message.
-     *
      *
      * <p>The following {@link net.dv8tion.jda.core.requests.ErrorResponse ErrorResponses} are possible:
      * <ul>
@@ -217,7 +215,7 @@ public interface TextChannel extends Channel, MessageChannel, Comparable<TextCha
      *         If the currently logged in account does not have
      *         {@link net.dv8tion.jda.core.Permission#MESSAGE_MANAGE Permission.MESSAGE_MANAGE} in this channel.
      * @throws java.lang.IllegalArgumentException
-     *         If the provided unicode emoji is null or empty.
+     *         If the provided {@code id} is {@code null} or empty.
      *
      * @return {@link net.dv8tion.jda.core.requests.RestAction RestAction} - Type: {@link java.lang.Void}
      */
@@ -225,9 +223,7 @@ public interface TextChannel extends Channel, MessageChannel, Comparable<TextCha
 
     /**
      * Attempts to remove all reactions from a message with the specified {@code messageId} in this TextChannel
-     *
      * <br>This is useful for moderator commands that wish to remove all reactions at once from a specific message.
-     *
      *
      * <p>The following {@link net.dv8tion.jda.core.requests.ErrorResponse ErrorResponses} are possible:
      * <ul>
@@ -248,8 +244,6 @@ public interface TextChannel extends Channel, MessageChannel, Comparable<TextCha
      * @throws net.dv8tion.jda.core.exceptions.PermissionException
      *         If the currently logged in account does not have
      *         {@link net.dv8tion.jda.core.Permission#MESSAGE_MANAGE Permission.MESSAGE_MANAGE} in this channel.
-     * @throws java.lang.IllegalArgumentException
-     *         If the provided unicode emoji is null or empty.
      *
      * @return {@link net.dv8tion.jda.core.requests.RestAction RestAction} - Type: {@link java.lang.Void}
      */
--- a/src/main/java/net/dv8tion/jda/core/entities/TextChannel.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/TextChannel.java
@@ -192,6 +192,72 @@ public interface TextChannel extends Channel, MessageChannel, Comparable<TextCha
     RestAction<Void> deleteWebhookById(String id);
 
     /**
+     * Attempts to remove all reactions from a message with the specified {@code messageId} in this TextChannel
+     *
+     * <br>This is useful for moderator commands that wish to remove all reactions at once from a specific message.
+     *
+     *
+     * <p>The following {@link net.dv8tion.jda.core.requests.ErrorResponse ErrorResponses} are possible:
+     * <ul>
+     *     <li>{@link net.dv8tion.jda.core.requests.ErrorResponse#MISSING_ACCESS MISSING_ACCESS}
+     *     <br>The clear-reactions request was attempted after the account lost access to the {@link net.dv8tion.jda.core.entities.TextChannel TextChannel}
+     *         due to {@link net.dv8tion.jda.core.Permission#MESSAGE_READ Permission.MESSAGE_READ} being revoked, or the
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/core/entities/impl/TextChannelImpl.java
```diff
@@ -386,6 +386,23 @@ public JDA getJDA()
     }
 
     @Override
+    public RestAction<Void> clearReactionsByMessageId(String messageId)
+    {
+        checkPermission(Permission.MESSAGE_MANAGE);
+        return new RestAction<Void>(getJDA(), Route.Messages.REMOVE_ALL_REACTIONS.compile(getId(), messageId), null)
```
MinnDevelopment: Move the `Route.Messages.REMOVE_ALL_REACTIONS.compile(getId(), messageId)` into its own variable

src/main/java/net/dv8tion/jda/core/entities/TextChannel.java
```diff
@@ -192,6 +192,72 @@
...
+     *
+     *     <li>{@link net.dv8tion.jda.core.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
+     *         The clear-reactions request was attempted after the Message had been deleted.</li>
+     * </ul>
+     *
+     * @throws net.dv8tion.jda.core.exceptions.PermissionException
+     *         If the currently logged in account does not have
+     *         {@link net.dv8tion.jda.core.Permission#MESSAGE_MANAGE Permission.MESSAGE_MANAGE} in this channel.
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided unicode emoji is null or empty.
```
MinnDevelopment: This never throws

src/main/java/net/dv8tion/jda/core/entities/TextChannel.java
```diff
@@ -192,6 +192,72 @@
...
+     */
+    RestAction<Void> clearReactionsByMessageId(String messageId);
+
+    /**
+     * Attempts to remove all reactions from a message with the specified {@code messageId} in this TextChannel
+     *
+     * <br>This is useful for moderator commands that wish to remove all reactions at once from a specific message.
+     *
+     *
+     * <p>The following {@link net.dv8tion.jda.core.requests.ErrorResponse ErrorResponses} are possible:
```
MinnDevelopment: There is too much whitespace here don't make a line between the first and second sentence and only one line between the first and second paragraph

src/main/java/net/dv8tion/jda/core/entities/TextChannel.java
```diff
@@ -192,6 +192,72 @@
...
+     *
+     * @throws net.dv8tion.jda.core.exceptions.PermissionException
+     *         If the currently logged in account does not have
+     *         {@link net.dv8tion.jda.core.Permission#MESSAGE_MANAGE Permission.MESSAGE_MANAGE} in this channel.
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided unicode emoji is null or empty.
+     *
+     * @return {@link net.dv8tion.jda.core.requests.RestAction RestAction} - Type: {@link java.lang.Void}
+     */
+    RestAction<Void> clearReactionsByMessageId(String messageId);
```
MinnDevelopment: Better `clearReactionsById`

src/main/java/net/dv8tion/jda/core/entities/TextChannel.java
```diff
@@ -192,6 +192,72 @@
...
+     *
+     *     <li>{@link net.dv8tion.jda.core.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
+     *         The clear-reactions request was attempted after the Message had been deleted.</li>
+     * </ul>
+     *
+     * @throws net.dv8tion.jda.core.exceptions.PermissionException
+     *         If the currently logged in account does not have
+     *         {@link net.dv8tion.jda.core.Permission#MESSAGE_MANAGE Permission.MESSAGE_MANAGE} in this channel.
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided unicode emoji is null or empty.
```
MinnDevelopment: This should be if its empty or null

src/main/java/net/dv8tion/jda/core/entities/impl/TextChannelImpl.java
MinnDevelopment: Also missing a `Args.notEmpty(messageId, "Message ID");`

MinnDevelopment:
Even when those show as outdated they are still correct.

src/main/java/net/dv8tion/jda/core/entities/TextChannel.java
MinnDevelopment: This was already done.

DreadMoirai:
ok, fixed all of it. accidentally made 3 commits when I only wanted to make one oops

src/main/java/net/dv8tion/jda/core/entities/impl/MessageImpl.java
```diff
@@ -142,7 +142,8 @@ else if (reaction.isSelf())
     public RestAction<Void> clearReactions()
     {
         checkPermission(Permission.MESSAGE_MANAGE);
-        return new RestAction<Void>(getJDA(), Route.Messages.REMOVE_ALL_REACTIONS.compile(getChannel().getId(), getId()), null)
+        Route.CompiledRoute route = Route.Messages.REMOVE_ALL_REACTIONS.compile(getChannel().getId(), getId());
+        return new RestAction<Void>(getJDA(), route, null)
```
MinnDevelopment: This method should use TextChannel#clearReactions and throw an exception if this message was not sent in a TextChannel

DreadMoirai: ok, used a `IllegalStateException` with message "Cannot clear reactions from a message in a Group or PrivateChannel."

src/main/java/net/dv8tion/jda/core/entities/impl/MessageImpl.java
```diff
@@ -141,19 +141,10 @@ else if (reaction.isSelf())
...
-                if (response.isOk())
-                    request.onSuccess(null);
-                else
-                    request.onFailure(response);
-            }
-        };
+        if (!isFromType(ChannelType.TEXT)) {
+            throw new IllegalStateException("Cannot clear reactions from a message in a Group or PrivateChannel.");
+        }
+        return ((TextChannel) channel).clearReactionsById(getId());
```
MinnDevelopment: Remove the brackets (they don't follow our style anyway) and use getTextChannel()

MinnDevelopment: Also add this exception to the docs

DV8FromTheWorld:
We need to establish a rule here:
Do we do `long` -> `string` or `string` -> `long` methods?
Currently, everywhere else is `string` -> `long`

This is `long` -> `string`

Other than that, this is good to squash.

MinnDevelopment:
It depends on the method. The methods that do `String` -> `long` usually access the cache. This method however only constructs a RestAction which needs a `String` to compile anyway.

DreadMoirai:
Do you want me to switch it around?

DreadMoirai:
I did `long` -> `string` because that's how i saw MessageChannel#getMessageById worked

MinnDevelopment:
@DreadMoirai You did good, dw.

PR #343, opened 2017-05-04
MessagePaginationAction and some adjustments to MessageHistory

MinnDevelopment:


src/main/java/net/dv8tion/jda/core/events/guild/voice/GuildVoiceJoinEvent.java
```diff
@@ -20,18 +20,12 @@
 import net.dv8tion.jda.core.entities.Member;
 import net.dv8tion.jda.core.entities.VoiceChannel;
 
-public class GuildVoiceJoinEvent extends GenericGuildVoiceEvent
+public class GuildVoiceJoinEvent extends GuildVoiceMoveEvent
```
Almighty-Alpaca: I don't really like this, join != move

src/main/java/net/dv8tion/jda/core/events/guild/voice/GuildVoiceLeaveEvent.java
```diff
@@ -20,18 +20,10 @@
 import net.dv8tion.jda.core.entities.Member;
 import net.dv8tion.jda.core.entities.VoiceChannel;
 
-public class GuildVoiceLeaveEvent extends GenericGuildVoiceEvent
+public class GuildVoiceLeaveEvent extends GuildVoiceMoveEvent
```
Almighty-Alpaca: same as above

MinnDevelopment:
I'm not sure I like combining the voice events, either

MinnDevelopment:
Changes made in this PR related to the PaginationIterator cache need to be squashed before audit log development can continue.

PR #348, opened 2017-05-08
Fix WS Sending race condition.

DV8FromTheWorld:
Fixed bug where packets sent by the WS ratelimit system were sent before IDENTIFY due to racecondition.


DV8FromTheWorld:
I don't really see the difference here. We don't need millisecond level accuracy

PR #349, opened 2017-05-10
Replaced Unirest with OkHttp for less static abuse

Almighty-Alpaca:
and made a few other enhancements


src/main/java/net/dv8tion/jda/core/entities/TextChannel.java
```diff
@@ -246,7 +250,11 @@
      * </ul>
      *
      * @param  messageId
+<<<<<<< master
      *         The message id
+=======
+     *         The target message's id
+>>>>>>> Squashed and rebased onto master
```
MinnDevelopment: This needs to go.

src/main/java/net/dv8tion/jda/core/entities/impl/VoiceChannelImpl.java
```diff
@@ -1,133 +1,133 @@
...
-        return this;
-    }
-
-    // -- Map Getters --
-
-    public TLongObjectMap<Member> getConnectedMembersMap()
-    {
-        return connectedMembers;
-    }
-}
```
MinnDevelopment: What happened here?

src/main/java/net/dv8tion/jda/core/requests/Response.java
```diff
@@ -16,30 +16,80 @@
...
+            }
+            else if (begin == '[')
+            {
+                object = new JSONArray(new JSONTokener(reader));
+            }
+        }
+        catch (Exception e)
+        {
+            // TODO rethrow?
+            e.printStackTrace();
```
MinnDevelopment: I'd suggest to only rethrow here, not print the trace

src/main/java/net/dv8tion/jda/core/requests/Response.java
```diff
@@ -61,36 +111,11 @@ public boolean isRateLimit()
...
-    {
-        return responseText;
-    }
-
-    public String toString()
-    {
-        return exception == null ? "HTTPResponse[" + code + ": " + responseText + ']'
-                : "HTTPException[" + exception.getMessage() + ']';
+        return (JSONArray) object;
     }
```
MinnDevelopment: Why remove `toString()`?
Before this change it would simply return `null` if there were no JSONArray now it throws ClassCastException

src/main/java/net/dv8tion/jda/core/requests/RestAction.java
```diff
@@ -616,18 +650,34 @@ public T completeAfter(long delay, TimeUnit unit)
...
+    RequestBody getData()
     {
-        return null;
+        return this.data;
+    }
+
+    CaseInsensitiveMap<String, String> getHeaders()
+    {
+        return this.headers;
     }
```
MinnDevelopment: Make getData and getHeaders `protected`

src/main/java/net/dv8tion/jda/core/requests/RestAction.java
```diff
@@ -679,4 +739,44 @@ public T complete(boolean shouldQueue)
...
+    /**
+     * Specialized form of {@link net.dv8tion.jda.core.requests.RestAction} that is used to provide information that
+     * has already been retrieved or generated so that another request does not need to be made to Discord.
+     * <br>Basically: Allows you to provide a value directly to the success returns.
+     *
+     * @param <T>
+     *        The generic response type for this RestAction
+     */
+    public static class FailedRestAction<T> extends RestAction<T>
```
MinnDevelopment: This javadoc is misleading 
Did you also add one of these to `AuditableRestAction`?

src/main/java/net/dv8tion/jda/core/utils/WidgetUtil.java
```diff
@@ -191,33 +190,44 @@ public static Widget getWidget(String guildId) throws RateLimitedException
...
             {
-                case 200: return new Widget(result.getBody().getObject()); // ok
+                case 200: // ok
+                    try (InputStream stream = connection.getInputStream())
+                    {
+                        return new Widget(new JSONObject(new JSONTokener(stream)));
+                    }
+                    catch (IOException ignored) {
+                        return null;
+                    }
```
MinnDevelopment: Add a `{}` scope to the case here, looks a bit weird imo and the other case has it.

src/main/java/net/dv8tion/jda/core/JDA.java
```diff
@@ -743,23 +734,29 @@ public boolean equals(Object o)
...
+     * 
+     * @see #shutdown(long, TimeUnit)
      */
     void shutdown();
 
     /**
-     * Shuts down JDA, closing all its connections.
+     * Force closes this {@link JDA} instance violently. This will also cancel all queued {@link net.dv8tion.jda.core.requests.RestAction RestActions}.
+     */
+    void forceShutdown(); // TODO: docs forceShutdown()
```
MinnDevelopment: `TODO: docs forceShutdown()`

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -119,52 +116,23 @@ public JDABuilder setToken(String token)
...
+    // on the other side this allows them to configure the behaviour more deeply without
+    // making JDABuilder overly complicated by adding too many methods 
+    public JDABuilder setHttpClientBuilder(OkHttpClient.Builder httpClientBuilder)
     {
-        if (jdaCreated)
-            throw new UnsupportedOperationException("You cannot change the proxy after a JDA object has been created. Proxy settings are global among all instances!");
-        this.proxy = proxy;
+        this.httpClientBuilder = httpClientBuilder;
         return this;
     }
```
MinnDevelopment: Incomplete documentation, need to decide if this is a good handling to allow setting the builder

src/main/java/net/dv8tion/jda/core/entities/impl/VoiceChannelImpl.java
Almighty-Alpaca: line ending change happened here, not sure why though

src/main/java/net/dv8tion/jda/core/requests/RestAction.java
Almighty-Alpaca: added that now

src/main/java/net/dv8tion/jda/core/JDABuilder.java
MinnDevelopment: Don't remove setWebSocketTimeout and setProxy. Make them deprecated with a notice of removal in a later version

Almighty-Alpaca: I can't keep `setProxy` as the `HttpHost` class doesn't exist anymore. I laso don't really like the idea with the `WebSocketOptions` class anymore as it would just be a clone of the WebSocketFactory in the end (same for the http stuff).

@DV8FromTheWorld do you have any idea what we should do with this?

src/main/java/net/dv8tion/jda/core/utils/WidgetUtil.java
```diff
@@ -191,33 +190,47 @@ public static Widget getWidget(String guildId) throws RateLimitedException
...
-                case 200: return new Widget(result.getBody().getObject()); // ok
+                case 200: // ok
+                {
+                    try (InputStream stream = connection.getInputStream())
+                    {
+                        return new Widget(new JSONObject(new JSONTokener(stream)));
+                    }
+                    catch (IOException e)
+                    {
+                        throw e;
```
MinnDevelopment: Throw RuntimeException here already not later to have the more accurate trace

src/main/java/net/dv8tion/jda/core/requests/RateLimiter.java
```diff
@@ -74,10 +72,16 @@ public boolean isRateLimited(CompiledRoute route)
...
         isShutdown = true;
 
+        pool.setKeepAliveTime(time, unit);
+        pool.allowCoreThreadTimeOut(true);
+    }
+
+    public void forceShutdown()
+    {
         pool.shutdownNow();
     }
```
MinnDevelopment: Why are you not setting isShutdown here?

src/main/java/net/dv8tion/jda/core/JDA.java
```diff
@@ -743,34 +735,42 @@ public boolean equals(Object o)
 
     /**
      * Shuts down JDA, closing all its connections.
-     *
-     * <p>This is the same as calling {@link #shutdown(boolean) shutdown(true)}.
+     * After this command is issued the JDA Instance can not be used anymore.
+     * Already enqueued {@link net.dv8tion.jda.core.requests.RestAction RestActions} are still be executed.
      */
     void shutdown();
```
MinnDevelopment: Are they actually still executed?
Requester#request rejects new requests after the rateLimiter is shutdown

Also `are still be executed` makes no sense as a statement.

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
```diff
@@ -524,30 +525,36 @@ protected void handleResponse(Response response, Request<Message> request)
...
+        MultipartBody.Builder builder = new okhttp3.MultipartBody.Builder()
+                .setType(MultipartBody.FORM);
+        try
+        {
+            builder.addFormDataPart("file", fileName, RequestBody.create(MediaType.parse("application/octet-stream"), ByteString.read(data, Integer.MAX_VALUE)));
+        }
+        catch (IOException e)
+        {
+            return new RestAction.FailedRestAction<Message>(e);
+        }
```
Almighty-Alpaca: In that case this would be the only RestAction that could throw an Exception. That's why I decided to use the error consumer the user already provides instead.

src/main/java/net/dv8tion/jda/core/JDA.java
Almighty-Alpaca: A long as any non-deamon thread still keeps the JVM alive yes. `Requester#request()` is called when queueing RestActions, that's why only ***already enqueued** RestActions are still executed.*

The only problem is that JDA doesn't have any non-deamon threds at that time which could be fixed by making the executor(s) non-deamon.

src/main/java/net/dv8tion/jda/core/entities/MessageHistory.java
```diff
@@ -156,12 +156,11 @@ public MessageChannel getChannel()
...
 
-        Route.CompiledRoute route;
-        if (history.isEmpty())
-            route = Route.Messages.GET_MESSAGE_HISTORY.compile(channel.getId(), Integer.toString(amount));
-        else
-            route = Route.Messages.GET_MESSAGE_HISTORY_BEFORE.compile(channel.getId(), Integer.toString(amount), String.valueOf(history.lastKey()));
-        return new RestAction<List<Message>>(getJDA(), route, null)
+        Route.CompiledRoute route = Route.Messages.GET_MESSAGE_HISTORY.compile(channel.getId()).withQueryParams("limit", Integer.toString(amount));
+        if (!history.isEmpty())
+            route.withQueryParams("before",String.valueOf(history.lastKey()));
```
Almighty-Alpaca: fixed

src/main/java/net/dv8tion/jda/core/requests/RateLimiter.java
Almighty-Alpaca: Because it is already set by `shutdown()` as`JDA#shutdownNow()` calls `JDA#shutdown()`

src/main/java/net/dv8tion/jda/core/requests/Requester.java
```diff
@@ -99,58 +105,67 @@ public JDAImpl getJDA()
...
+                        url,
+                        response.code(), attempt));
                 try
                 {
                     Thread.sleep(50 * attempt);
                 }
                 catch (InterruptedException ignored) {}
             }
-            while (attempt < 3 && response.getStatus() >= 500);
+            while (attempt < 3 && response.code() >= 500);
```
Almighty-Alpaca: Inside our outside the `while` block?

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
MinnDevelopment: Why would the RestAction cause the IOException if you catch it inside the method then the method can also throw it right there before you execute it, no?

src/main/java/net/dv8tion/jda/core/requests/Requester.java
MinnDevelopment: Outside the `while` makes more sense I would say

src/main/java/net/dv8tion/jda/core/JDA.java
MinnDevelopment: Oh my bad I mixed up Requester#execute and Requester#request

src/main/java/net/dv8tion/jda/core/requests/RestAction.java
```diff
@@ -615,18 +652,44 @@ public T completeAfter(long delay, TimeUnit unit)
...
+
+    protected RequestBody getData()
+    {
+        return this.data;
+    }
+
+    protected CaseInsensitiveMap<String, String> getHeaders()
+    {
+        return this.headers;
+    }
```
Almighty-Alpaca: Good catch, didn't think about this. But I think this is also problematic for the fields. Fixed it in the next commit.

src/main/java/net/dv8tion/jda/core/requests/RestAction.java
```diff
@@ -646,13 +709,14 @@ protected void finalizeRoute() { }
...
         {
-            super(api, null, null);
+            super(api, null);
             this.returnObj = returnObj;
         }
 
         @Override
         public void queue(Consumer<T> success, Consumer<Throwable> failure)
         {
+
```
Almighty-Alpaca: who knows ¯\_(ツ)_/¯

src/main/java/net/dv8tion/jda/core/utils/WidgetUtil.java
Almighty-Alpaca: done

src/main/java/net/dv8tion/jda/core/requests/Response.java
```diff
@@ -16,81 +16,112 @@
...
 
+    @Override
     public String toString()
     {
-        return exception == null ? "HTTPResponse[" + code + ": " + responseText + ']'
-                : "HTTPException[" + exception.getMessage() + ']';
+        return this.exception == null
+                ? "RestActionResponse[" + this.code + (this.object == null ? "" : ", " + this.object.toString()) + ']'
+                : "RestActionException[" + this.exception.getMessage() + ']';
```
Almighty-Alpaca: I'm not sure anymore why I changed this, will revert it in the next commit.

src/main/java/net/dv8tion/jda/core/requests/Route.java
```diff
@@ -212,16 +192,12 @@
...
+        public static final Route REMOVE_ALL_REACTIONS =     new Route(DELETE, "channels/{channel_id}/messages/{message_id}/reactions",                           "channel_id");
+        public static final Route GET_REACTION_USERS =       new Route(GET,    "channels/{channel_id}/messages/{message_id}/reactions/{reaction_code}",           "channel_id");
 
-        public static final Route GET_MESSAGE_HISTORY =        new Route(GET, "channels/{channel_id}/messages?limit={}",           "channel_id");
-        public static final Route GET_MESSAGE_HISTORY_BEFORE = new Route(GET, "channels/{channel_id}/messages?limit={}&before={}", "channel_id");
-        public static final Route GET_MESSAGE_HISTORY_AFTER =  new Route(GET, "channels/{channel_id}/messages?limit={}&after={}",  "channel_id");
-        public static final Route GET_MESSAGE_HISTORY_AROUND = new Route(GET, "channels/{channel_id}/messages?limit={}&around={}", "channel_id");
+        public static final Route GET_MESSAGE_HISTORY =        new Route(GET, "channels/{channel_id}/messages", "channel_id");
```
Almighty-Alpaca: :+1:

src/main/java/net/dv8tion/jda/core/requests/ratelimit/BotRateLimiter.java
```diff
@@ -70,24 +69,31 @@ protected void queueRequest(Request request)
...
             {
-                String global = headers.getFirst("X-RateLimit-Global");
-                String retry = headers.getFirst("Retry-After");
+                String global = headers.get("X-RateLimit-Global");
+                String retry = headers.get("Retry-After");
                 if (retry == null || retry.isEmpty())
                 {
-                    JSONObject limitObj = new JSONObject(response.getBody());
-                    retry = limitObj.get("retry_after").toString();
+                    try (Reader reader = response.body().charStream()) {
```
Almighty-Alpaca: fixed

src/main/java/net/dv8tion/jda/core/requests/ratelimit/ClientRateLimiter.java
```diff
@@ -65,26 +65,34 @@ protected void queueRequest(Request request)
...
+            int code = response.code();
             if (code == 429)
             {
-                JSONObject limitObj = new JSONObject(response.getBody());
-                long retryAfter = limitObj.getLong("retry_after");
-                if (limitObj.has("global") && limitObj.getBoolean("global"))    //Global ratelimit
-                {
-                    globalCooldown = now + retryAfter;
+
+                try (Reader reader = response.body().charStream()) {
```
Almighty-Alpaca: fixed

src/main/java/net/dv8tion/jda/core/requests/Response.java
```diff
@@ -16,81 +16,112 @@
...
+            if (begin == '{')
+                this.object = new JSONObject(new JSONTokener(reader));
+            else if (begin == '[')
+                this.object = new JSONArray(new JSONTokener(reader));
+            else
+                this.object = null;
+        }
+        catch (final Exception e)
+        {
+            throw new RuntimeException("An error occured while parsing a RestAction the response", e);
```
MinnDevelopment: occurred

src/main/java/net/dv8tion/jda/core/requests/Response.java
```diff
@@ -16,81 +16,112 @@
...
-    protected Response(int code, String response, long retryAfter)
+    protected Response(final int code, final String message, final okhttp3.Response response, final long retryAfter)
     {
         this.code = code;
-        this.responseText = response;
+        this.message = message;
         this.exception = null;
         this.retryAfter = retryAfter;
+
+        if (response == null || response.body().contentLength() == 0)
```
MinnDevelopment: What about `response.body() == null`? Is that 100% impossible?

src/main/java/net/dv8tion/jda/core/requests/Response.java
MinnDevelopment: What happens if the response is neither JSONObject or JSONArray, like an xml response at some outage?

Almighty-Alpaca: Right... totally forgot that is possible too. I'll just save it as a String in that case

src/main/java/net/dv8tion/jda/core/requests/Response.java
Almighty-Alpaca: From the OkHttp javadoc

> Returns a non-null value if this response was passed to Callback.onResponse or returned from Call.execute(). Response bodies must be closed and may be consumed only once.

src/main/java/net/dv8tion/jda/core/requests/Requester.java
Almighty-Alpaca: Turned out this doesn't fully work because request and response bodies are represented as Input-/OutputStreams and as such can only be read once. I can still add a trace log for the other request properties though

src/main/java/net/dv8tion/jda/core/requests/Route.java
```diff
@@ -393,12 +369,36 @@ public String toString()
...
+            this(baseRoute, ratelimitRoute, compiledRoute, false);
+        }
+
+        public CompiledRoute withQueryParams(String name, String value)
+        {
+            return new CompiledRoute(baseRoute, ratelimitRoute, compiledRoute + (hasQueryParams ? '&' : '?') + name + '=' + value, true);
+        }
+
+        public CompiledRoute withQueryParams(String... params)
+        {
```
DV8FromTheWorld: What happens if you call it like: `withQueryParams()` ? Looks like it would set `hasQueryParams` to true, and use `&` from then on, which would be wrong.

Additionally, if we have a `String...` version, do we really need a `(String, String)` version? Especially since we're making sure there is a multiple of two params?

src/main/java/net/dv8tion/jda/core/requests/Route.java
```diff
@@ -393,12 +369,36 @@ public String toString()
...
+            Checks.check(params.length % 2 == 0, "param length has to be even");
+
```
DV8FromTheWorld: Change message to 'params length must be a multiple of 2`.
Basically the same, but more clear.

src/main/java/net/dv8tion/jda/core/requests/Route.java
Almighty-Alpaca: Adding a check for the first thing.

I've added the `(String, String)` version because it theoretically is faster (no loop). Micro-optimization :pray:. 
If you want I can also just simply remove it

src/main/java/net/dv8tion/jda/core/requests/Route.java
Almighty-Alpaca: done

src/main/java/net/dv8tion/jda/core/requests/Route.java
DV8FromTheWorld: That's micro-opt to the max. It isn't needed.

Almighty-Alpaca: removed it :smile:

src/main/java/net/dv8tion/jda/core/utils/WidgetUtil.java
```diff
@@ -191,35 +190,63 @@ public static Widget getWidget(String guildId) throws RateLimitedException
...
-            {
-                case 200: return new Widget(result.getBody().getObject()); // ok
-                case 400:              // not valid snowflake
-                case 404: return null; // guild not found
-                case 403: return new Widget(guildId); // widget disabled
-                case 429: // ratelimited
+            connection = (HttpURLConnection) new URL(String.format(WIDGET_URL, guildId)).openConnection();
+            connection.setRequestMethod("GET");
+            connection.addRequestProperty("user-agent", Requester.USER_AGENT);
+            connection.connect();
```
DV8FromTheWorld: Why are we using HttpUrlConnection here instead of OkHttp?
Also, you need to add the `accept-encoding` header here just like in requester.

src/main/java/net/dv8tion/jda/core/entities/Icon.java
```diff
@@ -112,9 +112,15 @@ public static Icon from(InputStream stream) throws IOException
...
-        String encoding = StringUtils.newStringUtf8(Base64.getEncoder().encode(data));
-        return new Icon(encoding);
+        try
+        {
+            return new Icon(new String(Base64.getEncoder().encode(data), "UTF-8"));
+        }
+        catch (UnsupportedEncodingException e)
+        {
+            throw new RuntimeException(e); // thanks JDK 1.4
+        }
```
MinnDevelopment: `org.apache.commons.codec.binary` is not available anymore since we removed Unirest and its dependencies.
This is essentially the same code though, just inlined.

src/main/java/net/dv8tion/jda/core/requests/ratelimit/BotRateLimiter.java
```diff
@@ -70,24 +66,31 @@ protected void queueRequest(Request request)
...
+            int code = response.code();
             if (timeOffset == null)
                 setTimeOffset(headers);
 
             if (code == 429)
             {
-                String global = headers.getFirst("X-RateLimit-Global");
-                String retry = headers.getFirst("Retry-After");
+                String global = headers.get("X-RateLimit-Global");
+                String retry = headers.get("Retry-After");
```
MinnDevelopment:
```java
  private static String get(String[] namesAndValues, String name) {
    for (int i = namesAndValues.length - 2; i >= 0; i -= 2) {
      if (name.equalsIgnoreCase(namesAndValues[i])) {
        return namesAndValues[i + 1];
      }
    }
    return null;
  }
```
this is the code which searches for the header.

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
```diff
@@ -545,30 +549,40 @@ protected void handleResponse(Response response, Request<Message> request)
...
         Route.CompiledRoute route = Route.Messages.SEND_MESSAGE.compile(getId());
-        MultipartBody body = Unirest.post(Requester.DISCORD_API_PREFIX + route.getCompiledRoute())
-                .fields(null); //We use this to change from an HttpRequest to a MultipartBody
-
-        body.field("file", data, fileName);
+        MultipartBody.Builder builder = new okhttp3.MultipartBody.Builder()
+                .setType(MultipartBody.FORM);
+        try
+        {
+            byte[] bytes = IOUtil.readFully(data);
```
MinnDevelopment: https://github.com/DV8FromTheWorld/JDA/blob/experimental/message-rw/src/main/java/net/dv8tion/jda/core/entities/Message.java#L977-L1006 this can easily be made public in `experimental/message-rw` later on.

src/main/java/net/dv8tion/jda/core/entities/Message.java
```diff
@@ -789,19 +787,9 @@ public boolean download(File file)
...
-                else
-                {
-                    con = url.openConnection(new Proxy(Proxy.Type.HTTP,
-                            new InetSocketAddress(jda.getGlobalProxy().getAddress(), jda.getGlobalProxy().getPort())));
-                }
-                con.addRequestProperty("user-agent", Requester.USER_AGENT);
-                in = con.getInputStream();
+                Request request = new Request.Builder().addHeader("user-agent", Requester.USER_AGENT).url(getUrl()).build();
+                Response response = jda.getRequester().getHttpClient().newCall(request).execute();
+                in = response.body().byteStream();
```
MinnDevelopment: https://github.com/DV8FromTheWorld/JDA/blob/experimental/message-rw/src/main/java/net/dv8tion/jda/core/entities/Message.java#L977-L1006 this can easily be made public in `experimental/message-rw` later on.

src/main/java/net/dv8tion/jda/core/requests/ratelimit/ClientRateLimiter.java
```diff
@@ -55,8 +52,6 @@ public Long getRateLimit(Route.CompiledRoute route)
     @Override
     protected void queueRequest(Request request)
     {
-        if (isShutdown)
-            throw new RejectedExecutionException("Cannot queue a request after shutdown");
```
MinnDevelopment: I was able to apply everything except for this, @Almighty-Alpaca should comment his intentions and apply changes as necessary.

src/main/java/net/dv8tion/jda/core/requests/RestAction.java
```diff
@@ -615,18 +655,27 @@ public T completeAfter(long delay, TimeUnit unit)
...
-        Args.notNull(unit, "TimeUnit");
+        Checks.notNull(executor, "Scheduler");
+        Checks.notNull(unit, "TimeUnit");
         return executor.schedule(() -> queue(success, failure), delay, unit);
     }
 
-    protected void finalizeData() { }
+    protected RequestBody finalizeData() { return data; }
+    protected Route.CompiledRoute finalizeRoute() { return route; }
+    protected CaseInsensitiveMap<String, String> finalizeHeaders() { return headers == null ? null : new CaseInsensitiveMap<>(headers); } // maps are mutable
```
MinnDevelopment: I reverted this to only have the method which returns a map of headers, the field addition as of now is not needed and a pointless change. In the case of this being needed we can adjust this as fits.

src/main/java/net/dv8tion/jda/core/requests/Requester.java
```diff
@@ -159,70 +180,41 @@ public JDAImpl getJDA()
...
             if (!rays.isEmpty())
                 LOG.debug("Received response with following cf-rays: " + rays);
+
             if (retryAfter == null)
-                apiRequest.getRestAction().handleResponse(new Response(response.getStatus(), response.getBody(), -1), apiRequest);
+                apiRequest.handleResponse(new Response(response, -1, rays));
+            else if (handleOnRatelimit)
+                apiRequest.handleResponse(new Response(response, retryAfter, rays));
+            else
+                response.close();
```
MinnDevelopment: I have adjusted this to close all responses in an additional `finalize` block. For that I created a new `okhttp3.Response[4]` which holds all possible responses then closes them before exiting the method. Once one of the elements is `null` the iteration of the array is intercepted and code-flow continues.

src/main/java/net/dv8tion/jda/core/requests/Requester.java
```diff
@@ -92,65 +98,80 @@ public JDAImpl getJDA()
...
+        String method = apiRequest.getRoute().getMethod().toString();
+        RequestBody body = apiRequest.getBody();
+
+        if (body == null && HttpMethod.requiresRequestBody(method))
+            body = EMPTY_BODY;
+
+        builder.method(method, body);
+
+        builder.header("user-agent", USER_AGENT);
+
```
MinnDevelopment: This required a minor change in the code to wrap the response body in a `GZIPInputStream` when needed.

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
MinnDevelopment: I have not found a different way of doing this in the okhttp documentation/codebase. If someone can suggest a better approach I am open for input.

src/main/java/net/dv8tion/jda/core/entities/Message.java
DV8FromTheWorld: :+1:

src/main/java/net/dv8tion/jda/core/requests/ratelimit/ClientRateLimiter.java
Almighty-Alpaca: I removed it because `Requester#request()` already checks for it and it's unnecessary to do it twice:
```java
public <T> void request(Request<T> apiRequest)
{
    if (rateLimiter.isShutdown) 
        throw new IllegalStateException("The Requester has been shutdown! No new requests can be requested!");
    if (apiRequest.shouldQueue())
        rateLimiter.queueRequest(apiRequest);
    else
        execute(apiRequest, true);
}
```
This is the only place `RateLimiter#queueRequest()` is called from.

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
Almighty-Alpaca: I found a way, but we'll have to live with the fact that we can't check the size anymore.

MinnDevelopment:
If nothing major happens, this can be squashed within the upcoming days.

MinnDevelopment:
@Almighty-Alpaca whenever you're ready, this can be squashed.

Issue #344, opened 2017-05-04
Audit Logs


clevertoucan:
Would love to see an update that allows JDA to retrieve audit logs from a guild

Almighty-Alpaca:
audit logs are in progress, see [`feature/audit-logs`](https://github.com/DV8FromTheWorld/JDA/tree/feature/audit-logs)

MinnDevelopment:
We have gotten fairly far into audit logs development in JDA.
Here is a small sneak-peak on what you can expect to be able to do
```java
guild.getController().setNickname(guild.getSelfMember(), "Awesome Bot").reason("Wasn't aweseome 'nough").queue()
guild.getController().kick("86699011792191488").reason("Not awesome enough").queue()
```
> Here we add reasons to the `AuditableRestAction`s which are returned by most GuildController methods
```java
AuditLogPaginationAction logs = guild.getAuditLogs();
for (AuditLogEntry entry : logs)
{
    System.out.println("Type: " + entry.getActionType() + "\nReason: " + entry.getReason());
}
```
Output:
```
Type: MEMBER_UPDATE
Reason: Wasn't aweseome 'nough
Type: KICK
Reason: Not awesome enough

MinnDevelopment:
This has been introduced in commit e909ca9

PR #351, opened 2017-05-12
Audit Logs and Reasons

MinnDevelopment:
This added:
-  `AuditableRestAction` / `AuditableRestAction.EmptyRestAction`
    Allowing to add a reason to any auditable request via `reason(String)`
- New reason routes for kick/ban to be able to use in case client accounts may not be able to use the `X-Audit-Log-Reason` header
- `AuditLogEntry` / `AuditLogPaginationAction`
    Allowing to view and iterate over all audit log entries via pagination
- `AuditLogChange`
    POJO containing changes for a single field/key 
- `ActionType` / `TargetType`
   Identifies what kind of action an `AuditLogEntry` represents

This directly addresses #344


src/main/java/net/dv8tion/jda/core/entities/impl/TextChannelImpl.java
```diff
@@ -132,15 +94,54 @@ protected void handleResponse(Response response, Request<List<Webhook>> request)
...
+                if (response.isOk())
+                    request.onSuccess(null);
+                else
+                    request.onFailure(response);
+            }
+        };
+    }
+
+    @Override
+    public AuditableRestAction<Void> deleteWebhookById(String id)
```
DV8FromTheWorld: Why did all of this get shifted? O.o

src/main/java/net/dv8tion/jda/core/managers/GuildController.java
```diff
@@ -377,19 +387,25 @@ protected void handleResponse(Response response, Request<Integer> request)
      *             <br>See {@link net.dv8tion.jda.core.utils.PermissionUtil#canInteract(Member, Member) PermissionUtil.canInteract(Member, Member)}</li>
      *         </ul>
      *
-     * @return {@link net.dv8tion.jda.core.requests.RestAction RestAction}
+     * @return {@link net.dv8tion.jda.core.requests.restaction.AuditableRestAction AuditableRestAction}
      *         Kicks the provided Member from the current Guild
      */
-    public RestAction<Void> kick(Member member)
+    public AuditableRestAction<Void> kick(Member member, String reason)
```
DV8FromTheWorld: Is this reason different from the AuditLog reason?

MinnDevelopment: Client accounts cannot use the `X-Audit-Log-Reason` header that is why I added kick/ban with the query parameter.

MinnDevelopment:
I believe this is now ready to get merged.

PR #356, opened 2017-05-15
Reorder stripping of mentions

ScoreUnder:
You can `/nick everyone`, and the default `stripMentions` will "strip" a mention of yourself to become an unfiltered `@everyone`.

This PR is to fix that. I've locally tested that it works, and written a note to future editors of those functions so that people think twice before rewriting it.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/MessageBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/MessageBuilder.java
@@ -372,7 +372,9 @@ public class MessageBuilder implements Appendable
      */
     public MessageBuilder stripMentions(JDA jda)
     {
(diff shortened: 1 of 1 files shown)


ScoreUnder:
Oh for crying out loud...


Shall I save this one for another commit or fix it here?
(On a related note: If this doesn't end up with the mention stripping code being rewritten entirely, should the mention order be forced to something sensible just in case the user specifies their own in the "wrong" order?)

ScoreUnder:
If you want a real code review challenge:

https://github.com/DV8FromTheWorld/JDA/compare/master...ScoreUnder:mentions_fix_two_electric_boogaloo

This attempts to fix the problem highlighted in the screenshot above, as well as the `@everyone` problem.

I'm not expecting that branch to "make it" as-is, on the grounds of how complex it is, but if you could point me to a better way while still correctly handling every case, that would also be great.

DV8FromTheWorld:
@MinnDevelopment @Almighty-Alpaca Lets deal with this PR.

PR #363, opened 2017-05-23
Message update

Sanduhr32:
Since Member#getAsMention() exists i thought about adding Message#isMentioned(Member) / Message#getMentionedMembers() and now i realized my plans and so i hope its good enough (docs and code) and helps some users. Your Sanduhr <3


MinnDevelopment:
This is something we want to address in the message rewrite specifically.

Sanduhr32:
Ah ok Minn hopefully soon™ <3

MinnDevelopment:
This is now tracked on [`experimental/message-rw`](https://github.com/DV8FromTheWorld/JDA/tree/experimental/message-rw)

PR #373, opened 2017-06-02
Update of audit log reasons / allow utf-8

MinnDevelopment:
Note that I made TextChannel#deleteMessages only return RestAction according to hammerandchisel/discord-api-docs#265
And I did the same for TextChannel#clearReactionsById due to there not being an action type I suspect there not to be one in the near future


MinnDevelopment:
This needs to be tested once the changes are deployed

MinnDevelopment:
This has now been deployed and works. The `+` which is generated for spaces within `MiscUtil.encodeUTF8(String)` has to be replaced as discord does not decode it back to a space instead we just send space itself.

PR #381, opened 2017-06-18
Fix cleaning up of the keepAliveRunnable

### LINKED ISSUE #380: Memory leak of audio connections
My bots appear to be suffering from memory leaks. I took a heap dump of my patron bot, which has been running for 30 days with JDA `3.0.0_200`. Looking at the heap dump I see that `java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue` is retaining 53% of my heap.

**Clarification:** The 30 days is with a relatively small bot. For my bot with over 1400x as many guilds this problem is about 10x-15x faster to take up the heap to the point of impacting audio playback.

image
The problem appears to be that we have numerous `AudioConnection`s which still holds references to my send handler, which in my case is called `GuildPlayer`:

image

I'm not sure why these connections are being retained, but it doesn't look like these connections are getting references from my code without going through JDA code. 

This could possibly be related to https://github.com/DV8FromTheWorld/JDA/issues/357

I use JDA-NAS

Shredder121:
Fixes #380 

The issue of the rogue memory leak is this line:

```java
keepAlivePool.remove(keepAliveRunnable);
```

Because the specific executor service adapts the `Runnable` to a different class, it's never contained in the internal queue datastructure, (as in `Collection.indexOf()`).

Lot of sad times, since it's never removed.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
+++ b/src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
@@ -337,7 +337,7 @@ public class AudioWebSocket extends WebSocketAdapter
         }
         if (keepAliveHandle != null)
         {
-            keepAliveHandle.cancel(true);
+            keepAliveHandle.cancel(false);
             keepAliveHandle = null;
         }
 
--- a/src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
+++ b/src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
@@ -35,10 +35,7 @@ import java.nio.ByteBuffer;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.RejectedExecutionException;
-import java.util.concurrent.ScheduledThreadPoolExecutor;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
 
 public class AudioWebSocket extends WebSocketAdapter
@@ -65,7 +62,7 @@ public class AudioWebSocket extends WebSocketAdapter
     private boolean connected = false;
     private boolean ready = false;
     private boolean shutdown = false;
-    private Runnable keepAliveRunnable;
+    private Future<?> keepAliveHandle;
     private String wssEndpoint;
     private boolean shouldReconnect;
 
@@ -338,10 +335,10 @@ public class AudioWebSocket extends WebSocketAdapter
                 );
             api.getClient().send(obj.toString());
         }
-        if (keepAliveRunnable != null)
+        if (keepAliveHandle != null)
         {
-            keepAlivePool.remove(keepAliveRunnable);
-            keepAliveRunnable = null;
+            keepAliveHandle.cancel(true);
+            keepAliveHandle = null;
         }
 
         if (audioConnection != null)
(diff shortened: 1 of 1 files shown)


Shredder121:
We also have to decide, whether to `.cancel(true)` or `.cancel(false)`, so whether to interrupt the running thread if it's running (it's just a keep alive, but up to you)

src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
```diff
@@ -521,7 +517,7 @@ private void setupKeepAlive(final int keepAliveInterval)
 
         try
         {
-            keepAlivePool.scheduleAtFixedRate(keepAliveRunnable, 0, keepAliveInterval, TimeUnit.MILLISECONDS);
+            keepAliveHandle = keepAlivePool.scheduleAtFixedRate(keepAliveRunnable, 0, keepAliveInterval, TimeUnit.MILLISECONDS);
```
Shredder121: Hmmmm just realized this actually overwrites when it's called a second time.

How do you want me to fix it instead?

MinnDevelopment: setupKeepAlive should only be called once

Shredder121: Hmm okay, but it doesn't recover still, and we'll have a dangling runnable, which was the problem in the first place.
Just letting you know.
But if you think it's fine disregard my comment.

src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
```diff
@@ -338,10 +335,9 @@ public void close(ConnectionStatus closeStatus)
                 );
             api.getClient().send(obj.toString());
         }
-        if (keepAliveRunnable != null)
+        if (keepAliveHandle != null)
         {
-            keepAlivePool.remove(keepAliveRunnable);
-            keepAliveRunnable = null;
+            keepAliveHandle.cancel(true);
```
Shredder121: Note to self, set it to `null`.

Frederikam:
Did some testing with JProfiler with this commit. Seems like the problem has been solved 👍

Running in production now.

Shredder121:
maybe we can pull this particular commit off of the other commit in the original `patch-speed` and apply it.
So if you @DV8FromTheWorld want to include it, we can do that instead.
That, or just cherry pick it.
But let me know first what you think, or if something needs to change first. 👍

src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
DV8FromTheWorld: like he said, setupKeepAlive doesn't fire twice, and if it did we would have a lovely FATAL log message to tell us it happened. Should be fine

Shredder121:
>We also have to decide, whether to .cancel(true) or .cancel(false), so whether to interrupt the running thread if it's running (it's just a keep alive, but up to you)

Just pointing out this wasn't addressed yet? Or is the code as it is fine?

MinnDevelopment:
This seems to work well. 
I will let some time pass so @Frederikam can retrieve some data on this change.

I would like to use `.cancel(false)` to avoid possible unneeded exceptions that cause console clutter. It is not a problem if it finished sending the keepAlive. Though I doubt it makes a difference in this case.

I will squash this PR into `patch-speedup` and proceed with merging `patch-speedup` into master once some time has passed.

Shredder121:
Good to hear about that point, I updated that.
I can't amend and force push though, since otherwise possibly the [build of FredBoat will stop working](https://github.com/Frederikam/FredBoat/blob/92ef431e97d8e71f89b85671a38fa8f7446e52f7/FredBoat/pom.xml#L40-L44).
But seeing as you're eventually going to squash the commits, I don't think that's going to be an issue.

MinnDevelopment:
The build fred is using will continue to work unless you force delete the build from https://jitpack.io as jitpack builds artifacts on pulling by reference but doesn't delete them when the commit disappears.

Shredder121:
Good to know that, neat.

PR #382, opened 2017-06-21
Do not remove SelfUser from user cache

MinnDevelopment:


src/main/java/net/dv8tion/jda/core/handle/MessageUpdateHandler.java
```diff
@@ -152,7 +151,7 @@ private Long handleDefaultMessage(JSONObject content)
 
     private Long handleMessageEmbed(JSONObject content)
     {
-        EntityBuilder builder = api.getEntityBuilder();;
```
kantenkugel: my favorite

DV8FromTheWorld: 👍

PR #386, opened 2017-07-04
Improve event cache handling

MinnDevelopment:
Reference for changes to CHANNEL_CREATE handling: hammerandchisel/discord-api-docs#184

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/handle/ChannelCreateHandler.java
+++ b/src/main/java/net/dv8tion/jda/core/handle/ChannelCreateHandler.java
@@ -49,33 +49,41 @@ public class ChannelCreateHandler extends SocketHandler
             case TEXT:
             {
                 api.getEventManager().handle(
-                        new TextChannelCreateEvent(
-                                api, responseNumber,
-                                api.getEntityBuilder().createTextChannel(content, guildId)));
+                    new TextChannelCreateEvent(
+                        api, responseNumber,
+                        api.getEntityBuilder().createTextChannel(content, guildId)));
                 break;
             }
             case VOICE:
             {
                 api.getEventManager().handle(
-                        new VoiceChannelCreateEvent(
-                                api, responseNumber,
-                                api.getEntityBuilder().createVoiceChannel(content, guildId)));
+                    new VoiceChannelCreateEvent(
+                        api, responseNumber,
+                        api.getEntityBuilder().createVoiceChannel(content, guildId)));
                 break;
             }
             case PRIVATE:
             {
+                // When a bot receives a message from a PrivateChannel
+                //the channel will first be sent using a CHANNEL_CREATE due to changes made on June 4th 2017
+                //see: https://github.com/hammerandchisel/discord-api-docs/issues/184#issuecomment-312960055
+                //in this case we now add a check for existing private channels to stop from re-creating channels over and over
+                //when the channel already exists we don't have to deal with event cache replay due to it already existing we know
+                //that no events can be cached here in the first place
+                if (api.getPrivateChannelById(content.getLong("id")) != null)
+                    return null;
                 api.getEventManager().handle(
-                        new PrivateChannelCreateEvent(
-                                api, responseNumber,
-                                api.getEntityBuilder().createPrivateChannel(content)));
+                    new PrivateChannelCreateEvent(
+                        api, responseNumber,
+                        api.getEntityBuilder().createPrivateChannel(content)));
                 break;
             }
             case GROUP:
             {
                 api.getEventManager().handle(
-                        new GroupJoinEvent(
-                                api, responseNumber,
-                                api.getEntityBuilder().createGroup(content)));
+                    new GroupJoinEvent(
+                        api, responseNumber,
+                        api.getEntityBuilder().createGroup(content)));
                 break;
             }
             default:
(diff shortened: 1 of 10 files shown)


src/main/java/net/dv8tion/jda/core/handle/ChannelCreateHandler.java
```diff
@@ -49,33 +49,41 @@ protected Long handleInternally(JSONObject content)
...
             }
             case GROUP:
             {
                 api.getEventManager().handle(
-                        new GroupJoinEvent(
-                                api, responseNumber,
-                                api.getEntityBuilder().createGroup(content)));
+                    new GroupJoinEvent(
+                        api, responseNumber,
+                        api.getEntityBuilder().createGroup(content)));
```
DV8FromTheWorld: Groups might be considered PrivateChannels in this case. Please doublecheck this.

src/main/java/net/dv8tion/jda/core/handle/EventCache.java
```diff
@@ -77,6 +77,19 @@ public int size()
...
+    {
+        TLongObjectMap<List<Runnable>> cache = eventCache.get(type);
+        if (cache == null)
+            return;
+        List<Runnable> events = cache.get(triggerId);
+        if (events == null)
+            return;
+        LOG.debug("Clearing cache for type " + type + " with ID " + triggerId + " (Size: " + events.size() + ')');
+        events.clear();
+        cache.remove(triggerId);
```
DV8FromTheWorld: change `cache.get` to `cache.remove`. Remove has the item removed returned. This will also prevent possible raceconditions.

src/main/java/net/dv8tion/jda/core/handle/GuildDeleteHandler.java
```diff
@@ -144,6 +144,7 @@ else if (api.getAccountType() == AccountType.CLIENT)
                 new GuildLeaveEvent(
                         api, responseNumber,
                         guild));
+        api.getEventCache().clear(guild.getIdLong(), EventCache.Type.GUILD);
```
DV8FromTheWorld: Clear cache for all channels  and roles in this guild too?

src/main/java/net/dv8tion/jda/core/handle/MessageDeleteHandler.java
```diff
@@ -58,8 +58,9 @@ protected Long handleInternally(JSONObject content)
         }
         if (channel == null)
         {
-            api.getEventCache().cache(EventCache.Type.CHANNEL, channelId, () -> handle(responseNumber, allContent));
-            EventCache.LOG.debug("Got message delete for a channel/group that is not yet cached. ChannelId: " + channelId);
+//            api.getEventCache().cache(EventCache.Type.CHANNEL, channelId, () -> handle(responseNumber, allContent));
+//            EventCache.LOG.debug("Got message delete for a channel/group that is not yet cached. ChannelId: " + channelId);
+            // Don't cache message events for channels unless they are MESSAGE_CREATE
```
DV8FromTheWorld: Added reference to the specific comment in the Github issue as to *why* we don't cache anymore

src/main/java/net/dv8tion/jda/core/handle/MessageUpdateHandler.java
```diff
@@ -88,9 +88,12 @@ private Long handleDefaultMessage(JSONObject content)
...
                 {
-                    final long channelId = content.getLong("channel_id");
-                    api.getEventCache().cache(EventCache.Type.CHANNEL, channelId, () -> handle(responseNumber, allContent));
-                    EventCache.LOG.debug("Received a message update for a channel that JDA does not currently have cached");
+//                    final long channelId = content.getLong("channel_id");
+//                    api.getEventCache().cache(EventCache.Type.CHANNEL, channelId, () -> handle(responseNumber, allContent));
+//                    EventCache.LOG.debug("Received a message update for a channel that JDA does not currently have cached");
+                    // When we receive message events from a channel that is not in our cache we are expected to discard them
+                    //the only events that should get cached are MESSAGE_CREATE because in the case of private channels
+                    //we can expect a CHANNEL_CREATE to be fired eventually (see Eventually Consistent)
```
DV8FromTheWorld: This is actually an invalid change. We can still get MESSAGE_CREATE and CHANNEL_CREATE out of order.
Additionally, this will effect more than just PrivateMessages.

src/main/java/net/dv8tion/jda/core/handle/ChannelCreateHandler.java
MinnDevelopment: Why is this relevant for this change? Groups shouldn't be available to bots in the first place

MinnDevelopment:
I will open a new PR for these changes together with the `experimental/disposable-cache` changes.

src/main/java/net/dv8tion/jda/core/handle/ChannelCreateHandler.java
DV8FromTheWorld: it isn't relative for this specific line, but there was nothing I could tag it on.
Forgot that the change was only for bots.

MinnDevelopment:
This has been replaced by #393

PR #391, opened 2017-07-12
Release for JDA 3.2 / finalization

MinnDevelopment:
Cleanup for some documentation and removal of deprecated methods.
JDABuilder will no longer throw an exception for `useSharding(0, 1)` as it is a valid configuration. (tested)
The exceptions thrown by JDABuilder are now documented and no longer simple RuntimeExceptions!

- [x] Compiles
- [x] No javadoc errors
- [ ] Version bump (will be final release commit)

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/JDABuilder.java
@@ -20,6 +20,7 @@ import net.dv8tion.jda.core.JDA.Status;
 import net.dv8tion.jda.core.audio.factory.IAudioSendFactory;
 import net.dv8tion.jda.core.entities.Game;
 import net.dv8tion.jda.core.entities.impl.JDAImpl;
+import net.dv8tion.jda.core.exceptions.AccountTypeException;
 import net.dv8tion.jda.core.exceptions.RateLimitedException;
 import net.dv8tion.jda.core.hooks.IEventManager;
 import net.dv8tion.jda.core.managers.impl.PresenceImpl;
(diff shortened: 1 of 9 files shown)


src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -448,23 +449,29 @@ public JDABuilder setMaxReconnectDelay(int maxReconnectDelay)
...
      * @see    net.dv8tion.jda.core.JDA#getShardInfo() JDA.getShardInfo()
      */
     public JDABuilder useSharding(int shardId, int shardTotal)
     {
-        if (shardId < 0 || shardTotal < 2 || shardId >= shardTotal)
-        {
-            throw new RuntimeException("This configuration of shardId and shardTotal is not allowed! 0 <= shardId < shardTotal with shardTotal > 1");
-        }
+        if (accountType != AccountType.BOT)
+            throw new AccountTypeException(AccountType.BOT);
```
DV8FromTheWorld: Is it actually not possible to shard a User account?

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -448,23 +449,29 @@ public JDABuilder setMaxReconnectDelay(int maxReconnectDelay)
      * <br>The shardId that receives all stuff related to given bot is calculated as follows: shardId == (guildId {@literal >>} 22) % shardTotal;
      * <br><b>PMs are only sent to shard 0.</b>
      *
-     * <p>Please note, that a shard will not even know about guilds not assigned to.
+     * <p>Please note, that a shard will not even know about guilds which are not assigned to it.
```
DV8FromTheWorld: On this line, `even` isn't needed.

src/main/java/net/dv8tion/jda/core/JDABuilder.java
MinnDevelopment: It does nothing on AccountType.CLIENT and I don't see a reason to allow it then. It can only cause confusion for users.

PR #393, opened 2017-07-14
Cascading Cache Cleanup (CCC)

MinnDevelopment:
- Introduced new Disposable and DisposableState interfaces to indicate entities which can be untracked by JDA cache
- Added new checks for disposed entities to prevent unexpected errors on cache accessing
- Added convenience methods to check whether an entity is tracked or not
- Now cleaning up event cache when an entity has been disposed
- Properly handling events for the CHANNEL_CREATE for cached PrivateChannels
- No longer handling MESSAGE_UPDATE for uncached PrivateChannels
- Regions are now stored as raw strings and converted to enum constants on read through `getRegion`

When JDA has fired the `ShutdownEvent` it will now properly invalidate cache once the event has been handled.


MinnDevelopment:
This is the first stage of the cache improvements that I am planning.
After this I want to finish up `feature/cache-view` and possibly something else later for improvements with role adding to members.

MinnDevelopment:
These changes should be ready for testing. 
I am overall done with my planned changes for this PR.

Reviews and testing required for squash.

MinnDevelopment:
This PR needs to wait for #410 to be completed in order to properly handle invalidation with audio managers!

MinnDevelopment:
This will be redone at a later point. Too much has changed and I want to reconsider some of the changes I made.

PR #400, opened 2017-07-20
Add JSR-305 annotations

Shredder121:
These annotations are picked up by all modern IDEs and prevent the forgetting of `.queue()` and friends.

I see the patch has deviated a bit from current master, I'll rebase but first want to know what you think.

should check method return value


MinnDevelopment:
I am 100% supporting this if it has no downside for other users. The amount of questions regarding RestAction operations that can be resolved this this is a big win for us.

Shredder121:
To my knowledge, this doesn't have any downside.
If at runtime the class file for an annotation type is not present, it's regarded as an unknown annotation.
So even if people manually download JDA, lavaplayer, trove4j, etc and forget these annotations [they should be dropped](https://stackoverflow.com/a/3568041/3485069).

DV8FromTheWorld:
I want to load this up in my IDE to see how it behaves.

Side note: You can apparently apply these annotations at class level to incorporate all methods in a class and specifically makes some as unneeded.

http://errorprone.info/bugpattern/CheckReturnValue

Not sure if that is better than marking every method individually. Thoughts?

Shredder121:
> Side note: You can apparently apply these annotations at class level to incorporate all methods in a class and specifically makes some as unneeded.

That's if we are going to use errorprone annotations.
I'm not sure if that is standard though.

Avarel:
Also, how about `@NotNull` and `@Nullable` on applicable methods?

GrandPanda:
That would be a massive task do it properly.

GrandPanda:
I feel it should be noted that the findbugs project is dead. Development has been continued under [spotbugs](https://github.com/spotbugs/spotbugs).

MinnDevelopment:
@Avarel that might be something we can do in a future version, this PR should only focus on one thing for now.

PR #403, opened 2017-07-22
Added Invite#getURL()

MinnDevelopment:
This needs to wait for #400 due to possible conflicts.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/Invite.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/Invite.java
@@ -103,13 +103,13 @@ public interface Invite
 
     /**
      * The invite URL for this invite in the format of:
-     * {@code "discord.gg/" + getCode()}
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/core/entities/Invite.java
```diff
@@ -102,6 +102,17 @@
...
 
     /**
+     * The invite URL for this invite in the format of:
+     * {@code "discord.gg/" + getCode()}
+     *
+     * @return Invite URL for this Invite
+     */
+    default String getURL()
+    {
+        return "discord.gg/" + getCode();
```
Almighty-Alpaca: valid URLs need to have a protocol

MinnDevelopment: I suppose that makes sense, was thinking for generally posting in channels but you're right.

PR #404, opened 2017-07-22
Implemented NSFW flag

Almighty-Alpaca:
now that it's fixed...

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/impl/TextChannelImpl.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/impl/TextChannelImpl.java
@@ -427,7 +427,7 @@ public class TextChannelImpl extends AbstractChannelImpl<TextChannelImpl> implem
         return this;
     }
 
-    public TextChannelImpl setNsfw(boolean nsfw)
+    public TextChannelImpl setNSFW(boolean nsfw)
     {
         this.nsfw = nsfw;
         return this;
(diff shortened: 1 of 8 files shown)


src/main/java/net/dv8tion/jda/core/managers/ChannelManagerUpdatable.java
```diff
@@ -173,6 +174,28 @@ public Guild getGuild()
...
+     * on the returned {@link net.dv8tion.jda.core.managers.fields.ChannelField ChannelField} instance.
+     *
+     * <p><b>This is only available to {@link net.dv8tion.jda.core.entities.TextChannel TextChannels}</b>
+     *
+     * @throws UnsupportedOperationException
+     *         If the selected {@link net.dv8tion.jda.core.entities.Channel Channel}'s type is not {@link net.dv8tion.jda.core.entities.ChannelType#TEXT TEXT}
+     *
+     * @return {@link net.dv8tion.jda.core.managers.fields.ChannelField ChannelField} - Type: {@code String}
+     */
+    public ChannelField<Boolean> getNSFWField()
```
MinnDevelopment: `Type: {@code boolean}`

src/main/java/net/dv8tion/jda/core/entities/impl/TextChannelImpl.java
```diff
@@ -426,6 +427,12 @@ public TextChannelImpl setLastMessageId(long id)
         return this;
     }
 
+    public TextChannelImpl setNsfw(boolean nsfw)
```
MinnDevelopment: for consistency this should be `setNSFW`

PR #406, opened 2017-07-23
Fix User#getDiscrimator Javadoc

meowingtwurtle:


DV8FromTheWorld:
I would appreciate it if you would also update the line above about friends list.
Since they have been released, this can be removed.

meowingtwurtle:
The commit has been updated

PR #409, opened 2017-07-26
Remove lang3 dependency

MinnDevelopment:
This should sort lots of compatibility issues with frameworks such as bukkit or spigot.

I'm open for suggestions to name the `Helpers` class.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/utils/Helpers.java
+++ b/src/main/java/net/dv8tion/jda/core/utils/Helpers.java
@@ -82,7 +82,7 @@ public final class Helpers
         if (input == null)
             return null;
         Checks.notNegative(maxWidth, "maxWidth");
-        if (input.length() == maxWidth)
+        if (input.length() <= maxWidth)
             return input;
         if (maxWidth == 0)
             return "";
--- a/src/main/java/net/dv8tion/jda/core/utils/Helpers.java
+++ b/src/main/java/net/dv8tion/jda/core/utils/Helpers.java
@@ -1,0 +1,123 @@
+/*
+ *     Copyright 2015-2017 Austin Keener & Michael Ritter & Florian Spieß
+ *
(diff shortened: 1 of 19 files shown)


src/main/java/net/dv8tion/jda/core/utils/Helpers.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+        return count;
+    }
+
+    public static String truncate(final String input, final int maxWidth)
+    {
+        if (input == null)
+            return null;
+        Checks.notNegative(maxWidth, "maxWidth");
+        if (input.length() == maxWidth)
```
kantenkugel: would <= be better?

MinnDevelopment: Thank you for pointing that out, apparently I haven't tested this with too long width.

src/main/java/net/dv8tion/jda/core/utils/tuple/Pair.java
```diff
@@ -0,0 +1,117 @@
...
+ * <p>Subclass implementations may be mutable or immutable.
+ * However, there is no restriction on the type of the stored objects that may be stored.
+ * If mutable objects are stored in the pair, then the pair itself effectively becomes mutable.</p>
+ *
+ * @param <L> the left element type
+ * @param <R> the right element type
+ *
+ * @since Lang 3.0
+ */
+public abstract class Pair<L, R> implements /*Map.Entry<L, R>, Comparable<Pair<L, R>>,*/ Serializable {
```
DV8FromTheWorld: Why is there a commented Map.Entry here? O.o

MinnDevelopment: I wanted to keep it as close to the original as possible. This is just commented out to show that it had this but we stripped it off since we don't need it.

PR #410, opened 2017-07-27
Implement Voice Resume

MinnDevelopment:
This should make voice connection issues seem way less disrupting. A proper voice resume is almost unnoticeable.


src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
```diff
@@ -238,15 +236,29 @@ public void onDisconnected(WebSocket websocket, WebSocketFrame serverCloseFrame,
         {
             LOG.debug("Reason: " + serverCloseFrame.getCloseReason());
             LOG.debug("Close code: " + serverCloseFrame.getCloseCode());
+            int code = serverCloseFrame.getCloseCode();
+            VoiceCode.CloseCode closeCode = VoiceCode.CloseCode.from(code);
+            switch (closeCode) {
```
MinnDevelopment: put opening bracket on newline

MinnDevelopment:
It seems that the latest change is causing heartbeat timeouts after resuming.
When resuming we receive HELLO which creates a new keep alive task.

src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
```diff
@@ -157,35 +162,46 @@ public void onTextMessage(WebSocket websocket, String message)
...
+                LOG.trace("-> HEARTBEAT " + contentAll);
+                send(VoiceCode.HEARTBEAT, heartbeatSent = System.currentTimeMillis());
                 break;
             }
-            case HEARTBEAT_PING_RETURN:
+            case VoiceCode.HEARTBEAT_ACK:
             {
-                long timePingSent  = contentAll.getLong("d");
-                long ping = System.currentTimeMillis() - timePingSent;
+                LOG.trace("-> HEARTBEAT_ACK " + contentAll);
```
DV8FromTheWorld: This did not need to be changed. Please revert.

src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
```diff
@@ -157,35 +162,46 @@ public void onTextMessage(WebSocket websocket, String message)
...
+                final long ping = System.currentTimeMillis() - heartbeatSent;
```
DV8FromTheWorld: This did not need to be change. Please revert.

MinnDevelopment:
https://github.com/hammerandchisel/discord-api-docs/pull/323#issuecomment-319464672 should handle USER_DISCONNECT as well, can be implemented again then

MinnDevelopment:
As soon as discord updates the HELLO interval we need to remove all heartbeat_interval handling in READY handling

Sanduhr32:
Uh my god thats not a big deal :^)

Kodehawa:
Testing this branch for over a week, I've ran into the following issue(s) which have been confirmed to not happen in versions without this patch:

1.- Shards deadlock on SHUTTING_DOWN, adjunting an image of the bottom of the dump and the dump itself. {[Thread dump file](https://github.com/DV8FromTheWorld/JDA/files/1205853/deadlock.zip)}
Bottom of the file: https://im-ded-but.mantaro-is.life/34P6626-581.jpg
2.- After the shard deadlocks on SHUTTING_DOWN, the shard watcher sees the shard won't respond to anything and it basically kill -9's it internally, which somehow causes a leak of ~2GB of memory, which I'll attempt to analyze after I get out of uni.

Shredder121:
Word of caution, don't mix explicit lock objects and object monitors, it looks like at some point you're needing 2 locks, but can only acquire one
A different thread looks like it's holding the other lock, so two threads seem to be intertwined in a cold embrace, not relinquishing either lock.

kantenkugel:
From the dump submitted by @Kodehawa it looks like:
MainWS-ReadThread is trying to put something into a SynchronizedLongObjectMap
  at: WebSocketClient.queueAudioConnect(WebSocketClient.java:902)
while
MainWS-Sending Thread is attempting to get from a SynchronizedLongObjectMap
  at GuildImpl.getAudioManager(GuildImpl.java:600)
  at WebSocketClient.getNextAudioConnectRequest(WebSocketClient.java:928)

MinnDevelopment:
The deadlock fix is not confirmed but I have trust that this should be the issue resolved.
The HELLO handling has been updated to properly respect the heartbeat_interval mentioned in https://github.com/hammerandchisel/discord-api-docs/pull/323#issuecomment-319527028 to be future-proof. As of now this will simply resume on timeout (every heartbeat is timing out so don't use it until discord fixed it on their side).

On hold until discord has implemented the aforementioned fix.

MinnDevelopment:
I'm requesting a review from @DV8FromTheWorld as I believe this can be finished now.
The heartbeat_interval handling for HELLO will need to be updated in about a month once discord has rolled out the fixed value on their voice servers.

PR #411, opened 2017-07-27
Move event handling to worker thread to counter issues with shard deaths

MinnDevelopment:
This is also relevant for future plans of making member chunking optional on startup (lazy loading)


MinnDevelopment:
I'm gonna postpone this idea until we actually need this.

PR #415, opened 2017-07-27
Fix wrong variable in EmbedBuilder#addField() docs

napstr:


MinnDevelopment:
Can we change these to a `{@value ...}` tags?

napstr:
Done.

MinnDevelopment:
Could you do it for all the limit references in the other respective methods as well?

napstr:
Done.

src/main/java/net/dv8tion/jda/core/entities/MessageEmbed.java
```diff
@@ -156,8 +156,8 @@
      * <p>An Embed can only have, at max, {@value #EMBED_MAX_LENGTH_BOT} displayable text characters for {@link net.dv8tion.jda.core.AccountType#BOT AccountType.BOT}
      * accounts or {@value #EMBED_MAX_LENGTH_CLIENT} displayable text characters for {@link net.dv8tion.jda.core.AccountType#CLIENT AccountType.CLIENT} accounts.
      *
-     * <p>Both of these values are defined by {@link #EMBED_MAX_LENGTH_BOT EMBED_MAX_LENGTH_BOT} and
-     * {@link #EMBED_MAX_LENGTH_CLIENT EMBED_MAX_LENGTH_CLIENT} respectively.
+     * <p>Both of these values are defined by {@value #EMBED_MAX_LENGTH_BOT EMBED_MAX_LENGTH_BOT} and
+     * {@value #EMBED_MAX_LENGTH_CLIENT EMBED_MAX_LENGTH_CLIENT} respectively.
```
MinnDevelopment: Value tags don't support alias

napstr: Rip. Force push inc.

Almighty-Alpaca:
nice mentions :eyes:

napstr:
Reverted to the original commit which just fixes the error.

PR #423, opened 2017-08-04
Changed how RestActions are implemented

Almighty-Alpaca:
Makes most RestActions way smaller and much more beautiful. I hope I didn't accidentally break on, still would like still would like someone to test it first as I can't test all of them myself.


src/main/java/net/dv8tion/jda/core/requests/RestAction.java
```diff
@@ -212,14 +272,70 @@ public RestAction(JDA api, Route.CompiledRoute route, RequestBody data)
...
+     */
+    public RestAction(JDA api, Route.CompiledRoute route, RequestBody data, Function<Response, T> successTransformer, Function<Response, Throwable> failureTransformer)
+    {
+        Checks.notNull(api, "api");
+
+        this.api = (JDAImpl) api;
+        this.route = route;
+        this.data = data;
+        this.successTransformer = successTransformer;
+        this.failureTransformer = r -> Optional.ofNullable(failureTransformer.apply(r)).orElse(DEFAULT_FAILURE_TRANSFORMER.apply(r));
```
kantenkugel: What does this do exactly? it assumes that `failureTransformer != null`, then either returns `failureTransformer.apply(r)` or if that is `null` it returns `DEFAULT_FAILURE_TRANSFORMER.apply(r)`?

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -366,24 +366,24 @@ protected String getGateway()
...
-                        else
-                            request.onFailure(new Exception("Failed to get gateway url"));
-                    }
-                    catch (Exception e)
-                    {
-                        request.onFailure(e);
-                    }
-                }
-            };
+            RestAction<String> gateway = new RestAction<String>(api, Route.Misc.GATEWAY.compile(), r -> r.getObject().getString("url")); // TODO
```
kantenkugel: Still a TODO

Almighty-Alpaca: fml

DV8FromTheWorld:
This is prettier, but I feel like it opens the door to easy errors, which have been shown just in the conversion to this new, prettier, version.

MinnDevelopment:
Is there a good reason other than "looks prettier" to do this change? I personally don't believe this improves or enhances JDA in any shape or form which makes me question why we should merge this. This seems to just be a risk of introducing issues for no actual gain in either performance or code quality.

Almighty-Alpaca:
In my opinion this does not only look prettier but also improves readability as a lot of redundant code has been removed, for example all
```java
if (!response.isOk())
{
    request.onFailure(response);
    return;
}
```
checks which before had to be in **every** single RestAction implementation.

I also don't think that it opens that many easy errors as I actually didn't change the code of all the RestActions itself but rather just copied the code from an inner class into a consumer. I do agree though that it needs much testing as it touches core parts of the JDA.

The intention behind this pr was that I really don't think having to write all the error handling over and over again is a good approach and I feel like this improves it a lot.

DV8FromTheWorld:
A decision has been made not to go through with these changes.

I appreciate the concept of removing code duplicity, but not at the cost of hiding what is happening.

Issue #418, opened 2017-07-28
Provide access to decrypted AudioPackets, regardless of sequence number


mentlerd:
For applications that wish to do per-user voice recording an API like this would be really useful. 

The responsibility of having to reassemble the audio stream from out-of-order packets should be placed on the consumers of the interface.

SparklingComet:
It would be really useful, but recording people without them knowing in a (group)voice chat is illegal in many countries. It should be restricted to the currently logged in user's voice, if possible. (Not sure whether it is)

Almighty-Alpaca:
People can and will always abuse features. Restricting recording to yourself (which is implementation wise something entirely different from recording others btw) just because someone might use on other people without their knowledge is really bad imo.

SparklingComet:
> Restricting recording to yourself (which is implementation wise something entirely different from recording others btw)
I thought so
> Restricting recording to yourself just because someone might use on other people without their knowledge is really bad imo.
True, but we are allowing them to do so by adding this. If I am speaking with someone and I want to record only myself, I can use a mic/recording software. A bot will allow abusing and nothing else.

MinnDevelopment:
JDA already supports receiving the audio of users when they speak. I don't understand how this discussion is relevant to the thread.

SparklingComet:
> JDA already supports receiving the audio of users when they speak. I don't understand how this discussion is relevant to the thread.

It's relevant to the suggestion - whether or not it makes sense.

MinnDevelopment:
The suggestion has nothing to do with the fact that you can already implement such a functionality with the current state of JDA. It is requested to give access to unhandled frames to users. 
You are talking about generally having voice receive support which is already a thing thus discussing whether it should be restricted or not is nonsense.

SparklingComet:
> The suggestion has nothing to do with the fact that you can already implement such a functionality with the current state of JDA. It is requested to give access to unhandled frames to users.
You are talking about generally having voice receive support which is already a thing thus discussing whether it should be restricted or not is nonsense.

If it's already a thing, the whole issue can be closed. It's not my reply, but the issue itself that's redundant.

MinnDevelopment:
No you don't even understand what is requested, stop arguing about something that is completely irrelevant to this thread please.

DV8FromTheWorld:
@mentlerd I don't see the use case.

[...]

PR #425, opened 2017-08-07
Implement AudioPacketInterceptor

mentlerd:
Implemented a new interface for AudioPacketInterceptor.

This object can be registered just like an AudioReceiveHandler, but it provides a much more low level access to audio packets than the standard handleUserAudio callback.

Also allows discarding AudioPackets, effectively hiding them from the JDA backend.

This PR implements issue #418

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/audio/AudioConnection.java
+++ b/src/main/java/net/dv8tion/jda/core/audio/AudioConnection.java
@@ -57,6 +57,9 @@ public class AudioConnection
     public static final int OPUS_FRAME_TIME_AMOUNT = 20;//This is 20 milliseconds. We are only dealing with 20ms opus packets.
     public static final int OPUS_CHANNEL_COUNT = 2;     //We want to use stereo. If the audio given is mono, the encoder promotes it
                                                         // to Left and Right mono (stereo that is the same on both sides)
+    
+    private static final byte[] SILENCE_BYTES = new byte[] {(byte)0xF8, (byte)0xFF, (byte)0xFE};
+    
     private final TIntLongMap ssrcMap = new TIntLongHashMap();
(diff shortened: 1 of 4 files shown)


mentlerd:
This pull request belonds to issue #425

src/main/java/net/dv8tion/jda/core/audio/AudioConnection.java
```diff
@@ -258,28 +266,33 @@ else if (sendHandler == null && sendSystem != null)
 
     private synchronized void setupReceiveSystem()
     {
```
mentlerd: The current implementation initializes the receive thread even if there is only an AudioPacketInterceptor, but no ReceiveHandler.

I have tried to cleanup this setup routine as much as I could, but I would appreciate a second pair of eyes checking on it.

kantenkugel: With your changes, if there is a "consumer" but the socket is either null or closed, nothing happens.
In the old implementation, if the socket was null/closed (no matter if "consumer" is set, there was cleanup (your else branch)

mentlerd: Thanks, I did not spot that!

I think it does not make a difference in this case tough, the combined audio thread only gets setup when the udpSocket is not null, so there is nothing to clean up.

MinnDevelopment:
This PR is on hold until voice resume has been merged. On a side note, you need to update your code style to fit into JDA, the wiki explains this.

mentlerd:
I have fixed the teardown and style issues

MinnDevelopment:
I'm assigning this to @DV8FromTheWorld so he can decide how we should implement this.

MinnDevelopment:
Closing this, so @DV8FromTheWorld can use the issue to decide on an implementation. This PR is not satisfying.

PR #432, opened 2017-08-18
Ignore the first 4 bytes (int) of the answer package from discord whi…

Waishon:
…ch contains the SSRC. Otherwise the discovered IP is invalid, because Java can't parse the IP if there's a char in front of it

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
+++ b/src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
@@ -447,7 +447,7 @@ public class AudioWebSocket extends WebSocketAdapter
             // the last 2 bytes. Not sure why these exist.  The last 2 bytes are the port. More info below.
             String ourIP = new String(receivedPacket.getData());//Puts the entire byte array in. nulls are converted to spaces.
             ourIP = ourIP.substring(0, ourIP.length() - 2); //Removes the port that is stuck on the end of this string. (last 2 bytes are the port)
-            ourIP = ourIP.trim();                           //Removes the extra whitespace(nulls) attached to both sides of the IP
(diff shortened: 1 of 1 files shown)


Waishon:
This is the UDP answer package of the official Discord Client:
official

jvtrigueros:
@Waishon did this change fix it for you? Maybe I could build a local copy of JDA while this gets resolved.

Waishon:
@jvtrigueros
Yep that fixed it for me. 
Have you the same issue with UDP?
Would be interesting to know if this affects only very few users or if there're more.

src/main/java/net/dv8tion/jda/core/audio/AudioWebSocket.java
```diff
@@ -447,7 +447,7 @@ private InetSocketAddress handleUdpDiscovery(InetSocketAddress address, int ssrc
             // the last 2 bytes. Not sure why these exist.  The last 2 bytes are the port. More info below.
             String ourIP = new String(receivedPacket.getData());//Puts the entire byte array in. nulls are converted to spaces.
             ourIP = ourIP.substring(0, ourIP.length() - 2); //Removes the port that is stuck on the end of this string. (last 2 bytes are the port)
-            ourIP = ourIP.trim();                           //Removes the extra whitespace(nulls) attached to both sides of the IP
+            ourIP = ourIP.substring(4, ourIP.length()).trim();  //Removes the SSRC of the answer package and the extra whitespace(nulls) attached to both sides of the IP
```
MinnDevelopment: This could be in the line above as `substring(4, ourIP.length() - 2)` to only keep the trim on that line

Waishon: Fixed

PR #436, opened 2017-08-20
Removed GameImpl and replace with Game immutable

MinnDevelopment:
This only really changes internals of JDA.
The change that can be noticed by users is a different hashCode output due to it being changed to Objects.hash(Object...)


src/main/java/net/dv8tion/jda/core/entities/Game.java
```diff
@@ -15,40 +15,99 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o)
+    {
+        if (!(o instanceof Game))
+            return false;
+
+        Game oGame = (Game) o;
+        if (oGame.getType() != type)
```
kantenkugel: Isn't that If redundant? (see return statement below). Seems like this was in previous version too tho

kantenkugel: Never get used to how github posts snippets... (and not being able to set a range yourself)

MinnDevelopment: I just copied these methods from GameImpl

DV8FromTheWorld: I have to agree. The if statement is redundant and should be removed.

src/main/java/net/dv8tion/jda/core/entities/Game.java
```diff
@@ -100,15 +159,15 @@ static Game of(String name, String url)
      *
      * @return True if the provided url is valid for triggering Discord's streaming status
      */
-    static boolean isValidStreamingUrl(String url)
+    public static boolean isValidStreamingUrl(String url)
     {
-        return url != null && url.matches("^https?:\\/\\/(www\\.)?twitch\\.tv\\/.+");
+        return url != null && url.matches("^https?://(www\\.)?twitch\\.tv/.+");
```
DV8FromTheWorld: I see that this regex changed. Was it supposed to?

src/main/java/net/dv8tion/jda/core/entities/impl/GameImpl.java
```diff
@@ -17,63 +17,11 @@
 
 import net.dv8tion.jda.core.entities.Game;
 
-public class GameImpl implements Game
+@Deprecated
+public class GameImpl extends Game
```
DV8FromTheWorld: Why is this Deprecated instead of removed?
This is an internal class, and as such, we shouldn't be deprecating. It should be removed wholesale.

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -86,6 +86,11 @@ public SelfUser createSelfUser(JSONObject self)
                 .setBot(self.has("bot") && self.getBoolean("bot"));
     }
 
+    public Game createGameInternal(String name, String url, Game.GameType type)
```
DV8FromTheWorld: This does not need `Internal` appended to it.
It is implied that anything being built by the EntityBuilder is being done by an internal JDA system.

src/main/java/net/dv8tion/jda/core/entities/impl/GameImpl.java
MinnDevelopment: Even as it is a whole class, it was public and removing a public class is a breaking change. I can imagine users that didn't read the docs and used this class directly in setGame methods.

src/main/java/net/dv8tion/jda/core/entities/Game.java
MinnDevelopment: Yes this was intended.

src/main/java/net/dv8tion/jda/core/entities/Game.java
```diff
@@ -81,15 +140,15 @@ static Game of(String name)
      *
      * @see    #isValidStreamingUrl(String)
```
kantenkugel: I would also add a note to the url param and maybe reformulate it (the invalid part doesn't make that much sense as user can't specify type anyway)

src/main/java/net/dv8tion/jda/core/entities/Game.java
```diff
@@ -100,25 +159,33 @@ static Game of(String name, String url)
...
+    public enum GameType
     {
         /**
          * The GameType used to represent a normal {@link net.dv8tion.jda.core.entities.Game Game} status.
          */
         DEFAULT(0),
         /**
          * Used to indicate that the {@link net.dv8tion.jda.core.entities.Game Game} is a stream, specifically for
+         * <br>This type is displayed as "Streaming" in the discord client.
+         */
```
kantenkugel: Broken doc (probs due to copy-paste)

src/main/java/net/dv8tion/jda/core/entities/Game.java
```diff
@@ -100,25 +159,33 @@ static Game of(String name, String url)
...
+        STREAMING(1),
+        /**
+         * Used to indicate that the {@link net.dv8tion.jda.core.entities.Game Game} is a stream, specifically for
          * <a href="https://www.twitch.tv">https://www.twitch.tv</a>.
          * <br>This type is displayed as "Streaming" in the discord client.
+         *
+         * @deprecated This might be removed when discord introduces more streaming methods. Use {@link #STREAMING} instead.
          */
+        @Deprecated
         TWITCH(1);
```
kantenkugel: Could actually be removed as users weren't and still aren't able to use it anyway (other than for == checks). 
Might be worth to keep tho in case Discord actually adds different types of streaming. 
Deprecation message could also be reformulated, as removal is not really dependent on discord adding other methods imo (removal due to no-use in lib, just like GameImpl).

src/main/java/net/dv8tion/jda/core/managers/impl/PresenceImpl.java
```diff
@@ -231,7 +231,7 @@ private JSONObject getGameJson(Game game)
         JSONObject gameObj = new JSONObject();
         gameObj.put("name", game.getName());
         gameObj.put("type", game.getType().getKey());
-        if (game.getType() == Game.GameType.TWITCH && game.getUrl() != null)
+        if (game.getUrl() != null)
             gameObj.put("url", game.getUrl());
```
kantenkugel: Does it do anything if we send a url when type is not STREAMING? (Game with url but DEFAULT type is created, when specifying a invalid streaming url to `Game.of(String, String)`)

src/main/java/net/dv8tion/jda/core/entities/Game.java
MinnDevelopment: I don't want to remove it for now. I want to decide later when discord makes their changes (which I expect to happen)

src/main/java/net/dv8tion/jda/core/entities/impl/GameImpl.java
DV8FromTheWorld: It doesn't matter. This is an Impl class, and as such, we don't deprecate them.
Please remove.

src/main/java/net/dv8tion/jda/core/entities/Game.java
```diff
@@ -161,7 +161,7 @@ public static Game of(String name, String url)
      */
     public static boolean isValidStreamingUrl(String url)
     {
-        return url != null && url.matches("^https?://(www\\.)?twitch\\.tv/.+");
```
DV8FromTheWorld: Why was the `^` removed?

MinnDevelopment: It doesn't matter when you use `matches`

PR #437, opened 2017-08-20
Enhanced exception UX

MinnDevelopment:
- Replaced a lot of raw RuntimeExceptions with more specific throwables
- Split up PermissionException into InsufficientPermissionException and HierarchyException for userfriendly multi-catch

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -119,38 +119,39 @@ javadoc {
 
     //### excludes ###
     //jda internals
-    exclude("net/dv8tion/jda/bot/entities/impl")
-    exclude("net/dv8tion/jda/bot/handle")
-    exclude("net/dv8tion/jda/bot/managers/impl")
+    exclude('net/dv8tion/jda/bot/entities/impl')
+    exclude('net/dv8tion/jda/bot/handle')
+    exclude('net/dv8tion/jda/bot/managers/impl')
 
-    exclude("net/dv8tion/jda/client/entities/impl")
-    exclude("net/dv8tion/jda/client/handle")
-    exclude("net/dv8tion/jda/client/managers/impl")
+    exclude('net/dv8tion/jda/client/entities/impl')
+    exclude('net/dv8tion/jda/client/handle')
+    exclude('net/dv8tion/jda/client/managers/impl')
 
-    exclude("net/dv8tion/jda/core/audio/AudioConnection.java",
-            "net/dv8tion/jda/core/audio/AudioWebSocket.java",
-            "net/dv8tion/jda/core/audio/AudioPacket.java",
-            "net/dv8tion/jda/core/audio/Decoder.java",
-            "net/dv8tion/jda/core/audio/hooks/ListenerProxy.java")
-    exclude("net/dv8tion/jda/core/entities/impl")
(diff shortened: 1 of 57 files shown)


kantenkugel:
I would only merge this for the next minor/major version update due to it very likely breaking user-code

MinnDevelopment:
This does not affect user code in any way as all new exceptions are children of the previous

kantenkugel:
Oh right, i somehow thought InsufficientPermEx was a rename of normal PermEx :)

kantenkugel:
That last commit: the most random changes :)

build.gradle
```diff
@@ -119,38 +119,39 @@ javadoc {
...
+    exclude('net/dv8tion/jda/core/requests/GuildLock.java',
+            'net/dv8tion/jda/core/requests/WebSocketClient.java',
+            'net/dv8tion/jda/core/requests/RateLimiter.java',
+            'net/dv8tion/jda/core/requests/Request.java',
+            'net/dv8tion/jda/core/requests/Route.java',
+            'net/dv8tion/jda/core/requests/Requester.java',
+            'net/dv8tion/jda/core/requests/Response.java',
+            'net/dv8tion/jda/core/requests/ratelimit',
+            'net/dv8tion/jda/core/requests/restaction/CompletedFuture.java',
+            'net/dv8tion/jda/core/requests/restaction/RequestFuture.java')
```
DV8FromTheWorld: Not that it is a problem, but why did you feel it necessary to change these from `""` to `''`?

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -472,8 +472,8 @@ public JDABuilder useSharding(int shardId, int shardTotal)
     {
         if (accountType != AccountType.BOT)
             throw new AccountTypeException(AccountType.BOT);
-        if (shardId < 0 || shardTotal < 1 || shardId >= shardTotal)
-            throw new RuntimeException("This configuration of shardId and shardTotal is not allowed! 0 <= shardId < shardTotal with shardTotal > 0");
+        Checks.check(shardId >= 0 && shardTotal > 0 && shardId < shardTotal,
```
DV8FromTheWorld: Boolean logic was changed, but the error message was not updated.
Also, the error message should be clarified. It is a bit complicated.

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -43,6 +43,7 @@
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.io.IOError;
```
DV8FromTheWorld: Why the import?

build.gradle
MinnDevelopment: Because the `""` is a groovy string which has interpolation and `''` is a normal string afaik. As we don't need interpolation I changed it to a normal string

src/main/java/net/dv8tion/jda/core/JDABuilder.java
MinnDevelopment: The logic has actually not changed it was just inverted here

PR #440, opened 2017-08-24
Improve access for copy methods & fix missing copy of isNSFW value

MinnDevelopment:
- Fixed isNSFW value not being copied with GuildController#createCopyOfChannel
- Added createCopy(Guild) into entity interface
- Moved GuildController#createWebhook into TextChannel interface deprecating the old method (probably will not be removed)


kantenkugel:
I would add some `@CheckReturnValue` annotations to the RestActions (and subclasses) that are returned by most methods

src/main/java/net/dv8tion/jda/core/entities/impl/TextChannelImpl.java
```diff
@@ -225,6 +237,21 @@ public int getPosition()
...
+    {
+        Checks.notNull(guild, "Guild");
+        ChannelAction action = guild.getController().createTextChannel(name).setNSFW(nsfw).setTopic(topic);
+        for (PermissionOverride o : overrides.valueCollection())
+        {
+            if (o.isMemberOverride())
+                action.addPermissionOverride(o.getMember(), o.getAllowedRaw(), o.getDeniedRaw());
+            else
+                action.addPermissionOverride(o.getRole(), o.getAllowedRaw(), o.getDeniedRaw());
+        }
```
DV8FromTheWorld: What happens when someone provides a different guild to copy into?
What if the member which the override is for isn't the new guild?
Also, if a Guild is provided that isn't the guild that this channel belongs to, then it isn't going to have roles that have role overrides for this channel.

I think the implications of this need to be rethought related to permission copying related to overrides.

src/main/java/net/dv8tion/jda/core/entities/impl/VoiceChannelImpl.java
```diff
@@ -75,6 +75,21 @@ public int getPosition()
...
+    {
+        Checks.notNull(guild, "Guild");
+        ChannelAction action = guild.getController().createVoiceChannel(name).setBitrate(bitrate).setUserlimit(userLimit);
+        for (PermissionOverride o : overrides.valueCollection())
+        {
+            if (o.isMemberOverride())
+                action.addPermissionOverride(o.getMember(), o.getAllowedRaw(), o.getDeniedRaw());
+            else
+                action.addPermissionOverride(o.getRole(), o.getAllowedRaw(), o.getDeniedRaw());
+        }
```
DV8FromTheWorld: Same statement as above.

src/main/java/net/dv8tion/jda/core/entities/impl/TextChannelImpl.java
MinnDevelopment: The handling of these cases has not changed, however we could do special checks to only copy users that are in the target guild if it isn't the same guild.

DV8FromTheWorld: I'm not quite sure how to feel about this, specifically because it has a ton of side effects that you would only realize by diving into the code.
Certain situations copy member overrides, sometimes it doesn't. Within a guild role overrides are copied, "but it didn't when i cross-copied to a different Guild that has the same Role names!!"

Seems like this could easily cause headache.

MinnDevelopment: I see it as "it copies everything it can"

PR #444, opened 2017-08-25
Adding SessionReconnectQueue

MinnDevelopment:
This can be used to stop sessions from reconnecting simultaneously.
It will prevent them from sending IDENTIFY payloads on reconnect until they come up in the queue.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/requests/SessionReconnectQueue.java
+++ b/src/main/java/net/dv8tion/jda/core/requests/SessionReconnectQueue.java
@@ -54,7 +54,7 @@ public class SessionReconnectQueue
             while (!reconnectQueue.isEmpty())
             {
                 final WebSocketClient client = reconnectQueue.poll();
-                client.init();
+                client.sendIdentify();
                 try
                 {
                     if (!reconnectQueue.isEmpty())
--- a/src/main/java/net/dv8tion/jda/core/requests/SessionReconnectQueue.java
+++ b/src/main/java/net/dv8tion/jda/core/requests/SessionReconnectQueue.java
@@ -16,20 +16,20 @@
 
 package net.dv8tion.jda.core.requests;
 
-import java.util.Queue;
+import java.util.concurrent.BlockingQueue;
 
-class SessionReconnectQueue
+public class SessionReconnectQueue
 {
-    private final Object lock = new Object();
-    private final Queue<WebSocketClient> reconnectQueue;
-    private volatile Thread reconnectThread;
+    protected final Object lock = new Object();
+    protected final BlockingQueue<WebSocketClient> reconnectQueue;
+    protected volatile Thread reconnectThread;
 
-    SessionReconnectQueue(final Queue<WebSocketClient> reconnectQueue)
+    public SessionReconnectQueue(final BlockingQueue<WebSocketClient> reconnectQueue)
     {
         this.reconnectQueue = reconnectQueue;
     }
 
-    void appendSession(final WebSocketClient client)
+    protected void appendSession(final WebSocketClient client)
     {
         reconnectQueue.add(client);
         synchronized (lock)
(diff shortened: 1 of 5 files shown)


src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -415,9 +417,16 @@ public void onConnected(WebSocket websocket, Map<String, List<String>> headers)
         messagesSent = 0;
         ratelimitResetTime = System.currentTimeMillis() + 60000;
         if (sessionId == null)
-            sendIdentify();
+        {
+            if (!firstInit && reconnectQueue != null)
+                reconnectQueue.appendSession(this);
```
DV8FromTheWorld: Won't this run into an issue where if a lot of shards attempt to reconnect, and they are all placed in this queue, then shards at the end of the queue have to wait for a significant amount of time, thus their period for identify/resume will expire.

What happens if the MainWS DCs and then the reconnectQueue attempts to sendIdentify?

MinnDevelopment: I didn't think much about the implementation because I wanted a fix fast for now, I'm looking further into it.

MinnDevelopment: The sessions are only added to the reconnect queue if they cannot resume. Active sessions will not be queued but full new setup will require them to wait their turn.
The system is designed to first come first served using a blocking queue that is handled when it is first populated.
Reconnect will happen on the queue worker causing it to lose the identification string which is appended to the client's threads, instead it logs the shard information when it reconnects the session.
### Exmaple
```
[06:20:39] [Warning] [JDASocket]: Got disconnected from WebSocket (Internet?!)... Appending session to reconnect queue
[06:20:39] [Warning] [JDASocket]: Got disconnected from WebSocket (Internet?!)... Appending session to reconnect queue
[06:20:39] [Warning] [JDASocket]: Queue is attempting to reconnect a shard... Shard: [0 / 2]
[06:20:39] [Warning] [JDASocket]: Attempting to reconnect in 2s
[06:20:41] [Warning] [JDASocket]: Attempting to reconnect!
[06:20:41] [Info] [JDASocket]: Connected to WebSocket
[06:20:42] [Info] [JDA]: Finished (Re)Loading!
[06:20:46] [Warning] [JDASocket]: Queue is attempting to reconnect a shard... Shard: [1 / 2]
[06:20:46] [Warning] [JDASocket]: Attempting to reconnect in 2s
[06:20:48] [Warning] [JDASocket]: Attempting to reconnect!
[06:20:48] [Info] [JDASocket]: Connected to WebSocket
[06:20:48] [Info] [JDA]: Finished (Re)Loading!
```

src/main/java/net/dv8tion/jda/core/requests/SessionReconnectQueue.java
```diff
@@ -0,0 +1,86 @@
...
+
+                    if (!reconnectQueue.isEmpty())
+                        Thread.sleep(RECONNECT_DELAY);
+                }
+                catch (InterruptedException ex)
+                {
+                    throw new AssertionError(ex);
+                }
+            }
+            synchronized (lock)
```
DV8FromTheWorld: There is a race condition here.
1) While loop ends due to queue being empty
2) New item is added via appendSession
3) sync in appendSession returns because no new thread needs to be created
4) sync in run() kills the thread.

Now there is a session that will never reconnect.

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -331,6 +333,11 @@ public void close(int code)
         socket.sendClose(code);
     }
 
+    public void close(int code, String reason)
```
DV8FromTheWorld: Does this actually need to be public?

MinnDevelopment: It was useful for testing and can be useful for debugging in the future

PR #445, opened 2017-08-25
Implemented phone verification (only get) and 50 MiB file limit for nitro accounts

Almighty-Alpaca:
If anyone knows what the `mobile` is exactly used for let me know so that I can update the javadoc for it as it's currently a bit vague.


meowingtwurtle:
While you're there, could you fix: `File is to big! (MessageChannel:511)

src/main/java/net/dv8tion/jda/core/handle/UserUpdateHandler.java
```diff
@@ -43,54 +43,70 @@ protected Long handleInternally(JSONObject content)
...
+
         if (api.getAccountType() == AccountType.CLIENT && !Objects.equals(email, self.getEmail()))
         {
             String oldEmail = self.getEmail();
             self.setEmail(email);
-            api.getEventManager().handle(
-                    new SelfUpdateEmailEvent(
-                            api, responseNumber,
-                            oldEmail));
+            api.getEventManager().handle(new SelfUpdateEmailEvent(api, responseNumber, oldEmail));
```
MinnDevelopment: Why did you change these events here? They have a consistent setup...

MinnDevelopment:
Does having a registered phone number outweigh the other verification level requirements?

Sanduhr32:
Yes Minn. However the current verification level of a guild outweights the lower verification levels. And yes i tested it :)

MinnDevelopment:
If that is the case this swtich has to be changed. And when the phone number is not null it should not break but return true from what I can tell.

src/main/java/net/dv8tion/jda/core/entities/impl/GuildImpl.java
```diff
@@ -661,19 +661,22 @@ public boolean checkVerification()
             return true;
         if(canSendVerification)
             return true;
+
+        if (api.getSelfUser().getPhoneNumber() != null)
```
Almighty-Alpaca: @MinnDevelopment @Sanduhr32 is this what you mean?

src/main/java/net/dv8tion/jda/core/entities/impl/GuildImpl.java
```diff
@@ -661,19 +661,22 @@ public boolean checkVerification()
...
+            return canSendVerification = true;
+
         switch (verificationLevel)
         {
             case VERY_HIGH:
-                if(api.getSelfUser().getPhoneNumber() != null)
-                    break;
+                return false; // we already checked for a verified phone number
```
Dunciboy: Shouldn't this be a `break;` to be consistent?

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -80,6 +80,9 @@ public SelfUser createSelfUser(JSONObject self)
                 .setVerified(self.getBoolean("verified"))
                 .setMfaEnabled(self.getBoolean("mfa_enabled"))
                 .setEmail(!self.isNull("email") ? self.getString("email") : null)
+                .setMobile(self.has("mobile") ? self.getBoolean("mobile") : false) 
+                .setPremium(self.has("premium") ? self.getBoolean("premium") : false)
+                .setPhoneNumber(!self.isNull("phone") ? self.getString("phone") : null)
```
MinnDevelopment: This could be `!self.isNull(x) && self.getBoolean(x)` for the boolean values.

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
```diff
@@ -507,8 +507,8 @@ protected void handleResponse(Response response, Request<Message> request)
         Checks.notNull(file, "file");
         Checks.check(file.exists() && file.canRead(),
             "Provided file is either null, doesn't exist or is not readable!");
-        Checks.check(file.length() <= Message.MAX_FILE_SIZE,// TODO: deal with Discord Nitro allowing 50MB files.
-            "File is to big! Max file-size is 8MB");
+        Checks.check(file.length() <= (getJDA().getAccountType() == AccountType.CLIENT && getJDA().getSelfUser().isPremium() ? Message.MAX_FILE_SIZE_NITRO : Message.MAX_FILE_SIZE),
```
MinnDevelopment: This ternary is pretty ugly, can we move this outside the check call?

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
```diff
@@ -660,11 +660,11 @@ protected void handleResponse(Response response, Request<Message> request)
     @CheckReturnValue
     default RestAction<Message> sendFile(byte[] data, String fileName, Message message)
     {
-        Checks.notNull(data, "file data[]");
+        Checks.notNull(data, "data");
         Checks.notNull(fileName, "fileName");
 
-        Checks.check(data.length <= Message.MAX_FILE_SIZE,   //8MB
-                "Provided data is too large! Max file-size is 8MB (%d)", Message.MAX_FILE_SIZE);
+        Checks.check(data.length <= (getJDA().getAccountType() == AccountType.CLIENT && getJDA().getSelfUser().isPremium() ? Message.MAX_FILE_SIZE_NITRO : Message.MAX_FILE_SIZE),
```
MinnDevelopment: Same as above

src/main/java/net/dv8tion/jda/core/entities/SelfUser.java
```diff
@@ -54,6 +54,43 @@
...
+     * Used to get the phone number of the currently logged in account if a phone number has been attached to it.
+     * <br><b>NOTE:</b> this is a {@link net.dv8tion.jda.core.AccountType#CLIENT AccountType.CLIENT} method only!
+     *
+     * @throws AccountTypeException
+     *         If this method is called when {@link net.dv8tion.jda.core.JDA#getAccountType() JDA#getAccountType()} does not return
+     *         {@link net.dv8tion.jda.core.AccountType#CLIENT AccountType.CLIENT}. E.g: If the logged in account isn't a Client account!
+     *
+     * @return The phone of the currently logged in account or null if there's no number associated
+     */
+    String getPhoneNumber() throws AccountTypeException;
```
MinnDevelopment: Why do you specifically throw an `AccountTypeException` for these methods? I would suggest to actually just return w/e we parsed and document that for bots the value will most likely be false or null. It might be possible that discord will make some of these available to bots in the future (unlikely) so we should just go with that, imo.

src/main/java/net/dv8tion/jda/core/entities/impl/GuildImpl.java
```diff
@@ -661,20 +661,28 @@ public boolean checkVerification()
...
         if(canSendVerification)
             return true;
+
+        if (api.getSelfUser().getPhoneNumber() != null)
+            return canSendVerification = true;
+
         switch (verificationLevel)
         {
+            case VERY_HIGH:
+                return false; // we already checked for a verified phone number
```
MinnDevelopment: This should be a `break;` for consistency

src/main/java/net/dv8tion/jda/core/handle/UserUpdateHandler.java
MinnDevelopment: ^ I wanna reiterate this comment, please don't change these as we have a consistent style for these event handling calls.

src/main/java/net/dv8tion/jda/core/handle/UserUpdateHandler.java
```diff
@@ -43,54 +43,70 @@ protected Long handleInternally(JSONObject content)
...
+            self.setPremium(premium);
+//            api.getEventManager().handle(new SelfUpdatePremiumEvent(api, responseNumber, oldPremium));
+        }
+
+        if (api.getAccountType() == AccountType.CLIENT && !Objects.equals(phoneNumber, self.getPhoneNumber()))
+        {
+            JDAImpl.LOG.debug("");
+            String oldPhoneNumber = self.getPhoneNumber();
+            self.setPhoneNumber(phoneNumber);
+//            api.getEventManager().handle(new SelfUpdatePhoneNumberEvent(api, responseNumber, oldPhoneNumber));
```
MinnDevelopment: When do you intend to add these events?

DV8FromTheWorld:
I'd like to see some motion on this PR.

Almighty-Alpaca:
Right... I completely forgot this still exists.

src/main/java/net/dv8tion/jda/core/entities/SelfUser.java
Almighty-Alpaca: I chose this because we already have the same be behavior for `getEmail()`. I also highly doubt that any of these will be available for bots anytime in the future.

src/main/java/net/dv8tion/jda/core/entities/Message.java
```diff
@@ -63,7 +63,16 @@
...
-    int MAX_FILE_SIZE = 8 << 20; // 8mb
+    /**
+     * Useful constant for the max allowed file size for message attachments on normal accounts. Value: {@value #MAX_FILE_SIZE} Bytes (8 MiB)
+     */
+    int MAX_FILE_SIZE = 8 << 20; // 8 MiB
+
+    /**
+     * Useful constant for the max allowed file size for message attachments on nitro accounts. Value: {@value #MAX_FILE_SIZE_NITRO} Bytes (50 MiB)
+     */
+    int MAX_FILE_SIZE_NITRO = 50 << 20; // 50 MiB
```
DuncanCasteleyn: This isn't really consistent you are calling your Boolean premium to check if user is a Nitro users but here you are calling you constant MAX_FILE_SIZE_NITRO instead of MAX_FILE_SIZE_PREMIUM. I'd say either rename the Boolean to nitro or rename this constant.
Your comment is "//50 MiB" is wrong the current int value would be 50mb.

src/main/java/net/dv8tion/jda/core/entities/SelfUser.java
```diff
@@ -54,6 +54,55 @@
...
+     * The Discord Nitro status of this account.
+     * <br><b>NOTE:</b> this is a {@link net.dv8tion.jda.core.AccountType#CLIENT AccountType.CLIENT} method only!
+     *
+     * @throws AccountTypeException
+     *         If this method is called when {@link net.dv8tion.jda.core.JDA#getAccountType() JDA#getAccountType()} does not return
+     *         {@link net.dv8tion.jda.core.AccountType#CLIENT AccountType.CLIENT}. E.g: If the logged in account isn't a Client account!
+     *
+     * @return The Discord Nitro status of the currently logged in account.
+     */
+    boolean isPremium() throws AccountTypeException;
```
MinnDevelopment: We should call this `isNitro` to avoid unneeded confusion

MinnDevelopment:
@Almighty-Alpaca can you please update this PR

src/main/java/net/dv8tion/jda/core/events/self/SelfUpdateMobileEvent.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.core.events.self;
+
+import net.dv8tion.jda.core.JDA;
+
+public class SelfUpdateMobileEvent extends GenericSelfUpdateEvent
```
MinnDevelopment: This event needs documentation on class-level and on its methods.

src/main/java/net/dv8tion/jda/core/events/self/SelfUpdateNitroEvent.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.core.events.self;
+
+import net.dv8tion.jda.core.JDA;
+
+public class SelfUpdateNitroEvent extends GenericSelfUpdateEvent
```
MinnDevelopment: This event needs documentation on class-level and on its methods.

src/main/java/net/dv8tion/jda/core/events/self/SelfUpdatePhoneNumberEvent.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.core.events.self;
+
+import net.dv8tion.jda.core.JDA;
+
+public class SelfUpdatePhoneNumberEvent extends GenericSelfUpdateEvent
```
MinnDevelopment: This event needs documentation on class-level and on its methods.

src/main/java/net/dv8tion/jda/core/handle/UserUpdateHandler.java
```diff
@@ -44,53 +44,83 @@ protected Long handleInternally(JSONObject content)
...
+            boolean oldNitro = self.isNitro();
+            self.setNitro(nitro);
+            api.getEventManager().handle(new SelfUpdateNitroEvent(
+                    api, responseNumber,
+                    oldNitro));
+        }
+
+        if (api.getAccountType() == AccountType.CLIENT && !Objects.equals(phoneNumber, self.getPhoneNumber()))
+        {
+            JDAImpl.LOG.debug("");
```
MinnDevelopment: Uh?

PR #446, opened 2017-08-27
Rewrite decompression to use less allocations and regrows

Shredder121:
The old code used a `StringBuilder` with the default capacity and made a new String for each 128 byte chunk.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
+++ b/src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
@@ -43,12 +43,13 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.time.OffsetDateTime;
 import java.util.*;
 import java.util.zip.DataFormatException;
-import java.util.zip.Inflater;
+import java.util.zip.InflaterOutputStream;
 
 public class WebSocketClient extends WebSocketAdapter implements WebSocketListener
 {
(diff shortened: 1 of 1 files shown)


kantenkugel:
I am not entirely sure if this system is that much faster, as there might be a lot of array copies to enlarge the buffer of the ByteArrayOutputStream depending on the size of the decompressor output

Shredder121:
At least the greater initial size limits the regrows, as was seen in the old StringBuilder approach, and actually doing the string coding at the end is a lot better for if there are multi byte characters, and string coders that might do validation?
I've been running it for some time, but forgot to take screenshots of my sessions.
You can take a look at my branch `decompresser-test` if you wish.

MinnDevelopment:
Would you please update this to follow our code style in bracket placing.

Shredder121:
Sorry my bad, I will 😅

Shredder121:
With only 6 guilds and rebuilding the shard it's apparent all those calls to the string constructor contribute to 1) more cpu time than necessary and 2) more allocations.
I don't have any hard evidence to support whether multi-byte characters would cause issues or not, but I suppose it would be pretty bad luck if it poses a problem in the real world, if those happen to align with the 128 byte chunks.

| CPU call tree | Allocation call tree |
|---------------|----------------------|
| ![CPU Call Tree][cpu] | ![Allocation Call Tree][mem]

[cpu]: https://user-images.githubusercontent.com/4105066/29750882-0acb8b6e-8b48-11e7-9e3a-8256856293d8.png
[mem]: https://user-images.githubusercontent.com/4105066/29750886-1e26307e-8b48-11e7-894c-bc0115ded832.png

Shredder121:
(the last repush was just some import cleanup)

MinnDevelopment:
@Shredder121 Thank you for your contribution

Shredder121:
:+1:

PR #450, opened 2017-09-01
Added method to download attachment into byte array

Dunciboy:
Made it possible to download attachments into a byte array, so you are no longer forced to write them to a file.

For example when you want to send an attachment from one guild to another one.


MinnDevelopment:
This will already be addressed in `experimental/message-rw`

Dunciboy:
Oh, let me take a look at that.

Dunciboy:
Yeah, should grant me access to what I want in the future.

PR #451, opened 2017-09-02
Emit ShutdownEvent when adding to queue fails

MinnDevelopment:


src/main/java/net/dv8tion/jda/core/requests/SessionReconnectQueue.java
```diff
@@ -39,7 +39,8 @@ public SessionReconnectQueue(final BlockingQueue<WebSocketClient> reconnectQueue
 
     protected void appendSession(final WebSocketClient client)
     {
-        reconnectQueue.add(client);
+        if (!reconnectQueue.offer(client))
+            throw new IllegalStateException("Queue rejected session");
```
Shredder121: Nice, thanks!

PR #452, opened 2017-09-02
Version 3.3 Release

MinnDevelopment:


MinnDevelopment:
They were not deprecated in the 3.2 release thus we cannot remove them yet

PR #453, opened 2017-09-03
Message Rewrite

MinnDevelopment:
- Added MessageAction
- Added handling for other MessageTypes
- Added more methods to Message
- Made Message and MessageEmbed immutable
- Added DataMessage as output of MessageBuilder
- Improved some events / added more events


MinnDevelopment:
I would like to move forward with these changes.

DV8FromTheWorld:
I will review these changes again.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
```diff
@@ -0,0 +1,822 @@
...
+        boolean channel = false;
+        boolean role = false;
+        boolean user = false;
+        boolean emote = false;
+        for (MentionType type : types)
+        {
+            switch (type)
+            {
+                case EVERYONE:
+                case HERE:
```
Almighty-Alpaca: The returned list should also include `@here`/`@everyone` mentions.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
```diff
@@ -0,0 +1,822 @@
...
+                    if (!emote)
+                        mentions.addAll(getEmotes());
+                    emote = true;
+            }
+        }
+        return Arrays.asList(mentions.toArray(new IMentionable[mentions.size()]));
+    }
+
+    @Override
+    public boolean isMentioned(IMentionable mentionable, MentionType... types)
```
Almighty-Alpaca: Can't you get the `MentionType` from the `IMentionable`?

src/main/java/net/dv8tion/jda/core/entities/impl/MessageImpl.java
```diff
@@ -1,650 +0,0 @@
...
-import net.dv8tion.jda.core.utils.Helpers;
-import org.json.JSONObject;
-
-import java.io.IOException;
-import java.time.OffsetDateTime;
-import java.util.*;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
-public class MessageImpl implements Message
```
Almighty-Alpaca: This still exists as an empty file that has not been deleted yet.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
```diff
@@ -0,0 +1,822 @@
...
+            final Member member = getGuild().getMember((User) mentionable);
+            if (CollectionUtils.containsAny(getMentionedRoles(), member.getRoles()))
+                return true;
+        }
+        return false;
+    }
+
+    private boolean isMass(String s)
+    {
+        return hasPermission(Permission.MESSAGE_MENTION_EVERYONE) && content.contains(s);
```
Almighty-Alpaca: This should use the `mentionsEveryone` fields as this can give incorrect results if the permissions of the author have been changed since the message has been sent.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
```diff
@@ -0,0 +1,822 @@
...
+                            .replace("<@!" + user.getId() + '>', '@' + name);
+                }
+            }
+            for (Emote emote : getEmotes())
+            {
+                tmp = tmp.replace(emote.getAsMention(), ":" + emote.getName() + ":");
+            }
+            for (TextChannel mentionedChannel : getMentionedChannels())
+            {
+                tmp = tmp.replace("<#" + mentionedChannel.getId() + '>', '#' + mentionedChannel.getName());
```
Almighty-Alpaca: You could use `mentionedChannel.getAsMention()` here, just as with the emotes.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
```diff
@@ -0,0 +1,822 @@
...
+            }
+            for (Role mentionedRole : getMentionedRoles())
+            {
+                tmp = tmp.replace("<@&" + mentionedRole.getId() + '>', '@' + mentionedRole.getName());
```
Almighty-Alpaca: Same as above.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
```diff
@@ -0,0 +1,822 @@
...
+                        mentions.addAll(getMentionedRoles());
+                    role = true;
+                    break;
+                case EMOTE:
+                    if (!emote)
+                        mentions.addAll(getEmotes());
+                    emote = true;
+            }
+        }
+        return Arrays.asList(mentions.toArray(new IMentionable[mentions.size()]));
```
Almighty-Alpaca: What is the reason to create a completely new List? Why don't just return `Collections.unmodifiableList(mentions)` ?

src/main/java/net/dv8tion/jda/core/handle/GuildUpdateHandler.java
```diff
@@ -44,17 +44,16 @@ protected Long handleInternally(JSONObject content)
...
         Guild.NotificationLevel notificationLevel = Guild.NotificationLevel.fromKey(content.getInt("default_message_notifications"));
         Guild.MFALevel mfaLevel = Guild.MFALevel.fromKey(content.getInt("mfa_level"));
         Guild.ExplicitContentLevel explicitContentLevel = Guild.ExplicitContentLevel.fromKey(content.getInt("explicit_content_filter"));
         Guild.Timeout afkTimeout = Guild.Timeout.fromKey(content.getInt("afk_timeout"));
-        VoiceChannel afkChannel = !content.isNull("afk_channel_id")
-                ? guild.getVoiceChannelsMap().get(content.getLong("afk_channel_id"))
-                : null;
+        VoiceChannel afkChannel = content.isNull("afk_channel_id")
+                ? null : guild.getVoiceChannelsMap().get(content.getLong("afk_channel_id"));
         TextChannel systemChannel = !content.isNull("system_channel_id")
```
Almighty-Alpaca: As you've already removed the `!` in the three lines above could you also change it here?

src/main/java/net/dv8tion/jda/core/handle/MessageReactionBulkRemoveHandler.java
```diff
@@ -43,21 +40,10 @@ protected Long handleInternally(JSONObject content)
...
-        if (channel == null)
-        {
-            channel = api.getFakePrivateChannelMap().get(channelId);
-        }
-        if (channel == null)
         {
             api.getEventCache().cache(EventCache.Type.CHANNEL, channelId, () -> handle(responseNumber, allContent));
-            EventCache.LOG.debug("Received a reaction for a channel that JDA does not currently have cached");
+            EventCache.LOG.debug("Received a reaction for a channel that JDA does not currently have cached " +
+                "channel_id: " + channelId + " message_id: " + messageId);
```
Almighty-Alpaca: This should be adjusted to the new logging conventions.

src/main/java/net/dv8tion/jda/core/handle/MessageUpdateHandler.java
```diff
@@ -52,17 +52,17 @@ protected Long handleInternally(JSONObject content)
...
                 }
             }
             else
             {
-                //TODO: handle partial message update info. Example:
-                //From webhook/rich-embed.
-                //{"author":{"bot":true,"id":"233501884294365184","avatar":"27ae7496b3b30cddab2feaaed06d862a","username":"GitHub","discriminator":"0000"},"id":"234838126969880596","embeds":[{"color":15109472,"author":{"icon_url":"https://avatars.githubusercontent.com/u/2415829?v=3","name":"abalabahaha","proxy_icon_url":"https://images-ext-2.discordapp.net/eyJ1cmwiOiJodHRwczovL2F2YXRhcnMuZ2l0aHVidXNlcmNvbnRlbnQuY29tL3UvMjQxNTgyOT92PTMifQ.mMTBuOMUKYowcUU1H8Gzc7g4fFQ","url":"https://github.com/abalabahaha"},"description":"It was removed from the unofficial docs and other places because devs didn't want automated registration.","type":"rich","title":"[hammerandchisel/discord-api-docs] New comment on issue #148: Register API call","url":"https://github.com/hammerandchisel/discord-api-docs/issues/148#issuecomment-252524279"}],"channel_id":"168311874624946176"}
+//                WebSocketClient.LOG.warn("Received MESSAGE_UPDATE with no specified \"type\" field. JSON: " + content);
```
Almighty-Alpaca: Same as above.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
MinnDevelopment: The mention types are used in order to check if a user was mentioned by role for instance

src/main/java/net/dv8tion/jda/core/handle/MessageUpdateHandler.java
MinnDevelopment: This is commented out as it isn't going to be logged.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
Almighty-Alpaca: Oh, that makes sense I guess.

src/main/java/net/dv8tion/jda/core/handle/MessageUpdateHandler.java
Almighty-Alpaca: :man_facepalming:

kantenkugel:
Just wanna note that @TheMonitorLizard would appreciate it when he gets a heads-up when the merge is ready (might not matter anyway as it seems like he is waiting for shardmanager as well)

src/main/java/net/dv8tion/jda/core/handle/MessageReactionHandler.java
```diff
@@ -81,14 +81,14 @@ protected Long handleInternally(JSONObject content)
...
-            channel = api.asClient().getGroupById(channelId);
-        }
         if (channel == null)
         {
             channel = api.getFakePrivateChannelMap().get(channelId);
         }
+        if (channel == null && api.getAccountType() == AccountType.CLIENT)
+        {
+            channel = api.asClient().getGroupById(channelId);
+        }
```
DV8FromTheWorld: Why was this moved down?

src/main/java/net/dv8tion/jda/core/hooks/ListenerAdapter.java
```diff
@@ -651,13 +650,14 @@ else if (event instanceof GenericVoiceChannelEvent)
...
         else if (event instanceof GenericCategoryEvent)
             onGenericCategory((GenericCategoryEvent) event);
-        else if (event instanceof GenericGuildEvent)
-            onGenericGuild((GenericGuildEvent) event);
         else if (event instanceof GenericRoleEvent)
             onGenericRole((GenericRoleEvent) event);
         else if (event instanceof GenericEmoteEvent)
             onGenericEmote((GenericEmoteEvent) event);
 
+        if (event instanceof GenericGuildEvent)
```
DV8FromTheWorld: put a comment here specifying why this is by itself (generic events that have 2 levels of generic subclasses)

src/main/java/net/dv8tion/jda/core/hooks/ListenerAdapter.java
```diff
@@ -751,6 +749,11 @@ else if (event instanceof GenericCallVoiceEvent)
                 onGenericCallVoice((GenericCallVoiceEvent) event);
 
             //Client Only Generic Events
+            //Subclass of GenericGroupEvent must be in different if/else block
+            // cannot be in block above due to GenericGroupMessageReactionEvent being a child
+            if (event instanceof GenericGroupMessageEvent)
+                onGenericGroupMessage((GenericGroupMessageEvent) event);
+
             if (event instanceof GenericRelationshipEvent)
```
DV8FromTheWorld: should this be an `else if` ?

src/main/java/net/dv8tion/jda/core/MessageBuilder.java
```diff
@@ -703,15 +850,15 @@ public int lastIndexOf(CharSequence target, int fromIndex, int endIndex)
...
+     *
      * @return the created {@link net.dv8tion.jda.core.entities.Message Messages}
      */
     public Queue<Message> buildAll(SplitPolicy... policy)
     {
         if (this.isEmpty())
             throw new UnsupportedOperationException("Cannot build a Message with no content. (You never added any content to the message)");
 
-        LinkedList<Message> messages = new LinkedList<Message>();
+        LinkedList<Message> messages = new LinkedList<>();
```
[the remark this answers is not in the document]
MinnDevelopment: DataMessage is just an implementation of Message, we do return a queue of DataMessage but we are hiding the implementation detail from the end user.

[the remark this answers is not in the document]
DV8FromTheWorld: I know, I am saying that since `build()` returns a `DataMessage`, then we should stay consistent in what we return.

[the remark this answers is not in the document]
kantenkugel: build still returns Message from what i have seen... (method signature)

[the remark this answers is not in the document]
DV8FromTheWorld: You're right. I'm not sure why, but I thought I saw it returning `DataMessage`. Ignore this.

src/main/java/net/dv8tion/jda/core/entities/Message.java
```diff
@@ -64,56 +97,157 @@
...
+    boolean isMentioned(IMentionable mentionable, MentionType... types);
+
     /**
      * Indicates if this Message mentions everyone using @everyone or @here.
-     * In {@link net.dv8tion.jda.core.entities.PrivateChannel PrivateChannel's}, this always returns false.
      *
-     * @return True if message is mentioning everyone
+     * @return True, if message is mentioning everyone
      */
     boolean mentionsEveryone();
```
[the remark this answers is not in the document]
kantenkugel: Not so sure about that...
a) that would def. break backwards compatibility (with no prior deprecation)
b) the alternative namings i can come up with are `isMentioningEveryone` and `isEveryoneMention`, which IMO both don't really sound better.

I do get your point of starting boolean getters with `is` tho (convention)

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
```diff
@@ -79,7 +73,7 @@
      * <p>This value is updated on each {@link net.dv8tion.jda.core.events.message.MessageReceivedEvent MessageReceivedEvent}
      * and <u><b>will be reset to {@code null} if the message associated with this ID gets deleted</b></u>
      *
-     * @throws java.lang.IllegalStateException
+     * @throws java.util.NoSuchElementException
```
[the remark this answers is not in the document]
MinnDevelopment: The documentation for `NoSuchElementException` suggests that it can be used outside of collections.

>Thrown by various accessor methods to indicate that the element being requested does not exist.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
MinnDevelopment: I disagree

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
MinnDevelopment: Good question, don't remember

src/main/java/net/dv8tion/jda/core/handle/MessageReactionHandler.java
MinnDevelopment: Because I don't see a logical difference and I suspect CLIENT to be the least likely case.

src/main/java/net/dv8tion/jda/core/hooks/ListenerAdapter.java
MinnDevelopment: GenericGroupMessageEvent is a subclass of GenericGroupEvent which is handled in the following if/else block (see comment above the if)

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
```diff
@@ -0,0 +1,815 @@
...
+        {
+            final Member member = (Member) mentionable;
+            if (CollectionUtils.containsAny(getMentionedRoles(), member.getRoles()))
+                return true;
+        }
+        else if (isFromType(ChannelType.TEXT) && mentionable instanceof User)
+        {
+            final Member member = getGuild().getMember((User) mentionable);
+            if (CollectionUtils.containsAny(getMentionedRoles(), member.getRoles()))
+                return true;
```
kantenkugel: This might throw a NPE if the user given is not present in the guild the message was sent in

src/main/java/net/dv8tion/jda/core/entities/Message.java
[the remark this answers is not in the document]
DV8FromTheWorld: I ran into the same problem, but was curious if anyone else might have a better idea for naming.

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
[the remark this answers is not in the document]
DV8FromTheWorld: Yes, but like I said, this isn't an element, it is an attribute.
Either way, the use of "Element" makes one think of an Iterator or Collection, which is why I believe IllegalStateException is better, especially since it quite literally is a state. A channel could have had a LastMessage a moment ago, but the state changed when the last message in the channel was deleted.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
```diff
@@ -531,32 +531,25 @@ public String getContentDisplay()
...
-                    String name;
-                    if (getGuild().isMember(user))
-                        name = getGuild().getMember(user).getEffectiveName();
-                    else name = user.getName();
-                    tmp = tmp.replace("<@" + user.getId() + '>', '@' + name)
-                            .replace("<@!" + user.getId() + '>', '@' + name);
-                }
+                    name = user.getName();
+                tmp = tmp.replace("<@" + user.getId() + '>', '@' + name)
+                         .replace("<@!" + user.getId() + '>', '@' + name);
```
DV8FromTheWorld: Lets replace this double call to `replace` with a single call to `replaceAll` using a regex that covers both a nickname and non-nickname mention (make the `!` optional)

src/main/java/net/dv8tion/jda/core/entities/Message.java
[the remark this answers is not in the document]
MinnDevelopment: Can't think of a good name but I understand why you bring it up

MinnDevelopment:
@DV8FromTheWorld can you give your view on this PR? Some people have been using this branch and seem to have no issues

src/main/java/net/dv8tion/jda/core/entities/Message.java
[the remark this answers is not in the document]
MinnDevelopment: A poll in #experimental concluded to keeping the current name

src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -302,7 +301,7 @@ public StringBuilder getDescriptionBuilder()
      *
      * @return the builder after the description has been set
      */
-    public EmbedBuilder setDescription(CharSequence description)
+    public final EmbedBuilder setDescription(CharSequence description)
```
MinnDevelopment: Why was this needed?

src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -169,8 +169,7 @@ public int length()
         int length = description.length();
         synchronized (fields)
         {
-            for (MessageEmbed.Field f : fields)
-                length += f.getName().length() + f.getValue().length();
+            length = fields.stream().map(f -> f.getName().length() + f.getValue().length()).reduce(length, Integer::sum);
```
MinnDevelopment: Why was this needed?

src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
jagrosh: It's being called in the constructor, so it shouldn't be overridable

MinnDevelopment: But the overridden method would then be called which would be the intention of the override, no?

PR #457, opened 2017-09-06
Changed audio connection requests

MinnDevelopment:
Now all requests will work in the same system
- Connecting to a VoiceChannel
- Re-connecting to a VoiceChannel
- Disconnecting from a VoiceChannel

This should counter some unwanted race-conditions with the old system.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/handle/VoiceStateUpdateHandler.java
+++ b/src/main/java/net/dv8tion/jda/core/handle/VoiceStateUpdateHandler.java
@@ -124,6 +124,8 @@ public class VoiceStateUpdateHandler extends SocketHandler
             else if (channel == null)
             {
                 oldChannel.getConnectedMembersMap().remove(userId);
+                if (guild.getSelfMember().equals(member))
+                    api.getClient().updateAudioConnection(guildId, null);
                 api.getEventManager().handle(
                         new GuildVoiceLeaveEvent(
                                 api, responseNumber,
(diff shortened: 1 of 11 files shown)


src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -288,46 +289,44 @@ else if (audioRequest != null)
...
                         if (!needRatelimit)
                         {
-                            //If the connection is already established, then the packet just sent
-                            // was a move channel packet, thus, it won't trigger the removal from
-                            // queuedAudioConnections in VoiceServerUpdateHandler because we won't receive
-                            // that event just for a move, so we remove it here after successfully sending.
-                            if (audioManager.isConnected() && stage == ConnectionStage.CONNECT
-                                || stage == ConnectionStage.DISCONNECT)
+                            //If we didn't get RateLimited, Next allowed connect request will be 2 seconds from now
+                            audioRequest.setNextAttemptEpoch(System.currentTimeMillis() + 2000);
```
DV8FromTheWorld: Will moving this here effect the Audio Resume system? @MinnDevelopment

MinnDevelopment: No audio resume does not interact with the mWS in any way. However this will heavily affect the RECONNECT system of audio connections meaning it will send DISCONNECT then wait 2 seconds and then eventually send CONNECT which I don't agree is good. I recommend reverting this change to only be in effect for CONNECT.

src/main/java/net/dv8tion/jda/core/handle/VoiceStateUpdateHandler.java
```diff
@@ -144,9 +146,10 @@ else if (channel == null)
                     //If we have connected (VOICE_SERVER_UPDATE received and AudioConnection created (actual connection might still be setting up)),
                     // then we need to stop sending audioOpen/Move requests through the MainWS if the channel
                     // we have just joined / moved to is the same as the currently queued audioRequest
-                    // (handled by removeAudioConnection
+                    // (handled by updateAudioConnection)
                     if (mng.isConnected())
-                        api.getClient().removeAudioConnection(guildId, channel);
+                        api.getClient().updateAudioConnection(guildId, channel);
+                    //If we are not already connected this will be removed by VOICE_SERVER_UPDATE
```
DV8FromTheWorld: Technically this comment about VOICE_SERVER_UPDATE belongs in the above `if (oldChannel == null)` block

PR #458, opened 2017-09-07
Do not retain the audio managers after shutdown

Shredder121:
_DIsclaimer_ not yet tested.

Is there a reason we don't `clear()` this one, but other maps are cleared? (e.g. `opusDecoders`)


MinnDevelopment:
Why should we clear this map here?

DV8FromTheWorld:
Seems like a pointless change ¯\_(ツ)_/¯

Shredder121:
I merely noticed an abnormal amount of AudioManagers.
And also noticed that all other maps that were `foreach(*::close)`ed were also `clear()`ed.

There's a cycle in the references, which usually is no problem, but with the cycle being so big you can quickly lose gigabytes of RAM.
I merely recommend breaking it up this way.

PR #459, opened 2017-09-09
Step 0 busywork for category channels implementation

Avarel:
This is a groundwork implementation of category channels. __**It is not feature-complete, however it does currently resolves the warnings.**__
* New interface: `CategoryChannel` and its implementation `CategoryChannelImpl`.
  * `AbstractChannelImpl` were implemented by `CategoryChannelImpl`.
* JDA(Impl) and Guild(Impl) gained a new set of methods for categories (partially tested).
* New events relating to categories (partially tested).
* Updated `EntityBuilder.java` to build categories.

However, there are several problems that should be delegated to staff of JDA.
* Category channels JSON content and creation happens **after** text and voice channels so far from my tests.
* TextChannel and VoiceChannel has a nullable `parent_id` field that points to the category it belongs to.
* CategoryChannel has a `getTextChannel()` method but due to the previous two problems, it currently returns an empty list because I'm not sure what is the best way to add child text and voice channels to it (todo for the staff).
* Message also has a new `getCategoryChannel()` but it currently returns null as a limitation those 3 problems.
* TextChannelUpdatePositionEvent and VoiceChannelUpdatePositionEvent will also have to update to handle moving to categories.

Just wanted to help out <3

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/JDA.java
+++ b/src/main/java/net/dv8tion/jda/core/JDA.java
@@ -452,6 +452,71 @@ public interface JDA
     List<Role> getRolesByName(String name, boolean ignoreCase);
 
     /**
+     * An unmodifiable List of all {@link net.dv8tion.jda.core.entities.CategoryChannel CategoryChannels} of all connected
+     * {@link net.dv8tion.jda.core.entities.Guild Guilds}.
+     *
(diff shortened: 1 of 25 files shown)


MinnDevelopment:
This will be worked on in #460 
I had to implement this by myself instead of reviewing.

PR #460, opened 2017-09-09
Bootleg fix implementation for categories

MinnDevelopment:
- Missing docs
- Missing more research

Revisit after documentation update

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/impl/CategoryImpl.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/impl/CategoryImpl.java
@@ -1,0 +1,105 @@
+package net.dv8tion.jda.core.entities.impl;
+
+import gnu.trove.map.TLongObjectMap;
+import net.dv8tion.jda.core.entities.*;
+import net.dv8tion.jda.core.requests.RestAction;
+import net.dv8tion.jda.core.requests.restaction.AuditableRestAction;
+import net.dv8tion.jda.core.requests.restaction.InviteAction;
+import net.dv8tion.jda.core.utils.Checks;
+import net.dv8tion.jda.core.utils.MiscUtil;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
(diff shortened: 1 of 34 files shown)


src/main/java/net/dv8tion/jda/core/entities/impl/CategoryImpl.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public InviteAction createInvite()
+    {
+        throw new UnsupportedOperationException("Cannot create invites for category!");
+    }
+
+    @Override
+    public RestAction<List<Invite>> getInvites()
+    {
+        return new AuditableRestAction.EmptyRestAction<>(getJDA(), Collections.emptyList());
```
Almighty-Alpaca: When you can't create invites for categories, are you sure getting them works?

src/main/java/net/dv8tion/jda/core/events/channel/category/GenericCategoryEvent.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public Category getCategory()
+    {
+        return category;
+    }
+
+    public String getId()
+    {
+        return Long.toUnsignedString(getIdLong());
```
Almighty-Alpaca: why don't just use `catergory.getId()` ?

src/main/java/net/dv8tion/jda/core/entities/impl/CategoryImpl.java
MinnDevelopment: It is an empty rest action thus it is always an empty list.

src/main/java/net/dv8tion/jda/core/events/channel/category/GenericCategoryEvent.java
MinnDevelopment: Originally I only had the ids but then i realized that all events have a Category object anyway.

src/main/java/net/dv8tion/jda/core/entities/impl/CategoryImpl.java
Almighty-Alpaca: right...

PR #462, opened 2017-09-10
Updated Permission class to mirror the new system with categories

PhotonBursted:
Changed:
 - MESSAGE_READ is no longer a guild permission
 - VIEW_CHANNEL is a new permission for voice channels
 - READ_SEE_CHANNELS is a new permission for categories and guilds
All of these have offset 10 but have different purposes.
 - Permission#isVoice adjusted to include offset 10

Added:
 - ALL_CATEGORY_PERMISSIONS field and Permission#isCategory method

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/Permission.java
+++ b/src/main/java/net/dv8tion/jda/core/Permission.java
@@ -38,7 +38,7 @@ public enum Permission
     VIEW_AUDIT_LOGS(7, true, false, "View Audit Logs"),
 
     // Text Permissions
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
They actually used MESSAGE_READ for these permissions as well. It is after-all the same bit.
Will think about it.

MinnDevelopment:
This has been addressed in 004ce0b. Thank you for bringing attention to this change I was not aware that offset 10 was no also applicable to VoiceChannels, fortunately I predicted this in the permission rewrite and PermissionUtil was already able to handle this case.

PR #464, opened 2017-09-13
Add MessageChannel#editMessageById(long messageId, String newContent)

BreadMoirai:
Feels like this one is missing.


DV8FromTheWorld:
Pretty sure this is fine.

MinnDevelopment:
This has already been added in #453 and would cause conflicts. If this is an important change I'll consider merging it, otherwise I don't see why we should rush it and cause unwanted conflicts.

PR #465, opened 2017-09-13
Introducing CacheView 

MinnDevelopment:
The newly added `SnowflakeCacheView` and similar are read-only views of internal JDA cache allowing to retrieve information about size without an overhead of O(n) for creating a copy.
In addition this will now add a central place for all convenience methods such as `getXByName` or `getXById` in these implementations.

Instead of doing `jda.getUsers().size()` O(n) you can now do `jda.getUserCache().size()` O(1) to know exactly how many users are cached without creating a whole list copy first.


MinnDevelopment:
To better illustrate the structure and hierarchy of this implementation I created a class diagram:
image

src/main/java/net/dv8tion/jda/core/entities/impl/GuildImpl.java
```diff
@@ -53,12 +56,18 @@
...
-    private final TLongObjectMap<TextChannel> textChannels = MiscUtil.newLongMap();
-    private final TLongObjectMap<VoiceChannel> voiceChannels = MiscUtil.newLongMap();
-    private final TLongObjectMap<Member> members = MiscUtil.newLongMap();
-    private final TLongObjectMap<Role> roles = MiscUtil.newLongMap();
-    private final TLongObjectMap<Emote> emotes = MiscUtil.newLongMap();
+//    private final TLongObjectMap<TextChannel> textChannels = MiscUtil.newLongMap();
+//    private final TLongObjectMap<VoiceChannel> voiceChannels = MiscUtil.newLongMap();
+//    private final TLongObjectMap<Member> members = MiscUtil.newLongMap();
+//    private final TLongObjectMap<Role> roles = MiscUtil.newLongMap();
+//    private final TLongObjectMap<Emote> emotes = MiscUtil.newLongMap();
```
DV8FromTheWorld: Remove this unneeded commented out code

MinnDevelopment:
Here is an updated class diagram with the new terminology
image

PR #469, opened 2017-09-16
Switched to SLF4J for logging

MinnDevelopment:
This change will bump the version to 3.3.1

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/utils/SimpleLog.java
+++ b/src/main/java/net/dv8tion/jda/core/utils/SimpleLog.java
@@ -15,6 +15,7 @@
  */
 package net.dv8tion.jda.core.utils;
 
+import net.dv8tion.jda.core.JDA;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.event.Level;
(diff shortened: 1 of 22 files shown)


Shredder121:
Is it possible to add some more methods that accept format arguments?
So when TRACE is not enabled, there is no string concatenation done?
(When all events come in, the message is concatenated, but not actually printed, it will save resources and be faster to only do it on demand)

Also, (Simple)DateFormat isn't thread safe.
(I know it might not be possible to change it in this PR, but I'd recommend fixing that)

MinnDevelopment:
This is the first iteration of completely replacing logging.
In this step we are adding first support for SLF4J and in the second step we will be removing SimpleLog completely in favor of the Logger interface. In that step we can make use of the SLF4J formatting methods.

Shredder121:
👍 Just wanted to make sure, carry on.

MinnDevelopment:
This will be merged once categories have been released on the stable discord client to be in sync with the stable category release when bumping the version.

src/main/java/net/dv8tion/jda/core/utils/SimpleLog.java
```diff
@@ -290,53 +105,113 @@ public Level getLevel()
...
+        {
+            if (logger.isTraceEnabled())
+                return Level.TRACE;
+            if (logger.isDebugEnabled())
+                return Level.DEBUG;
+            if (logger.isInfoEnabled())
+                return Level.INFO;
+            if (logger.isWarnEnabled())
+                return Level.WARN;
+            return Level.ERROR;
```
DV8FromTheWorld: is there no OFF level?

MinnDevelopment: No there is not

PR #473, opened 2017-09-22
+ TextChannelUpdateParentEvent to ListenerAdapter

Scarsz:
Was missing previously so wasn't able to be overridden via ListenerAdapter subclasses


src/main/java/net/dv8tion/jda/core/hooks/ListenerAdapter.java
```diff
@@ -174,6 +174,7 @@ public void onTextChannelUpdateTopic(TextChannelUpdateTopicEvent event) {}
     public void onTextChannelUpdatePosition(TextChannelUpdatePositionEvent event) {}
     public void onTextChannelUpdatePermissions(TextChannelUpdatePermissionsEvent event) {}
     public void onTextChannelUpdateNSFW(TextChannelUpdateNSFWEvent event) {}
+    public void onTextChannelUpdateParentEvent(TextChannelUpdateParentEvent event) {}
```
MinnDevelopment: This should be `onTextChannelUpdateParent` to be consistent with the rest of this class.

Scarsz: whoops

PR #479, opened 2017-09-23
Adding ShardedRateLimiter to keep track of global REST rate limit

MinnDevelopment:
This change will not require users to add a ShardedRateLimiter to their JDABuilder if they have already been using the same builder for all shards.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/JDABuilder.java
@@ -29,10 +29,7 @@ import net.dv8tion.jda.core.utils.Checks;
 import okhttp3.OkHttpClient;
 
 import javax.security.auth.login.LoginException;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.*;
 
 /**
  * Used to create new {@link net.dv8tion.jda.core.JDA} instances. This is also useful for making sure all of
@@ -138,7 +135,8 @@ public class JDABuilder
      * or {@link net.dv8tion.jda.core.JDABuilder#buildBlocking() buildBlocking()}
      * is called.
      *
-     * <p><u><b>This will reset the prior provided {@link #setShardedRateLimiter(ShardedRateLimiter)} setting!</b></u>
+     * <p><u><b>This will reset the prior provided {@link #setShardedRateLimiter(ShardedRateLimiter)} setting
+     * if this token is different to the previously set token!</b></u>
      *
      * <p>For {@link net.dv8tion.jda.core.AccountType#BOT} accounts:
      * <ol>
@@ -164,12 +162,12 @@ public class JDABuilder
      */
     public JDABuilder setToken(String token)
     {
-        this.token = token;
         //Share ratelimit for the same token
         // when this builder is used to build different accounts this makes sure we don't use the
         // same ratelimiter on them as it would be inaccurate
-        if (accountType == AccountType.BOT)
+        if (accountType == AccountType.BOT && !Objects.equals(token, this.token))
             shardRateLimiter = new ShardedRateLimiter();
+        this.token = token;
         return this;
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -50,6 +50,7 @@
     protected final List<Object> listeners;
 
     protected SessionReconnectQueue reconnectQueue = null;
+    protected ShardedRateLimiter shardRateLimiter = new ShardedRateLimiter();
```
Almighty-Alpaca: I would 
1) add a check to ShardedRateLimiter to prevent usage across multiple accounts 
2) initialize this lazily (like httpClientBuilder) to prevent having the same ShardedRateLimiter across multiple accounts when using the same JDABuilder

MinnDevelopment: I could set it when the setToken method is invoked

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -104,11 +105,41 @@ public JDABuilder setReconnectQueue(SessionReconnectQueue queue)
...
+    }
+
+    /**
      * Sets the token that will be used by the {@link net.dv8tion.jda.core.JDA} instance to log in when
      * {@link net.dv8tion.jda.core.JDABuilder#buildAsync() buildAsync()}
      * or {@link net.dv8tion.jda.core.JDABuilder#buildBlocking() buildBlocking()}
      * is called.
      *
+     * <p><u><b>This will reset the prior provided {@link #setShardedRateLimiter(ShardedRateLimiter)} setting!</b></u>
+     *
```
Shredder121: Whoa! This is potentially dangerous.
I suggest a `shardedRateLimiterSet` boolean to account for user code calling that first, then `setToken`, which resets the rate limiter.
Or is this for if you reuse the builder for multiple tokens? 🤔😲

Anyway, if all else fails I would at least detect if users call `setShardedRateLimiter` and then `setToken` and log that, to notify that we're setting up a new rate limiter.

(setters with side effects are tricky, because principle-of-least-astonishment)

MinnDevelopment: This isn't dangerous, there is a reason this warning is bold-underlined in the documentation. Most users will not even use this setter in the first place.

Shredder121: I'm just saying, the few people who will use it *better* not do it in the wrong order.
Since it will ignore the rate limiter that they've set.

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -104,11 +105,41 @@ public JDABuilder setReconnectQueue(SessionReconnectQueue queue)
...
+     * <p>When you construct multiple JDABuilder instances to build shards it is recommended to use the same ShardedRateLimiter on
+     * all of them. But it is to be <u>avoided</u> to use the same ShardedRateLimiter for different accounts/tokens!
+     *
+     * @param  rateLimiter
+     *         ShardedRateLimiter used to keep track of cross-session rate limits
+     *
+     * @return The {@link net.dv8tion.jda.core.JDABuilder JDABuilder} instance. Useful for chaining.
+     */
+    public JDABuilder setShardedRateLimiter(ShardedRateLimiter rateLimiter)
```
DV8FromTheWorld: what if someone, themselves, provide `null` in a BOT situation?

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -134,6 +165,11 @@ public JDABuilder setReconnectQueue(SessionReconnectQueue queue)
     public JDABuilder setToken(String token)
     {
         this.token = token;
+        //Share ratelimit for the same token
+        // when this builder is used to build different accounts this makes sure we don't use the
+        // same ratelimiter on them as it would be inaccurate
+        if (accountType == AccountType.BOT)
+            shardRateLimiter = new ShardedRateLimiter();
```
DV8FromTheWorld: Check if the previous token is the same as the new one before deciding to change. If they are the same, you shouldn't be making a new ratelimiter.

src/main/java/net/dv8tion/jda/core/JDABuilder.java
MinnDevelopment: `this.shardRateLimiter = rateLimiter == null ? new ShardedRateLimiter() : rateLimiter;`

PR #482, opened 2017-09-24
Fix the presence change

WalshyDev:
Make sure the setPresence call has the right parameters so that it does update correctly :)

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/managers/impl/PresenceImpl.java
+++ b/src/main/java/net/dv8tion/jda/core/managers/impl/PresenceImpl.java
@@ -86,7 +86,7 @@ public class PresenceImpl implements Presence
     @Override
     public void setStatus(OnlineStatus status)
     {
-        setPresence(game, stats);
+        setPresence(game, status);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/core/managers/impl/PresenceImpl.java
```diff
@@ -86,19 +86,19 @@ public boolean isIdle()
     @Override
     public void setStatus(OnlineStatus status)
     {
-        setPresence(game, idle);
+        setPresence(game, stats);
```
binaryoverload: Woopsie @WalshyDev I think this is meant to be "status" not "stats"

kantenkugel:
code for reference:

https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/core/managers/impl/PresenceImpl.java#L92-L96

binaryoverload:
#483 @WalshyDev You got ninjad

WalshyDev:
@binaryoverload is it a ninja if that one is older? Also that doesn't fix the setStatus ;)

binaryoverload:
@WalshyDev It's not older xD He opened it after yours

kantenkugel:
Whoops... thought this was a Issue at first
Gonna drop my PR then...
Can you fix the setStatus one to use the 3-arg setStatus?

PR #490, opened 2017-09-30
First pass on GuildAction

MinnDevelopment:
I've not finished the documentation on these changes but I thought it might be good to let others review the logic while I finish documenting.


src/main/java/net/dv8tion/jda/core/requests/restaction/PermOverrideData.java
```diff
@@ -0,0 +1,49 @@
...
+
+package net.dv8tion.jda.core.requests.restaction;
+
+import org.json.JSONObject;
+import org.json.JSONString;
+
+class PermOverrideData implements JSONString
+{
+    public static final int ROLE_TYPE = 0;
+    public static final int MEMBER_TYPE = 1;
```
meowingtwurtle: Why the magic numbers? Could this be an enum?

MinnDevelopment: Those are actual types used in the api, putting them in an enum doesn't make sense here since it's entirely internal and requires method calls to retrieve the int from the enum constant: `OverrideType.ROLE.getKey()`

DV8FromTheWorld: As a note, they aren't really magic numbers since they are specified by named constants.

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
```diff
@@ -546,6 +547,22 @@ public void removeEventListener(Object... listeners)
         return Collections.unmodifiableList(eventManager.getRegisteredListeners());
     }
 
+    @Override
+    public GuildAction createGuild(String name)
+    {
+        switch (accountType)
+        {
+            case BOT:
+                if (guildCache.size() > 10)
```
DV8FromTheWorld: Shouldn't this be `>= 10` ?

src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
```diff
@@ -190,6 +186,10 @@ public ChannelAction addPermissionOverride(Role role, Collection<Permission> all
     {
         checkPermissions(allow);
         checkPermissions(deny);
+        if (allow != null)
+            Checks.noneNull(allow, "Granted Permissions");
+        if (deny != null)
+            Checks.noneNull(deny, "Denied Permissions");
```
DV8FromTheWorld: Both of these `noneNull` checks are already being handled by `checkPermissions`, no?

src/main/java/net/dv8tion/jda/core/requests/restaction/GuildAction.java
```diff
@@ -0,0 +1,733 @@
...
+     * @throws java.lang.IndexOutOfBoundsException
+     *         If the provided index is not in bounds
+     *
+     * @return The current GuildAction for chaining convenience
+     */
+    @CheckReturnValue
+    public ChannelData getChannel(int index)
+    {
+        return this.channels.get(index);
+    }
```
DV8FromTheWorld: If you are supporting `addChannel` and `getChannel`, probably need to support `removeChannel`

src/main/java/net/dv8tion/jda/core/requests/restaction/GuildAction.java
```diff
@@ -0,0 +1,733 @@
...
+     * Creates and add a new {@link net.dv8tion.jda.core.requests.restaction.GuildAction.RoleData RoleData} object
+     * representing a Role for the resulting Guild.
+     *
+     * <p>This can be used in {@link GuildAction.ChannelData#addPermissionOverride(GuildAction.RoleData, long, long) ChannelData.addPermissionOverride(...)}.
+     * <br>You may change any properties of this {@link net.dv8tion.jda.core.requests.restaction.GuildAction.RoleData RoleData} instance!
+     *
+     * @return RoleData for the new Role
+     */
+    @CheckReturnValue
+    public RoleData newRole()
```
DV8FromTheWorld: The difference between how one adds a new role vs a new channel feels odd.

Consider, at minimum, providing a `newChannel(ChannelType, name: String)`

src/main/java/net/dv8tion/jda/core/requests/restaction/RoleAction.java
```diff
@@ -43,7 +43,7 @@
 {
 
     protected final Guild guild;
-    protected long permissions = 0;
+    protected Long permissions;
```
DV8FromTheWorld: I don't understand the decision behind this change.
Additionally, there is no way to `null`-ify this value after it has been set.

MinnDevelopment: It makes a difference whether you send permissions 0 or not at all. I made this nullable to allow using the default values (everyone role permissions)

DV8FromTheWorld: Then you need to make it possible to actually set the Permissions to null then. You didn't modify any of the setters to support this.

src/main/java/net/dv8tion/jda/core/requests/restaction/PermOverrideData.java
meowingtwurtle: Fair enough

MinnDevelopment:
I will resolve conflicts and finish this Pull Request later today.

PR #492, opened 2017-10-01
SLF4J Logger rewrite part2

kantenkugel:
Completely removed SimpleLog in favor of a SLF4J Logger.
- Now using the Logger from slf4j-simple as fallback.
- new class JDALogger acts as JDA's Logger factory. It either returns the Logger via LoggerFactory if slf4j binding is present or creates a SimpleLogger instance (and cachinges it).
- Replaced all String concatenations of loggers with the lazy construction using parametrized logger methods and `{}`
- Added `JDALogger#getLazyString(LAMBDATERM)` to lazily construct String from expression as slf4j doesn't support that natively yet


kantenkugel:
Would suggest waiting with merge until next version bump

napstr:
May I suggest to take advantage of slf4js methods and instead of separately logging stack traces, append them to the message they belong to?

The output should look exactly the same in terms of information content, with the benefit of making life easier on log aggregators which may get confused by the current behaviour of doing two log calls for a single issue (concrete example of when things go wrong: using the logback appender for sentry)

Example:
https://github.com/DV8FromTheWorld/JDA/blob/30174cfa66e1952d1e98b7ddc5eb7803f74c535f/src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java#L975-L985

Instead do:
```java
        catch (JSONException ex)
        {
            LOG.warn("Got an unexpected Json-parse error. Please redirect following message to the devs:\n\t{}\n\t{} -> {}",
                ex.getMessage(), type, content, ex);
        }
        catch (Exception ex)
        {
            LOG.error("Got an unexpected error. Please redirect following message to the devs:\n\t{} -> {}", type, content, ex);
        }
```

Shredder121:
The only issue is that if you use the `Throwable` method, the format arguments cannot be used.

kantenkugel:
And that (together with the fact that i want the json data seperated) is why i did it the way i did

napstr:
> The only issue is that if you use the Throwable method, the format arguments cannot be used.

That's not true.
https://www.slf4j.org/faq.html#paramException

napstr:
>  the fact that i want the json data seperated

What are the benefits of doing that? There is no information lost by attaching the stacktrace to the source of it.

Shredder121:
~~If you want it separated-but-together I can also recommend `MDC` to bind variables. (which you can then immediately remove if needed)~~
~~Or use String.format before it.~~
~~You can do a lot if you put everything in an `if (logger.is(Warn/Error)Enabled())` statement.~~
~~I do vote in favor of keeping the calls together, as "just" the exception logged is a shame for log aggregation tools, or continuity when viewing log files~~
*WOAH* TIL! Thanks.
(keeping draft for posterity)

src/main/java/net/dv8tion/jda/core/audio/AudioConnection.java
```diff
@@ -514,7 +515,7 @@ else if (sample < Short.MIN_VALUE)
                 }
                 catch (Exception e)
                 {
-                    LOG.fatal(e);
+                    LOG.error("Something happened!", e);
```
MinnDevelopment: Can we make the comments in this file a bit more descriptive like "Unexpected Exception occurred while X"

src/main/java/net/dv8tion/jda/core/entities/impl/GuildImpl.java
```diff
@@ -313,6 +313,7 @@ public Role getPublicRole()
     }
 
     @Override
+    @Deprecated
     public TextChannel getPublicChannel()
```
MinnDevelopment: Since this is going to be in 3.4 we will be removing this method anyway FYI

src/main/java/net/dv8tion/jda/core/audio/AudioConnection.java
kantenkugel: :) Yea i didn't know what exactly happened there

src/main/java/net/dv8tion/jda/core/entities/impl/GuildImpl.java
kantenkugel: just added it because gradle build complained about implementing a deprecated method

src/main/java/net/dv8tion/jda/core/audio/AudioConnection.java
kantenkugel: Its funny how you didn't mention the other 2 exceptions above :)

MinnDevelopment: I used the plural "comments" for a reason 👀

src/main/java/net/dv8tion/jda/core/requests/ratelimit/BotRateLimiter.java
```diff
@@ -198,11 +198,10 @@ private void updateBucket(Bucket bucket, Headers headers)
...
             if (!bucket.getRoute().equals("gateway")
                     && !bucket.getRoute().equals("users/@me")
-                    && Requester.LOG.getEffectiveLevel().ordinal() >= Level.DEBUG.ordinal())
+                    && Requester.LOG.isDebugEnabled())
             {
-                Requester.LOG.debug("Encountered issue with headers when updating a bucket"
-                                  + "\nRoute: " + bucket.getRoute()
-                                  + "\nHeaders: " + headers);
+                Requester.LOG.debug("Encountered issue with headers when updating a bucket\nRoute: {}\nHeaders: {}",
+                    bucket.getRoute(), headers);
```
kantenkugel: Ok if i remove the isDebugEnabled part

MinnDevelopment: Yes, doubt that check is needed.

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -938,7 +938,7 @@ protected void handleEvent(JSONObject raw)
 //        }
 
         JSONObject content = raw.getJSONObject("d");
```
napstr: Maybe move this into the try catch block below? I just got a JSONException on 286 from this line, missing a print of the faulty JSON.

```
[2017-10-18 12:30:31,889] [ DEBUG] [JDA RateLimit-Queue Pool - Thread 2] n.d.jda.core.requests.Requester: Received response with following cf-rays: [3afad2a07e74646f-FRA]
[2017-10-18 12:30:47,770] [ ERROR] [JDA [1 / 2] MainWS-ReadThread] n.d.j.core.requests.WebSocketClient: Encountered an Exception 
org.json.JSONException: JSONObject["d"] is not a JSONObject.
        at org.json.JSONObject.getJSONObject(JSONObject.java:641)
        at net.dv8tion.jda.core.requests.WebSocketClient.handleEvent(WebSocketClient.java:947)
        at net.dv8tion.jda.core.requests.WebSocketClient.onTextMessage(WebSocketClient.java:666)
        at com.neovisionaries.ws.client.ListenerManager.callOnTextMessage(ListenerManager.java:352)
        at com.neovisionaries.ws.client.ReadingThread.callOnTextMessage(ReadingThread.java:260)
        at com.neovisionaries.ws.client.ReadingThread.callOnTextMessage(ReadingThread.java:238)
        at com.neovisionaries.ws.client.ReadingThread.handleTextFrame(ReadingThread.java:963)
        at com.neovisionaries.ws.client.ReadingThread.handleFrame(ReadingThread.java:746)
        at com.neovisionaries.ws.client.ReadingThread.main(ReadingThread.java:108)
        at com.neovisionaries.ws.client.ReadingThread.runMain(ReadingThread.java:64)
        at com.neovisionaries.ws.client.WebSocketThread.run(WebSocketThread.java:45)
[2017-10-18 12:31:21,579] [ DEBUG] [fredboat.FredBoat.main()] n.d.jda.core.requests.Requester: Received response with following cf-rays: [3afad3d719096361-FRA]
```

kantenkugel: Not sure if this is the right PR to do that in

kantenkugel: Seing as this is already fixed in another pr, not gonna implement here

MinnDevelopment: That cannot be moved into the try/catch block because the json is used within the catch

napstr: Aight. My intent on suggesting that was less concerned with the current bug, more of a general idea to improve future logging.
And Minn is correct, the change is more complicated than moving the two lines into the try catch, I'm sure whoever decides to add this in here or in another PR will figure out something appropriate. Maybe look at other similar places for possible uncaught JSONExceptions and have PR for that after this is merged.

src/main/java/net/dv8tion/jda/core/utils/package-info.java
```diff
@@ -18,7 +18,7 @@
  * Package which contains all utilities for the JDA library.
  * These are used by JDA itself and can also be useful for the library consumer!
  *
- * <p>This contains the JDA logger implementation {@link net.dv8tion.jda.core.utils.SimpleLog SimpleLog}
+ * <p>This contains the JDA-specific Logger factory {@link net.dv8tion.jda.core.utils.JDALogger JDALogger}
  * <br>Which is currently planned to be rewritten to be more end-user friendly.
```
Almighty-Alpaca: This is finished I guess

src/main/java/net/dv8tion/jda/core/utils/JDALogger.java
```diff
@@ -0,0 +1,152 @@
...
+         *
+         * @throws Exception
+         *         To allow lazy evaluation of methods that might throw exceptions
+         *
+         * @return The String for log message
+         */
+        String getString() throws Exception;
+    }
+
+    private JDALogger() {
```
Almighty-Alpaca: Shouldn't the inner class be at the end?

PR #494, opened 2017-10-01
Added handling for relationship presence updates

MinnDevelopment:


src/main/java/net/dv8tion/jda/core/handle/PresenceUpdateHandler.java
```diff
@@ -90,31 +102,29 @@ protected Long handleInternally(JSONObject content)
...
-                gameUrl = ( content.getJSONObject("game").isNull("url") ? null : content.getJSONObject("game").get("url").toString() );
+                gameName = game.get("name").toString();
+                gameUrl = game.isNull("url") ? null : game.get("url").toString();
                 try
                 {
-                    type = content.getJSONObject("game").isNull("type")
+                    type = game.isNull("type")
                             ? Game.GameType.DEFAULT
-                            : Game.GameType.fromKey(Integer.parseInt(content.getJSONObject("game").get("type").toString()));
+                            : Game.GameType.fromKey(Integer.parseInt(game.get("type").toString()));
```
DV8FromTheWorld: Wouldn't `game.getInt("type")` be the same here?

src/main/java/net/dv8tion/jda/core/handle/PresenceUpdateHandler.java
```diff
@@ -158,7 +168,37 @@ protected Long handleInternally(JSONObject content)
...
                 //In this case, this PRESENCE_UPDATE is for a Relation.
+                if (api.getAccountType() != AccountType.CLIENT)
+                    return null;
+                JDAClient client = api.asClient();
+                FriendImpl friend = (FriendImpl) client.getFriendById(userId);
 
+                if (friend == null)
+                {
+                    if (status != OnlineStatus.OFFLINE)
+                        return null;
```
DV8FromTheWorld: What is the point of this?

src/main/java/net/dv8tion/jda/core/handle/PresenceUpdateHandler.java
```diff
@@ -177,15 +217,8 @@ protected Long handleInternally(JSONObject content)
                 return null;
 
             //If this was for a Guild, cache it in the Guild for later use in GUILD_MEMBER_ADD
-            if (content.has("guild_id"))
-            {
-                GuildImpl guild = (GuildImpl) api.getGuildById(content.getLong("guild_id"));
+            if (guild != null)
```
DV8FromTheWorld: The removal of the `//cache in relationship stuff` isn't correct here.
We need to be caching this presence for a very soon Group add, similar to how we are doing for the Guild presence caching.

src/main/java/net/dv8tion/jda/core/handle/PresenceUpdateHandler.java
MinnDevelopment: In the past we've encountered numerous issues with game objects due to them not being checked by the API, I didn't touch that logic at all here.

src/main/java/net/dv8tion/jda/core/handle/PresenceUpdateHandler.java
MinnDevelopment: I'm not convinced we should be caching relationship presence updates, groups don't send these.

src/main/java/net/dv8tion/jda/core/handle/PresenceUpdateHandler.java
MinnDevelopment: ugh, looks like I forgot to remove this after I decided not to cache these presence updates

MinnDevelopment:
@DV8FromTheWorld can I get an update on your view on this PR?

src/main/java/net/dv8tion/jda/core/handle/PresenceUpdateHandler.java
```diff
@@ -158,7 +168,32 @@ protected Long handleInternally(JSONObject content)
...
+                                user, null, oldStatus));
+                    }
+                    if (!Objects.equals(friend.getGame(), nextGame))
+                    {
+                        Game oldGame = friend.getGame();
+                        friend.setGame(nextGame);
+                        api.getEventManager().handle(
+                            new UserGameUpdateEvent(
+                                api, responseNumber,
+                                user, null, oldGame));
```
DV8FromTheWorld: We should be firing FriendXEvents, not UserXEvents I think.
In that same vain though, we probably should rename the User events to Member events?

DV8FromTheWorld: Or, at a minimum, take a Friend object to these User events so that you can call `getFriend` like you can call `getMember`.

MinnDevelopment: That is already supported.

src/main/java/net/dv8tion/jda/core/handle/PresenceUpdateHandler.java
```diff
@@ -208,11 +208,9 @@ protected Long handleInternally(JSONObject content)
 
             //If the OnlineStatus is OFFLINE, ignore the event and return.
             OnlineStatus status = OnlineStatus.fromKey(content.getString("status"));
-            if (status == OnlineStatus.OFFLINE)
```
DV8FromTheWorld: We lost the ability to return in an `OFFLINE` non-guild situation.
Is that correct handling?

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -1319,8 +1321,11 @@ private void setupHandlers()
         handlers.put("VOICE_STATE_UPDATE",          new VoiceStateUpdateHandler(api));
 
         // Unused events
-        handlers.put("CHANNEL_PINS_UPDATE",         new SocketHandler.NOPHandler(api));
-        handlers.put("WEBHOOKS_UPDATE",             new SocketHandler.NOPHandler(api));
+        final SocketHandler.NOPHandler nopHandler = new SocketHandler.NOPHandler(api);
```
DV8FromTheWorld: Technically, since `NOPHandler` will never be implemented, we should have it be a `public static final` instance in `SocketHandler` instead of in-line creating one here.

MinnDevelopment: The reason is that the SocketHandler constructor expects a JDA instance and I don't want to add confusing situations where that JDA instance isn't present/null

src/main/java/net/dv8tion/jda/core/handle/PresenceUpdateHandler.java
MinnDevelopment: Can you elaborate? Not sure what you mean

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
DV8FromTheWorld: Okay, I can get behind that.
However, lets create the final variable at the top of the method, not in the middle.

src/main/java/net/dv8tion/jda/core/handle/PresenceUpdateHandler.java
DV8FromTheWorld: Nevermind. I see what has changed here. The fast escape logic is no longer needed.

PR #499, opened 2017-10-02
Improvement to MessageEmbed static value docs

Sanduhr32:
This PR should improve the docs for some constant fields so that butler bots can display the value.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/MessageEmbed.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/MessageEmbed.java
@@ -32,17 +32,17 @@ public interface MessageEmbed
 {
     /**
      * The character limit of the maximum title length is {@value} chacacters.<br>
-     * The title is the line under the author. [Image - More](https://imgur.com/a/yOb5n).
+     * The title usually contains the topic of the description or {@link net.dv8tion.jda.core.entities.MessageEmbed.Field Field} value. [Image - More](https://imgur.com/a/yOb5n).
      */
     int TITLE_MAX_LENGTH = 256;
     /**
      * The character limit of the maximum value length is {@value} chacacters.<br>
-     * The value is the text in a field . [Image - More](https://imgur.com/a/yOb5n).
+     * The value representes the text in a {@link net.dv8tion.jda.core.entities.MessageEmbed.Field Field}. Its called field value too. [Image - More](https://imgur.com/a/yOb5n).
      */
     int VALUE_MAX_LENGTH = 1024;
     /**
      * The character limit of the maximum text length is {@value} chacacters.<br>
-     * The text is the text of the description. [Image - More](https://imgur.com/a/yOb5n).
+     * The text equals the description. [Image - More](https://imgur.com/a/yOb5n).
      */
     int TEXT_MAX_LENGTH = 2048;
     /**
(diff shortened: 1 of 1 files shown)


kantenkugel:
You can use something like `{@value}` iirc... that way it isn't hardcoded in the docs and gets updated automatically once we change the value

Sanduhr32:
will change!

kantenkugel:
Would also be nice if you could explain a bit what each field is instead of just saying its name (what is value for example?)

kantenkugel:
Did you test building javadocs with latest changes? is that the correct usage of `@value`?

Sanduhr32:
👀 tell me how so i do it next time before i try

kantenkugel:
https://i.imgur.com/7ZX56zQ.png

You can just run `gradlew build` in the console, that will do everything (and build jars)

kantenkugel:
You don't have to explain for babies btw^^ "The title is the line under the author". Just a few clarifications because "VALUE" alone doesn't tell much for example. its the value of fields

Sanduhr32:
whups yee forgot the markdown translation from jdocs into hljs in discord

kantenkugel:
you can also just cut out the whole "The character limit of the maximum YYY length is XXXX chacacters." part. Just say something along the lines of "This field denotes the maximum length of a field value. The fields value is XXX" (less robotic)

src/main/java/net/dv8tion/jda/core/entities/MessageEmbed.java
```diff
@@ -30,11 +30,37 @@
...
+    /**
+     * The character limit of the maximum text length is {@value} chacacters.<br>
+     * The text equals the description.<br>
+     * {@see https://imgur.com/a/yOb5n}
+     */
     int TEXT_MAX_LENGTH = 2048;
+    /**
+     * The character limit of the maximum url length is {@value} chacacters.<br>
+     * The url is the optional url markdown of the title.<br>
+     * {@see https://imgur.com/a/yOb5n}
```
MinnDevelopment: The `@see` should be separated from the description by one line and not be wrapped in brackets.

src/main/java/net/dv8tion/jda/core/entities/MessageEmbed.java
```diff
@@ -30,11 +30,37 @@
  */
 public interface MessageEmbed
 {
+    /**
+     * The character limit of the maximum title length is {@value} chacacters.<br>
```
MinnDevelopment: `<br>` tags at the beginning of the next line

src/main/java/net/dv8tion/jda/core/entities/MessageEmbed.java
```diff
@@ -30,11 +30,37 @@
...
+     */
     int URL_MAX_LENGTH = 2000;
+    /**
+     * The character limit of the maximum bot embed length is {@value} chacacters.
+     */
     int EMBED_MAX_LENGTH_BOT = 4000;
+    /**
+     * The character limit of the maximum client embed length is {@value} chacacters.
+     */
     int EMBED_MAX_LENGTH_CLIENT = 2000;
```
MinnDevelopment: These limits need a better description as to what is included etc.

MinnDevelopment:
I want to see an update on this PR

MinnDevelopment:
As there hasn't been an update to this PR after my requested changes I'm closing this.

PR #502, opened 2017-10-03
Updated regions to use raw strings in Guild

MinnDevelopment:
This can be useful to debug which regions are not currently supported by JDA.


kantenkugel:
Why store region as string? Isn't an enum value one of the best things regarding storage?

PR #506, opened 2017-10-05
Removing FailedRestAction

Sanduhr32:
I got this stacktrace
```
[10:35:31] [ERROR] [net.dv8tion.jda.core.JDA]: One of the EventListeners had an uncaught exception
[10:35:31] [ERROR] [net.dv8tion.jda.core.JDA]: Encountered an Exception: 
java.lang.IllegalArgumentException: api may not be null
	at net.dv8tion.jda.core.utils.Checks.notNull(Checks.java:45)
	at net.dv8tion.jda.core.requests.RestAction.<init>(RestAction.java:199)
	at net.dv8tion.jda.core.requests.RestAction.<init>(RestAction.java:183)
	at net.dv8tion.jda.core.requests.RestAction$FailedRestAction.<init>(RestAction.java:741)
```
after creating a FailedRestAction. I looked into the constructor and saw that the api is null what threw the error.

The first commit was a bad merge from DV8/JDA into my forck. Sorry


kantenkugel:
I doubt that this compiles, seeing the only Constructor being changed parameter wise but no usage of it.

Gonna check the source

kantenkugel:
Ok... What did throw that error for you? I found no usage of that class when looking through code. Did you create a instance of FailedRestAction yourself?

Sanduhr32:
"I got this stacktrace [stacktrace] after creating an instance of FailedRestAction"

kantenkugel:
Why would you do that manually? That stuff is meant for JDA internals

kantenkugel:
So after some looking through source: `RestAction.FailedRestAction` has no internal usage (might be deleted in the future)... 

There is another `FailedRestAction` definition in `AuditableRestAction` that is used a single time (`JDA#installAuxiliaryCable(int)`). 
That one has the same problem of doing `super(null, null)` and causing an exception from the arg check. But if that is the only use, we might wanna remove that one too.

Either way: **DON'T USE JDA INTERNALS**

kantenkugel:
Requesting feedback from @MinnDevelopment

MinnDevelopment:
Considering we don't use this at all, I want to remove the FailedRestAction completely.

Sanduhr32:
Accoring to latest comment from Minn and my chat with Kantenkugel i considered to remove FailedRestAction. I removed the only usage too.

kantenkugel:
@MinnDevelopment didn't you have other plans with this? Something that also includes changes to EmptyRestAction?

Sanduhr32:
@kantenkugel EmptyRestAction is used by `JDA#retrieveUserById`
Are you sure that you want to remove it?

MinnDevelopment:
@kantenkugel those plans don't need to be pushed onto this one
@Sanduhr32 we talked about something internally but you don't have to worry about it

Sanduhr32:
@MinnDevelopment i dont worry :)
Have a nice day and dont battle about what to remove / optimize next

PR #507, opened 2017-10-05
Fix grammar.

mhdsmr2:
Removed "the" from the sentence: "[...] which divides Events for the you."


kantenkugel:
Please don't post 1-word fixes. We see your issue and will incorporate that into a bigger PR

kantenkugel:
Also don't create PRs against master. This would issue a rebuild with our current CI setup and we def. don't need that for Docs grammar fixes

PR #516, opened 2017-10-13
Fixed MessageHistory constructor's permission check for TextChannel

kantenkugel:


kantenkugel:
Gonna re-verify that READ perm is actually needed and split the if up into 2 inner ifs about perms. 
That way i can produce individual InsufficientPermissionExceptions instead of a single one for both cases of missing perms

MinnDevelopment:
According to https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/core/utils/PermissionUtil.java#L363-L368
It should be implicit that the MESSAGA_READ permission is granted when __any__ other permission is granted.

kantenkugel:
Fixed... didn't realize we had that check in place

PR #517, opened 2017-10-14
Add zlib compression support

napstr:
Related:
image
https://github.com/discordapp/discord-api-docs/pull/408

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
+++ b/src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
@@ -53,6 +53,7 @@ import java.util.*;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.zip.DataFormatException;
+import java.util.zip.Inflater;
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -72,6 +73,7 @@
     protected WebSocket socket;
     protected String gatewayUrl = null;
     protected String sessionId = null;
+    protected final Inflater zlibContext = new Inflater();
```
MinnDevelopment: This needs to be unique for every session, on reconnecting this will break.

napstr:
Done.

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -432,6 +432,7 @@ protected void connect()
                     throw new RuntimeException("Could not fetch WS-Gateway!");
                 }
             }
+            zlibContext = new Inflater();
```
MinnDevelopment: Move this into `onConnected`

Issue #519, opened 2017-10-14
Proposal for the ability to configure the websocket gateway URL 


Frederikam:
When a JDA instance is first connecting to a shard, it will request the gateway URL from the `/gateway` REST endpoint. I intend to rewrite this functionality in such a way that it is possible to override it from the `JDABuilder`, allowing the URL to be replaced.

You might be wondering why anyone would want to do this, as there are currently no available alternatives for the Discord WSS gateway. I am currently working on a project called Dike which behaves as a caching proxy that will keep shards connected even after a Discord library (like JDA) has shutdown. This would potentially allow us to have JDA send OP 2 for hundreds of shards in a very short time without affecting Discord's gateway.

I am opening this issue to discuss these changes and get your approval before writing any code. 

## Proposed changes
* Create the interface `net.dv8tion.jda.core.requests.IGatewayProvider` with a single method:
    * `#getGatewayUrl()` which replaces `WebSocketClient#getGateway()`
* Create the class `net.dv8tion.jda.core.requests.GatewayProviderImpl` which implements the interface as the current JDA behavior. This class would be default.
* Add the method `JDABuilder#setGatewayProvider(IGatewayProvider provider)` which allows replacing the default provider. Would be documented as an advanced feature in the javadoc.

## Additional thoughts
* It might be a good idea to invalidate the cached URL by default if a connection fails. JDA doesn't do this but it is described in the [Discord docs](https://discordapp.com/developers/docs/topics/gateway#get-gateway)
* It might be a good idea to statically cache the URL for all shards, as this is the behaviour that Discord recommends. JDA currently uses a nonstatic field in `WebSocketClient`.

adriantodt:
pinging @natanbc, who is also doing a persistent-gateway https://github.com/natanbc/discord-bot-gateway

DV8FromTheWorld:
I would be open to seeing a PR related to this.

MinnDevelopment:
Added in [3.5.0](https://github.com/DV8FromTheWorld/JDA/releases/tag/v3.5.0)

SecretAgent-YT:
Is this feature still implemented? Not seeing the method in the documentation for the JDABuilder interface...

PR #520, opened 2017-10-15
Added IGatewayProvider

natanbc:
Addresses #519.

The docs are really simple, will probably need to be rewritten.

Also fixed an accidental synchronization on a JDA.Status object instead of the JDA instance on JDAImpl#setStatus


src/main/java/net/dv8tion/jda/core/requests/DefaultGatewayProvider.java
```diff
@@ -0,0 +1,48 @@
...
+                            request.onFailure(new Exception("Failed to get gateway url"));
+                    }
+                    catch (Exception e)
+                    {
+                        request.onFailure(e);
+                    }
+                }
+            };
+
+            return gateway.complete(false) + "?encoding=json&v=" + WebSocketClient.DISCORD_GATEWAY_VERSION;
```
MinnDevelopment: This will break functionality later when we want to provide both ETF and JSON gateways

src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
```diff
@@ -145,7 +145,7 @@ public void login(String token, ShardInfo shardInfo, SessionReconnectQueue recon
 
     public void setStatus(Status status)
     {
-        synchronized (this.status)
+        synchronized (this)
```
MinnDevelopment: This was actually intended

Frederikam:
I mean, I actually intended to work on this myself but I guess this works too ¯\\\_(ツ)\_/¯

Shredder121:
I'd say also document when the method is going to be called, on login, on connection lost or not.
Maybe document on the class level that it's an advanced usage thing.
Right now it doesn't tell implementors much.

DV8FromTheWorld:
This should probably actually be implemented using the Builder pattern so that the JDA instance can be provided to any IGatewayProvider on creation. As it is, only the DefaultGatewayProvider would have the ability to be created with a JDA instance. No custom implementation would ever be able to be created to have a JDA instance.

As such, this PR is unacceptable in my opinion.

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -426,6 +427,14 @@ protected void connect()
         {
             if (gatewayUrl == null)
             {
+                if (gatewayProvider == null)
+                {
+                    gatewayProvider = gatewayProviderFactory.createGatewayProvider(api);
+                    if (gatewayProvider == null)
+                    {
+                        throw new RuntimeException("IGatewayProviderFactory returned a null IGatewayProvider!");
```
MinnDevelopment: Use IllegalArgumentException

src/main/java/net/dv8tion/jda/core/requests/IGatewayProviderFactory.java
```diff
@@ -0,0 +1,25 @@
+package net.dv8tion.jda.core.requests;
```
MinnDevelopment: Put this and the other interface in a separate package "...requests.factory" together with a package-info.java

src/main/java/net/dv8tion/jda/core/requests/IGatewayProviderFactory.java
```diff
@@ -0,0 +1,25 @@
...
+ * {@link net.dv8tion.jda.core.JDABuilder#setGatewayProviderFactory(IGatewayProviderFactory) JDABuilder.setAudioSendFactory(IAudioSendFactory)}.
+ */
+public interface IGatewayProviderFactory
+{
+    /**
+     * Called by JDA's websocket client when it's first connecting and needs an {@link IGatewayProvider IGatewayProvider}
+     * to tell it the URL it should connect to.
+     *
+     * @param jda
+     *            The JDA instance that needs an {@link IGatewayProvider IGatewayProvider}.
```
MinnDevelopment: This formatting is not inline with the JDA guidelines

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -84,6 +88,21 @@ public JDABuilder(AccountType accountType)
     }
 
     /**
+     * Sets the gateway provider that will be used to get the URL to connect the websocket to.
```
MinnDevelopment: This needs more information on what this is used for and why it can be changed.

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -538,7 +557,8 @@ public JDA buildAsync() throws LoginException, IllegalArgumentException, RateLim
                 .setCacheGame(game)
                 .setCacheIdle(idle)
                 .setCacheStatus(status);
-        jda.login(token, shardInfo, reconnectQueue);
+        IGatewayProviderFactory gatewayProviderFactory = this.gatewayProviderFactory == null ? new DefaultGatewayProviderFactory() : this.gatewayProviderFactory;
```
MinnDevelopment: This check should happen inside the WebSocketClient constructor

src/main/java/net/dv8tion/jda/core/requests/IGatewayProviderFactory.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Factory interface for the creation of new {@link net.dv8tion.jda.core.requests.IGatewayProvider IGatewayProvider} objects.
+ * <br>JDA, by default, uses {@link net.dv8tion.jda.core.requests.DefaultGatewayProviderFactory DefaultGatewayProviderFactory} for the
+ * creation of its gateway providers.
+ * <p>
+ * Implementations of this interface are provided to
+ * {@link net.dv8tion.jda.core.JDABuilder#setGatewayProviderFactory(IGatewayProviderFactory) JDABuilder.setAudioSendFactory(IAudioSendFactory)}.
+ */
+public interface IGatewayProviderFactory
```
MinnDevelopment: `JDABuilder.setAudioSendFactory` is incorrect.

src/main/java/net/dv8tion/jda/core/requests/IGatewayProvider.java
```diff
@@ -0,0 +1,21 @@
+package net.dv8tion.jda.core.requests;
+
+/**
+ * Tells JDA to which URL it should connect the websocket. Custom implementations can be used to have
```
MinnDevelopment: ~~Tells~~ Provides

src/main/java/net/dv8tion/jda/core/requests/IGatewayProviderFactory.java
```diff
@@ -0,0 +1,25 @@
...
+{
+    /**
+     * Called by JDA's websocket client when it's first connecting and needs an {@link IGatewayProvider IGatewayProvider}
+     * to tell it the URL it should connect to.
```
MinnDevelopment: ~~Tell~~ Provide

MinnDevelopment:
You need to fix your conflicts.

src/main/java/net/dv8tion/jda/core/requests/IGatewayProvider.java
DV8FromTheWorld: `Provides JDA a strategy to determine a URL which it should use when connecting the Main Websocket`

src/main/java/net/dv8tion/jda/core/JDABuilder.java
DV8FromTheWorld: I would actually like to see this handled in the same way that the `eventManager` and the `audioSendFactory` are handled above.

Frederikam:
@natanbc do you still intend to complete this PR? I depend on this feature for some of my work. If you don't intend to complete it soon, I will have to make a PR of my own based off of this code.

MinnDevelopment:
With all these cross-session system I think it's worth taking the time to centralize. We should make a single system that can be applied to multiple sessions like a `CrossSessionController` if you will.
This could take care of:
- Session IDENTIFY limit (both connect and reconnect)
- Global REST ratelimit
- Gateway retrieval (provider like in this PR)
- anything else that can be a cross-session state or behaviour

The idea would be to then just create one controller and implement the features (or use a default adapter) which can be set on the builder:
```java
jdaBuilder.setSessionController(new MySessionController());
```

Thoughts? @DV8FromTheWorld @Frederikam

MinnDevelopment:
I opened a proposing PR for the aforementioned system #562

DV8FromTheWorld:
I agree with centralizing this functionality.

MinnDevelopment:
This will be continued in #562

PR #521, opened 2017-10-15
Added tag to produce html5 docs

MinnDevelopment:
This will be ignored when using javadoc from jdk8 and will produce html5 documentation when using jdk9.

Additionally I updated gradle.

Edit: Jitpack produces html4 with their JDK 8 installation https://jitpack.io/com/github/DV8FromTheWorld/JDA/test-javadoc-html5-v3.3.0-g8b9ddff-31/javadoc/index.html


DV8FromTheWorld:
So long as this works with non-J9 compiling environments, then I am okay with this PR.
I will update the build server to Java9 ASAP.

PR #522, opened 2017-10-15
Allow SessionReconnectQueue extensions to call reconnect

napstr:
When implementing a custom ReconnectQueue there the issue that there is no way to call the reconnect method of a WebSocketClient because it is `protected`, so here's a method to do so from the `SessionReconnectQueue` class.
There's probably a less ugly way to achieve this like making WebSocketClient#reconnect public or something. I'm just not sure about how JDA wants to handle their method access levels, this is the less intrusive one.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/requests/SessionReconnectQueue.java
+++ b/src/main/java/net/dv8tion/jda/core/requests/SessionReconnectQueue.java
@@ -54,6 +54,12 @@ public class SessionReconnectQueue
         }
     }
 
+    //allows any custom extensions of this class to call reconnects on the WebSocketClient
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
Why do you need to change this behaviour? How different would your implementation be for this process

Shredder121:
Session**Re**connectQueue only does reconnects.
We want to hook it up to the initial connects as well.

MinnDevelopment:
I don't see how that is relevant to these changes.

napstr:
After having some rethonks I found a more elegant solution for this.
Previously I missed that one can provide a custom BlockingQueue implementation to the constructor of the SessionReconnectQueue, which results in a good enough:tm: handling of things from my pov.
Related: https://github.com/Frederikam/FredBoat/pull/346

PR #523, opened 2017-10-15
Hotfix zlib fragmentation

MinnDevelopment:


MinnDevelopment:
I've requested a user that encounters issues related to this to test these changes via jitpack.

DV8FromTheWorld:
Based purely on the code, it appears that this is attempting to buffer the binary data until it has all been sent before attempting to handle it.
Is that what this is attempting to solve?

MinnDevelopment:
That is indeed correct

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -1000,22 +1007,40 @@ protected void handleEvent(JSONObject raw)
...
         {
-            decompressor.write(binary);
+            readBuffer.writeTo(decompressor);
+            // send the inflated message to the TextMessage method
+            onTextMessage(websocket, out.toString("UTF-8"));
         }
         catch (IOException e)
         {
-            throw (DataFormatException) new DataFormatException("Malformed").initCause(e);
+            throw (IOException) new DataFormatException("Malformed").initCause(e);
```
Shredder121:
```
java.lang.ClassCastException: java.util.zip.DataFormatException cannot be cast to java.io.IOException
	at net.dv8tion.jda.core.requests.WebSocketClient.onBinaryMessage(WebSocketClient.java:1032)
	at com.neovisionaries.ws.client.ListenerManager.callOnBinaryMessage(ListenerManager.java:368)
	at com.neovisionaries.ws.client.ReadingThread.callOnBinaryMessage(ReadingThread.java:270)
	at com.neovisionaries.ws.client.ReadingThread.handleBinaryFrame(ReadingThread.java:990)
	at com.neovisionaries.ws.client.ReadingThread.handleFrame(ReadingThread.java:749)
	at com.neovisionaries.ws.client.ReadingThread.main(ReadingThread.java:108)
	at com.neovisionaries.ws.client.ReadingThread.runMain(ReadingThread.java:64)
	at com.neovisionaries.ws.client.WebSocketThread.run(WebSocketThread.java:45)
```

Shredder121: (line count doesn't match exactly, but this is the culprit)

MinnDevelopment: That change was unintended and is already fixed in the next commit I'm working on

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -1006,32 +1008,43 @@ protected void handleEvent(JSONObject raw)
         }
     }
 
+    public boolean onBufferMessage(byte[] binary) throws IOException
```
DV8FromTheWorld: This should be private.

DV8FromTheWorld: It should probably also be synchronized: Either on the entire method or on a semaphore controlling access to `readBuffer`

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -1006,32 +1008,43 @@ protected void handleEvent(JSONObject raw)
...
+    {
+        if (binary.length >= 4 && getInt(binary, binary.length - 4) == ZLIB_SUFFIX)
```
DV8FromTheWorld: is the length is greater than 4, but ends in the ZLIB suffix, why do we just throw away the data that came before the suffix in this packet?

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -1006,32 +1008,43 @@ protected void handleEvent(JSONObject raw)
...
-        ByteArrayOutputStream out = new ByteArrayOutputStream(readBuffer.size() * 2);
+        final int size = readBuffer != null ? readBuffer.size() : binary.length;
+        ByteArrayOutputStream out = new ByteArrayOutputStream(size * 2);
         try (InflaterOutputStream decompressor = new InflaterOutputStream(out, zlibContext))
         {
-            readBuffer.writeTo(decompressor);
+            if (readBuffer != null)
+                readBuffer.writeTo(decompressor);
+            else
+                decompressor.write(binary);
```
DV8FromTheWorld: Based on this code, it looks like you're relying onBufferMessage to return true when it has found the end of the ZLIB packet, however, take the following situation:
onBinaryMessage receives a partial zlib packet. We determine it is partial in onBufferMessage, place the partial in a buffer, and return false from onBufferMessage.
onBinaryMessage receives a second zlib packet, this one containing the rest of the partial packet AND the zlib suffix. onBufferMessage will return true without writting the data to the buffer.

Since onBufferMessage returned true, we attempt to decompress the data, however, the readBuffer will only have the partial packet data from the first packet.

Am I missing something here?

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
MinnDevelopment: Why should this be synchronized? This only called on one thread

napstr:
> I've requested a user that encounters issues related to this to test these changes via jitpack.

I'm not that user but I've been testing this (you shouldn't trust my testing skills) and seems to work. No fragmentation errors encountered so far on this branch (builds on master tend to get ripped about 20% of the time during shard creation aka can barely start five shards with it).

PR #530, opened 2017-10-18
Added WebhookClientBuilder(String) constructor

Frederikam:
This adds a constructor to `WebhookClientBuilder` which takes the URL for the webhook (as opposed to the id and token separately.)

Tested for the following URLs:
```
https://discordapp.com/api/webhooks/370264727500095498/token
http://discordapp.com/api/webhooks/370264727500095498/token
https://discordapp.com/api/6/webhooks/370264727500095498/token
https://canary.discordapp.com/api/webhooks/370264727500095498/token
https://discordapp.com/api/webhooks/370264727500095498/token/slack
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
+++ b/src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
@@ -25,6 +25,8 @@ import javax.annotation.Nullable;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Builder that creates a new {@link net.dv8tion.jda.webhook.WebhookClient WebhookClient} instance
@@ -32,6 +34,7 @@ import java.util.concurrent.ThreadFactory;
 public class WebhookClientBuilder
 {
     public static final OkHttpClient.Builder DEFAULT_HTTP_BUILDER = new OkHttpClient.Builder();
+    private static final Pattern WEBHOOK_PATTERN = Pattern.compile("webhooks\\/(\\d+)\\/([\\w-]+)");
 
     protected final long id;
     protected final String token;
@@ -59,6 +62,29 @@ public class WebhookClientBuilder
         Checks.noWhitespace(token, "Token");
         this.id = id;
         this.token = token;
+    }
+
+    /**
+     * Creates a new WebhookClientBuilder with the provided webhook URL
+     *
+     * @param  url
+     *         The URL of the webhook. May be directly copied from Discord's UI
+     *         <br>Example: {@code https://discordapp.com/api/webhooks/123456789012345678/my-webhook-token}
+     *         <br>This constructor also parses URLs pointing to subdomains of {@code discordapp.com}
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided token is {@code null}
+     *         or is incorrectly formatted
+     */
+    public WebhookClientBuilder(@Nonnull String url)
+    {
+        Matcher matcher = WEBHOOK_PATTERN.matcher(url);
+        if(!matcher.find()) {
+            throw new IllegalArgumentException("Failed to parse webhook URL");
+        }
+
+        this.id = Long.parseLong(matcher.group(1));
+        this.token = matcher.group(2);
     }
 
     /**
--- a/src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
+++ b/src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
@@ -80,7 +80,8 @@ public class WebhookClientBuilder
     public WebhookClientBuilder(@Nonnull String url)
     {
         Matcher matcher = WEBHOOK_PATTERN.matcher(url);
-        if (!matcher.matches()) {
+        if (!matcher.matches())
+        {
             throw new IllegalArgumentException("Failed to parse webhook URL");
         }
 
--- a/src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
+++ b/src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
@@ -35,7 +35,7 @@ import java.util.regex.Pattern;
 public class WebhookClientBuilder
 {
     public static final OkHttpClient.Builder DEFAULT_HTTP_BUILDER = new OkHttpClient.Builder();
-    private static final Pattern WEBHOOK_PATTERN = Pattern.compile("(?:https?://)?(?:\\w+.)?discordapp\\.com/api(?:/\\d+)?/webhooks/(\\d+)/([\\w-]+)(?:/(?:\\w+)?)?");
+    private static final Pattern WEBHOOK_PATTERN = Pattern.compile("(?:https?://)?(?:\\w+\\.)?discordapp\\.com/api(?:/v\\d+)?/webhooks/(\\d+)/([\\w-]+)(?:/(?:\\w+)?)?");
 
     protected final long id;
     protected final String token;
@@ -80,10 +80,8 @@ public class WebhookClientBuilder
     public WebhookClientBuilder(@Nonnull String url)
     {
         Matcher matcher = WEBHOOK_PATTERN.matcher(url);
-        if(!matcher.find()) {
+        if (!matcher.matches()) {
             throw new IllegalArgumentException("Failed to parse webhook URL");
-        } else if (!matcher.matches()) {
-            throw new IllegalArgumentException("Failed to match the entire string as a webhook URL");
         }
 
         this.id = MiscUtil.parseSnowflake(matcher.group(1));
--- a/src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
+++ b/src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
@@ -18,6 +18,7 @@ package net.dv8tion.jda.webhook;
 
 import net.dv8tion.jda.core.entities.Webhook;
 import net.dv8tion.jda.core.utils.Checks;
+import net.dv8tion.jda.core.utils.MiscUtil;
 import okhttp3.OkHttpClient;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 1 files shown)


kantenkugel:
Are you sure that the javadoc description of the `throws` statement is correct?

Frederikam:
I probably should have tested it, but it seems so:
```
Exception in thread "main" java.lang.IllegalArgumentException: Argument for @Nonnull parameter 'url' of net/dv8tion/jda/webhook/WebhookClientBuilder.<init> must not be null
	at net.dv8tion.jda.webhook.WebhookClientBuilder.$$$reportNull$$$0(WebhookClientBuilder.java)
	at net.dv8tion.jda.webhook.WebhookClientBuilder.<init>(WebhookClientBuilder.java)
	at scratch_7.main(scratch_7.java:14)
```
```
Exception in thread "main" java.lang.IllegalArgumentException: Failed to parse webhook URL
	at net.dv8tion.jda.webhook.WebhookClientBuilder.<init>(WebhookClientBuilder.java:83)
	at scratch_7.main(scratch_7.java:14)
```

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
```diff
@@ -62,6 +65,29 @@ public WebhookClientBuilder(final long id, final String token)
...
     /**
+     * Creates a new WebhookClientBuilder with the provided webhook URL
+     *
+     * @param  url
+     *         The URL of the webhook. May be directly copied from Discord's UI
+     *         <br>Example: {@code https://discordapp.com/api/webhooks/123456789012345678/my-webhook-token}
+     *         <br>This constructor also parses URLs pointing to subdomains of {@code discordapp.com}
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided token is {@code null}
```
kantenkugel: If provided **token** is null?

Frederikam: Oops

MinnDevelopment:
Why is this needed? I think this task should be done by the user of the library.

Frederikam:
Here's an example use-case:

https://github.com/FredBoat/Shooker/blob/master/src/main/java/com/fredboat/shooker/RestController.java#L53-L61

This application is meant to be configured with webhook URLs directly from a config file. A different use-case might be having a bot/website be given a webhook URL from a user.

I feel that this way of constructing a WebhookClientBuilder is much more intuitive than having the user think of how they can parse a URL for its ID and token. It is literally copy-paste from the webhook UI:
image

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
```diff
@@ -62,6 +65,29 @@ public WebhookClientBuilder(final long id, final String token)
...
+     *         <br>This constructor also parses URLs pointing to subdomains of {@code discordapp.com}
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided URL is {@code null}
+     *         or is incorrectly formatted
+     */
+    public WebhookClientBuilder(@Nonnull String url)
+    {
+        Matcher matcher = WEBHOOK_PATTERN.matcher(url);
+        if(!matcher.find()) {
```
MinnDevelopment: Why shouldn't this be a proper regex to ensure the URL is actually a discord url, then `pattern.matches(url)`

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
```diff
@@ -62,6 +65,29 @@ public WebhookClientBuilder(final long id, final String token)
...
+            throw new IllegalArgumentException("Failed to parse webhook URL");
+        }
+
+        this.id = Long.parseLong(matcher.group(1));
+        this.token = matcher.group(2);
```
MinnDevelopment: Use the existing constructor with the results, that should be `MiscUtil.parseSnowflake(...)`

Frederikam: There's no clean way of using the same constructor
image

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
Frederikam: Why the `pattern.matches(url)`? If #find() fails it wouldn't match anyways

MinnDevelopment: We are looking for a proper URL not some 2k char message with some url in it.

Frederikam: Also, do you want me to verify if the String indeed is a HTTP(S) URL, or should we ignore the schema? Eg do we consider `discordapp.com/api/webhooks/370264727500095498/token` valid?

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
MinnDevelopment: Understandable, still use the proper parse method.

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
MinnDevelopment: I would suggest to make it optional and use the matches method. `(https?://)?(www\.)?.......`

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
```diff
@@ -34,7 +35,7 @@
 public class WebhookClientBuilder
 {
     public static final OkHttpClient.Builder DEFAULT_HTTP_BUILDER = new OkHttpClient.Builder();
-    private static final Pattern WEBHOOK_PATTERN = Pattern.compile("webhooks\\/(\\d+)\\/([\\w-]+)");
+    private static final Pattern WEBHOOK_PATTERN = Pattern.compile("(?:https?://)?(?:\\w+.)?discordapp\\.com/api(?:/\\d+)?/webhooks/(\\d+)/([\\w-]+)(?:/(?:\\w+)?)?");
```
MinnDevelopment: Did you mean to add the version here? because that would be `/api/v{d}/` rather than `/api/{d}/`

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
```diff
@@ -34,7 +35,7 @@
 public class WebhookClientBuilder
 {
     public static final OkHttpClient.Builder DEFAULT_HTTP_BUILDER = new OkHttpClient.Builder();
-    private static final Pattern WEBHOOK_PATTERN = Pattern.compile("webhooks\\/(\\d+)\\/([\\w-]+)");
```
kantenkugel: Don't you wanna go with `(?:www\\.)?` ? At least escape the dot. Otherwise you match any prefix.

Also: I'm assuming `/api(?:/\\d+)?/` is to allow api version? If so: api version is prepended by `v` so like `/api/v6/` (see docs here: <https://discordapp.com/developers/docs/reference>)

Also: what is the last part (after id/token) about?

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
```diff
@@ -81,9 +82,11 @@ public WebhookClientBuilder(@Nonnull String url)
         Matcher matcher = WEBHOOK_PATTERN.matcher(url);
         if(!matcher.find()) {
             throw new IllegalArgumentException("Failed to parse webhook URL");
+        } else if (!matcher.matches()) {
+            throw new IllegalArgumentException("Failed to match the entire string as a webhook URL");
         }
```
kantenkugel: You can just drop the `matcher.find()` part completely in favor of `matcher.matches()`

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
Frederikam: >Don't you wanna go with (?:www\\.)?

No, that would exclude canary, ptb and whatnot

>what is the last part (after id/token) about?

This allows matching of routes that end with say `/slack` or `/github`

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
Frederikam: Looking at the javadoc I believe find() is still required to get the group data

kantenkugel: <https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#matches-->
`If the match succeeds then more information can be obtained via the start, end, and group methods.`

Frederikam: Ah, I'm blind. Nvm

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
kantenkugel: Oh right. other 2 points are still valid tho (escaping dot and adding `v` to version match)

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
kantenkugel: Actually didn't find any indication about only `find()` populating groups when going over class docs and comparing `find()` vs `matches()` docs

Frederikam: Just ran a test. Works fine with only `matches()`

src/main/java/net/dv8tion/jda/webhook/WebhookClientBuilder.java
```diff
@@ -62,6 +66,29 @@ public WebhookClientBuilder(final long id, final String token)
...
+     *         <br>This constructor also parses URLs pointing to subdomains of {@code discordapp.com}
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided URL is {@code null}
+     *         or is incorrectly formatted
+     */
+    public WebhookClientBuilder(@Nonnull String url)
+    {
+        Matcher matcher = WEBHOOK_PATTERN.matcher(url);
+        if (!matcher.matches()) {
```
MinnDevelopment: put the bracket on newline and we're good to go

PR #531, opened 2017-10-20
Correct reactions parsing

Vampire:
The field "self" actually is called "me"

https://discordapp.com/developers/docs/resources/channel#reaction-object


kantenkugel:
lgtm (current impl doesn't work)

PR #550, opened 2017-11-14
Added ContextException and checks for RestAction improvements

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

### Pull Request

1. This allows to pass a context cause to all exceptions inside the requester internals.
    The new `ContextException` will be used in one of two ways.
    1. Using `ContextException.here(Consumer<Throwable>)` directly
        ```java
        channel.sendMessage("test").queue(null, ContextException.here(Throwable::printStackTrace));
        ```
    2. Using `RestAction.setPassContext(true)` will make this an automated feature for all RestActions
2. Added last-second checks for the queue workers. These checks can be used to cancel RestActions before they execute, this is an alternative to using `submit().cancel(true)` which allows to have a central check function.
    Additionally this will allow the library to (for example) check permissions or states before making the HTTP request. This way we would be able to predict ErrorResponseExceptions and avoid an unnecessary HTTP call.


src/main/java/net/dv8tion/jda/core/requests/RestAction.java
```diff
@@ -161,13 +162,31 @@
             LOG.error("RestAction queue returned failure: [{}] {}", t.getClass().getSimpleName(), t.getMessage());
         }
     };
+    protected static boolean passContext = false;
```
DV8FromTheWorld: This will change functionality across all running instances of JDA.
Do we want that? Or do we want to provide per-JDA-instance support?

src/main/java/net/dv8tion/jda/core/requests/ratelimit/BotRateLimiter.java
```diff
@@ -311,6 +310,11 @@ public void run()
                         try
                         {
                             request = it.next();
+                            if (!request.runChecks())
+                            {
+                                it.remove();
+                                continue;
```
DV8FromTheWorld: This needs to actually complete the request somehow, not just drop it.
`request.onFailure`.

DV8FromTheWorld: This needs to actually complete the request somehow, not just drop it.
`request.onFailure`.

src/main/java/net/dv8tion/jda/core/requests/RestAction.java
DV8FromTheWorld: This will change functionality across all running instances of JDA.
Do we want that? Or do we want to provide per-JDA-instance support?

src/main/java/net/dv8tion/jda/core/requests/ratelimit/ClientRateLimiter.java
```diff
@@ -203,6 +203,11 @@ public void run()
                         try
                         {
                             request = it.next();
+                            if (!request.runChecks())
+                            {
+                                it.remove();
+                                continue;
```
DV8FromTheWorld: Same as above. (Complete the request)

src/main/java/net/dv8tion/jda/core/requests/RestAction.java
Almighty-Alpaca: I'm also more a fan of making this a setting in JDABuilder

MinnDevelopment: I'm unsure why this should be a per-JDA setting but DEFAULT_FAILURE and DEFAULT_SUCCESS should not

DV8FromTheWorld: 🤔 That is a good point. We should make a decision now as to whether we make both per-JDA, or both Global.

DV8FromTheWorld:
What is the status of this PR?

MinnDevelopment:
Currently not as important as the 3.4 PRs so I'm not working on it.

src/main/java/net/dv8tion/jda/core/requests/RestAction.java
MinnDevelopment: If we want to merge this, we will have to decide on something.

MinnDevelopment:
@DV8FromTheWorld can i get an update on this?

MinnDevelopment:
Example on using checks for internal permission checking:
```java
//public class MessageAction
@Override
public BooleanSupplier finalizeChecks()
{
    return () ->
    {
        checkPermission(Permission.MESSAGE_WRITE); //throws exception
        return true; // check passed without exception
    }
}
```
This will then throw an exception if the permission is missing. That exception is caught by the internal RateLimiter and forwarded to the failure consumer, then the request is removed from the queue.

User of the library can use a similar check system which is checked _after_ the internal checks have passed (`internal() && user()`)

```java
private volatile boolean isDone = false;
// someMethod(TextChannel channel) {
BooleanSupplier done = () -> isDone;
channel.getMembers().forEach( member -> {
     channel.sendMessageFormat("Hello, %s!", member).setChecks(done).queue();
});
// some time later
isDone = true; // all actions still in queue will be cancelled, this is forwarded to the queue failure consumer
```

PR #557, opened 2017-12-01
Added MDC

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

### Pull Request

Using MDC we are able to categorize our logging events per shard/jda instance.
You can set a custom context map by using `JDABuilder.setContextMap(ConcurrentMap<String, String>)` which will be set in all JDA controlled threads.

Additionally I added a new method to the IAudioSendSystem to set the context in custom implementations. This method is declared as default due to backward compatibility but is implemented in the DefaultSendSystem.

Using logback and log4j MDC can be used to filter by shard:
#### Logback
```java
public class MyFilter extends Filter<ILoggingEvent>
{
    @Override
    public FilterReply decide(ILoggingEvent event)
    {
        String shard = MDC.get("jda.shard");
        if (shard != null && shard.equals("[0 / 2]"))
            return FilterReply.DENY;
        return FilterReply.NEUTRAL;
    }
}
```
#### Log4j
```java
public class MyFilter extends Filter
{
    @Override
    public int decide(LoggingEvent event)
    {
        String shard = MDC.get("jda.shard");
        if (shard != null && shard.equals("[0 / 2]"))
            return DENY;
        return NEUTRAL;
    }
}
```


src/main/java/net/dv8tion/jda/core/entities/impl/JDAImpl.java
```diff
@@ -132,6 +135,13 @@ public void login(String token, ShardInfo shardInfo, SessionReconnectQueue recon
         if (token == null || token.isEmpty())
             throw new LoginException("Provided token was null or empty!");
 
+        if (shardInfo != null)
+        {
+            contextMap.put("jda.shard", shardInfo.getShardString());
+            contextMap.put("jda.shard.id", String.valueOf(shardInfo.getShardId()));
+            contextMap.put("jda.shard.total", String.valueOf(shardInfo.getShardTotal()));
+        }
+        MDC.setContextMap(contextMap);
```
Shredder121: This nukes any MDC values that client code might have set 😢 
In your own threads I don't mind manually managing the contextMap, but this is rather destructive
Since buildAsync isn't enough, you'd have to invoke buildAsync asynchronously and explain why that's done that way and hope no one refactors it away

MinnDevelopment: The client is able to control the map

MinnDevelopment: I understand your concern though. I could make use of MDC.MDCClosable and just put all the values in the existing context for the build process

MinnDevelopment: This seems like a fair alternative
```java
// set MDC metadata for build thread
Map<String, String> previousContext = MDC.getCopyOfContextMap();
contextMap.forEach(MDC::put);
// init code
MDC.setContextMap(previousContext);
```

Shredder121: As long as it's addressed I'm fine with it 👍

PR #558, opened 2017-12-01
Added ability to get a Guild's features (vip voice, verified, ...)

kantenkugel:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Pull Request

This PR adds the method `Guild#getFeatures()` that returns `List<Strings>` describing a Guild's features as given by Discord.

**Possible (currently known) features are:**
- VIP_REGIONS
- VANITY_URL
- INVITE_SPLASH
- VERIFIED
- MORE_EMOJI

Edit:
### Todo:
- [x] Add `Guild#getFeatures()`
- [x] Handle Guild update add `GuildUpdateFeaturesEvent`
- [x] Add `RestAction<String> Guild#getVanityUrl()`
- [x] Add higher bitrate support for VIP voice regions
- [x] Test all changes

### To test:
- [x] `Guild#getFeatures()`
- [x] `GuildUpdateFeaturesEvent` (via `ListenerAdapter`)
- [x] `Guild#getVanityUrl()`
- [x] High bitrate support for VIP voice regions

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
@@ -151,6 +151,17 @@ public class EntityBuilder
                 .setRequiredMFALevel(Guild.MFALevel.fromKey(guild.getInt("mfa_level")))
                 .setExplicitContentLevel(Guild.ExplicitContentLevel.fromKey(guild.getInt("explicit_content_filter")));
 
+
+        if(guild.isNull("features"))
+            guildObj.setFeatures(Collections.emptyList());
+        else
+        {
+            JSONArray features = guild.getJSONArray("features");
+            List<String> featureList = new ArrayList<>(features.length());
+            features.forEach(o -> featureList.add(String.valueOf(o)));
+            guildObj.setFeatures(featureList);
+        }
+
         JSONArray roles = guild.getJSONArray("roles");
         for (int i = 0; i < roles.length(); i++)
         {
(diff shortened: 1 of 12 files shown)


kantenkugel:
I could/should probably remove the json check for null... I think it is always present (possibly empty array).
Also not sure if there is a nicer way of getting that list of strings

WalshyDev:
I was literally working on this already 😂 
In mine I just left it as a String[] until I figured out what to do, i think an enum would be nice but it's how to handle that.

weeryan17:
Yegh from some testing i did with walshy we found out that it will allways have a features array. So the null check isn't needed

Almighty-Alpaca:
Always expect everything from Discord. When they send some weird shit again (such as not sending the array) we don't want our lib to blow up.

kantenkugel:
Regarding the enum: i specifically didn't do that because the set of used feature strings could just change at any time without us even noticing (the possible values are not documented by discord). Thats the reason I am writing `possible (currently known) features`

WalshyDev:
@kantenkugel  Yeah that's what I was thinking but an enum would just be a bit easier in my opinion for comparison, for example you want to check if a guild has a vanityUrl or more emojis, just check the enum and if it was to change string but have the same meaning it wouldn't break. I think the cons outweigh the pros though. Definitely think that we could add some checks internally in our bot and it would be easier for us to obviously update that rather than the entire lib update.

Thanks for making this anyway, I'm not home, if I was I would have done this already so kudos for doing it in my place.

kantenkugel:
Oh man... i guess i rushed this out to quickly... even forgot update events...

weeryan17:
If you going to add a getVanityUrl() you should probably add a hasVanityUrl(). I mean we chould check it our selves by check if the list contains `VANITY_URL`, but for ease of use I'd add a hasVanityUrl()

MinnDevelopment:
I don't think we need to add methods that just check a collection in JDA. If we started doing that we would have to add something for a lot of cases that can just as well be handled by the user of the library.

MinnDevelopment:
@kantenkugel Are you expecting me to test these changes?

PR #559, opened 2017-12-03
Added GuildVoiceUpdateEvent to combine leave and move

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

### Pull Request

Allows to do:
```java
public class MyListener extends ListenerAdapter {
    @Override
    public void onGuildVoiceUpdate(GuildVoiceUpdateEvent event) {
        // do stuff when user leaves channel
    }
}
```
instead of:
```java
public class MyListener extends ListenerAdapter {
    @Override
    public void onGuildVoiceLeave(GuildVoiceLeaveEvent event) {
        onGuildVoiceUpdate(event.getChannelLeft());
    }
    @Override
    public void onGuildVoiceMove(GuildVoiceMoveEvent event) {
        onGuildVoiceUpdate(event.getChannelLeft());
    }

    public void onGuildVoiceUpdate(VoiceChannel oldChannel) {
        // do stuff when user leaves channel
    }
}
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,9 @@
 # Preserve gradlew's line ending
-gradlew eol=lf
+.gitattributes text eol=crlf
+
+gradlew binary
+
+*.java text eol=crlf
+*.gradle text eol=crlf
+
+* text=auto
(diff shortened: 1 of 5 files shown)


.gitattributes
```diff
@@ -1,2 +1,9 @@
 # Preserve gradlew's line ending
-gradlew eol=lf
+.gitattributes text eol=crlf
+
+gradlew binary
+
+*.java text eol=crlf
+*.gradle text eol=crlf
+
+* text=auto
```
DV8FromTheWorld: What does this actually change? Why'd you set it?

MinnDevelopment: On linux git wants to replace the crlf with lf on all files so I had to set it, this also prevents git from sending .java/.gradle files with other line endings. the `* test=auto` is the default for git to choose the best line ending it sees fit if no other rule overweights it

PR #562, opened 2017-12-04
Introducing SessionController System

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

### Pull Request

This combines all the cross-session states/behaviours that we now have in JDA 3.4 into one object.
I hope this will allow us to scale a lot better and keep it user-friendly.
Additionally the SessionReconnectQueue (legacy) is now able to handle the initial build process which allows users to build all their shards asynchronously without having to wait between starting.

#### Example
```java
JDABuilder builder = new JDABuilder(AccountType.BOT);
builder.setToken(BOT_TOKEN).addEventListener(myListener);
builder.setSessionController(new SessionControllerAdapter());
for (int i = 0; i < 10; i++) {
    builder.useSharding(i, 10).buildAsync();
}
```
> Note how this does not wait between starting shards as it is handled by the session controller automatically.


src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -539,7 +558,10 @@ public JDA buildAsync() throws LoginException, IllegalArgumentException, RateLim
     {
         OkHttpClient.Builder httpClientBuilder = this.httpClientBuilder == null ? new OkHttpClient.Builder() : this.httpClientBuilder;
         WebSocketFactory wsFactory = this.wsFactory == null ? new WebSocketFactory() : this.wsFactory;
-        JDAImpl jda = new JDAImpl(accountType, httpClientBuilder, wsFactory, shardRateLimiter, autoReconnect, enableVoice, enableShutdownHook,
+
+        if (controller == null && (reconnectQueue != null || shardRateLimiter != null))
+            controller = new ProvidingSessionController();
```
MinnDevelopment: Unsure if this is worth it, might be better to completely drop the legacy system.

src/main/java/net/dv8tion/jda/core/utils/SessionController.java
```diff
@@ -0,0 +1,44 @@
+/*
+ *     Copyright 2016 - 2017 Florian Spieß
```
MinnDevelopment: mfw I didn't update my copyright profile, whoops

MinnDevelopment:
Note to self: Update the README to mention ShardManager and SessionController functionality, replacing existing section.

Frederikam:
The SessionController looks great for what I intend to use it for. I'll begin adapting FredBoat to use a custom implementation for my custom gateway and see if this works.

src/main/java/net/dv8tion/jda/core/utils/SessionController.java
```diff
@@ -0,0 +1,190 @@
...
+ * builder.addEventListener(myListener);
+ * builder.build();
+ * </code></pre>
+ */
+public interface SessionController
+{
+    /**
+     * The default delay (in seconds) to wait between running {@link net.dv8tion.jda.core.utils.SessionController.SessionConnectNode SessionConnectNodes}
+     */
+    int IDENTIFY_DELAY = 5;
```
Frederikam: Why is this a field rather than a method? As a field of an interface it is effectively `public static final`. With my gateway I would want to replace this with 0 as the OP 2 ratelimit would be handled by the gateway.

MinnDevelopment: It is a constant from discord. If you want to remove or change the delay you have to override the worker that starts the sessions.

MinnDevelopment: 0 delay:
```java
@Override
public void appendSession(SessionConnectNode node) {
    node.run(); // with try/catch
}
```

Frederikam: Fair enough. I just thought it would make more sense to just extend the adapter and change that value.

MinnDevelopment: I added additional constructors for QueueWorker to make this easier.
```java
new SessionControllerAdapter()
{
    @Override
    public void runWorker()
    {
        synchronized(lock)
        {
            if (workerHandle == null)
            {
                workerHandle = new QueueWorker(0); // or 0L for millis
                workerHandle.start();
            }
        }
    }
};
```

MinnDevelopment:
I am planning to merge this by the end of this week.

PR #563, opened 2017-12-06
Add missing null checks in HttpRequestEvent

napstr:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

### Pull Request

Response#rawResponse is a nullable field and some methods in the HttpRequestEvent don't treat it like it is (though some do, like HttpRequestEvent#getResponseHeaders). This adds the missing checks.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/events/http/HttpRequestEvent.java
+++ b/src/main/java/net/dv8tion/jda/core/events/http/HttpRequestEvent.java
@@ -65,12 +65,12 @@ public class HttpRequestEvent extends Event
 
     public Headers getRequestHeaders()
     {
-        return this.response.getRawResponse().request().headers();
+        return this.response == null ? null : this.response.getRawResponse() == null ? null : this.response.getRawResponse().request().headers();
     }
 
     public okhttp3.Request getRequestRaw()
     {
-        return this.response == null ? null : this.response.getRawResponse().request();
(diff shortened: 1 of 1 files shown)


kantenkugel:
isn't this a bit too many null-checks? can request/response itself be null?

MinnDevelopment:
okhttp3.Response can be null but I don't believe Request or Response can be null

napstr:
HttpRequestEvent#getRequestHeaders a ton of methods in the class are checking it for null so I assumed that its nullable.

napstr:
I can but then again my main goal here is to fix the NPE not spend time checking the behaviour of other variables. If you don't want the nullcheck even though it is in style with the rest of the class please let me know and I'll remove it - I can't tell from the feedback so far if that decision has been reached yet by JDA staff.

Personally I've adopted `@Nonnull` and `@Nullable` annotations at many places to deal with not having to assume things. If that's a thing the JDA team would be interested in I'd be happy to add those as they create lasting value. Otherwise, the next person touching this has to do the research again, etc. Feel free to ping me in the Discord if it's a thing that grabs your interest.

MinnDevelopment:
The jsr-305 annotations might be getting more attention in JDA 3.5 but we would make that decision when the time is right. In this PR this is not needed.

napstr:
How about this?

PR #573, opened 2017-12-15
Use proper value for shutdown hooks of managed shards

napstr:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

### Pull Request

There's a bug here where the DefaultShardManager is erroneously passing the `enableBulkDeleteSplitting` field as a parameter for enabling shutdown hooks. Instead the value should always be false, since the DefaultShardManager manages the shutdown hook itself.


kantenkugel:
lgtm, gonna let minn (or better: almighty) merge

kantenkugel:
might wanna change value to true tho (default for standard jda)... dunno

napstr:
The DefaultShardManager uses it's own (configurable) shutdown hook to manage the shards it creates - setting shutdown hooks on the managed shards would be redundant.

kantenkugel:
ok then just ignore my previous comment :)

PR #574, opened 2017-12-16
Renamed GameType to ActivityType and deprecated old functionality

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This will update the enum to the proper new name. The documentation has adjusted this to be now called `Activity Type` so I renamed and deprecated it accordingly.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
@@ -100,9 +100,9 @@ public class EntityBuilder
         return selfUser;
     }
 
-    public Game createGame(String name, String url, Game.GameType type)
+    public Game createGame(String name, String url, Game.ActivityType type)
     {
-        return new Game(name, url, type);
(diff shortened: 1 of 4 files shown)


MinnDevelopment:
We could also take the chance to rename Game to Activity in 3.5

kantenkugel:
@MinnDevelopment that rename Game -> Activity sounds good (for 3.5)

DV8FromTheWorld:
Does the API refer to it as `game` still?

MinnDevelopment:
They are unable to rename the field in v6 as it would break compatibility but the documentation is being updated to this new name so I'm assuming it will also change in v7.

DV8FromTheWorld:
Gotcha. Works for me then.

MinnDevelopment:
I will get back to this in a later version.

MinnDevelopment:
I don't think this is worth the effort and inconvenience for existing applications.

PR #581, opened 2017-12-22
First pass on animated emote support

TheMonitorLizard:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This PR adds a `Emote#isAnimated` member to check whether or not the `Emote` is an animated custom emoji or not, as well as proper handling for the new property.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/Emote.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/Emote.java
@@ -136,6 +136,15 @@ public interface Emote extends ISnowflake, IMentionable, IFakeable
     EmoteManagerUpdatable getManagerUpdatable();
 
     /**
+     * Whether or not this Emote is animated or not.
+     *
+     * <p>Animated Emotes are available only to Discord Nitro users.
+     *
+     * @return Whether the Emote is animated or not.
+     */
+    boolean isAnimated();
+
+    /**
      * A String representation of the URL which leads to image displayed within the official Discord&trade; client
      * when this Emote is used
      *
@@ -157,7 +166,7 @@ public interface Emote extends ISnowflake, IMentionable, IFakeable
     @Override
     default String getAsMention()
     {
-        return "<:" + getName() + ":" + getIdLong() + ">";
+        return (isAnimated() ? "<a:" : "<:") + getName() + ":" + getIdLong() + ">";
     }
 
     /**
--- a/src/main/java/net/dv8tion/jda/core/entities/Emote.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/Emote.java
@@ -136,9 +136,9 @@ public interface Emote extends ISnowflake, IMentionable, IFakeable
     EmoteManagerUpdatable getManagerUpdatable();
 
     /**
-     * Whether or not this Emote is animated or not.
+     * Whether or not this Emote is animated.
      *
-     * <p>Animated Emotes are available only to Discord Nitro users.
+     * <p>Animated Emotes are available to Discord Nitro users as well as Bot accounts.
      *
      * @return Whether the Emote is animated or not.
      */
@@ -152,7 +152,7 @@ public interface Emote extends ISnowflake, IMentionable, IFakeable
      */
     default String getImageUrl()
     {
-        return "https://cdn.discordapp.com/emojis/" + getId() + ".png";
+        return "https://cdn.discordapp.com/emojis/" + getId() + (isAnimated() ? ".gif" : ".png");
     }
 
     /**
--- a/src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
@@ -29,10 +29,7 @@ import net.dv8tion.jda.core.requests.Response;
 import net.dv8tion.jda.core.requests.Route;
 import net.dv8tion.jda.core.requests.restaction.AuditableRestAction;
 
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.*;
 
 /**
  * Represents a Custom Emote. (Emoji in official Discord API terminology)
@@ -46,7 +43,7 @@ public class EmoteImpl implements Emote
     private final long id;
     private final GuildImpl guild;
     private final JDAImpl api;
-    private final HashSet<Role> roles;
+    private final Set<Role> roles;
 
     private final Object mngLock = new Object();
     private volatile EmoteManager manager = null;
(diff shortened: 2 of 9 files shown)


src/main/java/net/dv8tion/jda/core/entities/Emote.java
```diff
@@ -135,6 +135,15 @@
      */
     EmoteManagerUpdatable getManagerUpdatable();
 
+    /**
+     * Whether or not this Emote is animated or not.
```
GrandPanda: Redundant

src/main/java/net/dv8tion/jda/core/entities/Emote.java
```diff
@@ -135,6 +135,15 @@
...
+     *
+     * <p>Animated Emotes are available only to Discord Nitro users.
```
GrandPanda: Not sure if this is relevant info.

src/main/java/net/dv8tion/jda/core/handle/MessageReactionHandler.java
```diff
@@ -58,6 +58,7 @@ protected Long handleInternally(JSONObject content)
 
         final Long emojiId = emoji.isNull("id") ? null : emoji.getLong("id");
         String emojiName = emoji.optString("name", null);
+        final boolean emojiAnimated = emoji.getBoolean("animated");
```
GrandPanda: Should this be optBoolean()?

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
```diff
@@ -651,10 +651,12 @@ public Guild getGuild()
                 {
                     final long emoteId = MiscUtil.parseSnowflake(matcher.group(2));
                     final String emoteName = matcher.group(1);
+                    // Check animated by verifying whether or not it starts with <a: or <:
+                    final boolean animated = matcher.group(0).startsWith("<a:");
```
GrandPanda: group(0) is equivalent to just group(). The pattern doesn't seem to group that part so you probably don't mean 1.

src/main/java/net/dv8tion/jda/core/entities/Emote.java
kmddd59: Actually its a good idea I actually use the emoji part and think adding support for emoji animation detection is needed

kmddd59:
Actually its a good idea I actually use the emoji part and think adding support for emoji animation detection is needed

src/main/java/net/dv8tion/jda/core/entities/Emote.java
TheMonitorLizard: tbh, I wasn't quite sure what or how to document this.
I figured it might be useful for people using `AccountType.CLIENT` to know, as they cannot use the emotes without having Nitro.
If anything, I should add that it's also available to all `AccountType.BOT` clients, but I can understand if someone sees it as irrelevant.

src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
TheMonitorLizard: I did mean group 0, however I don't see a real reason to change this other than for the sake of changing it.

src/main/java/net/dv8tion/jda/core/handle/MessageReactionHandler.java
TheMonitorLizard: Yes, I'll change that asap 😅

MinnDevelopment:
I have a few questions.
Are bots able to use animated emotes?
Can animated emotes be used in the same server without nitro?

This behaviour needs to be documented if it changed from the previous behaviour.

TheMonitorLizard:
yes and yes

MinnDevelopment:
We should add checks for the emote upload limit in this PR as well.

src/main/java/net/dv8tion/jda/core/entities/impl/EmoteImpl.java
```diff
@@ -147,6 +148,12 @@ public EmoteManagerUpdatable getManagerUpdatable()
         return m;
     }
 
+    @Override
+    public boolean isAnimated()
+    {
+        return animated;
+    }
```
MinnDevelopment: The URL needs to be updated for animated emotes.

src/main/java/net/dv8tion/jda/core/entities/Emote.java
MinnDevelopment: What happens if you try to use an animated emote with a non-nitro client account?

src/main/java/net/dv8tion/jda/core/entities/Emote.java
MinnDevelopment: `Whether this emote is animated.`

src/main/java/net/dv8tion/jda/core/entities/Emote.java
TheMonitorLizard: They will not display as far as I can tell.
the new format is not supported for non supported accounts

MinnDevelopment: Then it should be checked in Member#canInteract(Emote) etc.

TheMonitorLizard: I was thinking the same thing!

src/main/java/net/dv8tion/jda/core/managers/GuildController.java
```diff
@@ -2117,6 +2117,10 @@ public RoleAction createCopyOfRole(Role role)
         Checks.notNull(icon, "Emote icon");
         Checks.notNull(roles, "Roles");
 
+        // Check for emote upload limits
+        Checks.check(getGuild().getEmoteCache().stream().filter(emote -> !emote.isAnimated()).count() < 50, "Guild normal emote limit has already been reached!");
+        Checks.check(getGuild().getEmoteCache().stream().filter(Emote::isAnimated).count() < 50, "Guild animated emote limit has already been reached!");
```
MinnDevelopment: If one bucket is full we just don't allow adding to the other?

TheMonitorLizard: whoops, wasn't thinking there

TheMonitorLizard: Actually now that I am looking at it, I am not quite sure how to do this.
We have no way of telling whether or not the `Icon` is animated before the request.

MinnDevelopment: Technically we can, discord does check it as well. Maybe this is too much to check, just document the error responses then.

PR #587, opened 2017-12-30
Nullability annotations and cleanup

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

Parameters annotated with `@NonNull` will display an IDE warning whenever a possibly-null value is passed. Due to `@ParamtersAreNonnullByDefault` we don't have to add this to every parameter, instead we only need to add the special cases where it is nullable using `@Nullable`.

Additionally I fixed a few javadoc issues.


napstr:
Are there plans to annotate return values too?

MinnDevelopment:
Return values turn out to be a lot more work.

MinnDevelopment:
I've left out most of the events from these changes for now as I needed a break from all of the copy-pasting.

MinnDevelopment:
Return value annotations generate checks and should thus be tested thoroughly before merging.

meowingtwurtle:
~~or just use Kotlin~~

MinnDevelopment:
This is not worth the effort at the moment. I might consider doing this when a bigger portion of the library needs to be rewritten.

PR #611, opened 2018-02-03
Store color with raw RGB value

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This replaces all use of `java.awt.Color` with raw int primitives.
You may still use the old front-facing api for the color class, we just extended it to also allow access to the raw value. The new default is defined as constant in `Role.DEFAULT_COLOR_RAW` to keep consistency between role colors and embed colors.

The default is specified as `0x1FFFFFFF` due to the `Color(int)` constructors behaviour.
If we used `-1` then `new Color(0xFFFFFF)` would yield `-1` as well which breaks compatibility.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
@@ -401,6 +401,8 @@ public class EmbedBuilder
      *         or {@code null} to use no color
      *
      * @return the builder after the color has been set
+     *
+     * @see    #setColor(int)
      */
     public EmbedBuilder setColor(Color color)
     {
@@ -408,6 +410,18 @@ public class EmbedBuilder
         return this;
     }
 
+    /**
+     * Sets the raw RGB color value for the embed.
+     *
+     * <a href="http://i.imgur.com/2YnxnRM.png" target="_blank">Example</a>
+     *
+     * @param  color
+     *         The raw rgb value, or {@link Role#DEFAULT_COLOR_RAW} to use no color
+     *
+     * @return the builder after the color has been set
+     *
+     * @see    #setColor(java.awt.Color)
+     */
     public EmbedBuilder setColor(int color)
     {
         this.color = color;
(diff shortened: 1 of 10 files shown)


src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
```diff
@@ -402,6 +403,12 @@ else if (temporal instanceof OffsetDateTime)
      * @return the builder after the color has been set
      */
     public EmbedBuilder setColor(Color color)
+    {
+        this.color = color == null ? Role.DEFAULT_COLOR_RAW : color.getRGB();
+        return this;
+    }
+
+    public EmbedBuilder setColor(int color)
```
MinnDevelopment: forgot docs here

MinnDevelopment:
This needs to be updated once the new managers have been merged. `RoleManager#setColor(int)` needs to be added later.

src/main/java/net/dv8tion/jda/core/events/role/update/RoleUpdateColorEvent.java
```diff
@@ -23,15 +23,20 @@
...
-    public RoleUpdateColorEvent(JDA api, long responseNumber, Role role, Color oldColor)
+    public RoleUpdateColorEvent(JDA api, long responseNumber, Role role, int oldColor)
     {
         super(api, responseNumber, role);
         this.oldColor = oldColor;
     }
 
     public Color getOldColor()
+    {
+        return oldColor != 0 ? new Color(oldColor) : null;
```
MinnDevelopment: this should check default color instead

DV8FromTheWorld:
What is the motivation behind this change?

MinnDevelopment:
We had someone in the JDA guild which had issues with awt as it created a window when using the color constructor. Additionally, I personally dislike that we use something made dor awt in JDA instead of just using the raw rgb value with the same effect

PR #614, opened 2018-02-05
Fixed Message#getMentionedX() methods returning non-mentioned entities

kantenkugel:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

Fixes message mention parsing to only return actually mentioned Users/Members/Roles via `getMentionedX()` methods instead of all entities that are found via regex matcher.

This removes strings resembling mentions but not actually being mentions (for example ones in code blocks or non-mentionable roles) from being treated as actually mentioning.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
@@ -17,6 +17,8 @@
 package net.dv8tion.jda.core.entities;
 
 import gnu.trove.map.TLongObjectMap;
+import gnu.trove.set.TLongSet;
+import gnu.trove.set.hash.TLongHashSet;
 import net.dv8tion.jda.bot.entities.ApplicationInfo;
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -921,12 +923,26 @@ public Message createMessage(JSONObject jsonObject, MessageChannel chan, boolean
...
+        }
+
         MessageType type = MessageType.fromId(jsonObject.getInt("type"));
         switch (type)
         {
             case DEFAULT:
-                return new ReceivedMessage(id, chan, type,
-                    fromWebhook, mentionsEveryone, tts, pinned,
+                return new ReceivedMessage(id, chan, type, fromWebhook,
+                    mentionsEveryone, mentionedUsers, mentionedRoles, tts, pinned,
```
MinnDevelopment: Can you also add these to system messages? Just in case.

PR #617, opened 2018-02-09
Implement speaking flags for the voice protocol v4

WalshyDev:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This adds in the new flag for the "speaking" field in the USER_SPEAKING_UPDATE event. This is for audio protocol 4 as described here: https://github.com/discordapp/discord-api-docs/issues/510


MinnDevelopment:
We are not currently planning to use v4. It is not a finished version and without doubt will only yield more problems than it could possibly "fix".

MinnDevelopment:
Thank you for your contribution but we will be adding this in #651

PR #620, opened 2018-02-10
Release 3.5.1

MinnDevelopment:


MinnDevelopment:
This will be merged and released on the 13th of February, 2018.

DV8FromTheWorld:
Why are we bothering to support a `getSize` variant for longs instead of just making the normal RichPresence getSize return a long?

MinnDevelopment:
The limit of the party size values are `2**63` (signed long) according to jake, this limit was used just yesterday and broke parsing for several bots which relied on 32 bits.
Changing `getSize()` to return int instead of long is a breaking change:
```java
int size = party.getSize(); // compiler error, incompatible types. required: int found: long
```

MinnDevelopment:
In addition to that, discord apparently provides these values in Strings once they exceed 32 bit ranges to be compatible with their javascript parts, this ultimately broke the JSONObject#getInt parsing.

PR #623, opened 2018-02-13
Add getSelfUser to ShardManager.

Blayyke:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This PR allows getting the SelfUser from ShardManager, similar to ShardManager#getApplicationInfo.

### DIFF
--- a/src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
+++ b/src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
@@ -120,7 +120,7 @@ public interface ShardManager
         return this.getShardCache().stream()
                 .findAny()
                 .orElseThrow(() -> new IllegalStateException("no active shards"))
-                .getSelfUser()
+                .getSelfUser();
     }
 
     /**
--- a/src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
+++ b/src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
@@ -30,6 +30,7 @@ import net.dv8tion.jda.core.requests.RestAction;
 import net.dv8tion.jda.core.utils.Checks;
 import net.dv8tion.jda.core.utils.cache.CacheView;
 import net.dv8tion.jda.core.utils.cache.SnowflakeCacheView;
+import net.dv8tion.jda.core.entities.SelfUser;
 
 /**
  * This class acts as a manager for multiple shards.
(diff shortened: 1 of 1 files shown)


kantenkugel:
This isn't as easy as you make it look like. While your code does return a `SelfUser` instance, its just ONE of many (different) `SelfUser` instances because every shard has its own `SelfUser`.

I personally also don't see what the benefit of this would be opposed to just getting the `SelfUser` from the JDA insance you are currently working in (as every JDA class has `getJDA()`)

Blayyke:
Ah, didn't realize shards had different SelfUsers. Assumed they were all the same provided the user wasn't different on any shards. A use-case I had for this was updating bot stats on the two of the bot lists, where i needed the bot's ID.

kantenkugel:
Well yea... they would equal() to each other and have same id, but they are different instances of SelfUser. Maybe Im just making it looks worse as it is. I'll wait what other devs say to this. might just be accepted, as long as docs note that you shouldn't use Shard specific methods like getJDA() on the SelfUser

MinnDevelopment:
Can I get a user-case for this?

src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
```diff
@@ -104,6 +105,22 @@ default int getShardsTotal()
...
+     *         If there is no running shard
+     *
+     * @return The {@link net.dv8tion.jda.core.entities.SelfUser} for the currently logged in account.
+     */
+    default SelfUser getSelfUser()
+    {
+        return this.getShardCache().stream()
+                .findAny()
+                .orElseThrow(() -> new IllegalStateException("no active shards"))
+                .getSelfUser()
```
MinnDevelopment: I somehow don't believe this would compile

Blayyke: Don't know how I missed that one, will fix.

Blayyke:
One use-case was in my message above

MinnDevelopment:
current way:
```java
updateStats(selfUser);

public void updateStats(SelfUser user) {
    ShardManager manager = user.getJDA().getShardManager();
}
```
vs suggested way:
```java
updateStats(manager);

public void updateStats(ShardManager manager) {
     SelfUser self = manager.getSelfUser();
}
```

I see no benefit here?

Blayyke:
I send a stat update on bot startup, so I didn't really have some proper way of getting the SelfUser. Also thought it may fit because getApplicationInfo exists.

WalshyDev:
Maybe instead of this just add getId() and getIdLong() for the SelfUser, this would still suite the use-case and mean you don't have to just get the first JDA instance every-time in order to get the ID.

Old way
```java
long id = shardManager.getShards().get(0).getSelfUser().getIdLong()
```

New way
```java
long id = shardManager.getIdLong()
```

I would like to see a getSelfUser though from the first connected shard it can find, the reason being if let's say shard 0 goes down and you're using the bots avatar in embeds then it can just die until the shard reconnects.

Something like
```java
public SelfUser getSelfUser() {
        for (JDA jda : getShardCache())
            if (jda.getStatus() == JDA.Status.CONNECTED)
                return jda.getSelfUser();
        return null;
    }
```

Would be nice but I'm not sure if this would be the best and most reliable way.

MinnDevelopment:
We decided that this is not needed. You should implement this yourself if you need this functionality.

```java
public static SelfUser getSelfUser(ShardManager manager) {
    return manager.getShardCache().stream()
            .filter( (shard) -> shard.getStatus() == JDA.Status.CONNECTED)
            .map(JDA::getSelfUser)
            .findAny().orElseThrow(IllegalStateException::new);
}
```

PR #627, opened 2018-02-23
Fallback body parsing for http response

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

Allow fallback for error response with non-json content again.
This fixes a small regression from 3.5.1 where the response is not parsed for error responses that aren't json. Additionally this removes an unused JSONObject in PermissionOverrideAction which isn't present for this endpoint and causes issues.


MinnDevelopment:
I have added `putPermissionOverride` here which allows to replace the existing overrides.

## Old way

```java
public void setPerms(Channel channel, Member member, long allow, long deny) {
    PermissionOverride override = channel.getPermissionOverride(member);
    if (override == null) {
        channel.createPermissionOverride(member).setAllow(allow).setDeny(deny).queue();
        return;
    }
    override.getManager().setAllow(allow).setDeny(deny).queue();
}
```

## New way

```java
public void setPerms(Channel channel, Member member, long allow, long deny) {
    channel.putPermissionOverride(member).setAllow(allow).setDeny(deny).queue();
}
```

PR #629, opened 2018-02-24
ISO 2 country codes

21Joakim:

## Description
Added ISO 3166-1 alpha-2 for the regions that is in a country or is a country.

This might be a pointless (maybe even bad) addition to the Region, see this as a suggestion.


MinnDevelopment:
I don't see why this would be needed considering the regions aren't really related to countries but _regions_ as the name suggests.

DV8FromTheWorld:
I appreciate the effort, but as Minn has mentioned, these are audio voice regions, and do not necessarily correspond to countries even if Discord has attached country labels to them.

PR #634, opened 2018-02-27
Make mentions distinct

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This makes sure we only add users/roles/channels once to the list


DV8FromTheWorld:
It would be cheaper memory-wise to do a `.contains` check before adding rather than making a `Set` _and_ making a `List`

PR #637, opened 2018-02-28
Restructure GenericMessageEvent

WalshyDev:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description
This is a simple PR, this just adds `getGuild()` to GenericMessageEvent since it is just added in implementations.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/events/message/GenericMessageEvent.java
+++ b/src/main/java/net/dv8tion/jda/core/events/message/GenericMessageEvent.java
@@ -17,7 +17,9 @@ package net.dv8tion.jda.core.events.message;
 
 import net.dv8tion.jda.core.JDA;
 import net.dv8tion.jda.core.entities.ChannelType;
+import net.dv8tion.jda.core.entities.Guild;
 import net.dv8tion.jda.core.entities.MessageChannel;
+import net.dv8tion.jda.core.entities.TextChannel;
 import net.dv8tion.jda.core.events.Event;
 
 /**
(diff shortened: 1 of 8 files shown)


src/main/java/net/dv8tion/jda/core/events/message/MessageBulkDeleteEvent.java
```diff
@@ -46,11 +46,6 @@ public TextChannel getChannel()
         return channel;
     }
 
-    public Guild getGuild()
```
MinnDevelopment: Do you realize that this event doesn't extend the GenericMessageEvent?

MinnDevelopment:
While you are at it, why not also move all the getXChannel methods up?

src/main/java/net/dv8tion/jda/core/events/message/MessageBulkDeleteEvent.java
WalshyDev: Oh whoops, reverted

src/main/java/net/dv8tion/jda/core/events/message/GenericMessageEvent.java
```diff
@@ -64,4 +68,59 @@ public ChannelType getChannelType()
...
+     * <br>If this Message was not received in a {@link net.dv8tion.jda.core.entities.TextChannel TextChannel},
+     * this will return {@code null}.
+     *
+     * @return The Guild the Message was received in or null if not from a TextChannel
+     *
+     * @see net.dv8tion.jda.core.events.message.GenericMessageEvent#isFromType(ChannelType)
+     */
+    public Guild getGuild()
+    {
+        return isFromType(ChannelType.TEXT) ? ((TextChannel) channel).getGuild() : null;
```
MinnDevelopment: change this to getTextChannel().getGuild()

PR #639, opened 2018-03-03
Allow EmbedBuilder to accept an Object field value

KingFaris10:
## Pull Request Etiquette

There are several guidelines you should follow in order for your Pull Request to be merged.

  can **rebase** your branch.

## Description

Allow the value type to be an Object because it won't break existing calls to this method and only add an extra method call (toString()) on any strings. An example of where this is useful is a field that displays an int like "Guilds" with value "9001". The developer can easily type: `embedBuilder.addField("Guilds", String.valueOf(getGuildCount()), false);`
But it's a lot cleaner for the developer to just type: `embedBuilder.addField("Guilds", getGuildCount(), false)`


MinnDevelopment:
Rejected, this is not a good design pattern.

PR #645, opened 2018-03-13
WebSocketClient optimization/enhancements

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

We now use the same buffer for decompression in order to use less heap and produce less garbage on event handling, additionally I replaced the rate-limit counter with an `AtomicInt`


napstr:
Been using this for ~36 hours
Here's a graph of message value (decent indicator of overall activity) vs garbage collection times on a 60k guilds bots:

websocket stats
1. was a restart including several dependency bumps (including then latest JDA) after about 20 days of uptime
2. was deployment of this branch (and nothing else).

Garbage collector time is down, both in comparison to continuos message volume, as well as in comparison to garbage collector times of the last week on the same weekdays.
NIce work!

PR #647, opened 2018-03-16
Allow plugging in custom ratelimiter pools

napstr:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

The current implementation of the ratelimiter creates 5 threads per shard to execute RestActions.
This results in enormous amount of threads for big bots, for example, FredBoat has almost 4k ratelimiter threads sitting around. There must be a less resource intensive way to output a maximum of 50 requests / second. Okhttp provides async callbacks, but that just shifts the problem, as they use an internal threadpool themselves. A super clean solution would be using an NIO based http client, but implementing that would be a lot of work given how JDA integrates with many okhttp classes.
So instead, this PR implements an unperfect, but rather cheap solution of the problem by allowing bot owners to plug in their own thread pools for the ratelimiters.

### DIFF
--- a/src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
+++ b/src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
@@ -39,6 +39,7 @@ import javax.security.auth.login.LoginException;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Function;
 import java.util.function.IntFunction;
 
 /**
@@ -207,6 +208,11 @@ public class DefaultShardManager implements ShardManager
     protected boolean enableMDC;
 
     /**
+     * Optional provider for threadpool(s) of ratelimiters
+     */
+    protected Function<JDA, ScheduledThreadPoolExecutor> ratelimiterPool;
+
+    /**
      * Creates a new DefaultShardManager instance.
      * @param  shardsTotal
      *         The total amount of shards or {@code -1} to retrieve the recommended amount from discord.
@@ -259,6 +265,8 @@ public class DefaultShardManager implements ShardManager
      *         Whether MDC should be enabled
      * @param  contextProvider
      *         The MDC context provider new JDA instances should use on startup
+     * @param  ratelimiterPool
+     *         Optional provider for threadpool(s) of ratelimiters
      */
     protected DefaultShardManager(final int shardsTotal, final Collection<Integer> shardIds,
                                   final SessionController controller, final List<Object> listeners,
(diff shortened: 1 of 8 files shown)


MinnDevelopment:
Do you have any benchmarks that show an improvement coming from this change? Threadcount by itself is not really resource intensive.

kantenkugel:
I do agree that just number of threads is pretty irrelevant... if they are just idle/parked, they don't take up any resources.

On the other hand, if a lot of them are active at the same time that just means that your bot has an immense throughput of RestActions and reducing thread count would just slow down/kill your bot.
At that point, it would be better to just split up the shards into multiple machines instead of trying to optimize a server overloaded by too many shards

napstr:
Alright. I did a bit of profiling, and also ran a bot that would usually use about 60 pools for a total of 300 ratelimiter threads on this branch with a single pool of 50 threads.

Profiling:
Turns out the threads have a negligible effect on memory consumption. Previous online source would talk about thread stacks consuming up to 1mb of memory, but apparently that's not the case. I ran a pool of 4k threads doing 50 http requests per second via okhttp that would be answered after about 300ms, and the whole JVM was just short of 40mb of memory for that after several minutes of activity.

Bot:
There were absolutely not negative effects measured on the response time running with a pool of 50 threads instead of 300. Didn't notice any positive ones though either which I was hoping for (see paragraph above about profiling).

I still think it's a bit of a smell to have (in FredBoats case) almost 4k threads idling just to manage 50 requests / second. I understand that the vast majority of users of this lib will never have to deal with that though, and this PR would expose yet another part of the inner workings that would need to be maintained and thought of when doing future changes, so not adjusting the current implementation is a perfectly fine decision. I'll leave it up to you to close this.

Maybe there's a compomise possible here: What about using okhttp's callbacks via enqueue() that use an internal threadpool (Dispatcher), which users of JDA could adjust via the OkHttpClient.Builder? Is this a thing that you have discussed already, and what was the conclusion?

MinnDevelopment:
>There were absolutely not negative effects measured on the response time running with a pool of 50 threads instead of 300.

You should notice a negative effect when using more than 5 different routes at once on one shard, the pool will notice that no threads are available and will create a new one, causing slowdown and allocations.
>I still think it's a bit of a smell to have (in FredBoats case) almost 4k threads idling just to manage 50 requests / seconds

Idle threads are no issue to the JVM because they are not given CPU time, the only affect they hold is their stack which is marginal 
>What about using okhttp's callbacks via enqueue() that use an internal threadpool (Dispatcher), which users of JDA could adjust via the OkHttpClient.Builder?

I do not feel the need to change our current system, nor do I have the time or interest.

napstr:
> the pool will notice that no threads are available and will create a new one
```
pool.allowCoreThreadTimeOut(false);
pool.prestartAllCoreThreads();
```

> I do not feel the need to change our current system, nor do I have the time or interest.

Fair enough. Can I assume this means that the JDA team is not interested in accepting a PR changing the requester/ratelimiter to make use of okhttp's enqueue()?

MinnDevelopment:
>Can I assume this means that the JDA team is not interested in accepting a PR changing the requester/ratelimiter to make use of okhttp's enqueue()?

I can only speak for myself, if you want to take this up with @DV8FromTheWorld feel free to do so, although I imagine we share the same opinion on this topic.

MinnDevelopment:
I'll close this due to inactivity, if you want to contact DV8 about a proposal to change the rate limiter internals you can do this via either an issue or discord.

PR #648, opened 2018-03-16
Basic Java 9 module support.

gudenau:
I added some extremely basic module support for Java 9.

The module-info file exports everything except for the impl packages and imports the javax annotations and java.desktop.


Sanduhr32:
You should not change the JVM Bytecode level that JDA uses because most people use J8 (52) and if JDA compiles to J9 Bytecode (53) many can't use JDA. Also join the support guild there is a spectra tag for the J9 setup.

GrandPanda:
Changing the minimum compatible Java version is a massive change that is likely not welcome, as pointed out previously. JDA can function just fine in the unnamed module for now.

MinnDevelopment:
1) We do not accept PRs against `master`
2) JDA already works on java 9 [example](https://github.com/JDA-Applications/jda-sample-jdk9)
3) We are not moving to java 9 as the majority of our user-base uses java 8.
4) I do not see a reason why we should move to java 9 at all.

gudenau:
Guess I will close this then, figured that would be okay since Java9 has been out for a while now and it does not work in my environment without the module info.

PR #659, opened 2018-04-09
Externalize opus binaries

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This should make the artifact easier to exclude and allows for a much smaller dependency.
Additionally we could offer audio support without the opus binaries (like lavaplayer which already does the opus encoding).

PS: ~~We have to also publish the opus artifact to jcenter.~~ https://bintray.com/minndevelopment/maven/opus-java


MinnDevelopment:
@DV8FromTheWorld Could you give some comments on this?

ArsenArsen:
You could also allow people to use the opus distributions that come with their systems (eg. `libopus0` on Debian repositories, which provides `/usr/lib/libopus.so.0`)

src/main/java/net/dv8tion/jda/core/audio/AudioNatives.java
```diff
@@ -22,12 +22,24 @@
...
     private static final Logger LOG = LoggerFactory.getLogger(AudioNatives.class);
     private static boolean initialized;
     private static boolean audioSupported;
 
+    private AudioNatives() {}
+
+    public static synchronized boolean loadFrom(String absolutePath)
+    {
+        if (initialized)
+            return false;
```
DV8FromTheWorld: should this return `audioSupported?`

MinnDevelopment: `ensureOpus()` returns `audioSupported` because it is used by JDA to check whether the library is available or not. `loadFrom()` returns true if the library is loaded from that method invokation and false if not (it was already loaded).

```java
if (AudioNatives.loadFrom("/lib/.../libopus.so")) {
    log.info("Loaded custom library at /lib/.../libopus.so");
} else {
    log.info("Opus binaries were already loaded, skipping.");
}
```

PR #660, opened 2018-04-10
Proper closing of resources in MessageAction

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

Added the following methods

* clearFiles(Consumer)
* clearFiles(BiConsumer)

This should allow users to close any of their provided resources,
additionally clearFiles() will now close any opened FileInputStreams!

I've added a note to the class-level documentation that the execution of MessageAction closes all provided resources.
You cannot send the same resource twice without copying it first!


src/main/java/net/dv8tion/jda/core/requests/restaction/MessageAction.java
```diff
@@ -49,6 +49,8 @@
  * <br>This can be used to remove existing embeds from a message:
  * <br>{@code message.editMessage("This message had an embed").override(true).queue()}
  *
+ * <p><u>When this RestAction has been executed all provided files will be closed.</u></p>
```
kantenkugel: I would rephrase this to something like: "If you provide Files to this builder but decide not to actually send this message, use clearFiles() to clean up their resources. See description of sendFile(File)"

src/main/java/net/dv8tion/jda/core/requests/restaction/MessageAction.java
```diff
@@ -337,6 +340,7 @@ public MessageAction appendFormat(final String format, final Object... args)
 
     /**
      * Adds the provided {@link java.io.InputStream InputStream} as file data.
+     * <br><u>The stream will be closed upon execution!</u>
```
kantenkugel: When are given input streams actually closed? in the sending code?

src/main/java/net/dv8tion/jda/core/requests/restaction/MessageAction.java
```diff
@@ -436,6 +440,7 @@ public MessageAction addFile(final File file)
 
     /**
      * Adds the provided {@link java.io.File File} as file data.
+     * <br><u>The file stream will be closed upon execution!</u>
```
kantenkugel: Would rephrase this similar to the suggestion of class definition ("this method creates a stream for reading the file, which should be cleaned up via clearFiles()")

src/main/java/net/dv8tion/jda/core/requests/restaction/MessageAction.java
```diff
@@ -480,13 +487,67 @@ public MessageAction addFile(final File file, final String name)
...
+     *
+     * @see    #clearFiles(Consumer)
+     * @see    #clearFiles(BiConsumer)
      */
     @CheckReturnValue
     public MessageAction clearFiles()
     {
         files.clear();
+        clearResources();
+        return this;
```
kantenkugel: Would by default close un-owned streams too (knowing ppl won't do it manually) and add an override with boolean parameter closeUnowned or something. same goes for other clearFiles() methods

src/main/java/net/dv8tion/jda/core/requests/restaction/MessageAction.java
```diff
@@ -517,6 +595,9 @@ protected RequestBody asMultipart()
         }
         if (!isEmpty())
             builder.addFormDataPart("payload_json", getJSON().toString());
+        // clear remaining resources, they will be closed after being sent
+        files.clear();
+        ownedResources.clear();
```
kantenkugel: not sure if this is only called by actual execution of the restaction (and if that will cause the MultipartBody to close all given streams)

src/main/java/net/dv8tion/jda/core/requests/restaction/MessageAction.java
MinnDevelopment: When they are read by the RequestBody it closes the stream.

src/main/java/net/dv8tion/jda/core/requests/restaction/MessageAction.java
MinnDevelopment: The body does close the stream, it is only called on execution.

src/main/java/net/dv8tion/jda/core/requests/restaction/MessageAction.java
MinnDevelopment: I don't want to close streams that were never used

kantenkugel: Yea but that would mean if a user uses channel.sendFile(someInputStream, "name") and never actually activates the restaction (or just calls clearFiles()), they would have to manually close the inputstream on their end. I think it would be nice if we would close any inputstream we get if they are either sent or the MessageAction is "canceled" (clearFiles).

Therefore i also suggested that rewrite of the docs (and probably add a big warning)

MinnDevelopment: If the user wants to close _all_ streams they can do so with `clearFiles(Consumer)`

kantenkugel: oh right... add that to some jdocs tho

PR #665, opened 2018-04-20
Permission#VOICE_MOVE_OTHERS does not grant access to a voice channel

21Joakim:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description
I noticed that my bot would sometimes not join a voice channel nor would it throw an exception and after investigating and testing I found out that the Move Members permission was causing this and that is what led me to this line of code.
As stated in the title the Move Members permission does not grant a user access to a channel only the Connect permission does.

**Note:** I have not tested this yet.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/managers/impl/AudioManagerImpl.java
+++ b/src/main/java/net/dv8tion/jda/core/managers/impl/AudioManagerImpl.java
@@ -35,8 +35,10 @@ import net.dv8tion.jda.core.managers.AudioManager;
 import net.dv8tion.jda.core.utils.Checks;
 import net.dv8tion.jda.core.utils.MiscUtil;
 import net.dv8tion.jda.core.utils.NativeUtil;
+import net.dv8tion.jda.core.utils.PermissionUtil;
 
 import java.io.IOException;
+import java.util.EnumSet;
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
When this was tested this was enough to join a voice channel when the userlimit is reached.

21Joakim:
What do you mean?

21Joakim:
The Move Members permission does give a member access to connect to a channel when the user limit is reached (Which is also tested for in AudioManagerImpl#checkUserlimit) but only if they have the base Connect permission first

MinnDevelopment:
Oh it looks like this permission check was moved to a wrong spot at some point...
This should only be checked for userlimit.

WalshyDev:
When the voice channel has hit the member limit (the channel is full or beyond full) then the VOICE_MOVE_OTHERS permission allows users and bots to still join. 

It seems like it is a good move to remove from here though as this permission is checked (correctly - checking the members currently there) is below.

MinnDevelopment:
It would be preferred to only do a permission check when the bot is not currently connected to the channel.

MinnDevelopment:
Would this [patch](https://gist.github.com/MinnDevelopment/7222373b5c62c43a5973340c07473e94) be a reasonable change?

PR #671, opened 2018-04-24
Add unicode flag emoji for each Region

jagrosh:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This PR adds the flag unicode for each Region, as well as a Region#getEmoji() method to retrieve said flag. Every voice region is represented by a flag within Discord's client:
Flags
(Some flags are from past versions of the client, as available voice regions change occasionally). Similar to how JDA offers the in-client name for Region (and other enums, like Permission), this would allow JDA to also offer the in-client 'graphic' for regions. The use case is similar as well; bot and application developers could include the flag in output for a guild's information (even outside of Discord, since it is unicode), similar to how they might include the in-client region name. Alternatively, the flag unicode could be used as a reaction to a message.
Flag unicode provided by @ithinkimcats .

Currently, UNKNOWN is set to the 🏳️ (`:white_flag:`) emoji, but this could be changed to `null` or an empty string.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/Region.java
+++ b/src/main/java/net/dv8tion/jda/core/Region.java
@@ -22,46 +22,48 @@ package net.dv8tion.jda.core;
  */
 public enum Region
 {
-    JAPAN("japan", "Japan", false),
-    AMSTERDAM("amsterdam", "Amsterdam", false),
-    BRAZIL("brazil", "Brazil", false),
-    EU_WEST("eu-west", "EU West", false),
-    EU_CENTRAL("eu-central", "EU Central", false),
-    FRANKFURT("frankfurt", "Frankfurt", false),
-    HONG_KONG("hongkong", "Hong Kong", false),
-    LONDON("london", "London", false),
-    RUSSIA("russia", "Russia", false),
-    SINGAPORE("singapore", "Singapore", false),
-    SYDNEY("sydney", "Sydney", false),
-    US_EAST("us-east", "US East", false),
-    US_WEST("us-west", "US West", false),
-    US_CENTRAL("us-central", "US Central", false),
-    US_SOUTH("us-south", "US South", false),
+    JAPAN("japan", "Japan", "\uD83C\uDDEF\uD83C\uDDF5", false),
+    AMSTERDAM("amsterdam", "Amsterdam", "\uD83C\uDDF3\uD83C\uDDF1", false),
+    BRAZIL("brazil", "Brazil", "\uD83C\uDDE7\uD83C\uDDF7", false),
+    EU_WEST("eu-west", "EU West", "\uD83C\uDDEA\uD83C\uDDFA", false),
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/core/Region.java
```diff
@@ -22,46 +22,48 @@
...
+    VIP_LONDON("vip-london", "London (VIP)", "\uD83C\uDDEC\uD83C\uDDE7", true),
+    VIP_SINGAPORE("vip-singapore", "Singapore (VIP)", "\uD83C\uDDF8\uD83C\uDDEC", true),
+    VIP_SYDNEY("vip-sydney", "Sydney (VIP)", "\uD83C\uDDE6\uD83C\uDDFA", true),
+    VIP_US_EAST("vip-us-east", "US East (VIP)", "\uD83C\uDDFA\uD83C\uDDF8", true),
+    VIP_US_WEST("vip-us-west", "US West (VIP)", "\uD83C\uDDFA\uD83C\uDDF8", true),
+    VIP_US_CENTRAL("vip-us-central", "US Central (VIP)", "\uD83C\uDDFA\uD83C\uDDF8", true),
+    VIP_US_SOUTH("vip-us-south", "US South (VIP)", "\uD83C\uDDFA\uD83C\uDDF8", true),
 
-    UNKNOWN("", "Unknown Region", false);
+    UNKNOWN("", "Unknown Region", "\uD83C\uDFF3", false);
```
MinnDevelopment: Question: What do we use when a new region is added that can't be associated with a flag?

MinnDevelopment: Ok, I think we should settle on using null instead for missing flags.

DV8FromTheWorld:
The code looks fine, but I'm not quite sure I feel about this PR in general. Technically the region isn't specifically represented by a geolocational area, and technically they could abandon using Emojis as the representation for the location at some point in the future (perhaps higher-res images or custom ones?) This PR has JDA take on information that isn't really specific to the API (pure information) and instead it is now specifically trying to make visualization of information "easier".

The question is: Is this really in the scope of the API? To me, this should be handled by the user.

Open to thoughts here. Very much 50/50, so opinions required before pull.

MinnDevelopment:
>This PR has JDA take on information that isn't really specific to the API (pure information)

This was already done before, see the "readable names for voice regions and permissions". 

>Technically the region isn't specifically represented by a geolocational area,

Currently all regions correspond to geolocational areas, if this would in-fact change at some point we use `null` to represent missing flags (see UNKNOWN). If by some means these regions are not at all geo-locations anymore in the future, we would deprecate this approach. However if that would happen we would have to do put work into this enum either way.

> they could abandon using Emojis as the representation for the location at some point in the future (perhaps higher-res images or custom ones?) 

Just because Discord decides to use custom images does not mean we have to stop supporting emoji flags as they are still correct.

PR #674, opened 2018-04-27
Improve gradle builds and debugging

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

- incremental compilation
- improved ABI stability
- async stack traces in intellij debugging

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,6 +52,7 @@ dependencies {
 
     //Code safety
     api 'com.google.code.findbugs:jsr305:3.0.2'
+    api 'org.jetbrains:annotations:16.0.1'
 
     //Logger
     api 'org.slf4j:slf4j-api:1.7.25'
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -200,20 +207,21 @@ publishing {
...
-    if (hasProperty(propertyName))
-    {
+    if (hasProperty(propertyName)) {
         property = this.properties[propertyName]
     }
     return property
 }
 
-wrapper.gradleVersion = '4.6'
+wrapper {
```
Almighty-Alpaca: Can we surround this with a check if the wrapper tasks exists?
```java
if (tasks.findByName('wrapper')) {
    wrapper {
        gradleVersion = '4.9'
    }
}
```
The reason for this is that the wrapper tasks only exists on root projects in gradle. If you include JDA as a subproject (I have it as subproject for my test bot for example) even loading the project already fails with 
```Could not get unknown property 'wrapper' for project ':JDA' of type org.gradle.api.Project.```

Issue #662, opened 2018-04-20
Parse webhooks in audit log entries


MinnDevelopment:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion

## General Troubleshooting
  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

### Issue Type

- [ ] Bug Report
- [x] Feature Request

### Description

The audit log pagination gives details about attached webhooks which is currently not parsed by JDA.
These webhooks are partial and need to be handled in a special way (no token -> no clients).

PR #675, opened 2018-04-28
Allow null user_id and provide partial Webhooks

jagrosh:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This PR fixes an issue where JDA required the user_id field in audit log entries to be non-null, despite Discord allowing null values for items such as webhook self-deletion. This PR also adds a getWebhook() method to AuditLogEntry to get a partial Webhook object that are the target of some Audit Log Entries. This fulfills #662.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/audit/AuditLogEntry.java
+++ b/src/main/java/net/dv8tion/jda/core/audit/AuditLogEntry.java
@@ -27,6 +27,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import net.dv8tion.jda.core.entities.Webhook;
+import net.dv8tion.jda.core.entities.impl.WebhookImpl;
 
 /**
  * Single entry for an {@link net.dv8tion.jda.core.requests.restaction.pagination.AuditLogPaginationAction
@@ -41,20 +43,22 @@ public class AuditLogEntry implements ISnowflake
     protected final long targetId;
     protected final GuildImpl guild;
     protected final UserImpl user;
+    protected final WebhookImpl webhook;
     protected final String reason;
 
     protected final Map<String, AuditLogChange> changes;
     protected final Map<String, Object> options;
     protected final ActionType type;
 
-    public AuditLogEntry(ActionType type, long id, long targetId, GuildImpl guild, UserImpl user, String reason,
-                         Map<String, AuditLogChange> changes, Map<String, Object> options)
+    public AuditLogEntry(ActionType type, long id, long targetId, GuildImpl guild, UserImpl user, WebhookImpl webhook, 
+                        String reason, Map<String, AuditLogChange> changes, Map<String, Object> options)
     {
         this.type = type;
         this.id = id;
         this.targetId = targetId;
         this.guild = guild;
         this.user = user;
+        this.webhook = webhook;
         this.reason = reason;
         this.changes = changes != null && !changes.isEmpty()
                 ? Collections.unmodifiableMap(changes)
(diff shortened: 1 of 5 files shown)


src/main/java/net/dv8tion/jda/core/audit/AuditLogEntry.java
```diff
@@ -27,6 +27,8 @@
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import net.dv8tion.jda.core.entities.Webhook;
+import net.dv8tion.jda.core.entities.impl.WebhookImpl;
```
MinnDevelopment: nit: these imports should be sorted properly.

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -1179,17 +1179,25 @@ public Webhook createWebhook(JSONObject object)
...
+            if (owner == null)
+            {
+                ownerJson.put("id", userId);
+                owner = createFakeUser(ownerJson, false);
+            }
+        }
+        
+        return new WebhookImpl(channel, id)
+                .setToken(token)
+                .setOwner(owner==null ? null : channel.getGuild().getMember(owner))
```
MinnDevelopment: nit: give the operators such as `==` some space

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -1375,7 +1383,8 @@ public AuditLogEntry createAuditLogEntry(GuildImpl guild, JSONObject entryJson,
         final JSONObject options = entryJson.isNull("options") ? null : entryJson.getJSONObject("options");
         final String reason = entryJson.optString("reason", null);
 
-        final UserImpl user = (UserImpl) createFakeUser(userJson, false);
+        final UserImpl user = userJson == null ? null : (UserImpl) createFakeUser(userJson, false);
+        final WebhookImpl webhook = webhookJson == null ? null : (WebhookImpl) createWebhook(webhookJson);
```
MinnDevelopment: since these two methods always return the specific implementation it would make more sense to just change the return type rather than doing a typecast

src/main/java/net/dv8tion/jda/core/entities/impl/WebhookImpl.java
```diff
@@ -155,6 +155,9 @@ public WebhookManager getManager()
     @Override
     public WebhookClientBuilder newClient()
     {
+        if (token == null)
+            throw new IllegalStateException("Webhooks without known tokens (such as those retrieved from Audit Logs) "
```
MinnDevelopment: add this to the documentation

src/main/java/net/dv8tion/jda/core/audit/AuditLogEntry.java
```diff
@@ -108,7 +122,7 @@ public Guild getGuild()
      * The {@link net.dv8tion.jda.core.entities.User User} responsible
      * for this action.
      *
-     * @return The User instance
+     * @return Possibly-null User instance
      */
     public User getUser()
```
MinnDevelopment: Maybe we should add an `@Nullable` to these

TheMonitorLizard: Addition of the `@Nullable` annotation breaks pre-existing kotlin code that assumes not-null.

MinnDevelopment: Bad code to assume things it shouldn't be assuming.

src/main/java/net/dv8tion/jda/core/entities/Webhook.java
```diff
@@ -57,11 +58,12 @@
     TextChannel getChannel();
 
     /**
-     * The owner of this Webhook.
+     * The owner of this Webhook. This will be null for partial Webhooks, such as those retrieved from Audit Logs.
```
MinnDevelopment: We should mark `Webhook` as `IFakable` for this scenario. Fake webhooks would not be able to use `newClient()` or `getManager()`.

src/main/java/net/dv8tion/jda/core/entities/impl/WebhookImpl.java
MinnDevelopment: `getManager()` should fail in the same way.

jagrosh: I think we should always throw an exception for a fake webhook for this method. I would suggest a `WebhookClientBuilder#newClient(String token)` but that would be confusing for non-fake webhooks.

Issue #666, opened 2018-04-21
Allow to retrieve approximate counts for invites


MinnDevelopment:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion

## General Troubleshooting
  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

### Issue Type

- [ ] Bug Report
- [x] Feature Request

### Description

Discord added approximate presence and member count information to invites, we should add an option to resolve that.

PR #678, opened 2018-04-29
Additional invite info

jagrosh:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This PR fulfills #666 by adding an additional Invite.resolve method to allow getting counts and verification level from invites.
Still needed at this time: docs, possibly exceptions if the resolved invite is not one that contains counts.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
@@ -28,6 +28,7 @@ import net.dv8tion.jda.core.WebSocketCode;
 import net.dv8tion.jda.core.audit.ActionType;
 import net.dv8tion.jda.core.audit.AuditLogChange;
 import net.dv8tion.jda.core.audit.AuditLogEntry;
+import net.dv8tion.jda.core.entities.Guild.VerificationLevel;
 import net.dv8tion.jda.core.entities.MessageEmbed.*;
 import net.dv8tion.jda.core.entities.impl.*;
 import net.dv8tion.jda.core.exceptions.AccountTypeException;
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -1260,8 +1261,11 @@ public Invite createInvite(JSONObject object)
         final long guildId = guildObject.getLong("id");
         final String guildName = guildObject.getString("name");
         final String guildSplashId = guildObject.optString("splash", null);
+        final VerificationLevel guildVerificationLevel = VerificationLevel.fromKey(guildObject.optInt("verification_level", -1));
+        final int presenceCount = object.optInt("approximate_presence_count", -1);
+        final int memberCount = object.optInt("approximate_member_count", -1);
```
MinnDevelopment: Should this be in the guild object if discord doesn't provide them in the guild object?

src/main/java/net/dv8tion/jda/core/entities/Invite.java
```diff
@@ -23,6 +23,7 @@
 
 import javax.annotation.CheckReturnValue;
 import java.time.OffsetDateTime;
+import net.dv8tion.jda.core.entities.Guild.VerificationLevel;
```
MinnDevelopment: import order

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
jagrosh: It's not provided in the guild object, but from a library user's standpoint I think it might make more sense. Either way would be fine from an implementation standpoint though.

MinnDevelopment: I'm fine with either, too

PR #680, opened 2018-05-05
Ability to add members to guilds

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This will allow bots to add members to guilds, assuming they acquired authentication via OAuth2 `guilds.join` scope.

I have tested this with [implicit grant](https://discordapp.com/developers/docs/topics/oauth2#implicit-grant).


src/main/java/net/dv8tion/jda/core/requests/restaction/MemberAction.java
```diff
@@ -0,0 +1,130 @@
+package net.dv8tion.jda.core.requests.restaction;
```
MinnDevelopment: missing copyright header

PR #703, opened 2018-07-03
Added Guild Features on Invite

Artuto:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This PR adds a method to get the features of a guild on an invite.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
@@ -1301,7 +1301,7 @@ public class EntityBuilder
         final VerificationLevel guildVerificationLevel = VerificationLevel.fromKey(Helpers.optInt(guildObject, "verification_level", -1));
         final int presenceCount = Helpers.optInt(object, "approximate_presence_count", -1);
         final int memberCount = Helpers.optInt(object, "approximate_member_count", -1);
-        final JSONArray guildFeatures = guildObject.getJSONArray("features");
+        final Set<String> guildFeatures = StreamSupport.stream(guildObject.getJSONArray("features").spliterator(), false).map(String::valueOf).collect(Collectors.toSet());
 
         final Invite.Guild guild = new InviteImpl.GuildImpl(guildId, guildIconId, guildName, guildSplashId, guildVerificationLevel, presenceCount, memberCount, guildFeatures);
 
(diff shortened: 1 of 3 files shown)


natanbc:
Why not create the set in EntityBuilder instead of creating a new one every time it's accessed?

Artuto:
Didn't thought of that, done.

kantenkugel:
Might be a good idea to wrap into a UnmodifiableSet and to make the json attribute optional (discord always gets us bad when deciding to just not send attributes sometimes)

Artuto:
Discord docs says its not optional, and should appear on every guild object, but yea, discord is.. discord

MinnDevelopment:
@Artuto the docs say in no way it should be in every guild object. Quite the opposite as their example does not include it. https://discordapp.com/developers/docs/resources/invite#invite-object-example-invite-object
The docs explicitly mention **partial** guild object

Artuto:
Oh. I'll check if its present in the guild object then.

kantenkugel:
And even if it would be in the docs... we had the case of a non-optional key being missing in the json payloads way to often, which broke JDA

Artuto:
Should be ok now.

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -1301,7 +1301,16 @@ public Invite createInvite(JSONObject object)
         final VerificationLevel guildVerificationLevel = VerificationLevel.fromKey(Helpers.optInt(guildObject, "verification_level", -1));
         final int presenceCount = Helpers.optInt(object, "approximate_presence_count", -1);
         final int memberCount = Helpers.optInt(object, "approximate_member_count", -1);
-        final Set<String> guildFeatures = StreamSupport.stream(guildObject.getJSONArray("features").spliterator(), false).map(String::valueOf).collect(Collectors.toSet());
+        final Set<String> guildFeatures;
+
+        if(guildObject.has("features"))
```
MinnDevelopment:
```java
if (guildObject.isNull("features"))
    guildFeatures = Collections.emptySet();
else
    guildFeatures = StreamSupport.stream(guildObject.getJSONArray("features").spliterator(), false).map(String::valueOf).collect(Collectors.toSet());
```

src/main/java/net/dv8tion/jda/core/entities/impl/InviteImpl.java
```diff
@@ -303,7 +304,7 @@ public GuildImpl(final long id, final String iconId, final String name, final St
             this.verificationLevel = verificationLevel;
             this.presenceCount = presenceCount;
             this.memberCount = memberCount;
-            this.features = features;
+            this.features = Collections.unmodifiableSet(features);
```
MinnDevelopment: You are possibly wrapping the `Collections.emptySet()` in an unmodifiable decorator for no reason. Instead you should do this when you actually create a set.

kantenkugel:
`isNull()` checks for non-existing as well, right? If so, this looks good to me.

Sanduhr32:
Yes @kantenkugel. It checks whether an object is null and non-existing due `Map#get(key)` of a not existing entry returns `null`.
I think you even explained me the difference between a normal null check and the null check of `isNull()` and mentioned its checks.

Issue #625, opened 2018-02-14
Specific shard stuck on LOADING_SUBSYSTEMS


Repulser:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

Hey there! Before you report a bug or suggest a new feature,
  please make sure to follow these steps first!
  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

> This is not the place to learn Java. Please refer to [StackOverflow][stack overflow]
  for your general programming questions.

### Issue Type

- [x] Bug Report
- [ ] Feature Request

> For questions join the official JDA discord server: https://discord.gg/0hMr4ce0tIl3SLv5

### Description

A specific shard is always stuck on LOADING_SUBSYSTEMS, it seems that its stuck processing the ready payload (the processingReady variable in WebSocketClient was true), tried restarting it a few times, nothing helped

WalshyDev:
Trace logs? Latest version?

Repulser:
Latest version indeed, and i have debug logs but i would rather not post them publicly

WalshyDev:
Staff can't help without more info, trace logs would be best. You can make sure to remove or censor any IPs or tokens that may be sent, what's needed is as much info as possible around and during the time of the shard failure.

Repulser:
its basically full of this with different events
```
[2018-02-14 00:22:27,549] [ DEBUG] [JDA [94 / 800] MainWS-ReadThread] n.d.j.core.requests.WebSocketClient: Caching PRESENCE_UPDATE event during init!

**Comment by Repulser** [comment]
I believe i have found the cause:
```
[ WARN ] [WebSocketClient] Got an unexpected Json-parse error. Please redirect following message to the devs:
	JSONObject["name"] not a string.
[...]

Issue #602, opened 2018-01-23
Got a GuildCreateEvent for a guild that already existed!


napstr:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

Hey there! Before you report a bug or suggest a new feature,
  please make sure to follow these steps first!
  
- [x] I have checked for similar issues.
- [ ] I have updated to the [latest JDA version] (not exactly but this was on 327)
- [x] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

> This is not the place to learn Java. Please refer to [StackOverflow][stack overflow]
  for your general programming questions.

### Issue Type

- [x] Bug Report
- [ ] Feature Request

> For questions join the official JDA discord server: https://discord.gg/0hMr4ce0tIl3SLv5

### Description

Got a bunch of
``` Got an unexpected error. Please redirect following message to the devs:
	GUILD_CREATE -> {giant ass json}
```
all at the same time. Happened right after the shard printed that it finished reloading.
Here's a log file: 
[log.2018-01-22.txt](https://github.com/DV8FromTheWorld/JDA/files/1658021/log.2018-01-22.txt)

Could be a one time discord thing since several shards disconnected and reconnected there in a short time frame.

kantenkugel:
What does the text exactly say? "reloaded"? "resumed"? "reconnected"?

napstr:
My bad, I thought I had added it to the file...had to reupload cause github wanted a .txt and used the wrong one. Here it is:
```
07:33:45.489 [JDA [2 / 35] MainWS-ReadThread] [INFO ] net.dv8tion.jda.core.JDA - Finished (Re)Loading!
07:33:45.586 [JDA [2 / 35] MainWS-ReadThread] [INFO ] space.npstr.aki.Main - user 260610606091075584 says !aki invite
07:33:45.592 [main-scheduler-t84]        [INFO ] space.npstr.aki.listings.Listing - Successfully posted bot stats to discordbots.org on attempt 1, code 200
07:33:45.606 [commandprocessor-260610606091075584] [INFO ] space.npstr.aki.Main - Executing InviteComm for user U:FunIsInfinite(260610606091075584)
[...]

Issue #512, opened 2017-10-08
Completed the setup for Guild without matching id in Ready Handler cache - no trace


napstr:
```
17:07:17.292 [JDA [1 / 2] MainWS-ReadThread] [ERROR] n.d.j.core.requests.WebSocketClient - Completed the setup for Guild: G:SNIP(SNIP) without matching id in Ready Handler cache
```

Got this error on 3.3.1_286 , while booting up shards.
There is no other log or stack trace to indicate the reason, just a bunch of these in the debuglog right before that:

```
17:07:09.482 [JDA [0 / 2] MainWS-ReadThread] [DEBUG] n.d.j.core.requests.WebSocketClient - Caching PRESENCE_UPDATE event during init!
17:07:10.533 [JDA [0 / 2] MainWS-ReadThread] [DEBUG] n.d.j.core.requests.WebSocketClient - Caching MESSAGE_DELETE event during init!
17:07:10.932 [JDA [1 / 2] MainWS-ReadThread] [DEBUG] n.d.j.core.requests.WebSocketClient - Caching PRESENCE_UPDATE event during init!
17:07:11.543 [JDA [0 / 2] MainWS-ReadThread] [DEBUG] n.d.j.core.requests.WebSocketClient - Caching MESSAGE_CREATE event during init!
17:07:11.544 [JDA [0 / 2] MainWS-ReadThread] [DEBUG] n.d.j.core.requests.WebSocketClient - Caching PRESENCE_UPDATE event during init!
17:07:11.572 [JDA [0 / 2] MainWS-ReadThread] [DEBUG] n.d.j.core.requests.WebSocketClient - Caching PRESENCE_UPDATE event during init!
17:07:14.197 [JDA [1 / 2] MainWS-ReadThread] [DEBUG] n.d.j.core.requests.WebSocketClient - Caching PRESENCE_UPDATE event during init!
17:07:15.113 [JDA [0 / 2] MainWS-ReadThread] [DEBUG] n.d.j.core.requests.WebSocketClient - Caching MESSAGE_CREATE event during init!
17:07:15.113 [JDA [0 / 2] MainWS-ReadThread] [DEBUG] n.d.j.core.requests.WebSocketClient - Caching PRESENCE_UPDATE event during init!
17:07:15.358 [JDA [0 / 2] MainWS-ReadThread] [DEBUG] n.d.j.core.requests.WebSocketClient - Caching MESSAGE_CREATE event during init!
[...]

PR #706, opened 2018-07-07
Rewrite for guild setup system

### LINKED ISSUE #691: Issues with init logic
Currently JDA requires all guilds and all of the associated entities to be loaded before starting to fire events. This has serious effects as entire shards can get stuck on a single guild refusing to connect to the gateway. In addition to that JDA can have serious cache issues during outages that can lead to memory issues due to duplicated cache entries that don't get removed.

- [ ] Guilds should not prevent startup
- [ ] Events should fire when all required entities are loaded
- [ ] Cache should be aware of duplicates
- [ ] Properly handle unavailability

We have ideas of adding a `GuildReadyEvent` in addition to `ReadyEvent` which would be useful to allow events to fire even if not all guilds are ready yet. The `ReadyEvent` would be fired when all guilds have been initialized, however events may already have fired for them.

I will redirect the issue of #625 to this issue because they are closely related to one-another but I will keep it open for now.

As I don't have a lot of free time at the moment, I cannot give any ETA on when I will find time to work on these proposals. This most likely requires a rewrite of JDA internals.

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

Fixes #691, #625, #602, #512 

These changes will make it possible to fire events for guilds without having all guilds loaded. For this I have added `GuildReadyEvent` which will fire when a guild that we received in the `READY` dispatch finished setting up. In most cases this is done as soon as a `GUILD_CREATE` arrives, others will be added for chunking.

These changes are not fully completed at this time, however I want everyone to have a chance to start looking through my changes in order to spot any issues that I might have missed so far.
 
I'm not sure what version I should bump for these changes, in all honesty I'm considering major as it affects every user of the library.


src/main/java/net/dv8tion/jda/core/handle/VoiceServerUpdateHandler.java
```diff
@@ -36,13 +36,15 @@ public VoiceServerUpdateHandler(JDAImpl api)
...
+        if (api.getGuildSetupController().isLocked(guildId))
+            return guildId;
         Guild guild = api.getGuildMap().get(guildId);
         if (guild == null)
             throw new IllegalArgumentException("Attempted to start audio connection with Guild that doesn't exist! JSON: " + content);
 
         api.getClient().updateAudioConnection(guildId, guild.getSelfMember().getVoiceState().getChannel());
 
-        if (api.getGuildLock().isLocked(guildId))
+        if (api.getGuildSetupController().isLocked(guildId))
```
MinnDevelopment: this check doesn't make sense

MinnDevelopment:
## Summary of changes

+ Added GuildSetupController/-Node
+ Events will now fire when all entities required for it are ready, not the entire cache
+ Added GuildReadyEvent to see when a guild finishes starting up (only for init, you can still use GuildJoinEvent for real joins)
+ Introduced CacheFlags for more configuration of cache, not all bots need info about user's online status
- ReadyEvent remains unchanged but is no longer the guaranteed first event to fire
- EventCache has been updated and improved, the events are now cached together with meta-information relevant to their behavior and can timeout when in cache for too long

MinnDevelopment:
I have added a status enum to `GuildSetupController` to allow for better debugging. All status changes for nodes will be tracked on `TRACE` logging level for `GuildSetupController`.

`[{GUILD_ID}] Updated status {OLD_STATUS}->{NEW_STATUS}`

Additionally you can now use `GuildSetupController#getSetupNodes(Status)` to see which nodes have which status allowing to do real runtime debugging without having to analyze logs.

kantenkugel:
After looking at the source, I think you have to add some public modifiers to methods in `GuildSetupNode` (and maybe make the class itself public?). Cuz currently they are all default visibility. So `GuildSetupController#getSetupNodes(Status)` is currently not really useful for non-internal code

MinnDevelopment:
The visibility is intentions. `GuildSetupController#getSetupNodes(Status)` is useful to see how many guilds are on that specific status.

MinnDevelopment:
I have now been able to test the edge case of guilds being unavailable during startup and coming back way later, they will properly recover and fire a `GuildReadyEvent` and will not prevent startup to complete for other guilds, the `ReadyEvent` is fired without that specific unavailable guild being ready (as intended).

MinnDevelopment:
After one more week of testing I plan to merge this on the 2nd of August (that's next sunday).

PR #709, opened 2018-07-10
Better group invite support 

Almighty-Alpaca:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

This PR adds better support for group invites.


src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -1281,38 +1281,72 @@ public Group createGroup(JSONObject groupJson)
...
-        final String guildName = guildObject.getString("name");
-        final String guildSplashId = guildObject.optString("splash", null);
-        final VerificationLevel guildVerificationLevel = VerificationLevel.fromKey(Helpers.optInt(guildObject, "verification_level", -1));
-        final int presenceCount = Helpers.optInt(object, "approximate_presence_count", -1);
-        final int memberCount = Helpers.optInt(object, "approximate_member_count", -1);
-        final Set<String> guildFeatures;
+            final JSONArray usernameArray = channelObject.optJSONArray("recipients");
+            final List<String> usernames;
+            if (usernameArray == null)
+                usernames = null;
```
MinnDevelopment: Use `Collection.emptyList()`

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -1281,38 +1281,72 @@ public Group createGroup(JSONObject groupJson)
...
+            else
+                usernames = Collections.unmodifiableList(StreamSupport.stream(usernameArray.spliterator(), false).map(String::valueOf).collect(Collectors.toList()));
 
-        if (guildObject.isNull("features"))
+            group = new InviteImpl.GroupImpl(groupIconId, groupName, groupId, usernames);
+        }
+        else if (channelType == ChannelType.TEXT || channelType == ChannelType.VOICE)
```
MinnDevelopment: Why not `channelType.isGuild()`?

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -1281,38 +1281,72 @@ public Group createGroup(JSONObject groupJson)
...
+
+            final Set<String> guildFeatures;
+            if (guildObject.isNull("features"))
+                guildFeatures = Collections.emptySet();
+            else
+                guildFeatures = Collections.unmodifiableSet(StreamSupport.stream(guildObject.getJSONArray("features").spliterator(), false).map(String::valueOf).collect(Collectors.toSet()));
+
+            guild = new InviteImpl.GuildImpl(guildId, guildIconId, guildName, guildSplashId, guildVerificationLevel, presenceCount, memberCount, guildFeatures);
+
+            final String channelName = channelObject.getString("name");
```
MinnDevelopment: Why are you using `optString` for the name when you handle the name for group and `getString` here?

src/main/java/net/dv8tion/jda/core/entities/Invite.java
```diff
@@ -367,4 +390,59 @@ default String getURL()
          */
         Set<String> getFeatures();
     }
+
+    /**
+     * POJO for the group information provided by an invite.
+     *
+     * @see #getChannel()
+     */
+    interface Group extends ISnowflake
```
MinnDevelopment: All the methods in this class seem to be nullable but not documented that way.

src/main/java/net/dv8tion/jda/core/entities/Invite.java
```diff
@@ -367,4 +390,59 @@ default String getURL()
...
+         */
+        List<String> getUsers();
+    }
+
+    /**
+     * Enum representing the type of an invite.
+     *
+     * @see #getType()
+     */
+    enum Type
```
MinnDevelopment: InviteType? I wouldn't use the name `Type` because it is also used by java reflections

MinnDevelopment:
Any update on this?

src/main/java/net/dv8tion/jda/core/entities/Invite.java
Almighty-Alpaca: You can already uniquely reference it by using `Invite.Type`. Having `Invite.InviteType` just feels wrong imo.

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
Almighty-Alpaca: I did this to indicate that the invite didn't contain any information about the group's users. This is also documented in `Invite.Group#getUsers()`
> If the users were not included in the invite, this will return `null`.

MinnDevelopment: Returning null is an inconvenience for the end user. What is the purpose of knowing whether this information was present or not?

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
Almighty-Alpaca: I just left in there what we used previously. It makes sense though because groups don't always have a name, channels in a guild however are required to have one.

src/main/java/net/dv8tion/jda/core/entities/Invite.java
```diff
@@ -367,4 +390,59 @@ default String getURL()
...
+
+        /**
+         * The names of all users in this group. If the users were not included in the
+         * invite, this will return {@code null}. Users will usually only be returned when resolving the invite via the
+         * {@link #resolve(net.dv8tion.jda.core.JDA, java.lang.String, boolean) Invite.resolve()} method with the
+         * {@code withCounts} boolean set to {@code true}.
+         *
+         * @return The names of the groups's users or null if not preset in the invite
+         */
+        List<String> getUsers();
```
MinnDevelopment: Ok, I understand your intentions here for making it return null. What is your opinion on doing `Optional<List<String>>` instead?

src/main/java/net/dv8tion/jda/core/entities/impl/InviteImpl.java
```diff
@@ -117,6 +121,9 @@ protected void handleResponse(final Response response, final Request<Void> reque
         if (this.expanded)
             return new RestAction.EmptyRestAction<>(getJDA(), this);
 
+        if (this.type != Invite.InviteType.GUILD)
+            throw new UnsupportedOperationException("Only guild invites can be expanded");
```
MinnDevelopment: I can't see where this exception is documented. I would also prefer using `IllegalStateException` since the state is what makes the difference here, not the operation.

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
DV8FromTheWorld: What cases are there in which you don't receive `recipients`? What causes it?

src/main/java/net/dv8tion/jda/core/entities/impl/InviteImpl.java
Almighty-Alpaca: [`Invite#expand()`](https://github.com/DV8FromTheWorld/JDA/pull/709/files#diff-ed6e83754ad768ec7d551332e8cbed3cR102):
>As bots can't be in groups this is only available for guild invites and will throw an _UnsupportedOperationException_ for other types.

>@throws java.lang.UnsupportedOperationException
>If this is a group invite

Also you can get more info about a group invite if you're in the group (`created_at`, `max_age`) so it technically is an unsupported operation. Have no problems  it to an`IllegalStateException` though.

MinnDevelopment: Oh looks like I misread `expand` as `resolve`, my bad. But yeah I would rather use `IllegalStateException` for this.

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
Almighty-Alpaca: Discord only sends them if you use the `with_counts=true` query parameter. This is already dosumented in `Invite.Group#getUsers()`.

src/main/java/net/dv8tion/jda/core/entities/Invite.java
DV8FromTheWorld: Do we have any other places in JDA where we return null or optional for a collection?
I feel like most of our non-collection returns that indicate missing data use `null`

MinnDevelopment: I can't think of any example

DV8FromTheWorld: Basically my point here is that we (for better or worse) don't use `Optional` in JDA when we lack data.

Are there any other places that we, upon not receiving a collection, just decide to do `List.emptyList`? For the most part I feel like we trying to not return null from collection related getters, so im conflicted here.

MinnDevelopment: I generally never return null when a collection is defined, so I can't think of any case where this would be done in JDA at the moment. Some indication of this being optional or possibly missing is definitely needed in my opinion, even if just an `@Nullable`.

Almighty-Alpaca: Personally I don't like switching to returning an empty object instead of `null` just because it's a collection. It's just feels unintuitive and inconsistent having if you have to make null-checks everywhere except for this specific case.

PR #714, opened 2018-07-16
Add getGuildsByName to ShardManager

duncte123:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

Added getGuildsByName to the ShardManager interface because I was missing the method.

### DIFF
--- a/src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
+++ b/src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
@@ -374,6 +374,22 @@ public interface ShardManager
     }
 
     /**
+     * An unmodifiable list of all {@link net.dv8tion.jda.core.entities.Guild Guilds} that have the same name as the one provided.
+     * <br>If there are no {@link net.dv8tion.jda.core.entities.Guild Guilds} with the provided name, then this returns an empty list.
+     *
+     * @param  name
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
```diff
@@ -373,6 +373,22 @@ default Guild getGuildById(final String id)
...
+     * <br>If there are no {@link net.dv8tion.jda.core.entities.Guild Guilds} with the provided name, then this returns an empty list.
+     *
+     * @param  name
+     *         The name of the requested {@link net.dv8tion.jda.core.entities.Guild Guilds}.
+     * @param  ignoreCase
+     *         Whether to ignore case or not when comparing the provided name to each {@link net.dv8tion.jda.core.entities.Guild#getName()}.
+     *
+     * @return Possibly-empty list of all the {@link net.dv8tion.jda.core.entities.Guild Guilds} that all have the same name as the provided name.
+     */
+    default List<Guild> getGuildsByName(String name, boolean ignoreCase)
```
MinnDevelopment: Since all other parameters in this interface are marked as final it would only make sense to do the same here. Same for using `this.getGuildCache()` like everywhere else.

PR #719, opened 2018-07-17
Send SSRC alongside speaking update

MinnDevelopment:
This behavior can be observed in the official discord client

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

Apparently this is now something we need to concern ourselves with. I'm not sure what implications not supporting it will entail in the future, for now I see no difference.


MinnDevelopment:
This is also mentioned in the documentation https://discordapp.com/developers/docs/topics/voice-connections#speaking

Although it doesn't actually break anything not to send it. I would assume the server should be aware of our SSRC already due to the fact it actually sends it to us in READY.

PR #721, opened 2018-07-21
Improve use of MDC

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

Now we use copies of the initial map in order to restore thread locality
for MDC, in addition we now set MDC keys in the onThreadStart methods
for websocket adapters


src/main/java/net/dv8tion/jda/core/audio/AudioConnection.java
```diff
@@ -174,9 +169,9 @@ public void setChannel(VoiceChannel channel)
         this.channel = channel;
     }
 
-    public JDA getJDA()
+    public JDAImpl getJDA()
     {
-        return channel.getJDA();
+        return (JDAImpl) channel.getJDA();
```
kantenkugel: Can't this be replaced with just `return api;`? (see line 91)

MinnDevelopment:
This should wait on #705 so we can disable MDC for rate-limiter when a custom pool is provided.

src/main/java/net/dv8tion/jda/core/audio/AudioConnection.java
MinnDevelopment: No it can't be replaced with that, the api reference is local in the constructor.

PR #726, opened 2018-07-25
Make use of WeakReference for upstream references

MinnDevelopment:

## Pull Request Etiquette

## Description

This should make the referencing within JDA represent more of a tree
rather than a bunch of circular references. The GC Root for JDA now is `WebSocketClient` which is considered "upstream". JDAImpl weakly refers to the `WebSocketClient` while strongly referring to its own cache.

"=>" strong reference
"<~" weak reference

```
WebSocketClient => JDAImpl
WebSocketClient <~ JDAImpl
JDAImpl => users
JDAImpl <~ user
JDAImpl => guilds
JDAImpl <~ guild
GuildImpl => members
GuildImpl <~ member
etc.
```

This then ends up as a tree of strong references, while each node may keep a weak reference to its referents (parents). For instance, if a thread holds a reference to a `Member` instance, it will not have a strong reference to the respective `GuildImpl` because `MemberImpl` only keeps a weak reference to `GuildImpl` which allows the garbage collector to clean up the guild object.


MinnDevelopment:
This builds and works. Ready for merge if no more objections are brought up.

I plan to merge this by saturday this week.

PR #745, opened 2018-08-23
Add new annotations for common documentation concerns

MinnDevelopment:

## Pull Request Etiquette

  Pull Request to be merged.

  can **rebase** your branch.

## Description

The updated README explains my intentions, questions and opinions welcome.

### DIFF
--- a/src/main/java/net/dv8tion/jda/annotations/ForRemoval.java
+++ b/src/main/java/net/dv8tion/jda/annotations/ForRemoval.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2015-2018 Austin Keener & Michael Ritter & Florian Spieß
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.annotations;
+
+import java.lang.annotation.*;
+
+/**
+ * Functionality annotated with ForRemoval will no longer be supported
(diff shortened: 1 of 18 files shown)


src/main/java/net/dv8tion/jda/annotations/ForRemoval.java
```diff
@@ -0,0 +1,30 @@
...
+package net.dv8tion.jda.annotations;
+
+import java.lang.annotation.*;
+
+/**
+ * Functionality annotated with ForRemoval will no longer be supported
+ * and should not be used anymore in new code.
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
```
kantenkugel: I'm pretty sure we don't need them at Runtime... For most of these, SOURCE would probably be enough, and if not, CLASS. I don't see why those would be needed at runtime for eg reflection

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -716,6 +718,7 @@ public JDA buildAsync() throws LoginException
      *         Use {@link #build()} and {@link JDA#awaitStatus(Status)} instead
      */
     @Deprecated
+    @ReplaceWith("build().awaitReady()")
```
kantenkugel: Wouldn't the better ReplaceWith suggestion be `build().awaitStatus(Status)` here?

src/main/java/net/dv8tion/jda/annotations/ForRemoval.java
MinnDevelopment: You could write runtime inspections this way

MinnDevelopment:
I have added a custom javadoc tag for incubating features.

image

src/main/java/net/dv8tion/jda/annotations/ForRemoval.java
Shengaero: I would agree with @kantenkugel here.
If these annotations are really for "documentation" they do not need to be available at runtime via reflection.

Shengaero:
It might also be worthwhile to have rules for applying these annotations documented in the [structure guide](https://github.com/DV8FromTheWorld/JDA/wiki/6%29-JDA-Structure-Guide) section of the wiki, like how to reference methods to use with `@ReplaceWith`.

PR #762, opened 2018-09-04
Prevent using complete() in callbacks

MinnDevelopment:

## Pull Request Etiquette

## Description

```java
channel.sendMessage("hello there").queue( msg -> {
    msg.editMessage("general kenobi").queue(); // ok
    msg.delete().complete(); // throws
});
```

This way we prevent any deadlocks from occurring in a fixed-pool causing everything to halt for a shard (or all shards if one shared pool is used)

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/requests/CallbackContext.java
+++ b/src/main/java/net/dv8tion/jda/core/requests/CallbackContext.java
@@ -18,7 +18,7 @@ package net.dv8tion.jda.core.requests;
 
 class CallbackContext implements AutoCloseable
 {
-    private static ThreadLocal<Boolean> callback = ThreadLocal.withInitial(() -> false);
+    private static final ThreadLocal<Boolean> callback = ThreadLocal.withInitial(() -> false);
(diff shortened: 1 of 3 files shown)


Athem24:
Can't complete rest action be private/package-private ?

napstr:
But then you wouldn't be able to use it at all. `complete()` does have its (rare) use cases though.

Btw Minn, what about `submit().get()`?
Also shouldn't the main read threads also stay clear of blocking calls, or is this crackdown on `complete` related to something specific with the callback threads?

MinnDevelopment:
This is to prevent deadlocks, as I have described in the comment.
I don't want to inject anything into the `submit()` future, use at own risk.

MinnDevelopment:
If we wanted to stop people from using any blocking operations on the main read thread we could simply interrupt the thread.

DV8FromTheWorld:
Java not having an `await` keyword makes this suck a _lot_

natanbc:
closest i've ever found was ea-async

PR #769, opened 2018-09-09
Fix granting / denying permissions

### LINKED ISSUE #767: Can't deny a granted permission, PermissionOverride resource left in bad state
## General Troubleshooting

Hey there! Before you report a bug or suggest a new feature,
  please make sure to follow these steps first!
  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

### Issue Type

- [x] Bug Report

### Description

There seems to be a bug in how permissions are denied with the PermOverrideManager.

Creating a PO, then granting a permission (example: Attach Files), and then denying it again leaves the PermissionOverride in a bad state where the raw values of denied and allowed permissions are both containing the raw value of the permission in question. The Discord API will happily accept a raw value in both denied and allowed bitmasks, but will assume that the permission is allowed, making it impossible to deny the permission via JDA in this way.

Here's a JS eval to reproduce this:
```js
var result = "";
var shardManager = context.getJda().asBot().getShardManager();
var tc = shardManager.getTextChannelById("488315931383169066");
var role = shardManager.getRoleById("488314982552043520");
var po = tc.createPermissionOverride(role).complete();
result += po.getAllowedRaw() + " " + po.getDeniedRaw() + "\n";
po.getManager().grant(Java.type("net.dv8tion.jda.core.Permission").MESSAGE_ATTACH_FILES.getRawValue()).complete();
Java.type("java.lang.Thread").sleep(1000);
po = tc.getPermissionOverride(role);
result += po.getAllowedRaw() + " " + po.getDeniedRaw() + "\n";
po.getManager().deny(Java.type("net.dv8tion.jda.core.Permission").MESSAGE_ATTACH_FILES.getRawValue()).complete();
Java.type("java.lang.Thread").sleep(1000);
po = tc.getPermissionOverride(role);
result += po.getAllowedRaw() + " " + po.getDeniedRaw() + "\n";
return result;
```
Giving the following output:
```
0 0
32768 0
32768 32768
```

The correct output should have been
```
0 0
32768 0
0 32768
```

I traced this issue down to the `PermOverrideManager#finalizeData` method.
When entering that method, the internal fields `allowed` and `denied` have the correct values (`allowed` being 0 and `denied` being 32768), but then this method calls `PermOverrideManager#setupValues`, which changes the value of the `allowed` field to contain 32768 instead of 0.

napstr:
Resolves #767, please see that issue for details.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/managers/PermOverrideManager.java
+++ b/src/main/java/net/dv8tion/jda/core/managers/PermOverrideManager.java
@@ -202,18 +202,9 @@ public class PermOverrideManager extends ManagerBase
         if (permissions == 0)
             return this;
         setupValues();
-        if ((allowed & permissions) == 0)
-        {
-            this.allowed |= permissions;
-            this.set |= ALLOWED;
-        }
-
-        if ((denied & permissions) != 0)
-        {
-            this.denied &= ~permissions;
-            this.set |= DENIED;
-        }
-
+        this.allowed |= permissions;
+        this.denied &= ~permissions;
+        this.set |= PERMISSIONS;
         return this;
     }
 
@@ -274,18 +265,9 @@ public class PermOverrideManager extends ManagerBase
         if (permissions == 0)
             return this;
         setupValues();
-        if ((allowed & permissions) != 0)
-        {
-            this.allowed &= ~permissions;
-            this.set |= ALLOWED;
-        }
-
-        if ((denied & permissions) == 0)
-        {
-            this.denied |= permissions;
-            this.set |= DENIED;
-        }
-
+        this.denied |= permissions;
+        this.allowed &= ~permissions;
+        this.set |= PERMISSIONS;
         return this;
     }
 
--- a/src/main/java/net/dv8tion/jda/core/managers/PermOverrideManager.java
+++ b/src/main/java/net/dv8tion/jda/core/managers/PermOverrideManager.java
@@ -202,9 +202,18 @@ public class PermOverrideManager extends ManagerBase
         if (permissions == 0)
             return this;
         setupValues();
-        this.allowed |= permissions;
-        this.denied &= ~permissions;
-        this.set |= ALLOWED;
+        if ((allowed & permissions) == 0)
+        {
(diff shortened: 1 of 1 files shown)


napstr:
```java
PermissionOverride po = tc.createPermissionOverride(role).complete();
//wait
print(po.getAllowedRaw() + " " + po.getDeniedRaw());
po.getManager().grant(Permission.MESSAGE_ATTACH_FILES.getRawValue()).complete();
//wait
print(po.getAllowedRaw() + " " + po.getDeniedRaw());
po.getManager().deny(Permission.MESSAGE_ATTACH_FILES.getRawValue()).complete();
//wait
print(po.getAllowedRaw() + " " + po.getDeniedRaw());
po.getManager().grant(Permission.MESSAGE_ATTACH_FILES.getRawValue()).complete();
//wait
print(po.getAllowedRaw() + " " + po.getDeniedRaw());
po.getManager().clear(Permission.MESSAGE_ATTACH_FILES.getRawValue()).complete();
//wait
print(po.getAllowedRaw() + " " + po.getDeniedRaw());
```
results in 
```
0 0
32768 0
0 32768
32768 0
0 0
```
which looks correct to me.

src/main/java/net/dv8tion/jda/core/managers/PermOverrideManager.java
```diff
@@ -202,9 +202,18 @@ public PermOverrideManager grant(long permissions)
...
             return this;
         setupValues();
-        this.allowed |= permissions;
-        this.denied &= ~permissions;
-        this.set |= ALLOWED;
+        if ((allowed & permissions) == 0)
+        {
+            this.allowed |= permissions;
+            this.set |= ALLOWED;
+        }
```
MinnDevelopment: So what this code now does: if **none** of the granted permissions are currently granted we set them.
This is really not how this is supposed to go, we want to update if **any** of the permissions are newly granted.

My suggestion was to only change the `this.set |= ALLOWED`.

PR #776, opened 2018-09-13
Configuration for mainWs sending pools

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Since these threads spend most of their time sleeping we should share the resources between shards.


src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
```diff
@@ -770,4 +764,28 @@ protected ScheduledExecutorService createExecutor(ThreadFactory threadFactory)
 
         return Executors.newSingleThreadScheduledExecutor(factory);
     }
+
+    protected static <E extends ExecutorService> ExecutorPaid<E> getExecutor(ThreadPoolProvider<? extends E> provider, int shardId)
```
DV8FromTheWorld: `ExecutorPair` instead of `ExecutorPaid` perhaps?

MinnDevelopment: :man_facepalming:

DV8FromTheWorld:
Can you also clean this up to use the same `exclude` please?
https://github.com/DV8FromTheWorld/JDA/blob/c6162b87b737404338621f59f33f21531f24aae3/build.gradle#L176-L198

MinnDevelopment:
Do you mean that all `core/*` use the same `exclude`?

MinnDevelopment:
Note to self: update readme version of okhttp

MinnDevelopment:
With the new pools you can configure JDA to 2 threads per shard, half of what is currently possible on master. This would mean you can reduce a 100 shard bot from 400 threads to 201 threads by simply changing the pools to a single threaded scheduler.

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManagerBuilder.java
```diff
@@ -983,13 +1035,13 @@ public DefaultShardManagerBuilder setWebsocketFactory(WebSocketFactory factory)
...
-            this.listeners, this.listenerProviders, this.token, this.eventManager,
-            this.audioSendFactory, this.gameProvider, this.statusProvider,
-            this.httpClientBuilder, this.httpClient, this.rateLimitPoolProvider, this.callbackPoolProvider, this.wsFactory, this.threadFactory,
-            this.maxReconnectDelay, this.corePoolSize, this.enableVoice, this.enableShutdownHook, this.enableBulkDeleteSplitting,
-            this.autoReconnect, this.idleProvider, this.retryOnTimeout, this.useShutdownNow, this.enableContext,
-            this.contextProvider, this.cacheFlags, this.enableCompression);
+                this.shardsTotal, this.shards, this.sessionController,
+                this.listeners, this.listenerProviders, this.token, this.eventManager,
+                this.audioSendFactory, this.gameProvider, this.statusProvider,
+                this.httpClientBuilder, this.httpClient, this.rateLimitPoolProvider, mainWsPoolProvider, this.callbackPoolProvider, this.wsFactory, this.threadFactory,
```
DV8FromTheWorld: Missing `this.`

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -365,19 +367,57 @@ public JDABuilder setRateLimitPool(ScheduledThreadPoolExecutor pool)
...
+     * the main WebSocket connection for workers. These workers spend most of their lifetime
+     * sleeping because they only activate for sending messages over the gateway.
+     * <br><b>Only change this pool if you know what you're doing.
+     * <br>This automatically disables the automatic shutdown of the main-ws pool, you can enable
+     * it using {@link #setMainWsPool(ScheduledExecutorService, boolean) setMainWsPool(pool, true)}</b>
+     *
+     * @param  pool
+     *         The thread-pool to use for WebSocket workers
+     *
+     * @return THre JDABuilder instance. Useful for chaining.
```
DV8FromTheWorld: `THre` -> `The`

src/main/java/net/dv8tion/jda/core/JDABuilder.java
```diff
@@ -365,19 +367,57 @@ public JDABuilder setRateLimitPool(ScheduledThreadPoolExecutor pool)
...
+     * the main WebSocket connection for workers. These workers spend most of their lifetime
+     * sleeping because they only activate for sending messages over the gateway.
+     * <br><b>Only change this pool if you know what you're doing.</b>
+     *
+     * @param  pool
+     *         The thread-pool to use for WebSocket workers
+     * @param  automaticShutdown
+     *         Whether {@link JDA#shutdown()} should shutdown this pool
+     *
+     * @return THre JDABuilder instance. Useful for chaining.
```
DV8FromTheWorld: `THre` -> `The`

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -283,136 +281,7 @@ protected boolean send(String message, boolean skipQueue)
...
-    protected JSONObject newVoiceOpen(AudioManager manager, long channel, long guild)
-    {
-        return new JSONObject()
-            .put("op", WebSocketCode.VOICE_STATE)
-            .put("d", new JSONObject()
-                .put("guild_id", guild)
-                .put("channel_id", channel)
-                .put("self_mute", manager.isSelfMuted())
-                .put("self_deaf", manager.isSelfDeafened()));
+        ratelimitThread = executor.scheduleAtFixedRate(new WebSocketSendingThread(this), 0, 500, TimeUnit.MILLISECONDS);
```
DV8FromTheWorld: This is not entirely accurate. We try to send as fast as possible, only sleeping when:
1) We havent sent auth info
2) We didn't try to send a packet during the latest iteration

By only running once every 500ms, we would take close to a minute to fully connect to 100 voice channels if we were reconnecting.

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
```diff
@@ -791,7 +641,12 @@ protected void invalidate()
         sessionId = null;
         sentAuthInfo = false;
 
-        chunkSyncQueue.clear();
+        locked("Interrupted while trying to invalidate queue", () -> {
+            chunkSyncQueue.clear();
+            ratelimitQueue.clear();
```
DV8FromTheWorld: We weren't clearing the ratelimit queue before.
Is there a specific reason we want to clear it now?

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
```diff
@@ -626,24 +640,25 @@ protected JDAImpl buildInstance(final int shardId) throws LoginException, Interr
...
+        ScheduledExecutorService rateLimitPool = rateLimitPair.executor;
+        boolean shutdownRateLimitPool = rateLimitPair.automaticShutdown;
+
+        ExecutorPair<ScheduledExecutorService> mainWsPair = resolveExecutor(mainWsPoolProvider, shardId);
+        ScheduledExecutorService mainWsPool = mainWsPair.executor;
+        boolean shutdownMainWsPool = mainWsPair.automaticShutdown;
+
+        ExecutorPair<ExecutorService> callbackPair = resolveExecutor(callbackPoolProvider, shardId);
+        ExecutorService callbackPool = callbackPair.executor;
+        boolean shutdownCallbackPool = callbackPair.automaticShutdown;
```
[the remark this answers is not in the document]
MinnDevelopment: Its easier to look at

src/main/java/net/dv8tion/jda/bot/sharding/DefaultShardManager.java
```diff
@@ -749,4 +764,28 @@ protected ScheduledExecutorService createExecutor(ThreadFactory threadFactory)
...
     }
+
+    protected static <E extends ExecutorService> ExecutorPair<E> resolveExecutor(ThreadPoolProvider<? extends E> provider, int shardId)
+    {
+        E executor = null;
+        boolean automaticShutdown = true;
+        if (provider != null)
+        {
+            executor = provider.provide(shardId);
+            automaticShutdown = provider.shouldShutdownAutomatically(shardId);
```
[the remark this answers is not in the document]
MinnDevelopment: We default to an executor in JDAImpl in-case it is null

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
MinnDevelopment: The queue might contain lifecycle events that should not be sent again

src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java
MinnDevelopment: We could change it to re-schedule itself depending on needed timeout, however we still need some kind of idle time when nothing was sent. How about:

- Nothing happened -> wait 500ms
- Sent message -> wait 10ms
- Rate-limit -> wait 60s

MinnDevelopment: Since we plan to share threads its fair to give other sessions a window to use it, even this small

PR #781, opened 2018-09-19
Replace ScheduledThreadPoolExecutor with ScheduledExecutorService

napstr:

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

## Description

Some existing delegators out there, for example Spring's `DelegatingSecurityContextScheduledExecutorService`, only implement the `ScheduledExecutorService` interface, and therefore can't be used for JDA's ratelimit pool.
The only reason JDA requires a `ScheduledThreadPoolExecutor` for the ratelimit pool is due to the optional customizing of the pool's shutdown behavior, therefore I think moving to using the more generic `ScheduledExecutorService` would be reasonable.


MinnDevelopment:
I have applied these changes to #776 instead since I am working on threading configuration there already. Thank you for the proposition.

PR #795, opened 2018-10-04
Add Guild#getChannels for convenience

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Convenience method to retrieve list of channels in a guild. The returned list is ordered the same way the client orders channels.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/Guild.java
@@ -733,6 +733,58 @@ public interface Guild extends ISnowflake
     SnowflakeCacheView<VoiceChannel> getVoiceChannelCache();
 
     /**
+     * Populated list of {@link net.dv8tion.jda.core.entities.Channel channels} for this guild.
+     * This includes all types of channels, such as category/voice/text.
+     * <br>This includes hidden channels by default.
+     *
+     * <p>The returned list is ordered in the same fashion as it would be by the official discord client.
+     * <ol>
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/core/entities/Guild.java
```diff
@@ -732,6 +732,58 @@ default VoiceChannel getVoiceChannelById(long id)
...
+     *
+     *
+     * @param  includeHidden
+     *         Whether to include channels with denied {@link Permission#VIEW_CHANNEL View Channel Permission}
+     *
+     * @return Immutable list of channels for this guild
+     *
+     * @see    #getChannels()
+     */
+    List<Channel> getChannels(boolean includeHidden);
```
Almighty-Alpaca: Would this be better as separate method `#getVisibleChannels()`?

DV8FromTheWorld: I think it would separate the functionality unnecessarily.
A boolean here is completely appropriate.

MinnDevelopment:
```java
List<Channel> channels;
if (hidden)
     channels = guild.getChannels();
else
     channels = guild.getVisibleChannels();
```

compared to

```java
List<Channels> channels = guild.getChannels(hidden);
```

PR #802, opened 2018-10-10
Added getUsersByPrefix methods

BenjaminUrquhart:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Added `getUsersByPrefix` methods to both `JDA` and `ShardManager`. This can probably be made neater by using streams, but I am a noob when it comes to those.

Note: I'm probably going to rename these to `getUsersWithPrefix` or something like that

### DIFF
--- a/src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
+++ b/src/main/java/net/dv8tion/jda/bot/sharding/ShardManager.java
@@ -825,7 +825,29 @@ public interface ShardManager
     {
         return this.getUserCache().getElementById(id);
     }
-
+  
+     /**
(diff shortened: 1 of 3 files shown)


MinnDevelopment:
Something like this should be done on user-end instead. It does not make sense for us to implement common list operations in JDA. Thank you for the proposal but we won't be accepting it.

DV8FromTheWorld:
You might find something like https://github.com/JDA-Applications/JDA-Utilities/blob/master/commons/src/main/java/com/jagrosh/jdautilities/commons/utils/FinderUtil.java helpful. JDA Utils provides a lot of these user-specific helpful functions.

PR #817, opened 2018-11-03
Add convenience purgeMessages overloads

JohnnyJayJay:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds three new method overloads for `MessageChannel#purgeMessages` for convenience.

`MessageChannel#purgeMessages(int)`: purges the amount of messages specified

`MessageChannel#purgeMessages(int, Consumer<? super Void>)`: Purges the amount of messages specified and accepts the consumer afterwards as a callback

`MessageChannel#purgeMessages(int, Consumer<? super Void>, Consumer<Throwable>)`: Purges the amount of messages specified, accepts the callback consumer afterwards and accepts the Throwable Consumer if any Exception occurs while purging.

I added this because many people get confused by `channel.getIterableHistory().takeAsync(amount).thenAccept(channel::purgeMessages)` and I believe the new methods are more convenient if you just want to purge the messages and not much more.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
@@ -29,6 +29,7 @@ import org.json.JSONArray;
 import javax.annotation.CheckReturnValue;
 import java.io.*;
 import java.util.*;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
I have personally not seen a single case where a user got confused by the current approach.
I also don't like how much information is abstracted away here and how errors are handled.

JohnnyJayJay:
I understand your concern, though I don't think the little information is a big deal. I've honestly never seen any code that uses more from purgeMessages than the new methods provide. And even if so - the other methods still exist and one could still utilise them.
Could you explain the issue with the error handling here?

napstr:
This looks like it belongs into user code. It is useless for anyone using the proper way of async in Java which are CompletableFutures.

MinnDevelopment:
I think the  missing information is a big deal and can be confusing. From looking at it you would think that this is an atomic operation, meaning all it would do is send a DELETE request with an integer. That however is not the case here, it has to fetch all messages in chunks of 100 each and then delete them in chunks of 100 each. Having it split up into 2 operations gives the user more of an idea about what is actually happening.

The error handling is just horrible, you have a single consumer for failures for something that can fail in numerous locations such as:

- Fetching messages
- Deleting single messages
- Deleting bulks of messages

And for that matter can fail multiple times not a single time. The other methods provide a list of futures representing each task and can provide a failure callback for each failure.

src/main/java/net/dv8tion/jda/core/entities/MessageChannel.java
```diff
@@ -184,6 +185,88 @@ default String getLatestMessageId()
...
+     * <p>For possible ErrorResponses see {@link #purgeMessagesById(long...)}.
+     *
+     * @param  amount
+     *         The amount of messages to delete
+     *
+     * @param  callback
+     *         The action to run after all of the messages have been deleted
+     *
+     * @param  exceptionally
+     *         A {@link java.util.function.Consumer<java.lang.Throwable>} that will accept any Exceptions
```
MinnDevelopment: This throws javadoc warnings

DV8FromTheWorld:
This doesn't follow a lot of what we've laid out about how JDA interacts with network requests. This functionality most likely belongs in either a Util library of UserSpace.

MinnDevelopment:
Based on our concerns expressed here, i am closing this.

Issue #818, opened 2018-11-06
Specify Position when creating a guild channel


DV8FromTheWorld:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

## Issue
Apparently the Discord API allows you to provide a channel position on channel creation based on this issue: https://github.com/discordapp/discord-api-docs/issues/720

We need to add this functionality to the [ChannelAction](https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java)
### Issue Type

- [ ] Bug Report
- [x] Feature Request

PR #819, opened 2018-11-06
Add ChannelAction#setPosition(Integer)

kantenkugel:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #818 

## Description

This PR will allow users to set the raw position value used when creating a new (Guild-)Channel.

While the actual change was not yet made on Discords end, this PR is on hold.

It will probably also need the doc addition about what happens when a position is set that already exists on another Channel with same Parent (including null parent). But since that depends on testing, it also has to wait until this feature is allowed on Discords end.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
+++ b/src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
@@ -46,6 +46,7 @@ public class ChannelAction extends AuditableRestAction<Channel>
     protected final ChannelType type;
     protected String name;
     protected Category parent;
+    protected Integer position;
 
     // --text only--
     protected String topic = null;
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/core/requests/restaction/ChannelAction.java
```diff
@@ -127,6 +128,29 @@ public ChannelAction setParent(Category category)
...
+     * @param  position
+     *         The raw position value that should be used for the new Channel
+     *
+     * @throws IllegalArgumentException
+     *         If the provided position value is {@code <0}
+     *
+     * @return The current ChannelAction, for chaining convenience
+     */
+    @CheckReturnValue
+    public ChannelAction setPosition(Integer position)
```
BenjaminUrquhart: Just wondering, why `Integer` instead of `int`?

kantenkugel: cuz nullable

kantenkugel: The alternative would've been using `-1` for "not set" since that wouldn't be in the accepted range anyway... but i like `null` more... It also goes with how `slowmode`, `bitrate` and `userlimit` are handled already.

BenjaminUrquhart: 👌

kantenkugel:
Should be good for (final) review and merge now... Feature was released and I tested how clash of positions are being handled (they aren't)

PR #834, opened 2018-11-30
Use RW-Lock for more efficient cache view access

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Changes internals for `CacheView` to synchronize with a read-write lock rather than using an implicit mutex.

Example usage:

```java
List<Member> matches = new ArrayList<>(5);
MemberCacheView view = guild.getMemberCache();
try (ClosableIterator<Member> it = view.lockedIterator()) { // acquires read-lock until closed or drained
    while (it.hasNext() && matches.size() < 5) {
        Member m = it.next();
        if (m.getRoles().size() >= 3)
            matches.add(m);
    }
}

AtomicInteger totalCount = new AtomicInteger();
view.forEach(m -> { // acquires read-lock until done
    if (m.getRoles().size() >= 3) totalCount.incrementAndGet()
});
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/JDA.java
+++ b/src/main/java/net/dv8tion/jda/core/JDA.java
@@ -358,6 +358,11 @@ public interface JDA
      * <p>If the developer is sharding, then only users from guilds connected to the specifically logged in
      * shard will be returned in the List.
      *
+     * <p>This copies the backing store into a list. This means every call
+     * creates a new list with O(n) complexity. It is recommended to store this into
+     * a local variable or use {@link #getUserCache()} and use its more efficient
+     * versions of handling these values.
+     *
      * @return List of all {@link net.dv8tion.jda.core.entities.User Users} that are visible to JDA.
      */
     default List<User> getUsers()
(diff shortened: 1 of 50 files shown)


DV8FromTheWorld:
Lets not do things like this: https://github.com/DV8FromTheWorld/JDA/pull/834/files#diff-3cd56d2bba5a9fa8c43fa87c7cb3b2caR281

It makes it harder to follow the source.
Start the stream on the first line, and bring the stream mutators to the next line
```
 return myStuff.stream()
     .filter(...)
     .map(...)
     .collect(...)
```

MinnDevelopment:
The latest commit also prevents things like 

```java
userCache.forEach(user -> userCache.remove(user.getIdLong()));
```

MinnDevelopment:
It looks like `iterator()` and `forEach(...)` have a different iteration order due to the changes I've made.
In `SortedSnowflakeCacheView` the `iterator()` is made to iterate the sorted array of values while `forEach` iterates the arbitrary order of the backing hash map. Now it is time to make a decision what we deem more important, efficiency or order.

PR #837, opened 2018-12-04
Make names consistent for timestamp getters

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

More consistent names for getters to increase usability. This also makes auto-complete a little better since `getTime` will now display all timestamp getters.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/events/DisconnectEvent.java
+++ b/src/main/java/net/dv8tion/jda/core/events/DisconnectEvent.java
@@ -113,7 +113,7 @@ public class DisconnectEvent extends Event
      *
      * @return Time of closure
      */
-    public OffsetDateTime getTimeDisconnect()
+    public OffsetDateTime getTimeDisconnected()
     {
         return disconnectTime;
     }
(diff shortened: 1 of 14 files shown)


src/main/java/net/dv8tion/jda/core/events/DisconnectEvent.java
```diff
@@ -113,7 +113,7 @@ public boolean isClosedByServer()
      *
      * @return Time of closure
      */
-    public OffsetDateTime getDisconnectTime()
+    public OffsetDateTime getTimeDisconnect()
```
MinnDevelopment: Maybe `getTimeDisconnected`?

DV8FromTheWorld: Agreed.

DV8FromTheWorld:
We should also rename the `getCreationTime` methods in MiscUtil

PR #848, opened 2018-12-15
Message deletion days validation on member ban

Fabricio20:
## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Library interface (affecting end-user code) [Non-breaking]
- [X] Documentation
- [ ] Other: \_____

Closes Issue: NaN

## Description

When banning a member and specifying a delDays bigger than 7, Discord will return a ErrorResponseException with code 400.

```
[ForkJoinPool.commonPool-worker-1] ERROR RestAction - RestAction queue returned failure: [ErrorResponseException] 400: {"delete-message-days":["int value should be less than or equal to 7."]}
```

This PR adds validation for that scenario following the same principle of the lower than 0 validation, which will throw an IllegalArgumentException.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/managers/GuildController.java
+++ b/src/main/java/net/dv8tion/jda/core/managers/GuildController.java
@@ -534,6 +534,7 @@ public class GuildController
      * @throws java.lang.IllegalArgumentException
      *         <ul>
      *             <li>If the provided amount of days (delDays) is less than 0.</li>
+     *             <li>If the provided amount of days (delDays) is bigger than 7.</li>
      *             <li>If the provided member is {@code null}</li>
      *         </ul>
      *
@@ -587,6 +588,7 @@ public class GuildController
      * @throws java.lang.IllegalArgumentException
      *         <ul>
      *             <li>If the provided amount of days (delDays) is less than 0.</li>
+     *             <li>If the provided amount of days (delDays) is bigger than 7.</li>
      *             <li>If the provided user is null</li>
      *         </ul>
      *
@@ -602,6 +604,8 @@ public class GuildController
             checkPosition(getGuild().getMember(user));
 
         Checks.notNegative(delDays, "Deletion Days");
+
+        Checks.notBiggerThan(delDays, 7, "Deletion Days");
 
         final String userId = user.getId();
 
--- a/src/main/java/net/dv8tion/jda/core/utils/Checks.java
+++ b/src/main/java/net/dv8tion/jda/core/utils/Checks.java
@@ -147,4 +147,10 @@ public class Checks
             throw new IllegalArgumentException(name + " may not be negative");
     }
 
+    public static void notBiggerThan(final int n, final int bound, String name)
+    {
+        if (n > bound)
+            throw new IllegalArgumentException(name + " may not be bigger than " + bound);
+    }
+
 }


src/main/java/net/dv8tion/jda/core/managers/GuildController.java
```diff
@@ -603,6 +605,8 @@ protected void handleResponse(Response response, Request<Void> request)
 
         Checks.notNegative(delDays, "Deletion Days");
 
+        Checks.notBiggerThan(delDays, 7, "Deletion Days");
```
Fabricio20: Not entirely sure if this would be the best approach, maybe change it to a simple if statement to avoid creating a new method on `Checks.java`.

kantenkugel: yea, we have `Checks.check(Predicate check, String message)` iirc

MinnDevelopment: An alternative is `Checks.notPositive(delDays - 7, ...)`

src/main/java/net/dv8tion/jda/core/managers/GuildController.java
```diff
@@ -603,6 +605,8 @@ protected void handleResponse(Response response, Request<Void> request)
 
         Checks.notNegative(delDays, "Deletion Days");
 
+        Checks.check(delDays <= 7, "Deletion Days may not be bigger than 7.");
```
kantenkugel: Is the correct phrasing "may not" or "must not" here? Or are both valid?

DV8FromTheWorld: _Technically_ they're the same in general conversation, however if we abide by [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt), then we `must` is more correct.

MinnDevelopment: We don't really abide to that RFC within our documentation

Fabricio20: Since may could imply a scenario where it is acceptable to use a number above 7, I've decided to change the phrasing to must (abiding to RFC 2119).

src/main/java/net/dv8tion/jda/core/managers/GuildController.java
```diff
@@ -661,18 +665,29 @@ protected void handleResponse(Response response, Request<Void> request)
      *         <br>See {@link net.dv8tion.jda.core.utils.PermissionUtil#canInteract(Member, Member) PermissionUtil.canInteract(Member, Member)}
      * @throws IllegalArgumentException
      *         If the provided amount of days (delDays) is less than 0.
```
kantenkugel: This is now mentioned twice

src/main/java/net/dv8tion/jda/core/managers/GuildController.java
```diff
@@ -661,18 +665,29 @@ protected void handleResponse(Response response, Request<Void> request)
...
+     * @throws java.lang.IllegalArgumentException
+     *         <ul>
+     *             <li>If the provided amount of days (delDays) is less than 0.</li>
+     *             <li>If the provided amount of days (delDays) is bigger than 7.</li>
+     *             <li>If the provided user is null</li>
```
kantenkugel: Using "user id" is probably better here

src/main/java/net/dv8tion/jda/core/managers/GuildController.java
```diff
@@ -821,8 +838,12 @@ else if (response.code == 404)
      * @throws net.dv8tion.jda.core.exceptions.HierarchyException
      *         If the logged in account cannot ban the other user due to permission hierarchy position.
      *         <br>See {@link net.dv8tion.jda.core.utils.PermissionUtil#canInteract(Member, Member) PermissionUtil.canInteract(Member, Member)}
-     * @throws IllegalArgumentException
-     *         If the provided amount of days (delDays) is less than 0.
+     * @throws java.lang.IllegalArgumentException
+     *         <ul>
+     *             <li>If the provided amount of days (delDays) is less than 0.</li>
+     *             <li>If the provided amount of days (delDays) is bigger than 7.</li>
+     *             <li>If the provided user is {@code null}</li>
```
kantenkugel: "user id" instead of just "user" here too i guess

PR #849, opened 2018-12-15
[Feature] Create User#getAsTag

duncte123:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

## Description

Created the method User#getAsTag because it makes code look cleaner than using String#format.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/User.java
@@ -111,7 +111,7 @@ public interface User extends ISnowflake, IMentionable, IFakeable
 
     /**
      * The "tag" for this user
-     * <p>This is the equivalent of calling {@link java.lang.String#format(String, Object...) String#format}("%#s", {@link net.dv8tion.jda.core.entities.User User})
+     * <p>This is the equivalent of calling {@link java.lang.String#format(String, Object...) String.format}("%#s", {@link net.dv8tion.jda.core.entities.User User})
      *
      * @return Never-null String containing the tag for this user, for example DV8FromTheWorld#6297
      */
--- a/src/main/java/net/dv8tion/jda/core/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/User.java
@@ -111,7 +111,7 @@ public interface User extends ISnowflake, IMentionable, IFakeable
 
     /**
      * The "tag" for this user
-     * This is the equivalent of calling {@link java.lang.String#format(String, Object...) String#format("%#s", {@link net.dv8tion.jda.core.entities.User User})}
+     * <p>This is the equivalent of calling {@link java.lang.String#format(String, Object...) String#format}("%#s", {@link net.dv8tion.jda.core.entities.User User})
      *
      * @return Never-null String containing the tag for this user, for example DV8FromTheWorld#6297
      */
--- a/src/main/java/net/dv8tion/jda/core/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/User.java
@@ -110,6 +110,13 @@ public interface User extends ISnowflake, IMentionable, IFakeable
     String getEffectiveAvatarUrl();
 
     /**
+     * The "tag" for this user, this is a shortcut to {@link java.lang.String#format(String, Object...) String#format("%#s", user)}
+     *
+     * @return The tag for this user
+     */
+    String getAsTag();
+
+    /**
      * Whether or not the currently logged in user and this user have a currently open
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/core/entities/impl/UserImpl.java
```diff
@@ -92,6 +92,11 @@ public String getEffectiveAvatarUrl()
         return getAvatarUrl() == null ? getDefaultAvatarUrl() : getAvatarUrl();
     }
 
+    @Override
+    public String getAsTag()
+    {
+        return String.format("%#s", this);
```
MinnDevelopment: Using a formatter just for this is pretty overkill

kantenkugel: Yea... i would just copy the code of the format method and then replace the one in the format method by a getAsTag() method call

kantenkugel: i mean i guess we use `String.format()` in `UserImpl#formatTo()` too... https://github.com/DV8FromTheWorld/JDA/blob/a7bfefe45f35d96b9d5b313f8b55b6ea926a873e/src/main/java/net/dv8tion/jda/core/entities/impl/UserImpl.java#L240-L256

MinnDevelopment: I think it would be reasonable to replace these with `getAsTag` similar to the `getAsMention` here

src/main/java/net/dv8tion/jda/core/entities/User.java
```diff
@@ -109,6 +109,13 @@
      */
     String getEffectiveAvatarUrl();
 
+    /**
+     * The "tag" for this user, this is a shortcut to {@link java.lang.String#format(String, Object...) String#format("%#s", user)}
+     *
+     * @return The tag for this user
+     */
+    String getAsTag();
```
MinnDevelopment: Add an example here for better understanding of what this means

src/main/java/net/dv8tion/jda/core/entities/impl/UserImpl.java
duncte123: I'm unsure what to do with the upper flag, or should I just ditch that?

MinnDevelopment: `toUpperCase()`?

duncte123: 🤦

src/main/java/net/dv8tion/jda/core/entities/User.java
```diff
@@ -109,6 +109,14 @@
      */
     String getEffectiveAvatarUrl();
 
+    /**
+     * The "tag" for this user
+     * This is the equivalent of calling {@link java.lang.String#format(String, Object...) String#format("%#s", {@link net.dv8tion.jda.core.entities.User User})}
```
MinnDevelopment: The formatting for this will look like this:

> The "tag" for this user This is the equivalent of calling String#format("%#s", User)

Either add a linefeed or use a period to separate the sentences. Maybe add a note that this is unique for every user.
Two nested link tags are not supported by javadoc and won't render correctly. Besides, I don't think it makes sense to link to the same class in the method docs.

kantenkugel: If you are writing the javadocs in Intellij, you should be able to just bring up the javadoc viewer from the method (ctrl+q by default) to get an approximate representation of how it will look btw

src/main/java/net/dv8tion/jda/core/entities/User.java
kantenkugel: I'm not sure if we should also clarify that this is not for mentioning and that IMentionable#getAsMention() should be used for that instead.

Also just FYI: the String#format() link doesn't render i think, since that method is not part of JDA and we are not rendering docs for java itself (so they are not linked). It isn't really an error (it will just render to blank text without any linking/jdoc syntax), but i guess it could be removed as it doesn't really do a lot.

duncte123: The link doesn't render in javadocs indeed, but it does in ide's (tested in intellij) so it might be useful to leave it there
javadocs
ide

MinnDevelopment: We use the call operator consistently in the docs, so the proper alias would be `String.format(...)`

PR #856, opened 2018-12-26
Make member a user delegate

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Member is simply an extension of what a user represents. It is a user with guild-specific properties such as roles and thus should be extending the User type. We can accomplish this by using a delegate implementation which forwards user specific methods to the user contained by the MemberImpl type.


MinnDevelopment:
Although we could now remove methods like `GuildController#ban(Member, int)` due to the Liskov substitution principle but I kept them for the sake of readability in the docs.

PR #857, opened 2018-12-27
Don't send presence update when reconnect is queued

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This prevents users from adding to the send queue while waiting for a reconnect.
It doesn't make sense to send 100 presence updates right on startup once the queue reaches the shard.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/managers/PresenceImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/managers/PresenceImpl.java
@@ -34,7 +34,6 @@ import org.json.JSONObject;
  */
 public class PresenceImpl implements Presence
 {
-
     private final UpstreamReference<JDAImpl> api;
     private boolean idle = false;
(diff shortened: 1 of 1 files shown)


DV8FromTheWorld:
Do we want to provide some user feedback when this happens? Or store the most recently package to send?

Currently, this will just silently drop the packages.

MinnDevelopment:
They are not dropped silently, the presence is updated in cache and will be sent with the [identify](https://github.com/DV8FromTheWorld/JDA/blob/48652d541a3ac0e8a5f377376c5ab92d83069ed7/src/main/java/net/dv8tion/jda/core/requests/WebSocketClient.java#L623)

DV8FromTheWorld:
Whoops, you right, you right.

PR #860, opened 2018-12-28
Add VoiceDispatchInterceptor

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This can be useful for integrations like [lavalink](https://github.com/FredBoat/Lavalink-Client) that currently depend on JDA internals.

### Example

```java
public class LavalinkInterceptor implements VoiceDispatchInterceptor
{
    private final JdaLavalink lavalink;

    public LavalinkInterceptor(JdaLavalink lavalink)
    {
        this.lavalink = lavalink;
    }

    @Override
    public void interceptServerUpdate(VoiceServerUpdate update)
    {
        JdaLink link = lavalink.getLink(update.getGuildId());
        link.onVoiceServerUpdate(update.getJSON(), update.getSessionId());
    }

    @Override
    public boolean interceptStateUpdate(VoiceStateUpdate update)
    {
        JdaLink link = lavalink.getLink(update.getGuildId());
        if (update.getChannel() == null && link.getState() == Link.State.DESTROYED)
            link.onDisconnected();
        else if (update.getChannel() != null)
            link.setChannel(update.getChannel().getId());
        return link.getState() == Link.State.CONNECTED;
    }
}
```

-----

```java
   protected void queueAudioDisconnect() {
        Guild g = getJda().getGuildById(guild);

        if (g != null) {
            g.getJDA().getAudioController().disconnect(g);
        } else {
            log.warn("Attempted to disconnect, but guild {} was not found", guild);
        }
    }

    protected void queueAudioConnect(long channelId) {
        VoiceChannel vc = getJda().getVoiceChannelById(channelId);
        if (vc != null) {
            vc.getJDA().getAudioController().connect(vc);
        } else {
            log.warn("Attempted to connect, but voice channel {} was not found", channelId);
        }
    }
```


MinnDevelopment:
There needs to be some way to access the audio queue without using internals

MinnDevelopment:
I would like to merge this after #876

src/main/java/net/dv8tion/jda/api/hooks/VoiceDispatchInterceptor.java
```diff
@@ -0,0 +1,246 @@
...
+ */
+public interface VoiceDispatchInterceptor
+{
+    /**
+     * Handles the <b>VOICE_SERVER_UPDATE</b>.
+     *
+     * @param update
+     *        The {@link VoiceServerUpdate} to handle
+     */
+    void interceptServerUpdate(@Nonnull VoiceServerUpdate update);
```
DV8FromTheWorld:
```suggestion
    void onVoiceServerUpdate(@Nonnull VoiceServerUpdate update);
```

src/main/java/net/dv8tion/jda/api/hooks/VoiceDispatchInterceptor.java
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Handles the <b>VOICE_STATE_UPDATE</b>.
+     * <br>This indicates the user might have moved to a new voice channel.
+     *
+     * @param  update
+     *         The {@link VoiceStateUpdate} to handle
+     *
+     * @return True, if a connection was previously established
+     */
+    boolean interceptStateUpdate(@Nonnull VoiceStateUpdate update);
```
DV8FromTheWorld:
```suggestion
    boolean onVoiceStateUpdate(@Nonnull VoiceStateUpdate update);
```

src/main/java/net/dv8tion/jda/api/hooks/VoiceDispatchInterceptor.java
```diff
@@ -0,0 +1,246 @@
...
+        {
+            return getJDA().getAudioController();
+        }
+
+        /**
+         * Shortcut to access the guild id
+         *
+         * @return The guild id
+         */
+        default long getGuildIdLong()
```
DV8FromTheWorld: I feel like this and the `getGuildId` shortcuts aren't really worth it.

src/main/java/net/dv8tion/jda/api/managers/DirectAudioController.java
```diff
@@ -0,0 +1,86 @@
...
+     * <br>This is done to retry the voice updates in case of a partial service failure.
+     *
+     * <p>Should be called when:
+     * <ol>
+     *     <li>Receiving a Voice State Update for the current account and we were previously connected (moved or disconnected)</li>
+     *     <li>Receiving a Voice Server Update (initial connect or region change)</li>
+     * </ol>
+     *
+     * Note that the voice state update will always be received prior to a voice server update.
+     * <br>THe internal dispatch handlers already call this when needed, a library end-user never needs to call this method.
```
DV8FromTheWorld:
```suggestion
     * <br>The internal dispatch handlers already call this when needed, a library end-user never needs to call this method.
```

src/main/java/net/dv8tion/jda/internal/handle/VoiceStateUpdateHandler.java
```diff
@@ -183,5 +188,10 @@ else if (channel == null)
                                 member, oldChannel));
             }
         }
+        if (voiceInterceptor != null)
```
DV8FromTheWorld: This implicitly means "If this is for the selfUser _and_ we have a voiceInterceptor.
Instead, we should be explicit in checking `selfUser && voiceInterceptor != null`.

As such this retrieval should not be selfUser guarded: https://github.com/DV8FromTheWorld/JDA/pull/860/files#diff-daf701d8ea1b39bcdb9e6e159a9c79fbR103

src/main/java/net/dv8tion/jda/internal/managers/DirectAudioControllerImpl.java
```diff
@@ -0,0 +1,77 @@
...
+        return api.get();
+    }
+
+    @Override
+    public void connect(VoiceChannel channel)
+    {
+        Checks.notNull(channel, "Voice Channel");
+        JDAImpl jda = getJDA();
+        WebSocketClient client = jda.getClient();
+        client.queueAudioConnect(channel);
```
DV8FromTheWorld:
```java
Checks.notNull(channel, "Voice Channel");
getJDA().getClient().queueAudioConnect(channel);
```

src/main/java/net/dv8tion/jda/api/JDA.java
```diff
@@ -316,6 +317,17 @@ default JDA awaitReady() throws InterruptedException
...
+    /**
+     * Direct access to audio (dis-)connect requests.
+     * <br>This should not be used when normal audio operation is desired.
+     *
+     * <p>The correct way to open and close an audio connection is through the {@link Guild Guild's}
+     * {@link AudioManager}.
+     *
+     * @return The {@link DirectAudioController} for this JDA instance
+     */
+    DirectAudioController getAudioController();
```
DV8FromTheWorld: We should probably call this `getDirectAudioController` so as to prevent mixups with `Guild#getAudioController()`

MinnDevelopment:
>Is there any use for DirectAudioController#update outside of JDA's internals?

I'll be honest with you, I added it thinking there would be a reason for it. I haven't figured out if it should never be used yet. I'm a bit split on whether it should be removed or not.

src/main/java/net/dv8tion/jda/internal/managers/DirectAudioControllerImpl.java
MinnDevelopment: Why does this matter?

src/main/java/net/dv8tion/jda/api/hooks/VoiceDispatchInterceptor.java
MinnDevelopment: Worth it? I don't feel like they hurt by being there

src/main/java/net/dv8tion/jda/internal/managers/DirectAudioControllerImpl.java
DV8FromTheWorld: Cleanliness.
I put it as a comment, not a `request changes` for a reason.

DV8FromTheWorld:
It is basically just to tell the audio system that we've finally connected right (the WS finally got to it)?
Seems like that probably shouldn't be exposed to users.

As a side note, does `reconnect` fall into the same category (should a user have access)?

MinnDevelopment:
`reconnect` is used when the audio web socket disconnected due to some voice server issue.

DV8FromTheWorld:
Okay, right. I remember now. thanks.

MinnDevelopment:
I removed `update` from the interface for now, if the need for it arises we can always add it back anyway.

MinnDevelopment:
This should be ready for merge

PR #863, opened 2019-01-01
Improve IPermissionHolder and PermissionOverrideAction

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This prevents users from creating overrides that allow and deny permissions at the same time.

Additionally I added a few more convenience methods to both IPermissionHolder and PermissionOverrideAction. This includes making Member extend ISnowflake to reduce verbosity such as 

```java
long id = permHolder instanceof Member ? ((Member) permHolder).getUser().getIdLong() : ((Role) permHolder).getIdLong();
```


MinnDevelopment:
Oh hold for #852

src/main/java/net/dv8tion/jda/api/requests/restaction/PermissionOverrideAction.java
```diff
@@ -65,23 +63,6 @@
...
-
-    /**
-     * Whether this override is for a {@link Role}
-     *
-     * @return True, if this override is for a role
-     */
-    default boolean isRoleOverride()
-    {
-        return !isMemberOverride();
-    }
```
MinnDevelopment: These were removed because we already had `isRole` and `isMember` in this interface.

src/main/java/net/dv8tion/jda/api/entities/IPermissionHolder.java
```diff
@@ -26,8 +26,57 @@
...
+     * <br>If this is not a role, this will return {@code null} instead.
+     *
+     * @return This permission holder as a concrete Role, or null
+     *
+     * @see    #isRole()
+     * @see    #asMember()
+     */
+    default Role asRole()
+    {
+        return isRole() ? (Role) this : null;
```
DV8FromTheWorld: This should throw, not return null.

src/main/java/net/dv8tion/jda/api/entities/IPermissionHolder.java
```diff
@@ -26,8 +26,57 @@
...
+    /**
+     * Converts this abstraction to the concrete {@link net.dv8tion.jda.api.entities.Member Member} type.
+     * <br>If this is not a member, this will return {@code null} instead.
+     *
+     * @return This permission holder as a concrete Member, or null
+     *
+     * @see    #isMember()
+     * @see    #asRole()
+     */
+    default Member asMember()
```
DV8FromTheWorld: This means that `Role#asMember()` is a thing now, yes?

Feel like this is an anti-pattern of some kind, but I've nothing more to say on the topic.

src/main/java/net/dv8tion/jda/api/managers/ChannelManager.java
```diff
@@ -160,7 +160,7 @@ default Guild getGuild()
     /**
      * Adds an override for the specified {@link net.dv8tion.jda.api.entities.IPermissionHolder IPermissionHolder}
      * with the provided raw bitmasks as allowed and denied permissions. If the permission holder already
-     * had an override on this channel it will be updated instead.
+     * had an override on this channel it will be replaced instead.
```
DV8FromTheWorld: Not that it is wrong, but why the change in vernacular?

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
```diff
@@ -106,15 +106,12 @@ public JDA getJDA()
     }
 
     @Override
-    public PermissionOverride getPermissionOverride(Member member)
+    public PermissionOverride getPermissionOverride(IPermissionHolder permissionHolder)
     {
-        return member != null ? overrides.get(member.getUser().getIdLong()) : null;
-    }
+        if (permissionHolder == null)
+            return null;
```
DV8FromTheWorld: Do we want to maintain this functionality? Or should we throw `IllegalArgumentException` if the holder is null?

DV8FromTheWorld: I believe we _should_ throw.

src/main/java/net/dv8tion/jda/internal/requests/restaction/PermissionOverrideActionImpl.java
```diff
@@ -97,19 +63,13 @@ public GuildChannel getChannel()
     @Override
     public Role getRole()
     {
-        return role;
+        return isRole() ? (Role) permissionHolder : null;
```
DV8FromTheWorld: Thoughts on throwing instead of returning null?

src/main/java/net/dv8tion/jda/api/entities/IPermissionHolder.java
MinnDevelopment: I actually don't know why these were added anymore, I believe the intention was to make retrieval of IDs simpler. However, since I decided to make Member extend ISnowflake that should no longer be needed

src/main/java/net/dv8tion/jda/api/managers/ChannelManager.java
MinnDevelopment: The reason here is that it seems like "updated" implies it only changes he set values, but in reality we completely replace the entity.

src/main/java/net/dv8tion/jda/internal/requests/restaction/PermissionOverrideActionImpl.java
MinnDevelopment: I think its better to return null, this makes it easier to check whether an override is for a specific entity.

Example: 
`guild.getPublicRole().equals(override.getRole())`

vs

`override.isRole() && guild.getPublicRole().equals(override.getRole())`

src/main/java/net/dv8tion/jda/api/managers/ChannelManager.java
DV8FromTheWorld: Makes sense.

src/main/java/net/dv8tion/jda/internal/requests/restaction/PermissionOverrideActionImpl.java
DV8FromTheWorld: I do feel like the latter adds clarity, and throwing an error would result in people writing code that is less prone to having silent errors.

I don't feel strongly enough to forcibly require the change though.

PR #865, opened 2019-01-02
Remove RequestFuture interface and rename LockIterator

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

The RequestFuture interface is no longer used as it involved unwanted verbosity and unneeded encapsulation. We have switched to CompletableFuture in v4.

Since `ClosableIteratorImpl` sounds like an internal class I renamed it to `LockIterator`. This class
can be used by the library end-user if the need for is present.


src/main/java/net/dv8tion/jda/api/entities/GuildVoiceState.java
```diff
@@ -16,12 +16,37 @@
...
 {
+    /**
+     * Returns the {@link net.dv8tion.jda.api.JDA JDA} instance of this VoiceState
+     *
+     * @return The corresponding JDA instance
+     */
+    JDA getJDA();
+
+    /**
+     * Returns whether the {@link net.dv8tion.jda.api.entities.Member Member}/{@link net.dv8tion.jda.api.entities.User User}
```
DV8FromTheWorld: This does not need reference to `User`

src/main/java/net/dv8tion/jda/api/entities/GuildVoiceState.java
```diff
@@ -16,12 +16,37 @@
...
+     */
+    boolean isSelfMuted();
+
+    /**
+     * Returns whether the {@link net.dv8tion.jda.api.entities.Member Member}/{@link net.dv8tion.jda.api.entities.User User}
+     * deafened themselves.
+     *
+     * @return The User's self-deaf status
+     */
+    boolean isSelfDeafened();
```
DV8FromTheWorld: This does not need reference to `User`

Issue #838, opened 2018-12-04
Changing Message#getContentStripped


Andre601:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

### Issue Type

- [ ] Bug Report
- [x] Feature Request

### Description

At the moment does `Message#getContentStripped();` remove symbols like `*` if they format the text (For example `*text*` becomes `text`).
But sadly is that also the case when the symbol is escaped, what also makes it look weird (`\*text*` becomes `\text`)
Would it be possible, to change that in a way, that it won't remove the symbols, if they are escaped.

BenjaminUrquhart:
Adding spoilers to Message#getContentStripped would also be nice

MinnDevelopment:
I'm currently working on something that could be used here: [MarkdownSanitizer](https://github.com/DV8FromTheWorld/JDA/blob/feature/v4/markdown/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java). This is still a WIP class but its probably a better approach to have a dedicated state-machine for sanitizing markdown formats.

MinnDevelopment:
Turns out markdown is even more complicated than I imagined. A simple state-machine won't work because something like "\`abc \*\*x\*\*" requires a backtracking algorithm.

BenjaminUrquhart:
Markdown is the bane of every chat client in existence

MinnDevelopment:
This is available as of [4.ALPHA.0_55](https://ci.dv8tion.net/job/JDA4-Alpha/55/)

Andre601:
But would that now allow to make `\*Your text* *And formatted text*` become \*Your text* *And formatted text* in `Message#getContentStripped();`?

MinnDevelopment:
That would have to be `\*Your text\* *And formatted text*` instead to be escaped correctly

PR #872, opened 2019-01-06
Mini modification to ReceivedMessage#getContentStripped in order to p…

RefractedShadow:
…reserve escaped markdown characters

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #838 

## Description

Added an if statement at the start of ReceivedMessage#getContentStripped in order to test if a message started with "\" and if so, returns everything literally, preserving the markdown characters.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java
@@ -464,6 +464,8 @@ public class ReceivedMessage extends AbstractMessage
             if (strippedContent != null)
                 return strippedContent;
             String tmp = getContentDisplay();
+            if(tmp.startsWith("\\"))
+                return tmp;
             //all the formatting keys to keep track of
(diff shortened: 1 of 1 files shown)


kantenkugel:
I don't really think this is expected behavior... it may be a "feature" but def. not something users would expect. This is also very easily done on the lib users end like

```
return msg.getContentDisplay().startsWith("\\") ? msg.getContentDisplay().substring(1) : msg.getContentStripped();
```

RefractedShadow:
> I don't really think this is expected behavior... it may be a "feature" but def. not something users would expect. This is also very easily done on the lib users end like
> 
> ```
> return msg.getContentDisplay().startsWith("\\") ? msg.getContentDisplay().substring(1) : msg.getContentStripped();
> ```

Fair point. Also, I did not acocunt for the fact that users could place the \ character in front of markdown text *in the middle of a message*. I will close this PR for now because I didn't do enough research, and re-open when I manage to preserve all the characters that were intended to be preserved.

PR #876, opened 2019-01-15
Use configs for JDA constructor

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This drastically improves the readability and usability of the JDAImpl constructor. ~~I plan to do~~ I have done the same for the DefaultShardManager constructor ~~later~~.


src/main/java/net/dv8tion/jda/internal/requests/Requester.java
```diff
@@ -66,16 +68,17 @@
...
+        this(api, ((JDAImpl) api).getAuthorizationConfig());
     }
 
-    public Requester(JDA api, AccountType accountType)
+    public Requester(JDA api, AuthorizationConfig authConfig)
     {
-        if (accountType == null)
-            throw new NullPointerException("Provided accountType was null!");
+        if (authConfig == null)
+            throw new NullPointerException("Provided config was null!");
```
DV8FromTheWorld: These are not equivalent. There are now no checks to ensure that AccountType was provided.

MinnDevelopment: These were not intended to be equivalent in the first place, the checks should be done in the configs now. I haven't gotten to that though.

MinnDevelopment:
With the latest changes you can start using JDAImpl with only this code:

```java
AuthorizationConfig conf = new AuthorizationConfig(AccountType.BOT, TOKEN);
new JDAImpl(conf).login();
```

MinnDevelopment:
I've decided to get rid of `setCorePoolSize` since all it does is increase the thread count of the rate limiter which can be done by changing the rate limit pool instead.

MinnDevelopment:
Creating a basic shard manager is even simpler than a JDAImpl instance

```java
new DefaultShardManager(BOT_TOKEN).login();
```

MinnDevelopment:
This also improves maintainability, as changes to the constructor no longer create configs as easily. Changes to different configs can be merged without conflicts needing to be resolved. An example of this is present in #860 where a change to both constructors is needed but could be reduced to a config change instead.

src/main/java/net/dv8tion/jda/internal/utils/config/ThreadingConfig.java
```diff
@@ -0,0 +1,130 @@
...
+
+public class ThreadingConfig
+{
+    private ScheduledExecutorService rateLimitPool;
+    private ScheduledExecutorService gatewayPool;
+    private ExecutorService callbackPool = ForkJoinPool.commonPool();
+
+    private boolean shutdownRateLimitPool = true;
+    private boolean shutdownGatewayPool = true;
+    private boolean shutdownCallbackPool = false;
```
DV8FromTheWorld: Use a parameterless constructor to set these. We may need to do more than this in the future, so lets do what the other configs are doing and use a constructor.

src/main/java/net/dv8tion/jda/internal/utils/config/MetaConfig.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public boolean isUseShutdownHook()
+    {
+        return useShutdownHook;
+    }
+
+    public static MetaConfig getDefault()
+    {
+        return defaultConfig;
```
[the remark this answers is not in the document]
MinnDevelopment: Because this is the only one (for JDA instances) that does not get updated later.

- EventConfig adds/removes listeners
- SessionConfig sets timeouts
- AuthorizationConfig doesn't have a default because that would be nonsense

src/main/java/net/dv8tion/jda/internal/utils/config/ThreadingConfig.java
MinnDevelopment: Yeah why not

PR #879, opened 2019-01-18
JSONException issue in description variable

stachu540:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

Discord last time assign [those variables](https://discordapp.com/developers/docs/topics/oauth2#get-current-application-information) to nullabe. Since [this one](https://pastebin.com/57Dw1wh5). Those minimal changes only change nullable `description` to be a `null`

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
@@ -1286,7 +1286,7 @@ public class EntityBuilder
 
     public ApplicationInfo createApplicationInfo(JSONObject object)
     {
-        final String description = object.getString("description");
+        final String description = object.optString("description", null);
         final boolean doesBotRequireCodeGrant = object.getBoolean("bot_require_code_grant");
         final String iconId = object.optString("icon", null);
         final long id = object.getLong("id");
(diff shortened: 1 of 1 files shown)


Sanduhr32:
I would like to know if it's consistent `null` or an empty string from Discord. In any case it should be a consistent null or empty string in jda.

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
```diff
@@ -1286,7 +1286,7 @@ else if (channelType.isGuild())
 
     public ApplicationInfo createApplicationInfo(JSONObject object)
     {
-        final String description = object.getString("description");
+        final String description = object.optString("description", null);
```
MinnDevelopment: This should be `optString("description, "")`

kantenkugel:
If this is in any way similar to the channel description changes they did a while back, then yes, it can be either null or an empty string... and i feel like the empty string is the more feasible default in those cases then

src/main/java/net/dv8tion/jda/core/entities/EntityBuilder.java
stachu540: Wish granted

stachu540:
Confirm updated referenced PR discordapp/discord-api-docs#814
image
Closing this PR

Issue #886, opened 2019-01-21
Permission category inconsistencies


averen:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

### Issue Type

- [x] Bug Report
- [ ] Feature Request

### Description

By using `isText()` on `MESSAGE_ADD_REACTION`, false is returned. However `isChannel()` returns true. This implies that reactions can be added to voice channels, but they can not.

PR #887, opened 2019-01-21
Add MESSAGE_ADD_REACTION to text only permissions

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #886 

## Description

You can only use this permissions on text channels

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/Permission.java
+++ b/src/main/java/net/dv8tion/jda/core/Permission.java
@@ -188,7 +188,7 @@ public enum Permission
      */
     public boolean isVoice()
     {
-        return offset == 10 || offset == 8 || (offset > 19 && offset < 26);
+        return offset == 8 || offset == 10 || (offset > 19 && offset < 26);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


DV8FromTheWorld:
It feels like the pattern of determining text vs voice based on offset ranges here is becoming an anti-pattern that we have to revisit every time permissions change to validate that it is still correct.

Perhaps we should consider a different way to determining if it is text or voice specific

MinnDevelopment:
@DV8FromTheWorld I agree but I feel like that would be more suited as a version 4 change.

DV8FromTheWorld:
Perhaps, but it wouldn't be a breaking change from the api perspective. It'd probably involve additional constructor params though.

I'm fine with with a v3 or v4 change, but I'd like there to be a change in the codebase.

src/main/java/net/dv8tion/jda/core/Permission.java
```diff
@@ -188,7 +188,7 @@ public boolean isText()
      */
     public boolean isVoice()
     {
-        return offset == 10 || offset == 8 || offset > 19 && offset < 26;
+        return offset == 10 || offset == 8 || (offset > 19 && offset < 26);
```
DV8FromTheWorld:
```suggestion
        return offset == 8 || offset == 10 || (offset > 19 && offset < 26);
```

PR #888, opened 2019-01-21
Invert dependencies of text/voice only property

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Make text/voice only property depend on the bitmask for the specific type.

I also cleaned up the class a tiny bit.


DV8FromTheWorld:
We will need to be diligent in remembering to add the permission to one of the "all" collections.

MinnDevelopment:
Luckily new permissions are not added frequently which makes this the best solution, in my opinion.

DV8FromTheWorld:
Agreed.

PR #891, opened 2019-01-23
Rename methods used to retrieve entities

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

I believe this to be a more intuitive naming conventions.


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -127,7 +127,7 @@
     /**
      * Pattern used to find instant invites in messages.
      *
-     * @see #getInvites() getInvites()
+     * @see #getInvites() retrieveInvites()
```
Sanduhr32: You might have missed renaming the method here in the `@see`

Issue #893, opened 2019-01-24
EmbedBuilder has alpha value for color making MessageEmbed.equals defect


BalintGergely:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [ ] I have updated to the [latest JDA version][download].
- [ ] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

## Issue

MessageEmbed.equals() mistakenly returns false for equal embeds if the color is set from a java.awt.Color object due to the color having an alpha value.

### Issue Type

- [x] Bug Report
- [ ] Feature Request

### Description

`EmbedBuilder.setColor(Color)` doesn't ignore the color's alpha. Uses `this.color = color.getRGB()`.
After this, `MessageEmbed`s built with the builder have color values with alpha, and their `equals` method always returns false for `MessageEmbed`s retrieved from `Message`s that were received by `JDA`. (Those never have alpha values.)

### Possible fix

`EmbedBuilder.setColor(Color color)` should ignore alpha: `this.color = color.getRGB() & 0xffffff`
`EmbedBuilder.setColor(int color)` should also ignore alpha: `this.color = color & 0xffffff`

PR #896, opened 2019-01-26
Ignore alpha when comparing colors in MessageEmbed

babbaj:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #893

## Description

see title and #893

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/entities/MessageEmbed.java
+++ b/src/main/java/net/dv8tion/jda/core/entities/MessageEmbed.java
@@ -398,7 +398,7 @@ public class MessageEmbed
             && Objects.equals(videoInfo, other.videoInfo)
             && Objects.equals(footer, other.footer)
             && Objects.equals(image, other.image)
-            && Objects.equals(color & 0x00FFFFFF, other.color & 0x00FFFFFF)
+            && (color & 0xFFFFFF) == (other.color & 0xFFFFFF)
             && Objects.equals(timestamp, other.timestamp)
             && Helpers.deepEquals(fields, other.fields);
     }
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/core/entities/MessageEmbed.java
```diff
@@ -398,7 +398,7 @@ public boolean equals(Object obj)
             && Objects.equals(videoInfo, other.videoInfo)
             && Objects.equals(footer, other.footer)
             && Objects.equals(image, other.image)
-            && Objects.equals(color, other.color)
+            && Objects.equals(color & 0x00FFFFFF, other.color & 0x00FFFFFF)
```
MinnDevelopment: Can we change this to `==` instead of equals?

MinnDevelopment:
```suggestion
            && (color & 0xFFFFFF) == (other.color & 0xFFFFFF)
```

babbaj: the extra zeroes are there for readability

babbaj: it makes it more clear that youre zeroing the first byte

MinnDevelopment: I disagree with that, besides that its also inconsistent with how we do this everywhere in JDA. We never do padding with 0's for this and its not needed. That isn't the important part of the suggested change, the part that actually matters is the change from `equals` to `==`.

PR #897, opened 2019-01-26
Entity Proxies

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Introduces proxy classes that indirectly access the JDA cache by only storing ID and JDA references.

### Proxies

- [x] User
- [x] SelfUser
- [x] Guild
- [x] Role
- [x] Member
- [x] GuildVoiceState
- [x] GuildChannel
- [x] MessageChannel
- [x] Emote
- [x] PermissionOverride

[Package for a complete list](https://github.com/DV8FromTheWorld/JDA/tree/feature/v4/proxy/src/main/java/net/dv8tion/jda/api/entities/proxy)

Let me know if I missed any entities that rely on the cache. Things like Webhooks are not included as they are not actually cached and thus cannot be proxied.

Currently we tell people to store the id of entities and resolve those ids to local variables. With this we can change and start saying "if it implements ProxySubject, use the proxy". 

Since all the proxies actually implement the interfaces of the subjects they can easily be used in existing code.

### Example

```diff
public class NameManager extends ListenerAdapter {
    private final Guild guild;

    public NameManager(Guild guild) {
-        this.guild = guild;
+        this.guild = guild.getProxy();
    }
...
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Category.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Category.java
@@ -16,6 +16,7 @@
 
 package net.dv8tion.jda.api.entities;
 
+import net.dv8tion.jda.api.entities.proxy.CategoryProxy;
 import net.dv8tion.jda.api.requests.restaction.ChannelAction;
 import net.dv8tion.jda.api.requests.restaction.order.CategoryOrderAction;
 import net.dv8tion.jda.api.requests.restaction.order.ChannelOrderAction;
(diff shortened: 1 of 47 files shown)


MinnDevelopment:
We now have to keep in mind here that entities are no longer one-to-one mapped to implementations and that our habit of simply casting interfaces without checking types would be an issue with this pattern. I have already done some work to remove casts like this from the API.

MinnDevelopment:
I added overrides for equals that just call equals on the subject of the proxy, that however violates the equals contract, specifically symmetry. Either we only check for equality of the proxy by id or we add the proxies to the equality contracts of the subject.

DV8FromTheWorld:
I understand the original motivation behind this proposal, but I'm not in support of the change.

Since it has been a while, lets reopen conversation about it or close down this PR.

PR #899, opened 2019-01-27
Use OkHTTP3 snapshot

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: Update dependency

Closes Issue: NaN

## Description

It seems like OkHTTP 3.13.0 is not coming any closer to being released, this is a solution that would work for us if we decide to just give up on waiting.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,6 +39,7 @@ sourceCompatibility = targetCompatibility = 1.8
 
 repositories {
     jcenter()
+    maven { url 'https://oss.sonatype.org/content/repositories/snapshots/' }
 }
 
 sourceSets {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -60,12 +61,14 @@ dependencies {
...
     api 'com.neovisionaries:nv-websocket-client:2.5'
-    api 'com.squareup.okhttp3:okhttp:3.12.0'
+    api 'com.squareup.okhttp3:okhttp:3.13.0-SNAPSHOT'
 
     //Opus library support
     api ('club.minnced:opus-java:1.0.4@pom') {
         transitive = true
     }
+    
+    compileOnly 'net.java.dev.jna:jna:4.4.0'
```
DV8FromTheWorld: Why this?

MinnDevelopment: Because its a runtime dependency of opus-java-api but we also use it for the Decoder class in JDA

DV8FromTheWorld: Was it not in the dependency tree before? This diff makes it look like it was added, not moved.

DV8FromTheWorld:
This PR should be accepted if work isn't progressing on OkHTTP by Feb 1, 2019.

build.gradle
MinnDevelopment: Because it was added, we need it at compile-time for our class. Gradle only recently started actually enforcing this rule and it seems a bit inconsistent. Sometimes compilation fails and sometimes it works. This only ensures that JNA is present at compile-time but doesn't add it as an explicit dependency of JDA. We only need it for Decoder when opus-java is present anyway.

DV8FromTheWorld: `Gradle only recently started actually enforcing this rule and it seems a bit inconsistent.`

Okay, this makes sense. Thanks for the explanation.

PR #902, opened 2019-01-29
Add ClientType enum for client_status support

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This adds support for the new `client_status` property. See discordapp/discord-api-docs#829 for more details.

Some decisions have yet to be made, for instance how to determine the "active status" of a user when multiple clients have a different status set.


src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
```diff
@@ -53,7 +55,9 @@ public MemberImpl(GuildImpl guild, User user)
         this.user = user;
         JDAImpl jda = (JDAImpl) getJDA();
         boolean cacheState = jda.isCacheFlagSet(CacheFlag.VOICE_STATE) || user.equals(jda.getSelfUser());
+        boolean cacheOnline = jda.isCacheFlagSet(CacheFlag.CLIENT_STATUS);
         this.voiceState = cacheState ? new GuildVoiceStateImpl(this) : null;
+        this.clientStatus = cacheOnline ? new ConcurrentHashMap<>(5) : null;
```
DV8FromTheWorld: Where does this `5` come from?

DV8FromTheWorld: Shouldn't this be something more like `ClientType.values().length` ?

MinnDevelopment: The default capacity is usually 16 so this is intended lower the memory footprint since we know its not more then 3 types but there might come a few more in the future.

PR #903, opened 2019-01-29
Add takeUntilAsync

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Greatly simplifies retrieving a bunch of elements until a certain condition is met.

### Example

```java
private static CompletableFuture<List<Message>> getMessagesBetween(long startId, long endId, MessageChannel channel) {
    return channel.getIterableHistory()
            .cache(false).skipTo(endId)
            .takeUntilAsync(it -> Long.compareUnsigned(it.getIdLong(), startId) < 0);
}
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/requests/restaction/pagination/PaginationAction.java
+++ b/src/main/java/net/dv8tion/jda/api/requests/restaction/pagination/PaginationAction.java
@@ -355,7 +355,7 @@ public interface PaginationAction<T, M extends PaginationAction<T, M>> extends R
         List<T> result = new ArrayList<>();
         CompletableFuture<List<T>> future = new CompletableFuture<>();
         CompletableFuture<?> handle = forEachAsync((element) -> {
-            if (!rule.test(element))
+            if (rule.test(element))
                 return false;
             result.add(element);
             return limit == 0 || limit > result.size();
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/requests/restaction/pagination/PaginationAction.java
```diff
@@ -261,6 +263,36 @@
...
+
+    default CompletableFuture<List<T>> takeUntilAsync(int limit, Predicate<T> rule)
+    {
+        Checks.notNull(rule, "Rule");
+        Checks.notNegative(limit, "Limit");
+        List<T> result;
+        if (limit == 0)
+            result = new ArrayList<>();
+        else
+            result = new ArrayList<>(limit);
```
MinnDevelopment: This might be a stupid idea

src/main/java/net/dv8tion/jda/api/requests/restaction/pagination/PaginationAction.java
```diff
@@ -261,6 +263,112 @@
...
+     * @throws IllegalArgumentException
+     *         If the provided rule is {@code null} or the limit is negative
+     *
+     * @return {@link CompletableFuture} - Type: {@link List List}
+     *         <br>Future representing the fetch task, the list will be sorted most recent to oldest
+     *
+     * @see    #takeWhileAsync(Predicate)
+     * @see    #takeUntilAsync(int, Predicate)
+     */
+    default CompletableFuture<List<T>> takeUntilAsync(int limit, final Predicate<? super T> rule)
```
DV8FromTheWorld: The `limit` should come after the `rule` as there is a method that only takes the rule.

MinnDevelopment: This is for better kotlin support. Having the limit be the second parameter would make it look like this in kotlin `action.takeUntilAsync({ condition(it) }, 10)` compared to now `action.takeUntilAsync(10) { condition(it) }`.

```kotlin
action.takeUntilAsync(1000) {
    it.idLong > startId
}
```

PR #908, opened 2019-02-02
Don't catch rejection of worker schedule

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

We should not swallow rejected executions of the startup worker in DefaultShardManager.


MinnDevelopment:
@Almighty-Alpaca I'm fairly certain this can only happen when `shutdown()` was called anyway which means the shards in the queue will just never be started.

Almighty-Alpaca:
Didn't think about that

PR #913, opened 2019-02-04
Add attachment options to apply to files

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds options to `addFile` overloads allowing to specify how the attachments should be used.
Currently the only option is `SPOILER` which marks the attachment as a spoiler. This is best for maintainability in the future, compared to simply adding a boolean flag.

### Example

```java
channel.sendMessage("SPOILER FOR XYZ:")
    .addFile(image, AttachmentOption.SPOILER)
    .queue();
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageActionImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageActionImpl.java
@@ -24,6 +24,7 @@ import net.dv8tion.jda.api.exceptions.InsufficientPermissionException;
 import net.dv8tion.jda.api.requests.Request;
 import net.dv8tion.jda.api.requests.Response;
 import net.dv8tion.jda.api.requests.restaction.MessageAction;
+import net.dv8tion.jda.api.utils.AttachmentOption;
 import net.dv8tion.jda.api.utils.MiscUtil;
 import net.dv8tion.jda.internal.requests.Method;
 import net.dv8tion.jda.internal.requests.Requester;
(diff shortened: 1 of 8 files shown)


src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageActionImpl.java
```diff
@@ -258,6 +263,22 @@ public MessageActionImpl override(final boolean bool)
         return this;
     }
 
+    private String applyOptions(String name, AttachmentOption[] options)
```
MinnDevelopment: This could certainly be done differently if we choose to add more options in the future. This change is mostly supposed to just support that kind of change without breaking the interface further in the future.

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageActionImpl.java
```diff
@@ -181,29 +182,33 @@ public MessageActionImpl append(final char c)
...
         return this;
     }
 
     @Override
     @CheckReturnValue
-    public MessageActionImpl addFile(final File file, final String name)
+    public MessageActionImpl addFile(final File file, String name, AttachmentOption... options)
     {
         Checks.notNull(file, "File");
+        Checks.notNull(options, "Options");
```
DV8FromTheWorld: This will need to check that each entry in `options` isn't null.

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageActionImpl.java
```diff
@@ -258,6 +263,22 @@ public MessageActionImpl override(final boolean bool)
         return this;
     }
 
+    private String applyOptions(String name, AttachmentOption[] options)
+    {
+        if (options.length > 0)
```
DV8FromTheWorld: We don't need to check if the options has any length. The loop will just skip if none were provided.

DV8FromTheWorld:
Also, this isn't a breaking change, right? Already existing code should still compile even given the new `AttachmentOption...` options parameter

MinnDevelopment:
This is a breaking ABI change because it changes the method signature. Code that was compiled against a previous version would no longer work.

DV8FromTheWorld:
Ah, you are referring to ABI. Gotcha.

DV8FromTheWorld:
Are we ready to merge this in?

Issue #906, opened 2019-02-01
Strip spoiler tags in Message#getContentStripped


BenjaminUrquhart:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [X] I have checked for similar issues.
- [X] I have updated to the [latest JDA version][download].
- [X] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

### Issue Type

- [ ] Bug Report
- [X] Feature Request

### Description

I think the title says it all. `|| spoiler ||` should become `spoiler`.

![czmv1g85i1qoug](https://user-images.githubusercontent.com/24701883/52124044-0524ef80-25f6-11e9-9a02-0ed65f29be3a.png)

I'm pretty sure simply adding `||` to [ReceivedMessage#getContentStripped on line 465](https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/core/entities/impl/ReceivedMessage.java#L465) will accomplish this.

MinnDevelopment:
I think this should be combined with #838. Please post a comment there instead.

PR #914, opened 2019-02-04
Added spoilers to Message#getContentStripped

BenjaminUrquhart:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #906 

## Description

Message#getContentStripped should now strip spoiler tags

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -344,9 +344,9 @@ public interface Message extends ISnowflake, Formattable
     String getContentRaw();
 
     /**
-     * Gets the textual content of this message using {@link #getContentDisplay()} and then strips it of all spoiler tags 
-     * and markdown characters like {@literal *, **, __, ~~, ||} that provide text formatting. Any characters that match these 
-     * but are not being used for formatting are escaped to prevent possible formatting.
+     * Gets the textual content of this message using {@link #getContentDisplay()} and then strips it of markdown characters 
+     * like {@literal *, **, __, ~~, ||} that provide text formatting. Any characters that match these but are not being used 
+     * for formatting are escaped to prevent possible formatting.
      *
      * @return The textual content from {@link #getContentDisplay()} with all text formatting characters removed or escaped.
      */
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -344,9 +344,9 @@
     String getContentRaw();
 
     /**
-     * Gets the textual content of this message using {@link #getContentDisplay()} and then strips it of all markdown characters
-     * like {@literal *, **, __, ~~} that provide text formatting. Any characters that match these but are not being used
-     * for formatting are escaped to prevent possible formatting.
+     * Gets the textual content of this message using {@link #getContentDisplay()} and then strips it of all spoiler tags 
+     * and markdown characters like {@literal *, **, __, ~~, ||} that provide text formatting. Any characters that match these
```
MinnDevelopment: The spoiler tags are also markdown characters so this is redundant

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -344,9 +344,9 @@
     String getContentRaw();
 
     /**
-     * Gets the textual content of this message using {@link #getContentDisplay()} and then strips it of all markdown characters
-     * like {@literal *, **, __, ~~} that provide text formatting. Any characters that match these but are not being used
-     * for formatting are escaped to prevent possible formatting.
+     * Gets the textual content of this message using {@link #getContentDisplay()} and then strips it of all spoiler tags 
+     * and markdown characters like {@literal *, **, __, ~~, ||} that provide text formatting. Any characters that match these 
+     * but are not being used for formatting are escaped to prevent possible formatting.
```
MinnDevelopment: Why was this changed?

src/main/java/net/dv8tion/jda/api/entities/Message.java
BenjaminUrquhart: Will revert

src/main/java/net/dv8tion/jda/api/entities/Message.java
BenjaminUrquhart: I tried to preserve the line length

PR #916, opened 2019-02-06
Make reactions more intuitive

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Make usage of reactions a little more intuitive by having some redundancies with better names.


src/main/java/net/dv8tion/jda/internal/utils/EncodingUtil.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+        catch (UnsupportedEncodingException e)
+        {
+            throw new AssertionError(e); // thanks JDK 1.4
+        }
+    }
+
+    public static String encodeCodepointsUTF8(String input)
+    {
+        if (!input.startsWith("U+"))
```
DV8FromTheWorld: Can codepoints start with `u+`? or must it be `U+`?

MinnDevelopment: I have never seen it being lowercase at least. The common notation is `U+CODE`

DV8FromTheWorld: Lets include information about what constitutes a valid format in the message of the `IllegalArgumentException`.
`Invalid Format. Example: "U+1F44D", Provided: "{input}"`

MinnDevelopment:
For some reason I can't add a comment to that review comment, so here is my response instead:
`encodeCodepointsUTF8` is only called by `addReactionById` and only if the code starts with `U+`.

src/main/java/net/dv8tion/jda/internal/utils/EncodingUtil.java
MinnDevelopment: `encodeCodepointsUTF8` is only called by `addReactionById` and only if the code starts with `U+`.

PR #921, opened 2019-02-07
Use ByteBuffer for sending audio

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This change allows send handlers to re-use a single buffer for all packets.

### Example Usage

```java
public class AudioSender implements AudioSendHandler {
    private AudioPlayer player;
    private AudioFrame frame;
    private ByteBuffer buffer = ByteBuffer.allocate(1024);

    public AudioSender(AudioPlayer player) {
        this.player = player;
    }

    @Override
    public boolean canProvide() {
        frame = player.provide();
        return frame != null;
    }

    @Override
    public ByteBuffer provide20MsAudio() {
        frame.getData(buffer.array(), frame.getDataLength()); // lavaplayer still had the old broken behavior where it used offset as length in arraycopy
        buffer.position(0);
        buffer.limit(frame.getDataLength());
        return buffer;
    }

    @Override
    public boolean isOpus() {
        return true;
    }
}
```

### DIFF
--- a/src/examples/java/AudioEchoExample.java
+++ b/src/examples/java/AudioEchoExample.java
@@ -212,10 +212,10 @@ public class AudioEchoExample extends ListenerAdapter
         @Override
         public void handleCombinedAudio(CombinedAudio combinedAudio)
         {
+            // we only want to send data when a user actually sent something, otherwise we would just send silence
             if (combinedAudio.getUsers().isEmpty())
                 return;
 
-            // we only want to send data when a user actually sent something, otherwise we just send silence
             byte[] data = combinedAudio.getAudioData(1.0f); // volume at 100% = 1.0 (50% = 0.5 / 55% = 0.55)
             queue.add(data);
         }
--- a/src/examples/java/AudioEchoExample.java
+++ b/src/examples/java/AudioEchoExample.java
@@ -1,0 +1,250 @@
+/*
+ * Copyright 2015-2019 Austin Keener, Michael Ritter, Florian Spieß, and the JDA contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import net.dv8tion.jda.api.JDABuilder;
+import net.dv8tion.jda.api.OnlineStatus;
+import net.dv8tion.jda.api.audio.AudioReceiveHandler;
+import net.dv8tion.jda.api.audio.AudioSendHandler;
+import net.dv8tion.jda.api.audio.CombinedAudio;
+import net.dv8tion.jda.api.audio.UserAudio;
+import net.dv8tion.jda.api.entities.*;
+import net.dv8tion.jda.api.events.message.guild.GuildMessageReceivedEvent;
+import net.dv8tion.jda.api.hooks.ListenerAdapter;
+import net.dv8tion.jda.api.managers.AudioManager;
+
+import javax.security.auth.login.LoginException;
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+public class AudioEchoExample extends ListenerAdapter
+{
+    public static void main(String[] args) throws LoginException
+    {
+        if (args.length == 0)
+        {
+            System.err.println("Unable to start without token!");
+            System.exit(1);
+        }
+        String token = args[0];
+
+        new JDABuilder(token)                            // Use provided token from command line arguments
+             .addEventListeners(new AudioEchoExample())  // Start listening with this listener
+             .setActivity(Activity.listening("to jams")) // Inform users that we are jammin' it out
+             .setStatus(OnlineStatus.DO_NOT_DISTURB)     // Please don't disturb us while we're jammin'
+             .build();                                   // Login with these options
+        // Note that its not needed to explicitly enable audio here
+    }
+
+    @Override
+    public void onGuildMessageReceived(GuildMessageReceivedEvent event)
+    {
+        Message message = event.getMessage();
+        User author = message.getAuthor();
+        String content = message.getContentRaw();
+        Guild guild = event.getGuild();
+
+        // Ignore message if bot
+        if (author.isBot())
+            return;
+
+        if (content.startsWith("!echo "))
+        {
+            String arg = content.substring("!echo ".length());
+            onEchoCommand(event, guild, arg);
+        }
+        else if (content.equals("!echo"))
+        {
+            onEchoCommand(event);
+        }
+    }
+
+    /**
+     * Handle command without arguments.
+     *
+     * @param event
+     *        The event for this command
+     */
+    private void onEchoCommand(GuildMessageReceivedEvent event)
+    {
+        Member member = event.getMember();                              // Member is the context of the user for the specific guild, containing voice state and roles
+        GuildVoiceState voiceState = member.getVoiceState();            // Check the current voice state of the user
+        VoiceChannel channel = voiceState.getChannel();                 // Use the channel the user is currently connected to
+        if (channel != null)
+        {
+            connectTo(channel);                                         // Join the channel of the user
+            onConnecting(channel, event.getChannel());                  // Tell the user about our success
+        }
+        else
+        {
+            onUnknownChannel(event.getChannel(), "your voice channel"); // Tell the user about our failure
+        }
+    }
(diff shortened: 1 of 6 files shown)


MinnDevelopment:
Both receive and send seem to be working as expected after these changes. This implies encoding/decoding of opus and encryption of audio are unaffected and work just like before.

The conversion for v3 users is actually a one-liner:

```diff
-    return data;
+    return data == null ? null : ByteBuffer.wrap(data);
```

DV8FromTheWorld:
Can you provide a full example of a v3 vs v4 implementation of AudioSendHandler?

Also, why does this not impact the AudioReceiveHandler?

MinnDevelopment:
>Can you provide a full example of a v3 vs v4 implementation of AudioSendHandler?

Its exactly the same as v3 except that `provide20MsAudio` should return a `ByteBuffer` with `position` pointing at the read-head of the buffer. For proper usage you would use a [`MutableAudioFrame`](https://github.com/sedmelluq/lavaplayer/blob/master/main/src/main/java/com/sedmelluq/discord/lavaplayer/track/playback/MutableAudioFrame.java) in lavaplayer which re-uses the buffer for each frame and populates it.
 `player.provide(mutableFrame)` returning true if something was written to the frame, else false.
This can be used for `canProvide()`. Then in `provide20MsAudio` you simply return the buffer after calling `flip()`.

```java
private final MutableAudioFrame frame = new MutableAudioFrame();
private final ByteBuffer buffer = ByteBuffer.allocate(1024);

public MyConstructor() {
    this.frame.setBuffer(buffer);
}

public ByteBuffer provide20MsAudio() {
    return buffer.flip();
}

public boolean canProvide() {
    return player.provide(frame);
}
```

The difference is that in v3 you have to return a fitting byte array instead which requires re-allocation for every frame.

>Also, why does this not impact the AudioReceiveHandler?

Because audio receive is not affected by changes to the send system.

DV8FromTheWorld:
>Because audio receive is not affected by changes to the send system.
Let me rephrase: Would there be any benefit we could eke out by using a buffer of some sort for the receiving system as well?

MinnDevelopment:
The receive system is generally not very memory optimized. Additionally, I only want to make changes to receive required to keep it working for now.

natanbc:
Currently, lavaplayer might [change](https://javascript.is-bad.com/4c823b.png) the buffer in MutableAudioFrame, so a safer approach for the send handler would be something like this (reflection is needed because there is no other way of knowing whether or not the buffer was replaced)

```java
public class MySendHandler implements AudioSendHandler {
    private static final Function<MutableAudioFrame, ByteBuffer> BUFFER_GETTER;

    static {
        try {
            Field f = MutableAudioFrame.class.getDeclaredField("frameBuffer");
            f.setAccessible(true);
            BUFFER_GETTER= frame -> {
                try {
                    return (ByteBuffer)f.get(frame);
                } catch(Exception impossible) {
                    throw new AssertionError(impossible);
                }
            };
        } catch(Exception e) {
            throw new ExceptionInInitializerError(e);
        }
    }

    //other fields and methods

    @Override
    public ByteBuffer provide20MsAudio() {
        if(BUFFER_GETTER.apply(frame) != buffer) {
            frame.getData(buffer.array(), frame.getDataLength());
            frame.setBuffer(buffer.limit(buffer.capacity()));
        }
        return buffer.position(0).limit(frame.getDataLength());
    }
}
```

MinnDevelopment:
@natanbc can't you just use `getData(buffer.array(), 0)` instead?

natanbc:
That would work as well, but would copy the data every time, this approach only copies when the buffer is replaced. After setting the buffer to the original one, new data will be written to it until LP changes the buffer again

MinnDevelopment:
It would be ideal if `MutableAudioFrame` would have a getter for the buffer it uses. Something like `provideBuffer(): ByteBuffer` or similar which would be safe to use.

MinnDevelopment:
In any case, this doesn't really matter for our interface change. We only need to provide a way to re-use one buffer properly. That should be possible regardless of lavaplayer usage.

src/main/java/net/dv8tion/jda/internal/audio/AudioConnection.java
```diff
@@ -663,8 +666,8 @@ public ByteBuffer getNextPacketRaw(boolean changeTalking)
                 cond: if (sentSilenceOnConnect && sendHandler != null && sendHandler.canProvide())
                 {
                     silenceCounter = -1;
-                    byte[] rawAudio = sendHandler.provide20MsAudio();
-                    if (rawAudio == null || rawAudio.length == 0)
+                    ByteBuffer rawAudio = sendHandler.provide20MsAudio();
+                    if (rawAudio == null || !rawAudio.hasRemaining() || !rawAudio.hasArray()) // no array = cannot encrypt
```
DV8FromTheWorld: `hasRemaining` can be true when there is no array?

src/main/java/net/dv8tion/jda/internal/audio/AudioPacket.java
```diff
@@ -227,24 +216,26 @@ public static AudioPacket decryptAudioPacket(AudioEncryption encryption, Datagra
...
 
-        byte[] encodedAudio;
+        ByteBuffer encodedAudio = encryptedPacket.encodedAudio;
+        int length = encodedAudio.remaining();
+        int offset = encodedAudio.arrayOffset();
         switch (encryption)
         {
             case XSALSA20_POLY1305:
-                encodedAudio = encryptedPacket.getEncodedAudio();
+//                length = encodedAudio.remaining();
```
DV8FromTheWorld: We should uncomment this for clarity.
Not sure what to do with the `int length` declaration above though.

src/main/java/net/dv8tion/jda/internal/audio/AudioConnection.java
MinnDevelopment: Yes, no array just means its not a heap buffer but a direct buffer. `hasRemaining()` is a short handle for `position() < limit()`.

src/main/java/net/dv8tion/jda/internal/audio/AudioPacket.java
MinnDevelopment: I kept it as a comment for clarity, assigning it again is not really needed here. The declaration above is the full length and the special cases subtract the nonce length from it.

src/examples/java/AudioEchoExample.java
```diff
@@ -0,0 +1,250 @@
...
+            return false;
+        }
+
+        @Override
+        public void handleCombinedAudio(CombinedAudio combinedAudio)
+        {
+            if (combinedAudio.getUsers().isEmpty())
+                return;
+
+            // we only want to send data when a user actually sent something, otherwise we just send silence
```
DV8FromTheWorld: The comment should be on the above if statement.

DV8FromTheWorld:
The AudioEcho example is a quality addition. Nice work.

DV8FromTheWorld:
LGTM

src/main/java/net/dv8tion/jda/api/audio/AudioSendHandler.java
```diff
@@ -50,9 +52,14 @@
...
-     * @return Should return a byte[] containing 20 Milliseconds of audio.
+     * @return Should return a {@link java.nio.ByteBuffer} containing 20 Milliseconds of audio.
+     *
+     * @see    #isOpus()
+     * @see    #canProvide()
+     * @see    java.nio.ByteBuffer#allocate(int)
+     * @see    java.nio.ByteBuffer#wrap(byte[])
      */
-    byte[] provide20MsAudio();
+    ByteBuffer provide20MsAudio();
```
napstr: Shouldn't this method have a ByteBuffer as an argument that the audio data should be written to?
Returning a buffer looks off because how does the implementor know when the buffer it returns here is safe to be used again? Or am I on the wrong track about how this would be used in an implementation with the goal of minimal allocations?

MinnDevelopment: If you would do provide(ByteBuffer) you get additional overhead for copying over data. It would also severely limit usability since the buffer could not be re-allocated correctly. The buffer should be updated by the next provide cycle. Meaning canProvide signals the buffer was sent and the next one will be requested. If you are worried about concurrency you an always use `buffer.duplicate()` instead or have a queue of buffers. Since we immediately read the buffer and write the encrypted data into a different buffer of the packet provider it should almost immediately be usable again.

MinnDevelopment: Additionally, the goal of these changes was *not* to force users into single-allocation which your suggestion would imply. The goal was to make it possible in the first place. Since returning a buffer with a position and limit you can create a fitting requirement without re-allocation compared to using a byte[]. This is due to the fact that byte[] needs to be allocated with the exact size needed for the audio frame while the buffer can be adjusted for it without any allocations. To achieve 0 allocations you simply need to have one buffer that is filled by canProvide() and then returned by provide() as was shown by the examples in the comments of this PR.

napstr: Thanks for clarifying!

PR #935, opened 2019-02-18
Add Message#refresh to get latest version of a message instance

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Add convenience method to get a fresh version of an existing message.

```java
 public CompletableFuture<List<MessageReaction>> retrieveReactions(Message m)
 {
     return m.refresh().submit().thenApply(Message::getReactions);
 }

 public void deleteIfExists(Message m)
 {
     m.refresh().queue(
         existing -> existing.delete().queue(), // the message still exists
         failure -> {} // the message has already been removed
     );
 }
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -739,6 +739,33 @@ public interface Message extends ISnowflake, Formattable
     AuditableRestAction<Void> delete();
 
     /**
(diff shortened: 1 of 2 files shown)


MinnDevelopment:
This change would encourage storing messages (and its transitive entities!) for longer periods of time which is not something we should be doing.

PR #936, opened 2019-02-18
Improve attachment retrieval by only offering async variants

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This removes all blocking methods from `Message.Attachment` in favor of async variants. Downloading files should not happen on the event thread at any point in time.

- `download()` -> `downloadToFile()`
- `getAsIcon()` -> `retrieveAsIcon()`

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -19,26 +19,16 @@ import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.requests.RestAction;
 import net.dv8tion.jda.api.requests.restaction.AuditableRestAction;
 import net.dv8tion.jda.api.requests.restaction.MessageAction;
-import net.dv8tion.jda.api.utils.IOConsumer;
 import net.dv8tion.jda.internal.JDAImpl;
 import net.dv8tion.jda.internal.requests.Requester;
-import net.dv8tion.jda.internal.utils.Checks;
-import net.dv8tion.jda.internal.utils.IOUtil;
-import okhttp3.OkHttpClient;
-import okhttp3.Request;
-import okhttp3.Response;
+import okhttp3.*;
 
 import javax.annotation.CheckReturnValue;
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.nio.file.Files;
-import java.nio.file.Paths;
+import java.io.*;
 import java.time.OffsetDateTime;
 import java.util.Formattable;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.CompletableFuture;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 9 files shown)


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1113,114 +1103,95 @@ public String getFileName()
...
+                public void onFailure(Call call, IOException e)
+                {
+                    future.completeExceptionally(new IllegalStateException(e));
+                }
+
+                @Override
+                public void onResponse(Call call, Response response) throws IOException
+                {
+                    InputStream body = Requester.getBody(response);
+                    future.complete(body);
```
MinnDevelopment: We should check for the HTTP status here

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1114,113 +1109,192 @@ public String getFileName()
...
+         * Alternatively you can use {@link #retrieveData()} and use a continuation with a different executor.
          *
-         * @throws java.io.IOException
-         *         If an IO error occurs trying to read from the opened HTTP channel
+         * <h2>Example</h2>
+         * <pre>{@code
+         * public void saveLocally(Message.Attachment attachment)
+         * {
+         *     attachment.downloadToFile("/tmp/" + attachment.getFileName())
+         *         .thenAccept(file -> System.out.println("Saved attachment to " + file.getName())
```
MinnDevelopment: This is missing a parenthesis

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1114,113 +1109,192 @@ public String getFileName()
...
-            return icon.get();
+            CompletableFuture<InputStream> future = new CompletableFuture<>();
+            Request req = getRequest();
+            OkHttpClient httpClient = getJDA().getHttpClient();
+            httpClient.newCall(req).enqueue(new Callback()
+            {
+                @Override
+                public void onFailure(Call call, IOException e)
+                {
+                    future.completeExceptionally(new IllegalStateException(e));
```
MinnDevelopment: Should we use `UncheckedIOException` instead?

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1114,113 +1106,192 @@ public String getFileName()
         }
 
         /**
-         * Creates an {@link net.dv8tion.jda.api.entities.Icon Icon} instance for
-         * this attachment if {@link #isImage() isImage()} is {@code true}!
-         * <br>This is a convenience method that can be used to retrieve an Icon from an attachment image link which
-         * requires a set user-agent to be loaded.
+         * Enqueues a request to retrieve the contents of this Attachment.
+         * <br>The receiver is expected to close the retrieved {@link java.io.InputStream}.
```
DV8FromTheWorld:
```suggestion
         * <br><b>The receiver is expected to close the retrieved {@link java.io.InputStream}.</b>
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1114,113 +1106,192 @@ public String getFileName()
...
+                }
+
+                @Override
+                public void onResponse(Call call, Response response) throws IOException
+                {
+                    if (response.isSuccessful())
+                    {
+                        InputStream body = Requester.getBody(response);
+                        if (!future.complete(body))
+                            IOUtil.silentClose(response);
```
DV8FromTheWorld: I don't quite understand this code. Why _wouldn't_ this `.complete` call return `true`?
Why are we gating the response close behind whether the complete moved the future to completed?
What happens to the Response if we don't close it here?
What happens to the body InputStream if we close the Response before it is read completely?

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1114,113 +1106,192 @@ public String getFileName()
...
-         * @throws java.lang.IllegalStateException
-         *         If this is not an image attachment
-         *
-         * @return {@link net.dv8tion.jda.api.entities.Icon Icon} for this image attachment
-         *
-         * @since  3.4.0
+         * @return {@link java.util.concurrent.CompletableFuture} - Type: {@link java.io.InputStream}
          */
-        public Icon getAsIcon() throws IOException
+        public CompletableFuture<InputStream> retrieveData() // it is expected that the response is closed by the callback!
```
DV8FromTheWorld: Thoughts on changing this to `retrieveInputStream`?

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1114,113 +1106,192 @@ public String getFileName()
...
+         * <h2>Example</h2>
+         * <pre>{@code
+         * public void saveLocally(Message.Attachment attachment)
+         * {
+         *     attachment.downloadToFile()
+         *         .thenAccept(file -> System.out.println("Saved attachment to " + file.getName()))
+         *         .exceptionally(t ->
+         *     { // handle failure
+         *         t.printStackTrace();
+         *         return null;
```
DV8FromTheWorld: Indention of the example's `exceptionally` callback is bork.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1114,113 +1106,192 @@ public String getFileName()
...
+                        IOUtil.silentClose(response);
+                    }
+                }
+            });
+            return future;
         }
 
         /**
-         * Downloads this attachment to given File
+         * Downloads the attachment to a file with the {@link #getFileName()} into the current working directory.
```
DV8FromTheWorld:
```suggestion
        * Downloads the attachment into the current working directory using the file name provided by {@link #getFileName()}.
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1114,113 +1106,192 @@ public String getFileName()
...
          *
-         * @return InputStream copy of the response body for this Attachment
+         * @param  path
+         *         The path to save the file to
          *
-         * @since  3.4.0
+         * @return {@link java.util.concurrent.CompletableFuture} - Type: {@link java.io.File}
          */
-        public InputStream getInputStream() throws IOException
+        public CompletableFuture<File> downloadToFile(String path)
```
DV8FromTheWorld: We should have an overload of `downloadToFile(File file)`

src/main/java/net/dv8tion/jda/api/entities/Message.java
MinnDevelopment: `complete` returns `false` when the future was already completed (for example a user calling complete(null) on the future). If that is the case we know the user cannot have closed the stream since the future didn't actually trigger the callbacks/get methods which means the stream didn't even reach them. Since that is the case we close it ourself.

If we don't close the response we keep an open connection to the resource until the GC cleans up the response. `response.close()` just calls `response.body().byteStream().close()`. If we close the response before it was read the user wouldn't be able to use the data anymore.

src/main/java/net/dv8tion/jda/api/entities/Message.java
MinnDevelopment: This indentation is intentional, though

src/main/java/net/dv8tion/jda/api/entities/Message.java
MinnDevelopment: Should it replace this overload?

src/main/java/net/dv8tion/jda/api/entities/Message.java
MinnDevelopment: I'm indifferent about it

src/main/java/net/dv8tion/jda/api/entities/Message.java
DV8FromTheWorld: > (for example a user calling complete(null) on the future)

This was what I needed. I didn't realize the CompleteableFuture we were exposing gave the user the ability to complete it themselves.

src/main/java/net/dv8tion/jda/api/entities/Message.java
DV8FromTheWorld: Does it render differently than it appears in the jdoc?
Jdoc source has the callback's code inline with the method chain. It should be 1 level deeper.

src/main/java/net/dv8tion/jda/api/entities/Message.java
DV8FromTheWorld: No. I would have one feed into the other.

PR #938, opened 2019-02-20
Add ability to set speaking delay in audio manager

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This can be helpful for song transitions that take longer than 200ms, or buffered systems.
The delay can be specified in ms but will be aligned with the frame length (20ms) through integer division (no rest). It is set to a minimum of 10 frames = 200ms.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/audio/AudioConnection.java
+++ b/src/main/java/net/dv8tion/jda/internal/audio/AudioConnection.java
@@ -59,6 +59,7 @@ public class AudioConnection
                                                         // to Left and Right mono (stereo that is the same on both sides)
     public static final long MAX_UINT_32 = 4294967295L;
 
+    private static final int NOT_SPEAKING = 0;
     private static final byte[] silenceBytes = new byte[] {(byte)0xF8, (byte)0xFF, (byte)0xFE};
     private static boolean printedError = false;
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/internal/audio/AudioConnection.java
```diff
@@ -696,15 +700,18 @@ else if (silenceCounter > -1)
                     else
                         seq++;
 
-                    if (++silenceCounter > 10)
+                    silenceCounter++;
+                    if ((!sentSilenceOnConnect && silenceCounter > 10) || silenceCounter > speakingDelay)
                     {
+                        if (sentSilenceOnConnect)
+                            setSpeaking(0);
```
DV8FromTheWorld: Lets make this `0` a constant somewhere to better convey what it means.

src/main/java/net/dv8tion/jda/internal/audio/AudioConnection.java
```diff
@@ -696,15 +700,18 @@ else if (silenceCounter > -1)
                     else
                         seq++;
 
-                    if (++silenceCounter > 10)
+                    silenceCounter++;
+                    if ((!sentSilenceOnConnect && silenceCounter > 10) || silenceCounter > speakingDelay)
```
DV8FromTheWorld:
```suggestion
                    //If we have sent our 10 silent packets on initial connect, or if we have sent enough silent packets
                    // to satisfy the speaking delay, stop transmitting silence.
                    if ((!sentSilenceOnConnect && silenceCounter > 10) || silenceCounter > speakingDelay)
```

DV8FromTheWorld: Took me longer than I want to admit to understand this, thus I add the comment for clarity.

PR #941, opened 2019-02-22
Update README

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Update the README for the beta release of JDA version 4.


README.md
```diff
@@ -177,8 +212,85 @@ public static void main(String[] args) throws Exception
...
+An **Entity** is the term used to describe types such as **GuildChannel**/**Message**/**User** and other entities
+that Discord provides.
+Instances of these entities are created and deleted by JDA when Discord instructs it. This means
+the lifetime depends on signals provided by the Discord API which are used to create/update/delete entities.
+This is done through Gateway Events known as "dispatches" that are handled by the JDA WebSocket handlers.
+When Discord instructs JDA to delete entities they are simply removed from the JDA cache and lose their references.
+Once that happens nothing in JDA interacts or updates the instances of those entities and they become useless. Discord
+may instruct to delete these entities randomly for cache synchronization with the API.
+
+**It is not recommended to store _any_ of these entities for a longer period of time!**
```
DV8FromTheWorld: We should give guidance on what should be done instead of storing them.
And what storage actually means

e.g: It is fine to store in a local event-driven variable, but not in a long lived class variable or static variable

README.md
```diff
@@ -177,8 +212,85 @@ public static void main(String[] args) throws Exception
...
+
+### Fake Entities
+
+Some entities in JDA are marked through an interface called `IFakeable`. These entities can exist outside
+of the JDA cache and are inaccessible through the common `get...ById(id)` methods.
+Fake entities are essentially instances that are not directly referenced by the JDA cached and are only
```
DV8FromTheWorld:
```suggestion
Fake entities are essentially instances that are not directly referenced by the JDA cache and are only
```

README.md
```diff
@@ -177,8 +212,85 @@ public static void main(String[] args) throws Exception
...
+    
+    public UserLogger(User user)
+    {
+        this.userId = user.getIdLong();
+    }
+    
+    @Override
+    public void onMessageReceived(MessageReceivedEvent event)
+    {
+        if (event.getAuthor().getIdLong() == userId)
```
DV8FromTheWorld: Store the author in a local variable to show that doing so is okay.

MinnDevelopment: We actually do this below in `onGuildJoin` but I guess we can do it here as well.

README.md
DV8FromTheWorld: I've see enough people take the advice of "use ids instead of storing entities" and interpret it as "use id lookup literally every time you want to interact with an entity", which of course, isn't the case.

Your example below is a good example of how to properly lookup and store for local var execution.

PR #943, opened 2019-02-24
Add getUserByTag

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds shorthand to get user by tag.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDA.java
+++ b/src/main/java/net/dv8tion/jda/api/JDA.java
@@ -28,6 +28,7 @@ import net.dv8tion.jda.api.utils.cache.CacheView;
 import net.dv8tion.jda.api.utils.cache.SnowflakeCacheView;
 import net.dv8tion.jda.internal.requests.RestActionImpl;
 import net.dv8tion.jda.internal.requests.Route;
+import net.dv8tion.jda.internal.utils.Checks;
 import okhttp3.OkHttpClient;
 
 import javax.annotation.CheckReturnValue;
@@ -479,6 +480,17 @@ public interface JDA
     default User getUserById(long id)
     {
         return getUserCache().getElementById(id);
+    }
+
+    default User getUserByTag(String tag)
+    {
+        Checks.notNull(tag, "Tag");
+        Checks.check(User.USER_TAG.matcher(tag).matches(), "Invalid tag format!");
+        return getUserCache().applyStream(stream ->
+              stream.filter(it -> it.getAsTag().equals(tag))
+                    .findFirst()
+                    .orElse(null)
+        );
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/net/dv8tion/jda/api/JDA.java
```diff
@@ -481,6 +483,68 @@ default User getUserById(long id)
...
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided tag is null or not in the described format
+     *
+     * @return The {@link net.dv8tion.jda.api.entities.User} for the discord tag or null if no user has the provided tag
+     */
+    default User getUserByTag(String tag)
+    {
+        Checks.notNull(tag, "Tag");
+        Checks.check(User.USER_TAG.matcher(tag).matches(), "Invalid tag format!");
```
DV8FromTheWorld: Since we are running the tag through the Regex matcher, perhaps we should capture the groups instead of matching _and_ substringing.

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -408,6 +408,62 @@ default Member getMemberById(long userId)
...
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided arguments are null or not in the described format
+     *
+     * @return The {@link net.dv8tion.jda.api.entities.Member} for the discord tag or null if no member has the provided tag
+     */
+    default Member getMemberByTag(String username, String discriminator)
+    {
+        User user = getJDA().getUserByTag(username, discriminator);
+        return getMember(user);
```
DV8FromTheWorld: Same as above.

src/main/java/net/dv8tion/jda/api/JDA.java
DV8FromTheWorld: Since we are running the tag through the Regex matcher, perhaps we should capture the groups instead of matching _and_ substringing.

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -408,6 +408,62 @@ default Member getMemberById(long userId)
...
+     *         If the provided tag is null or not in the described format
+     *
+     * @return The {@link net.dv8tion.jda.api.entities.Member} for the discord tag or null if no member has the provided tag
+     *
+     * @see    net.dv8tion.jda.api.JDA#getUserByTag(String)
+     */
+    default Member getMemberByTag(String tag)
+    {
+        User user = getJDA().getUserByTag(tag);
+        return getMember(user);
```
DV8FromTheWorld: This will NPE if `getUserByTag` returns `null`

Also, we should probably add an `Check.notNull` in `getMember(user)`

PR #948, opened 2019-03-01
Add nullability annotations

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds `@Nonnull` and `@Nullable` annotations for explicit nullability. This is also very useful for kotlin which makes use of these annotations in its type system.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Activity.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Activity.java
@@ -19,6 +19,7 @@ import net.dv8tion.jda.annotations.Incubating;
 import net.dv8tion.jda.internal.entities.EntityBuilder;
 import net.dv8tion.jda.internal.utils.Checks;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.time.Instant;
 import java.time.temporal.TemporalUnit;
@@ -47,13 +48,15 @@ public interface Activity
      *
      * @return RichPresence or {@code null} if {@link #isRich()} returns {@code false}
      */
+    @Nullable
     RichPresence asRichPresence();
 
     /**
      * The displayed name of the {@link Activity Activity}. If no name has been set, this returns null.
      *
-     * @return Possibly-null String containing the Activity's name.
+     * @return String containing the Activity's name.
      */
+    @Nonnull
     String getName();
 
     /**
@@ -62,6 +65,7 @@ public interface Activity
      *
      * @return Possibly-null String containing the Activity's URL.
      */
+    @Nullable
     String getUrl();
 
     /**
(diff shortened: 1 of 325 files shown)


src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -80,6 +82,7 @@
      * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.Invite Invite}
      *         <br>The Invite object
      */
+    @Nonnull
     static RestAction<Invite> resolve(final JDA api, final String code, final boolean withCounts)
```
MinnDevelopment: Seems I forgot to add annotations to parameters here

MinnDevelopment:
We should consider changing how we handle conditional nullability in types such as `Message`. For instance:

```java
if (message.isFromType(ChannelType.TEXT)) {
    TextChannel channel = message.getTextChannel();
    channel.getManager().setTopic("text").queue();
}
```

In this code an IDE will warn about `channel` being possibly null, which should be impossible when checking for the type beforehand. This can be avoided by throwing an exception instead of returning null. Opinions?

src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -47,13 +48,15 @@
...
+    @Nullable
     RichPresence asRichPresence();
 
     /**
      * The displayed name of the {@link Activity Activity}. If no name has been set, this returns null.
      *
-     * @return Possibly-null String containing the Activity's name.
+     * @return String containing the Activity's name.
      */
+    @Nonnull
```
DV8FromTheWorld: This was originally `Possibly-null` but the `Notnull` annotation was add. Is that correct?

MinnDevelopment: The `EntityBuilder` uses `String.valueOf(json.get("name"))` so this can't be null.

DV8FromTheWorld:
> This can be avoided by throwing an exception instead of returning null. Opinions?

Throwing instead of returning `null` for things like `getTextChannel` when the message did not come from a `TextChannel` seems completely appropriate. I would not be against this change as it shows to the developer, immediately, what the problem is instead of receiving an NPE.

I wonder how many people have code like this though:
```
TextChannel tchan = message.getTextChannel();
if (tchan != null) {
  // ... do a thing 
}
```
Just curious.

I would go ahead with making the channel accessor methods on Message / Message Events throw.

DV8FromTheWorld:
Also, removing the proposal tag as this is good. Lets include this.

PR #950, opened 2019-03-02
Avoid losing entities skipped by forEachAsync iteration

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Previously it was possible that entities of the current chunk
would be lost when the iteration of forEachAsync ends.
This means repeated uses of forEachRemainingAsync could
result in skipping significant sections of the data set.

I have also removed silly attempts of trying to make the class thread-safe that didn't actually make the class thread-safe to begin with. Instead I have added a note about thread-safety to the class documentation.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/AuditLogPaginationActionImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/AuditLogPaginationActionImpl.java
@@ -158,4 +158,10 @@ public class AuditLogPaginationActionImpl
 
         request.onSuccess(list);
     }
+
+    @Override
+    protected long getKey(AuditLogEntry it)
+    {
+        return it.getIdLong();
+    }
 }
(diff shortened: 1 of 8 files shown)


src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/AuditLogPaginationActionImpl.java
```diff
@@ -158,4 +158,10 @@ protected void handleSuccess(Response response, Request<List<AuditLogEntry>> req
 
         request.onSuccess(list);
     }
+
+    @Override
+    protected long getKey(AuditLogEntry it)
```
DV8FromTheWorld: Will `getKey` ever be anything other than a Snowflake's ID?
Can we just build a built-in reliant on `Snowflake#getIdLong`?

PR #952, opened 2019-03-04
Add PermissionOverrideAction#clear 

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This was available in the `PermOverrideManager` but was overlooked in the transition.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/requests/restaction/PermissionOverrideAction.java
+++ b/src/main/java/net/dv8tion/jda/api/requests/restaction/PermissionOverrideAction.java
@@ -489,8 +489,11 @@ public interface PermissionOverrideAction extends AuditableRestAction<Permission
      */
     @Nonnull
     @CheckReturnValue
+    @SuppressWarnings("ConstantConditions")
     default PermissionOverrideAction setInherited(@Nonnull Collection<Permission> permissions)
     {
+        if (permissions == null || permissions.isEmpty())
+            return this;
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/requests/restaction/PermissionOverrideAction.java
```diff
@@ -457,6 +458,123 @@ default PermissionOverrideAction deny(Permission... permissions)
...
+     * @param  inheritedBits
+     *         The permissions to clear from the {@link net.dv8tion.jda.api.entities.PermissionOverride PermissionOverride}
+     *
+     * @return The current PermissionOverrideAction - for chaining convenience
+     */
+    @Nonnull
+    @CheckReturnValue
+    default PermissionOverrideAction clear(long inheritedBits)
+    {
+        return setInherited(getInherited() | inheritedBits);
```
DV8FromTheWorld: This, technically, is the same as `return setInherited(inheritedBits);` in functionality, isn't it?

I am confused on the difference in functionality between `setInherited` and `clear`

MinnDevelopment: They were basically just a redundancy, I removed them now since its more confusing than it is helpful

PR #954, opened 2019-03-08
Change limit for role names to 100

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Role names can be 100 characters long now.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/restaction/RoleActionImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/restaction/RoleActionImpl.java
@@ -70,7 +70,7 @@ public class RoleActionImpl extends AuditableRestActionImpl<Role> implements Rol
     public RoleActionImpl setName(String name)
     {
         Checks.check(name == null || name.length() > 0 && name.length() <= 100, "Name must be between 1-100 characters long");
-        this.name = name == null ? null : name;
+        this.name = name;
         return this;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/internal/requests/restaction/RoleActionImpl.java
```diff
@@ -69,7 +69,8 @@ public Guild getGuild()
     @CheckReturnValue
     public RoleActionImpl setName(String name)
     {
-        this.name = name;
+        Checks.check(name == null || name.isEmpty() || name.length() <= 100, "Name may not be longer than 100 characters");
```
DV8FromTheWorld: This is not the same as how the RoleManager handles the Checks for name. Is there a reason that we want to convert `""` to `null`?

Seems like that isn't great for a user experience.

MinnDevelopment: Creating roles without setting a name automatically sets the name to a default value, this doesn't work for the manager afaik

DV8FromTheWorld: Ah let me rephrase. We shouldn't let `""` abstract to `null` behind the scenes. If someone is calling the `setName` method, it should be expected that they are providing a valid name, or providing `null` to reset a previous setter.

If they want to opt-in to the "no name provided, thus default random name" functionality, they need to either not call the function, or call it with the value that represents "no name provided" (`null`).

We should surface if an erroneous name was provided to help developer debug.

PR #955, opened 2019-03-08
Add guild and channel context to InsufficientPermissionException

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This adds guild and channel for any permission exceptions. Can be useful for a catch-all permission handling system (like a command manager).


src/main/java/net/dv8tion/jda/api/exceptions/InsufficientPermissionException.java
```diff
@@ -16,17 +16,55 @@
...
+
+    public InsufficientPermissionException(Guild guild, GuildChannel channel, Permission permission)
     {
         super(permission, "Cannot perform action due to a lack of Permission. Missing permission: " + permission.toString());
+        this.guildId = guild.getIdLong();
+        this.channelId = channel == null ? 0 : channel.getIdLong();
     }
 
-    public InsufficientPermissionException(Permission permission, String reason)
+    public InsufficientPermissionException(Guild guild, GuildChannel channel, Permission permission, String reason)
```
DV8FromTheWorld: Permission should come first in these constructor as it is the more important bit.

DV8FromTheWorld: We should also separate the constructors out to add clarity to the calling code.
```
InsufficientPermissionException(Permission, Guild)
InsufficientPermissionException(Permission, Guild, String)
InsufficientPermissionException(Permission, GuildChannel)
InsufficientPermissionException(Permission, GuildChannel, String)
```

This will prevent us from having to pass in nulls or pass additional values when they can be retrieved via other passed in values.
Example: pre-retrieved guild off of text channel `text`
```
new InsufficientPermissionException(Permission, text.getGuild(), text)
```

DV8FromTheWorld:
As we are starting to move forward on the GuildController removal PR, lets get this one finished up

MinnDevelopment:
I want to merge this one after the controller.

MinnDevelopment:
This should be ready to go now

PR #970, opened 2019-03-24
Rename GuildMemberNickChangeEvent to GuildMemberUpdateNicknameEvent

averen:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR renames `GuildMemberNickChangeEvent` to `GuildMemberUpdateNicknameEvent` and some related methods. I [discussed this change](https://discordapp.com/channels/125227483518861312/541288051243089920/543471466361126922) in the JDA discord about a two months ago and forgot to create a PR.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/events/guild/member/GuildMemberUpdateNicknameEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/guild/member/GuildMemberUpdateNicknameEvent.java
@@ -24,9 +24,13 @@ import net.dv8tion.jda.api.events.UpdateEvent;
  * Indicates that a {@link net.dv8tion.jda.api.entities.Member Member} updated their {@link net.dv8tion.jda.api.entities.Guild Guild} nickname.
  *
  * <p>Can be used to retrieve members who change their nickname, triggering guild, the old nick and the new nick.
+ *
+ * <p>Identifier: {@code nick}
  */
 public class GuildMemberUpdateNicknameEvent extends GenericGuildMemberEvent implements UpdateEvent<Member, String>
 {
+    public static final String IDENTIFIER = "nick";
+
     private final String oldNick, newNick;
 
     public GuildMemberUpdateNicknameEvent(JDA api, long responseNumber, Member member, String oldNick, String newNick)
@@ -59,7 +63,7 @@ public class GuildMemberUpdateNicknameEvent extends GenericGuildMemberEvent impl
     @Override
     public String getPropertyIdentifier()
     {
-        return "nick";
+        return IDENTIFIER;
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


MinnDevelopment:
The naming convention is `EntityUpdatePropertyEvent`

src/main/java/net/dv8tion/jda/api/events/guild/member/GuildMemberUpdateNicknameEvent.java
```diff
@@ -24,14 +24,14 @@
  *
  * <p>Can be used to retrieve members who change their nickname, triggering guild, the old nick and the new nick.
  */
-public class GuildMemberNickChangeEvent extends GenericGuildMemberEvent
+public class GuildMemberUpdateNicknameEvent extends GenericGuildMemberEvent
```
MinnDevelopment: This should implement `UpdateEvent` as well then

averen: Should `getPropertyIdentifier()` return `nick` or `nickname`?

DV8FromTheWorld: Following the api docs, it should be `nick`

src/main/java/net/dv8tion/jda/api/events/guild/member/GuildMemberUpdateNicknameEvent.java
```diff
@@ -40,18 +41,48 @@ public GuildMemberNickChangeEvent(JDA api, long responseNumber, Member member, S
...
-    public String getNewNick()
+    public String getNewNickname()
+    {
+        return getNewValue();
+    }
+
+    @Override
+    public String getPropertyIdentifier()
+    {
+        return "nick";
```
MinnDevelopment: This should be a constant in this class, compare to other classes that implement the `UpdateEvent` interface. Additionally it should be in the class-level docs.

DV8FromTheWorld:
@MinnDevelopment do you think that we should go ahead and add the `GenericGuildMemberUpdateEvent` hierarchy in preparation for there possibly being other member update events?

There haven't been thus far, but to do so later would be a breaking change.

MinnDevelopment:
@DV8FromTheWorld yes that sounds reasonable, however we don't need to do that in this PR.

DV8FromTheWorld:
Agreed. I'll put an issue into the backlog so we handle before v4 alpha is done.

averen:
Just noticed I forgot to update the method name.

Issue #967, opened 2019-03-22
Excluding users from CombinedAudio


BalintGergely:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [X] I have checked for similar issues.
- [ ] I have updated to the [latest JDA version][download].
- [ ] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

### Issue Type

- [ ] Bug Report
- [X] Feature Request

### Description

The equivalent of muting a user client-side. Basically a setting to exclude specific users from the `CombinedAudio` received in `AudioReceiveHandler.handleCombinedAudio(CombinedAudio)`.

Additionally, a setting to automatically exclude other bot users would be neat.

BalintGergely:
An alternative to this would be setting up a user filter each time JDA combines audio data.

DV8FromTheWorld:
Probably the simplest implementation of this would be modifying `AudioReceiveHandler` interface to have a `getExcludedUsers` should return `List<User>` which would be used to determine whether `handleCombinedAudio` and `handleUserAudio` received audio from a user when they user spoke.

BalintGergely:
I suggest giving `AudioReceiveHandler` it's own filter method like the following:

`boolean filter(User user)`

This would make it easier for example to automatically filter out bot users instead of having to add them to a list whenever they join the `VoiceChannel`.

PR #971, opened 2019-03-24
AudioReceiveHandler#includeUserInCombinedAudio

DV8FromTheWorld:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #967

## Description
Adds support for filtering what users are included in in the packet created for `handleCombinedAudio(CombinedAudio)`

Example: Only include audio from users that are not bots.
```
public MyHandler implements AudioReceiveHandler {
  canReceiveCombined() {
    return true;
  }

  handleCombinedAudio(CombinedAudio audio) {
    //Handle audio
  }

  includeUserInCombinedAudio(User user) {
    return !user.isBot();
  }
}
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
+++ b/src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
@@ -16,6 +16,8 @@
 
 package net.dv8tion.jda.api.audio;
 
+import net.dv8tion.jda.api.entities.User;
+
 import javax.sound.sampled.AudioFormat;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
```diff
@@ -86,4 +88,26 @@
...
     void handleUserAudio(UserAudio userAudio);
+
+    /**
+     * This method is a filter predicate used by JDA to determine whether or not to include a
+     * {@link net.dv8tion.jda.api.entities.User User}'s audio when creating a CombinedAudio packet.
+     * <p>
+     * This method is especially useful in creating whitelist / blacklist functionality for receiving audio.
+     * <p>
+     * A few possible examples:
+     *  <li>Have this method always return false for Users that are bots.</li>
```
MinnDevelopment: Isn't this missing a `<ul>` somewhere?

src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
```diff
@@ -86,4 +88,26 @@
...
+     *  <li>Have this method return false for users who have been placed on a blacklist for abusing the bot's functionality.</li>
+     *  <li>Have this method only return true if the user is in a special whitelist of power users.</li>
+     *
+     * @param  user
+     *         The user whose audio was received
+     *
+     * @return If true, JDA will include the user's audio when merging audio sources when created packets
+     *         for {@link #handleCombinedAudio(CombinedAudio)}
+     */
+    default boolean includeUserInCombinedAudio(User user)
```
MinnDevelopment: This should have a `@Nonnull` on the user parameter

src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
DV8FromTheWorld: Oh. Yeah, you're right.
I guess the javadoc renderer just understood the lack of `<ul>` when I previewed it.

MinnDevelopment:
The name of this method seems pretty obnoxious to me, can't we come up with something more slim?

DV8FromTheWorld:
I am open to suggestions as the current name is straight forward in its usage

DV8FromTheWorld:
Is there a counter suggestion for method naming? If not, we should progress this PR

Issue #968, opened 2019-03-24
Proposal: MessageChannel#getHistoryFromBeginning


DV8FromTheWorld:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

## Issue
For new users, it would be useful to codify and expose the fact that `0L` (and any other low integer values) can be used to retrieve message history starting from the beginning of a channel.

### Issue Type

- [ ] Bug Report
- [x] Feature Request

### Description
In the `MessageChannel` interface, add a default method called `getHistoryFromBeginning` that calls and returns `getHistoryAfter(0L)`

Name of method is up for debate. Few possibilities:
- `getHistoryFromBeginning`
- `getHistoryFromStart`
- `getHistoryFromFirst`
- `getHistoryBeginning`
- `getHistoryStart`

I believe the first is the better naming, but I'm open to opinions.

BenjaminUrquhart:
More ideas:
getAllMessages
getAllHistory
 This does break the "getHistoryX" pattern but it makes more sense to me.

DV8FromTheWorld:
Problem with those namings is that it invokes an idea that you will get all messages after single execution.
We have chosen the naming methodology that we have now because it invokes a feeling of where you are to start in history, not a full retrieval in one execution.

PR #972, opened 2019-03-25
Add method to easily get the beginning history of a MessageChannel

DManstrator:
## Pull Request Etiquette

- [x] I have read the [contributing guidelines](https://github.com/DV8FromTheWorld/JDA/wiki/5%29-Contributing).

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____

Closes Issue: #968

## Description

Adds a new method `getHistoryFromBeginning(int limit)` to the MessageChannel Interface that simply returns `getHistoryAfter(0L, limit)`. As @DV8FromTheWorld explained in the Issue mentioned above, most people don't know that this is the way to get the beginning history of a MessageChannel so I added it to make it easier for Users.

As I did that, I also made some minor fixes to the JavaDoc of some other functions in the Interface when I was at it (I spoke to @MinnDevelopment before doing so).

(It's my first real PR ever, I hope it's fine. I was spending more time in reading your guidelines and setting up this PR than writing the Code and the Documentation for it. If there is anything wrong, just tell me, I'll try to fix it asap.)

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
@@ -971,7 +971,7 @@ public interface MessageChannel extends ISnowflake, Formattable
     }
 
     /**
-     * Creates a new {@link MessageHistory MessageHistory} object for each call of this method.
+     * Creates a new {@link net.dv8tion.jda.api.entities.MessageHistory MessageHistory} object for each call of this method.
      * <br>MessageHistory is <b>NOT</b> an internal message cache, but rather it queries the Discord servers for previously sent messages.
      *
      * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
@@ -1031,7 +1031,7 @@ public interface MessageChannel extends ISnowflake, Formattable
 
     /**
      * Uses the provided {@code id} of a message as a marker and retrieves messages sent around
-     * the marker. The {@code limit} determines the amount of message retrieved near the marker. Discord will
+     * the marker. The {@code limit} determines the amount of messages retrieved near the marker. Discord will
      * attempt to evenly split the limit between before and after the marker, however in the case that the marker is set
      * near the beginning or near the end of the channel's history the amount of messages on each side of the marker may
      * be different, and their total count may not equal the provided {@code limit}.
@@ -1069,7 +1069,7 @@ public interface MessageChannel extends ISnowflake, Formattable
      * @param messageId
      *        The id of the message that will act as a marker.
      * @param limit
-     *        The amount of message to be retrieved around the marker. Minimum: 1, Max: 100.
+     *        The amount of messages to be retrieved around the marker. Minimum: 1, Max: 100.
      *
      * @throws java.lang.IllegalArgumentException
      *         <ul>
@@ -1084,7 +1084,7 @@ public interface MessageChannel extends ISnowflake, Formattable
      *         </ul>
      *
      * @return {@link net.dv8tion.jda.api.entities.MessageHistory.MessageRetrieveAction MessageHistory.MessageRetrieveAction}
-     *         <br>Provides a MessageHistory object with message around the provided message loaded into it.
+     *         <br>Provides a {@link net.dv8tion.jda.api.entities.MessageHistory MessageHistory} object with messages around the provided message loaded into it.
      *
      * @see    net.dv8tion.jda.api.entities.MessageHistory#getHistoryAround(MessageChannel, String) MessageHistory.getHistoryAround(MessageChannel, String)
      */
@@ -1096,7 +1096,7 @@ public interface MessageChannel extends ISnowflake, Formattable
 
     /**
      * Uses the provided {@code id} of a message as a marker and retrieves messages around
-     * the marker. The {@code limit} determines the amount of message retrieved near the marker. Discord will
+     * the marker. The {@code limit} determines the amount of messages retrieved near the marker. Discord will
      * attempt to evenly split the limit between before and after the marker, however in the case that the marker is set
      * near the beginning or near the end of the channel's history the amount of messages on each side of the marker may
      * be different, and their total count may not equal the provided {@code limit}.
@@ -1134,7 +1134,7 @@ public interface MessageChannel extends ISnowflake, Formattable
      * @param messageId
      *        The id of the message that will act as a marker. The id must refer to a message from this MessageChannel.
      * @param limit
-     *        The amount of message to be retrieved around the marker. Minimum: 1, Max: 100.
+     *        The amount of messages to be retrieved around the marker. Minimum: 1, Max: 100.
      *
      * @throws java.lang.IllegalArgumentException
      *         <ul>
@@ -1149,7 +1149,7 @@ public interface MessageChannel extends ISnowflake, Formattable
      *         </ul>
      *
      * @return {@link net.dv8tion.jda.api.entities.MessageHistory.MessageRetrieveAction MessageHistory.MessageRetrieveAction}
-     *         <br>Provides a MessageHistory object with message around the provided message loaded into it.
+     *         <br>Provides a {@link net.dv8tion.jda.api.entities.MessageHistory MessageHistory} object with messages around the provided message loaded into it.
      *
      * @see    net.dv8tion.jda.api.entities.MessageHistory#getHistoryAround(MessageChannel, String) MessageHistory.getHistoryAround(MessageChannel, String)
      */
@@ -1161,7 +1161,7 @@ public interface MessageChannel extends ISnowflake, Formattable
 
     /**
      * Uses the provided {@link net.dv8tion.jda.api.entities.Message Message} as a marker and retrieves messages around
-     * the marker. The {@code limit} determines the amount of message retrieved near the marker. Discord will
+     * the marker. The {@code limit} determines the amount of messages retrieved near the marker. Discord will
      * attempt to evenly split the limit between before and after the marker, however in the case that the marker is set
      * near the beginning or near the end of the channel's history the amount of messages on each side of the marker may
      * be different, and their total count may not equal the provided {@code limit}.
@@ -1199,7 +1199,7 @@ public interface MessageChannel extends ISnowflake, Formattable
      *        The {@link net.dv8tion.jda.api.entities.Message Message} that will act as a marker. The provided Message
      *        must be from this MessageChannel.
      * @param limit
-     *        The amount of message to be retrieved around the marker. Minimum: 1, Max: 100.
+     *        The amount of messages to be retrieved around the marker. Minimum: 1, Max: 100.
      *
      * @throws java.lang.IllegalArgumentException
      *         <ul>
@@ -1214,7 +1214,7 @@ public interface MessageChannel extends ISnowflake, Formattable
      *         </ul>
      *
      * @return {@link net.dv8tion.jda.api.entities.MessageHistory.MessageRetrieveAction MessageHistory.MessageRetrieveAction}
-     *         <br>Provides a MessageHistory object with message around the provided message loaded into it.
+     *         <br>Provides a {@link net.dv8tion.jda.api.entities.MessageHistory MessageHistory} object with messages around the provided message loaded into it.
      *
      * @see    net.dv8tion.jda.api.entities.MessageHistory#getHistoryAround(MessageChannel, String) MessageHistory.getHistoryAround(MessageChannel, String)
      */
@@ -1227,7 +1227,7 @@ public interface MessageChannel extends ISnowflake, Formattable
 
     /**
      * Uses the provided {@code id} of a message as a marker and retrieves messages sent after
-     * the marker ID. The {@code limit} determines the amount of message retrieved near the marker.
+     * the marker ID. The {@code limit} determines the amount of messages retrieved near the marker.
      *
      * <p><b>Examples:</b>
      * <br>Retrieve 100 messages from the middle of history. {@literal >}100 message exist in history and the marker is {@literal >}50 messages
@@ -1257,7 +1257,7 @@ public interface MessageChannel extends ISnowflake, Formattable
      * @param messageId
      *        The id of the message that will act as a marker.
      * @param limit
-     *        The amount of message to be retrieved after the marker. Minimum: 1, Max: 100.
+     *        The amount of messages to be retrieved after the marker. Minimum: 1, Max: 100.
      *
      * @throws java.lang.IllegalArgumentException
      *         <ul>
@@ -1272,7 +1272,7 @@ public interface MessageChannel extends ISnowflake, Formattable
      *         </ul>
      *
      * @return {@link net.dv8tion.jda.api.entities.MessageHistory.MessageRetrieveAction MessageHistory.MessageRetrieveAction}
-     *         <br>Provides a MessageHistory object with message after the provided message loaded into it.
+     *         <br>Provides a {@link net.dv8tion.jda.api.entities.MessageHistory MessageHistory} object with messages after the provided message loaded into it.
      *
      * @see    net.dv8tion.jda.api.entities.MessageHistory#getHistoryAfter(MessageChannel, String) MessageHistory.getHistoryAfter(MessageChannel, String)
      */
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -1565,7 +1565,68 @@ default MessagePaginationAction getIterableHistory()
...
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#MISSING_PERMISSIONS MISSING_PERMISSIONS}
+     *     <br>The request was attempted after the account lost
+     *         {@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY} in the
+     *         {@link net.dv8tion.jda.api.entities.TextChannel TextChannel}.</li>
+     *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_CHANNEL UNKNOWN_CHANNEL}
+     *     <br>The request was attempted after the channel was deleted.</li>
+     * </ul>
+     *
+     * @param limit
```
DV8FromTheWorld:
```suggestion
     * @param  limit
```

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -1565,7 +1565,68 @@ default MessagePaginationAction getIterableHistory()
...
+     *        The amount of messages to be retrieved. Minimum: 1, Max: 100.
```
DV8FromTheWorld:
```suggestion
     *         The amount of messages to be retrieved. Minimum: 1, Max: 100.
```

DV8FromTheWorld:
@MinnDevelopment Didn't we discuss removing the `.limit(int)` calls? I don't remember the context though.

MinnDevelopment:
@DV8FromTheWorld I don't remember wanting to remove them. We talked about removing `MessageReaction#getUsers(limit)` which happened already.

DV8FromTheWorld:
Ah, yep, that was what I was thinking about. Thanks for the reminder.

DManstrator:
> Ah, yep, that was what I was thinking about. Thanks for the reminder.

You're welcome. And thanks for letting me PR it. :D 
If there are more things needed to be changed / adjusted, just let me know.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -1565,7 +1565,68 @@ default MessagePaginationAction getIterableHistory()
         Checks.notNull(message, "Message");
         return getHistoryBefore(message.getId(), limit);
     }
-
+    
+    /**
+     * Retrieves messages from the beginning of this MessageChannel.
+     * The {@code limit} determines the amount of messages being retrieved.
+     * 
+     * <p><b>Example:</b>
```
MinnDevelopment: You can use a header tag here like `<h2>Example</h2>`

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -1565,7 +1565,68 @@ default MessagePaginationAction getIterableHistory()
...
+     *         Provided {@code limit} is less than {@code 1} or greater than {@code 100}.
+     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
+     *         If this is a {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} and the logged in account does not have
+     *         <ul>
+     *             <li>{@link net.dv8tion.jda.api.Permission#MESSAGE_READ Permission.MESSAGE_READ}</li>
+     *             <li>{@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY}</li>
+     *         </ul>
+     *
+     * @return {@link net.dv8tion.jda.api.entities.MessageHistory.MessageRetrieveAction MessageHistory.MessageRetrieveAction}
+     *         <br>Provides a {@link net.dv8tion.jda.api.entities.MessageHistory MessageHistory} object with messages - beginning from the start - loaded into it.
```
MinnDevelopment: "beginning from the start" sounds redundant to me

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -1565,7 +1565,68 @@ default MessagePaginationAction getIterableHistory()
...
+     *
+     * @see    net.dv8tion.jda.api.entities.MessageChannel#getHistoryAfter(long, String) MessageChannel.getHistoryAfter(long, String)
+     * @see    net.dv8tion.jda.api.entities.MessageHistory#getHistoryAfter(MessageChannel, String) MessageHistory.getHistoryAfter(MessageChannel, String)
```
MinnDevelopment: Maybe add a `@see MessageHistory#retrieveFuture` since to progress in this instance you advance to more recent messages using that method

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -1565,7 +1565,68 @@ default MessagePaginationAction getIterableHistory()
...
+     */
+    @CheckReturnValue
+    default MessageHistory.MessageRetrieveAction getHistoryFromBeginning(int limit)
```
MinnDevelopment: There should be a `@Nonnull` annotation from `javax.annotations` here as well for the return value. This is a pattern we are starting to deploy in another PR and it should already be done here as well. (only for the new method to avoid conflicts)

MinnDevelopment:
A minor detail is missing here which is the fact all the other `getHistory*` methods are also factory methods in [MessageHistory](https://github.com/DV8FromTheWorld/JDA/blob/8ba7e37c4fb442edcdc14ed57edc66426940e62d/src/main/java/net/dv8tion/jda/api/entities/MessageHistory.java#L357).

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: Most methods also have it like that so I stuck with that.
I will replace all of those occurrences with a proper header tag if that's okay for you.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: Yes, true. But all the other methods also explain in the return tag which kind of Messages will be in the history and I tried to include that the first messages will be in it.

```suggestion
     *         <br>Provides a {@link net.dv8tion.jda.api.entities.MessageHistory MessageHistory} object with the first messages of the channel loaded into it.
```
Is that better?

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: Not sure if I understand you correctly, do you mean that:
```suggestion
    @CheckReturnValue
    @NonNull
    default MessageHistory.MessageRetrieveAction getHistoryFromBeginning(int limit)
```
?

Never worked with the NonNull annotation, I just know that attributes / parameters sometimes have them.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: Will do that. :+1:

DManstrator:
Can you explain that a little but more? I don't see what I have to change since I'm not really familiar with that pattern.

What I understood: I should also create that getHistoryFromBeginng-method in the MessageHistory Class where I redirect to the function you have linked (instead of having the same code again)?

And instead of returning
```JAVA
return getHistoryAfter(0L, limit);
```
I return it as
```JAVA
return MessageHistory.getHistoryFromBeginning(this).limit(limit);
```
?

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
MinnDevelopment: But the phrase "beginning from the start" doesn't really clarify anything. Its simply redundant and confusing to read, imo

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
MinnDevelopment: Sort them by length for better readability.

```java
@Nonnull
@CheckReturnValue
default MessageHistory.MessageRetrieveAction getHistoryFromBeginning(int limit)
{
....
```

MinnDevelopment:
@DManstrator yes, the logic for all of these methods is located in the `MessageHistory` class. Having a central place for it is crucial for maintainability.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
MinnDevelopment: It doesn't need to be changed everywhere, the resulting javadoc is the same. However, only old docs should have this style while the new ones use headers instead.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DV8FromTheWorld: Note: It isn't `@NotNull`. (which is from Jetbrains). It is `@Nonnull` which is from `javax.annotations`

DManstrator: @MinnDevelopment You're right, did it like you suggested it.
@DV8FromTheWorld Thanks, confused those two.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: My problem is that `Provides a MessageHistory object with messages loaded into it.` says nothing at all. Without the method name I would not know which messages I'll get.

My suggestion doesn't have this phrase in it anymore, is that better? If not, can you suggest me something because I don't know something better.

DManstrator:
Having the code in the MessageHistory class is way better since calling this is even better / easier to call in my opinion. And yeah, you're right with that maintainability argument, I really forgot about that class. Thanks for noticing my mistake!

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: @MinnDevelopment I still don't know if I shouldn't remove those other two @see tags since they are not relevant for this function in my opinion (the new function only takes a limit, not a Message Id). I even made a mistake with `getHistoryAfter(long, String)` since this function doesn't exist at all. What do you think about it?

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
MinnDevelopment: I would just say "the oldest messages in this channel" but first would be fine as well. Maybe add another sentence: "The messages are ordered from the most recent to oldest" for clarity

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
MinnDevelopment: The see tag is relevant in the sense that it is used by this method.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: > Maybe add another sentence: "The messages are ordered from the most recent to oldest" for clarity

Isn't it the other way round? The first element is the oldest message for sure, I tested that.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: Okay, then I'll throw out the wrong method and "replace" by your suggestion. I will also have a check the readability.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
MinnDevelopment: Nope, it definitely should be ordered most recent to oldest. I also tested this.

DManstrator: You're totally right. Back then when testing it, I just retrieved one message, that's why I thought `get(0)` returns the oldest message.. :man_facepalming:

src/main/java/net/dv8tion/jda/api/entities/MessageHistory.java
```diff
@@ -447,6 +447,36 @@ public static MessageRetrieveAction getHistoryAround(MessageChannel channel, Str
...
+     *         Or if the provided messageId contains whitespace
+     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
+     *         If this is a TextChannel and the currently logged in account does not
+     *         have the permission {@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY}
+     *
+     * @return {@link net.dv8tion.jda.api.entities.MessageHistory.MessageRetrieveAction MessageRetrieveAction}
+     *
+     * @see    net.dv8tion.jda.api.entities.MessageChannel#getHistoryFromBeginning(int)  MessageChannel.getHistoryFromBeginning(int)
+     */
+    public static MessageRetrieveAction getHistoryFromBeginning(MessageChannel channel)
```
MinnDevelopment: This should also have the `@Nonnull` and `@CheckReturnValue` as method annotations and a `@Nonnull` for the parameter.

src/main/java/net/dv8tion/jda/api/entities/MessageHistory.java
```diff
@@ -447,6 +447,36 @@ public static MessageRetrieveAction getHistoryAround(MessageChannel channel, Str
...
+     * <br>{@code MessageHistory history = MessageHistory.getHistoryFromBeginning(channel).limit(60).complete()}
+     * <br>Will return a MessageHistory instance with the first 60 messages of the given {@link net.dv8tion.jda.api.entities.MessageChannel MessageChannel}.
+
+     *
+     * @param  channel
+     *         The {@link net.dv8tion.jda.api.entities.MessageChannel MessageChannel}
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If any of the provided arguments is {@code null};
+     *         Or if the provided messageId contains whitespace
```
MinnDevelopment: There is no messageId here

DManstrator: Even with the `@Nonnull` Annotation, `null` should still be passable, shouldn't it? So only the line with the Message Id needs to be removed (and the first one needs to be adjusted)? I copied a similar JavaDoc for this function, thought I got rid of every wrong thing.. Thanks for noticing it and sorry.

MinnDevelopment: Yes you can still pass null, java doesn't have a typesystem that would allow denying this. The exception is thrown by `Checks.nonNull` when null is passed. Since there is no messageId parameter this part is unnecessary.

DManstrator:
I apologize that this little fix took me so long, it has be a rough time for me lately.. I hope it's okay now. If I messed something up, let me know it.

src/main/java/net/dv8tion/jda/api/entities/MessageHistory.java
```diff
@@ -447,6 +449,37 @@ public static MessageRetrieveAction getHistoryAround(MessageChannel channel, Str
...
+     *
+     * @return {@link net.dv8tion.jda.api.entities.MessageHistory.MessageRetrieveAction MessageRetrieveAction}
+     *
+     * @see    net.dv8tion.jda.api.entities.MessageChannel#getHistoryFromBeginning(int)  MessageChannel.getHistoryFromBeginning(int)
+     */
+    @Nonnull
+    @CheckReturnValue
+    public static MessageRetrieveAction getHistoryFromBeginning(@Nonnull MessageChannel channel)
+    {
+        return getHistoryAfter(channel, "0L");
```
MinnDevelopment: The postfix `L` will probably cause the request to fail because its an invalid id. The `L` is only used on literals to indicate whether it is a long or integer.

DManstrator: Before I was passing `0L` but since the function takes a String, I just put quote marks around it..

I know how `L`, `D` and `F` are used normally but I can't believe I really did that...

Issue #965, opened 2019-03-21
Invite.Group#getUsers() returns stringified json objects


mdashlw:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features/bug fixes.

### Issue Type

- [x] Bug Report
- [ ] Feature Request

### Description

`Invite.Group#getUsers()` returns a list of stringified json objects with following schema:

```json
{
    "username": "actual_users_name"
}
```

MinnDevelopment:
Fixed in [3.8.2_464](https://github.com/DV8FromTheWorld/JDA/commit/9ae990533fbf12e943fead126b339d44b36f51a2)

PR #974, opened 2019-03-31
Invite.Group#getUsers now returns usernames

BenjaminUrquhart:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #965 

## Description

EntityBuilder#createInvite now parses the username JSON instead of passing the list of JSON strings to the constructor of Invite.Group.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
@@ -1157,7 +1157,7 @@ public class EntityBuilder
             if (usernameArray == null)
                 usernames = null;
             else
-                usernames = Collections.unmodifiableList(StreamSupport.stream(usernameArray.spliterator(), false).map(String::valueOf).collect(Collectors.toList()));
+                usernames = Collections.unmodifiableList(StreamSupport.stream(usernameArray.spliterator(), false).map(json -> new JSONObject(String.valueOf(json)).getString("username")).collect(Collectors.toList()));
 
             group = new InviteImpl.GroupImpl(groupIconId, groupName, groupId, usernames);
         }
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1157,7 +1157,7 @@ public Invite createInvite(JSONObject object)
             if (usernameArray == null)
                 usernames = null;
             else
-                usernames = Collections.unmodifiableList(StreamSupport.stream(usernameArray.spliterator(), false).map(String::valueOf).collect(Collectors.toList()));
+                usernames = Collections.unmodifiableList(StreamSupport.stream(usernameArray.spliterator(), false).map(json -> new JSONObject(String.valueOf(json)).getString("username")).collect(Collectors.toList()));
```
MinnDevelopment:
```suggestion
                usernames = map(channelObject, "recipients", (json) -> json.getString("username"));
```

kantenkugel:
So i assume that the recipient array is not in the form of `["Username", "Username", ...]` but of form `[{"username": "Username", ...}, ...]`?  If so, the current mapping could be heavily improved by just casting to `JSONObject` or looping the `JSONArray` directly via `JSONArray#getJSONObject(int index)`

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
kantenkugel: yup, and the `usernameArray` variable can be removed, together with the `if` (the `map` function already checks for existence)

kantenkugel:
@MinnDevelopment i would still remove the variable and the `if`, or at least change if to `JSONObject#hasKey()`

MinnDevelopment:
@kantenkugel I believe @Almighty-Alpaca made this intentionally null when the key is also null/not present. The `map` method returns an empty collection instead of null.

kantenkugel:
Even if so: we don't need a variable here anymore... just use hasKey() instead of variable + != null check

Issue #580, opened 2017-12-21
JDALogger does not correctly detect presence of a SLF4J Implementation provider, starting from SLF4J 1.8


Enveed:
### Issue Type

- [X] Bug Report
- [ ] Feature Request

### Description

Starting from SLF4J version 1.8, the old `org.slf4j.impl.StaticLoggerBinder` method of providing an implementation for SLF4J is no longer working, as detailed here: https://www.slf4j.org/codes.html#noProviders

The new way to provide an implementation is via `SLF4JServiceProvider` and a [*service-provider configuration* file](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html).

This would not be a problem if JDA would be entirely based on the exposed side of SLF4J's API, but [`JDALogger`](https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/core/utils/JDALogger.java) does an additional check for the presence of the old `StaticLoggerBinder`, [here](https://github.com/DV8FromTheWorld/JDA/blob/e8374137764a09733b1be9f56a7654af85788edd/src/main/java/net/dv8tion/jda/core/utils/JDALogger.java#L46) and uses a `SimpleLogger` instead of the SLF4J provided Logger if the class is not found:

```Java
public static final boolean SLF4J_ENABLED;
    static
    {
        boolean tmp = false;
        try
        {
            Class.forName("org.slf4j.impl.StaticLoggerBinder");
            tmp = true;
        }
        catch (ClassNotFoundException e)
        {
            //prints warning of missing implementation
            LoggerFactory.getLogger(JDALogger.class);
        }
        SLF4J_ENABLED = tmp;
    }
```

This obviously fails starting from 1.8 since implementation providers do not include this class anymore, but instead a `SLF4JServiceProvider`.

This logic has to be fixed to work with SLF4J 1.8+.

I know JDA is currently using SLF4J [`1.7.25`](https://github.com/DV8FromTheWorld/JDA/blob/e8374137764a09733b1be9f56a7654af85788edd/build.gradle#L56), however, this is probably important for the future.

### Temporary solution

This may be useful for other people who want to use SLF4J 1.8 instead of the version provided by JDA in their project for some reason, until this is fixed. Note: **This is VERY EVIL, NOT recommended  and in no way guaranteed to work in your situation!** (It does for me tho)

```Java
    /**
[...]

PR #975, opened 2019-04-02
Modified JDALogger SLF4J detection to be able to detect both pre-1.8.x and 1.8.x+ providers

enveeed:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #580 

## Description

This pull request replaces the detection system in `JDALogger`, which currently is only able to detect an SLF4J provider via the SLF4J `1.7.x` `StaticLoggerBinder` mechanism with one which is able to detect both providers with the current mechanism as well as new `ServiceLoader`-based providers from SLF4J `1.8.x`.

This enables JDA to function correctly with any newer SLF4J API and provider version which may be present at runtime, even if the version it is compiled against is `1.7.x`. (`slf4j-api` has [always been binary compatible](https://www.slf4j.org/faq.html#compatibility) to previous versions, also see [this issue](https://github.com/DV8FromTheWorld/JDA/issues/580#issuecomment-479213460) for more information)


MinnDevelopment:
This change should be done on the v4 branch instead.

MinnDevelopment:
To make your life easier, you can fix your PR this way:

```sh
git reset @~2 --soft
git stash
git checkout -b patch-slf4j upstream/v4
git stash pop
git commit -am "..."
git push -f origin patch-slf4j:patch-580
```

Or alternatively just do the same changes again on the other branch and open a new PR. I'm fine with either as long as it goes to v4.

enveeed:
Ah thanks, figured I corrupted it somehow. I'll just make a new one for v4 then.

PR #977, opened 2019-04-04
Added EmbedBuilder#setFooter for convenience

Raepheles:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

I only added EmbedBuilder#setFooter without the icon url parameter for convenient use so that we wouldn't have to pass a null object when we don't want to use a footer icon.

### DIFF
--- a/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/core/EmbedBuilder.java
@@ -675,6 +675,22 @@ public class EmbedBuilder
     }
     
     /**
+     * Sets the Footer of the embed without icon.
+     * @param  text
+     *         the text of the footer of the embed. If this is not set, the footer will not appear in the embed.
+     * @throws java.lang.IllegalArgumentException
+     *         <ul>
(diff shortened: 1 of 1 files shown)


jagrosh:
I'm not sure that this is necessary (albeit I don't have a strong opinion either way). However, your implementation should be no more complicated than something like this:
```java
public EmbedBuilder setFooter(String text)
{
    return setFooter(text, null);
}
```

Raepheles:
Neccessity is subjective. For me this is needed qol addition. There is really no point at passing null object to a method. Also it's confusing. Fixed the code btw.

MinnDevelopment:
In any case, feature additions should only go to the v4 branch.

Raepheles:
Isn't v4 complete rewrite? People still use v3 and development branch is for v3. Besides this is not exactly new feature it's just a simple qol addition and there is no harm in merging it.

MinnDevelopment:
No it is not a rewrite it just does a lot of breaking changes, this can easily be added there. v3 only receives important bug fixes now.

MinnDevelopment:
Closed in favor of #980

PR #979, opened 2019-04-05
Modified JDALogger SLF4J detection to be able to detect both pre-1.8.x and 1.8.x+ providers

enveeed:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #580 

## Description

This pull request replaces the detection system in `JDALogger`, which currently is only able to detect an SLF4J provider via the SLF4J `1.7.x` `StaticLoggerBinder` mechanism with one which is able to detect both providers with the current mechanism as well as new `ServiceLoader`-based providers from SLF4J `1.8.x`.

This enables JDA to function correctly with any newer SLF4J API and provider version which may be present at runtime, even if the version it is compiled against is `1.7.x`. (`slf4j-api` has [always been binary compatible](https://www.slf4j.org/faq.html#compatibility) to previous versions, also see [this issue](https://github.com/DV8FromTheWorld/JDA/issues/580#issuecomment-479213460) for more information)

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/utils/JDALogger.java
+++ b/src/main/java/net/dv8tion/jda/internal/utils/JDALogger.java
@@ -42,7 +42,7 @@ public class JDALogger
     public static final boolean SLF4J_ENABLED;
     static
     {
-        boolean tmp;
+        boolean tmp = false;
 
         try
         {
--- a/src/main/java/net/dv8tion/jda/internal/utils/JDALogger.java
+++ b/src/main/java/net/dv8tion/jda/internal/utils/JDALogger.java
@@ -23,6 +23,7 @@ import org.slf4j.LoggerFactory;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.Map;
+import java.util.ServiceLoader;
 
 /**
  * This class serves as a LoggerFactory for JDA's internals.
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/utils/JDALogger.java
```diff
@@ -34,23 +35,43 @@
...
-     * Marks whether or not a SLF4J <code>StaticLoggerBinder</code> was found. If false, JDA will use its fallback logger.
+     * Marks whether or not a SLF4J <code>StaticLoggerBinder</code> (pre 1.8.x) or
+     * <code>SLF4JServiceProvider</code> implementation (1.8.x+) was found. If false, JDA will use its fallback logger.
      * <br>This variable is initialized during static class initialization.
      */
     public static final boolean SLF4J_ENABLED;
     static
     {
-        boolean tmp = false;
+        boolean tmp;
```
MinnDevelopment: This was set to false for clarity, I don't see an improvement in removing that.

enveeed: I agree actually, not sure why I removed it.

enveeed:
No problem, will do when I'm home on Sunday

MinnDevelopment:
>Could you attach some proof showing that this actually works? After that I'd be happy to approve and merge.

Any update on this?

PR #981, opened 2019-04-07
Add bag accessors for mentions

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Bags are used to get item cardinality instead of collections with duplicates.

### Example

```java
List<Emote> emotes = message.getEmotes();
Bag<Emote> cardinalities = message.getEmotesBag();

for (Emote e : emotes) {
    System.out.printf("%s was used %d time(s)%n", e.getName(), cardinalities.getCount(e));
}
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -129,7 +129,8 @@ public interface Message extends ISnowflake, Formattable
 
     /**
      * An immutable list of all mentioned {@link net.dv8tion.jda.api.entities.User Users}.
-     * <br>If no user was mentioned, this list is empty.
+     * <br>If no user was mentioned, this list is empty. Elements are sorted in order of appearance. This only
+     * counts direct mentions of the user and not mentions through roles or the everyone tag.
      *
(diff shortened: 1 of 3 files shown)


kantenkugel:
I think the idea is not bad but we could remove duplicate `Matcher` code here (now the `Matcher` is used in both `getXBag()` and `getX()`. Caching them just like we do for the `Lists` would be nice too. 

We could maybe also just cache the `Bag` and in the non-Bag method do something like
```java
public List<X> getX() {
  return getXBag().stream().distinct().collect(Collectors.toList());
}
```
...which doesn't cache the `List`, but due to using the cached `Bag` and not using regex again would still have an improvement.

MinnDevelopment:
@kantenkugel bags don't support order and have a bigger memory footprint.

DV8FromTheWorld:
Can we use this PR to remove the significant supplicate code that has been create in relation to mention resolution?

It should be possible to create a method that can take a message, a matcher, and a collection and add entities as it finds them. I don't like that we would now have like 10 methods with almost duplicate code.

PR #982, opened 2019-04-07
RFC 2119 on Error Messages

Fabricio20:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [X] Other: Error messages

Closes Issue: NaN

## Description

Minor change to make error messages properly comply with RFC 2119.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/utils/Checks.java
+++ b/src/main/java/net/dv8tion/jda/internal/utils/Checks.java
@@ -53,42 +53,42 @@ public class Checks
     public static void notNull(final Object argument, final String name)
     {
         if (argument == null)
-            throw new IllegalArgumentException(name + " may not be null");
+            throw new IllegalArgumentException(name + " must not be null");
     }
 
     public static void notEmpty(final CharSequence argument, final String name)
     {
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
I don't see a need to change this since "may not" and "must not" have the same meaning. "may" is a weak suggestion but "may not" is just as strong as "must not". The RFC you mention is designed for writing other RFCs and doesn't really apply to error messages. I don't want to comply to an unnecessary standard with error messages, they don't change meaning with this in any way.

Frederikam:
`may not` can be synonymous with `must not`, however, the latter is less ambiguous.

DV8FromTheWorld:
I'm going to recend my approval here on grounds of "we needn't abide by RFCs just because they exist".

The currently existing language is clear enough in terms of its intentions, and this PR technically doesn't cover all locations which would need to be changed.

I'm not sure that changing the library to be 2119 compliant would be necessarily more beneficial than the current language, and it would require us actively ensure we are compliant going forward in addition to revisiting all current document and exception messaging.

I appreciate the idea and the contribution.

PR #984, opened 2019-04-09
Abstraction of json parser

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Removes direct API references to org.json and replaces parser dependency.


MinnDevelopment:
This abstraction would also simplify adding support for ETF gateway connections.

```java
public static byte[] getPayload(JDA api, DataObject data) {
    switch (api.getMode()) {
    case PayloadType.ETF:
        return data.toETF();
    case PayloadType.JSON:
        return data.toString().getBytes();
    default: throw new IllegalStateException();
    }
}
```

MinnDevelopment:
This should be ready to go

src/main/java/net/dv8tion/jda/api/entities/MessageHistory.java
```diff
@@ -556,16 +556,16 @@ public MessageRetrieveAction limit(@Nullable Integer limit)
...
             for (int i = 0; i < array.length(); i++)
             {
                 try
                 {
-                    JSONObject obj = array.getJSONObject(i);
+                    DataObject obj = array.getObject(i);
                     result.history.put(obj.getLong("id"), builder.createMessage(obj, channel, false));
                 }
-                catch (JSONException | NullPointerException e)
+                catch (UncheckedIOException | NullPointerException e)
```
MinnDevelopment: This should also check for ParsingException

PR #995, opened 2019-04-23
Add new guild properties

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Discord has added these new properties a while back.


MinnDevelopment:
We also have to add the new boosting properties `premium_tier` and `premium_subscription_count`. Member also has `premium_since` now.

DV8FromTheWorld:
Do we want to use the client naming of "Boosted" or Premium?

MinnDevelopment:
We could probably remove `retrieveVanityUrl` with this as well. The code is now a guild property it seems.

DV8FromTheWorld:
Do we get an update event when it is added?

MinnDevelopment:
@DV8FromTheWorld I haven't confirmed it but I would think so.

DV8FromTheWorld:
If we get WS events for this, then let's remove the method.

Need to detail that it has been removed and replaced by the non network variant

src/main/java/net/dv8tion/jda/internal/handle/GuildMemberUpdateHandler.java
```diff
@@ -117,6 +120,17 @@ protected Long handleInternally(DataObject content)
                                 member, oldNick));
             }
         }
+        if (content.hasKey("premium_since"))
+        {
+            long boostDate = 0;
+            if (!content.isNull("premium_since"))
```
MinnDevelopment: Forgot the else here which sets it to 0 and removing the boostDate variable which is unused

MinnDevelopment:
Reminder: Add the events to ListenerAdapter and add documentation

src/main/java/net/dv8tion/jda/internal/handle/GuildUpdateHandler.java
```diff
@@ -99,7 +99,7 @@ protected Long handleInternally(DataObject content)
         {
             guild.setBannerId(bannerId);
         }
-        if (!Objects.equals(vanityCode, guild.getVanityUrl()))
+        if (!Objects.equals(vanityCode, guild.getVanityCode()))
```
DV8FromTheWorld: We need to add events for all of these new properties.

MinnDevelopment: I already did.

DV8FromTheWorld: Ah, I was looking at an older version of the changes.

PR #996, opened 2019-04-24
Use enum for compression algorithm

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Discord might add other algorithms to choose from like [zstd](https://github.com/luben/zstd-jni) in the future.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/JDABuilder.java
@@ -229,6 +229,9 @@ public class JDABuilder
      * @param  compression
      *         The compression algorithm to use with the gateway connection
      *
+     * @throws java.lang.IllegalArgumentException
+     *         If provided with null
+     *
      * @return The JDABuilder instance. Useful for chaining
      *
      * @see    <a href="https://discordapp.com/developers/docs/topics/gateway#transport-compression" target="_blank">Official Discord Documentation - Transport Compression</a>
--- a/src/main/java/net/dv8tion/jda/api/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/JDABuilder.java
@@ -23,6 +23,7 @@ import net.dv8tion.jda.api.exceptions.AccountTypeException;
 import net.dv8tion.jda.api.hooks.IEventManager;
 import net.dv8tion.jda.api.hooks.VoiceDispatchInterceptor;
 import net.dv8tion.jda.api.requests.RestAction;
+import net.dv8tion.jda.api.utils.Compression;
 import net.dv8tion.jda.api.utils.SessionController;
 import net.dv8tion.jda.api.utils.SessionControllerAdapter;
 import net.dv8tion.jda.api.utils.cache.CacheFlag;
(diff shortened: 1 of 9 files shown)


src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -216,25 +217,26 @@ public JDABuilder setContextEnabled(boolean enable)
     }
 
     /**
-     * Enable stream-compression on the gateway connection,
+     * Enable compression on the gateway connection,
```
DV8FromTheWorld:
```suggestion
     * Set compression type used with the gateway connection,
```

src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -216,25 +217,26 @@ public JDABuilder setContextEnabled(boolean enable)
...
      * this will decrease the amount of used bandwidth for the running bot instance
      * for the cost of a few extra cycles for decompression.
-     * <br><b>Default: true</b>
+     * <br><b>Default: {@link net.dv8tion.jda.api.utils.Compression#ZLIB}</b>
      *
      * <p><b>We recommend to keep this enabled unless you have issues with the decompression</b>
```
DV8FromTheWorld: The verbiage on this probably needs to be updated.

src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -216,25 +217,26 @@ public JDABuilder setContextEnabled(boolean enable)
...
      * <br>This mode might become obligatory in a future version, do not rely on this switch to stay.
```
DV8FromTheWorld: This comment is effectively useless. Lets nix it.

src/main/java/net/dv8tion/jda/api/JDABuilder.java
DV8FromTheWorld: Also explain how you would disable it now (`false` vs new `NONE` enum value)

src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
```diff
@@ -218,25 +219,25 @@ public DefaultShardManagerBuilder setContextEnabled(boolean enable)
...
+     * @param  compression
+     *         The compression algorithm to use for the gateway connection
      *
      * @return The DefaultShardManagerBuilder instance. Useful for chaining.
      *
      * @see    <a href="https://discordapp.com/developers/docs/topics/gateway#transport-compression" target="_blank">Official Discord Documentation - Transport Compression</a>
      */
     @Nonnull
-    public DefaultShardManagerBuilder setCompressionEnabled(boolean enable)
+    public DefaultShardManagerBuilder setCompressionEnabled(Compression compression)
```
DV8FromTheWorld: Missing `@Notnull` and `Check.notNull`

DV8FromTheWorld: Also method name needs to be renamed to `setCompression`

src/main/java/net/dv8tion/jda/api/JDABuilder.java
MinnDevelopment: I don't see how it is useless, they might still require a specific compression in the future.

MinnDevelopment:
I have moved all the decompression logic to a `ZlibDecompressor` which allows us to easily implement other compression algorithms later on.

PR #999, opened 2019-05-01
Add embed album

MinnDevelopment:


Andre601:
Just some thoughts from me about this:
- Maybe rename the images to have it more easily understand what they are about? `1.png`, `2.png`, etc. aren't really telling someone what they show and you have to click on every single one to see (I don't know for what it is/will be used tho.)
- Maybe mention that inline fields will be only 2 per row when a thumbnail is shown?
- ~~Why not dark themed embed?~~

I thought to share this here instead of on the discord because I don't think it really is that of a topic for there.
If it does fit there, then I can easily move it to there, if needed.

Sanduhr32:
They are assets for docs and will be a visual explanation for text above/bellow/surrounding the image.

MinnDevelopment:
We are moving all our assets to a new branch in this repository to not rely on third parties. Currently the embed image album is hosted on imgur which has served us well so far but is not a reliable enough source for the documentation. This allows us to replace the images easily without needing to update any documentation. The names are debatable but it really only serves as a reference for assets that will be used in the documentation of the [EmbedBuilder](https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/core/EmbedBuilder.html).

I'm thinking of making the names something like `${index}-${method}` but I opened a PR for this exact reason instead of directly pushing them to the branch.

DV8FromTheWorld:
@Andre601 is right in that we should have a more explicit naming convention than numbers.

Andre601:
Makes sense considering that you can't just update an image on imgur and always have to re-upload it, which results in a new link being used...

And I like the idea of the name-system. So it would f.e. be `EmbedBuilder-setFooter` if I understand `${index}-${method}` right?

PR #1000, opened 2019-05-06
Add voice kick handling that stops JDA from reconnecting

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Since voice kick was not a thing before, we were not aware that we can be removed from a channel by a moderator. Due to the auto-reconnect logic we then simply reconnect right away. This is changed by this PR which will not attempt to reconnect if we were kicked from the channel.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/audio/hooks/ConnectionStatus.java
+++ b/src/main/java/net/dv8tion/jda/api/audio/hooks/ConnectionStatus.java
@@ -58,6 +58,8 @@ public enum ConnectionStatus
      * that this audio connection was connected to, thus the connection was severed.
      */
     DISCONNECTED_REMOVED_FROM_GUILD(false),
+    /** Indicates that we were kicked from a channel by a moderator */
+    DISCONNECTED_KICKED_FROM_CHANNEL(false),
     /**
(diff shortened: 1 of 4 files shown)


MinnDevelopment:
Should we offer a switch on the audio manager to reconnect if kicked?

nikammerlaan:
Would this stop the AudioManager from still displaying its connectedChannel as the one it was kicked from?

nikammerlaan:
I just experimented with this branch but it does not appear to fix my issue. Does this warrant its own issue or is it related to this? 🤔

MinnDevelopment:
I have no idea what issue you are talking about. This branch only stops JDA from reconnecting after a voice kick. You can tell by reading the description of this pull request.

nikammerlaan:
Yeah, I realize that. I'm just unsure about if the AudioManager mistakenly displaying as connected to a channel which it was kicked from is caused by this or if it's something separate.

MinnDevelopment:
The connected channel is removed by the disconnection.

nikammerlaan:
That doesn't appear to be the case for me.

PR #1001, opened 2019-05-06
Opus Interceptor

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #418 

## Description

Provide a new receive handling for raw opus packets to lazy decode incoming packages.

### Example

```kotlin
class Recorder : AudioSendHandler, AudioReceiveHandler {
    var playback = false
    private val queue = LinkedBlockingQueue<OpusPacket>()

    override fun isOpus() = true // we can just passthrough

    override fun canProvide() = playback && queue.isNotEmpty()

    override fun provide20MsAudio(): ByteBuffer {
        val opus = queue.poll().getOpusAudio() // no decoding needed
        return ByteBuffer.wrap(opus)
    }

    override fun canReceiveEncoded() = !playback

    override fun handleEncodedAudio(packet: OpusPacket) {
        queue += packet
    }

// ignored
    override fun canReceiveCombined() = false
    override fun canReceiveUser() = false

    override fun handleCombinedAudio(combinedAudio: CombinedAudio) {}
    override fun handleUserAudio(userAudio: UserAudio) {}
}
```


MinnDevelopment:
Just as a reminder to myself: I haven't updated the `AudioEchoExample` to use opus because it needs combined audio which isn't supported when only listening for opus.

src/main/java/net/dv8tion/jda/api/audio/OpusPacket.java
```diff
@@ -48,38 +57,93 @@ public OpusPacket(@Nonnull AudioPacket packet, long userId, @Nullable Decoder de
...
     }
 
+    /**
+     * The synchronization source identifier (SSRC) for the user that sent this audio packet.
+     *
+     * @return The SSRC
+     *
+     * @see    <a href="http://www.rfcreader.com/#rfc3550_line548" target="_blank">RTP Header</a>
+     */
     public int getSsrc()
```
DV8FromTheWorld: Why is this lowercase, but RawPacket's is `getSSRC`

MinnDevelopment: Because these getters were auto-generated by intellij. I didn't try to keep any kind of naming pattern. I'm open for suggestions, though.

DV8FromTheWorld: Lets go with `getSSRC` to match our other namings.

src/main/java/net/dv8tion/jda/api/audio/OpusPacket.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    /**
+     * The synchronization source identifier (SSRC) for the user that sent this audio packet.
+     *
+     * @return The SSRC
+     *
+     * @see    <a href="http://www.rfcreader.com/#rfc3550_line548" target="_blank">RTP Header</a>
+     */
+    public int getSsrc()
```
MinnDevelopment:
```suggestion
    public int getSSRC()
```

PR #1002, opened 2019-05-10
Add RawGatewayEvent

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This would allow direct access to gateway dispatch events regardless of JDA handling.
Another use-case is handling events that JDA ignores due to missing cache for e.g. channels.

### Example

```java
public class Forwarding extends ListenerAdapter {
    private final ThirdParty service;

    public Forwarding(ThirdParty service) {
        this.service = service;
    }

    @Override
    public void onRawGateway(RawGatewayEvent event) {
        service.forward(event);
    }
}
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDA.java
+++ b/src/main/java/net/dv8tion/jda/api/JDA.java
@@ -1321,13 +1321,6 @@ public interface JDA
     boolean isAutoReconnect();
 
     /**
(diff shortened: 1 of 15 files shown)


DV8FromTheWorld:
If we add this, we need to look into the possibility of putting it behind a feature flag so as to not double the amount of events fired.

PR #1004, opened 2019-05-10
Improve UpstreamReference

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Allows entities to access newer versions of encapsulated entities. For example an old `TextChannel` could use a newer `Guild` from the current session if it was stored during reconnect (like with a delayed task).
Additionally some requests can be made without a cached encapsulating entity as only the ID is required. Assuming no permission checks have to be made.


src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java
```diff
@@ -94,26 +101,45 @@ public long getDeniedRaw()
...
+            case TEXT:
+                channel = jda.getTextChannelById(channelId);
+                break;
+            case VOICE:
+                channel = jda.getVoiceChannelById(channelId);
+                break;
+            case CATEGORY:
+                channel = jda.getCategoryById(channelId);
+                break;
+                //TODO: STORE
```
MinnDevelopment: Maybe we should add a `Guild#getChannelById(id): GuildChannel` for these things.

src/main/java/net/dv8tion/jda/internal/entities/RoleImpl.java
```diff
@@ -60,18 +62,20 @@
...
+        this.guildId = guild.getIdLong();
+        this.api = new UpstreamReference<>((JDAImpl) guild.getJDA());
     }
 
     @Override
     public int getPosition()
     {
-        if (this == getGuild().getPublicRole())
+        Guild guild = getGuild();
+        if (this == guild.getPublicRole())
```
MinnDevelopment: We should use `equals()` here

PR #1013, opened 2019-06-01
Move presence information into guild presence object

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This should reduce the memory footprint of `MemberImpl` a tiny bit when presence caching is disabled.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/GuildPresence.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/GuildPresence.java
@@ -70,6 +70,8 @@ public interface GuildPresence
 
     class Empty implements GuildPresence
     {
+        private Empty() {}
+
         @Nonnull
         @Override
         public List<Activity> getActivities()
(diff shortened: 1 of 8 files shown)


MinnDevelopment:
The reason this is labeled as breaking is because I also renamed the `CacheFlag` for activity which wasn't representing of its actual usage.

Andre601:
Does this affect any end user code?

MinnDevelopment:
@Andre601 only people who use v4 alpha and made use of `CacheFlag.PRESENCE`. It was renamed to `CacheFlag.ACTIVITY`.

MinnDevelopment:
Memory breakdown per member:

| Scenario | Memory Footprint (64bit) | Meaning |
|-|-|-|
| Full Cache | 72 = 48 + 24 | Member + GuildPresence |
| Disabled Presence | 48 | Member + Empty (amortized) |
| Before patch | 56 | Member |

> I used `java.lang.instrument.Instrumentation` to calculate these values

When the new variable is added and all cache is enabled we increase the total memory footprint of member by 16 bytes (object header). With the cache disabled we improve by 8 bytes.
Now we have to consider whether this is worth it or not.

PR #1018, opened 2019-06-08
Add IconType enum to support more mime types in Icon

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This change is for future compatibility. We should ensure that this is possible with JDA in the case that discord changes the behavior for this.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Icon.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Icon.java
@@ -24,7 +24,7 @@ import javax.annotation.Nonnull;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 
 /**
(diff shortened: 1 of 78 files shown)


src/main/java/net/dv8tion/jda/api/entities/Icon.java
```diff
@@ -117,17 +198,115 @@ public static Icon from(@Nonnull InputStream stream) throws IOException
...
+        @Nonnull
+        public String getMIME()
+        {
+            return mime;
+        }
+
+        /**
+         * The data header for the encoding of an image.
+         *
+         * @return THe data header
```
DV8FromTheWorld:
```suggestion
         * @return The data header
```

MinnDevelopment:
@DV8FromTheWorld I always put enum parameters first in methods. I feel like the sentiment that optional parameters are at the end of the argument list is subjective. On that note, I'm indifferent about the order.

DV8FromTheWorld:
@MinnDevelopment I _think_ the way that we've been doing it thus far is optional parameters last, but with the changes made to the `sendMessage` method for v4, I'm not sure anymore.

I do think it is a bit more clear having them come at the end as it follows what the JDK does with its methods, but I'm only at like a 4 / 10 on the Care O' Meter

PR #1020, opened 2019-06-13
Fix rate-limits for reactions (undocumented behavior)

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fixes rate-limits on reaction endpoints. All these endpoints seem to use the same bucket for rate-limiting. I've tested this before and after this change and it seems to resolve the issue.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/Route.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/Route.java
@@ -106,7 +106,6 @@ public class Route
         public static final Route BAN =                new Route(PUT,    "guilds/{guild_id}/bans/{user_id}",    "guild_id");
         public static final Route KICK_MEMBER =        new Route(DELETE, "guilds/{guild_id}/members/{user_id}", "guild_id");
         public static final Route MODIFY_MEMBER =      new Route(PATCH,  "guilds/{guild_id}/members/{user_id}", "guild_id");
-        // TODO: no headers
         public static final Route ADD_MEMBER =         new Route(PUT,    "guilds/{guild_id}/members/{user_id}", "guild_id");
         public static final Route MODIFY_SELF_NICK =   new Route(PATCH,  "guilds/{guild_id}/members/@me/nick",  "guild_id");
         public static final Route PRUNABLE_COUNT =     new Route(GET,    "guilds/{guild_id}/prune",             "guild_id");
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/requests/Route.java
```diff
@@ -517,4 +515,19 @@ public String getRatelimitRoute()
             return "channels/%s/messages/{message_id}/delete"; //the additional "/delete" forces a new bucket
         }
     }
+
+    // This endpoint shares the rate-limit bucket with REMOVE_ALL_REACTIONS
+    private static class ReactionRoute extends Route
+    {
+        private ReactionRoute(Method method)
+        {
+            super(method, new RateLimit(1, 250), "channels/{channel_id}/messages/{message_id}/reactions/{reaction_code}/{user_id}", "channel_id");
```
DV8FromTheWorld: Does the `REMOVE_REACTION` endpoint share the same `1 reaction / 250ms` limit that `ADD_REACTION` does?

Additionally, if that is the case, does `REMOVE_ALL_REACTIONS` also share that limit?

How does this actually effect the system now that they have conflicting route bucket keys?

src/main/java/net/dv8tion/jda/internal/requests/Route.java
```diff
@@ -202,9 +201,8 @@
         public static final Route ADD_PINNED_MESSAGE =    new Route(PUT,    "channels/{channel_id}/pins/{message_id}",     "channel_id");
         public static final Route REMOVE_PINNED_MESSAGE = new Route(DELETE, "channels/{channel_id}/pins/{message_id}",     "channel_id");
 
-        public static final Route ADD_REACTION =             new Route(PUT,    new RateLimit(1, 250),
-                                                                               "channels/{channel_id}/messages/{message_id}/reactions/{reaction_code}/@me",       "channel_id");
-        public static final Route REMOVE_REACTION =          new Route(DELETE, "channels/{channel_id}/messages/{message_id}/reactions/{reaction_code}/{user_id}", "channel_id");
+        public static final Route ADD_REACTION =             new ReactionRoute(PUT);
```
DV8FromTheWorld: I'm fine with this for now, but if this starts to become something that we notice (routes having route-buckets that are different from their full route) then we need to look at adding additional functionality to the `Route` object to support that usecase.

Don't want or need a bunch of Route subclasses.

src/main/java/net/dv8tion/jda/internal/requests/Route.java
MinnDevelopment: My tests indicate that all these routes have the same shared rate limit. This also includes the reset time but I'll check again.

src/main/java/net/dv8tion/jda/internal/requests/Route.java
MinnDevelopment: If this becomes problematic I will change the handling for this but for now this workaround should be sufficient

src/main/java/net/dv8tion/jda/internal/requests/Route.java
MinnDevelopment: According to my tests this did not encounter 429 once in 10 tries.
```
03:25:40.988 JDA-ThreadPool 9  RateLimiter DEBUG  Backing off 250 milliseconds on route /channels/163772719836430337/messages/{message_id}/reactions
03:25:41.401 JDA-ThreadPool 12 RateLimiter DEBUG  Backing off 250 milliseconds on route /channels/163772719836430337/messages/{message_id}/reactions
03:25:41.823 JDA-ThreadPool 12 RateLimiter DEBUG  Backing off 250 milliseconds on route /channels/163772719836430337/messages/{message_id}/reactions
03:25:42.248 JDA-ThreadPool 12 RateLimiter DEBUG  Backing off 249 milliseconds on route /channels/163772719836430337/messages/{message_id}/reactions
03:25:42.652 JDA-ThreadPool 2  RateLimiter DEBUG  Backing off 250 milliseconds on route /channels/163772719836430337/messages/{message_id}/reactions
03:25:43.068 JDA-ThreadPool 12 RateLimiter DEBUG  Backing off 249 milliseconds on route /channels/163772719836430337/messages/{message_id}/reactions
03:25:43.482 JDA-ThreadPool 1  RateLimiter DEBUG  Backing off 249 milliseconds on route /channels/163772719836430337/messages/{message_id}/reactions
03:25:43.903 JDA-ThreadPool 6  RateLimiter DEBUG  Backing off 250 milliseconds on route /channels/163772719836430337/messages/{message_id}/reactions
03:25:44.312 JDA-ThreadPool 2  RateLimiter DEBUG  Backing off 250 milliseconds on route /channels/163772719836430337/messages/{message_id}/reactions
03:25:44.729 JDA-ThreadPool 2  RateLimiter DEBUG  Backing off 250 milliseconds on route /channels/163772719836430337/messages/{message_id}/reactions```

PR #1023, opened 2019-06-17
Update link to emoji documentation

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description
I updated some descriptions in the Emotes.java
- `Emote#getGuild()` also returns null in the case that the Guild is in fact one the bot shares, but is on a different shard than where this method was called.
- https://discordapp.com/developers/docs/resources/guild#emoji-object is no longer a valid link (no longer exists) so it was removed. When the link changed, let me know so that I can update it.

This PR is from the feature/v4/new-guild-properties and should be pulled into this branch if it gets accepted.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Emote.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Emote.java
@@ -54,7 +54,7 @@ public interface Emote extends IMentionable, IFakeable
     /**
      * The {@link net.dv8tion.jda.api.entities.Guild Guild} this emote is attached to.
      *
-     * <p><b>This is null if the emote is fake (retrieved from a Message)</b>
+     * <p><b>This is null if the emote is fake (retrieved from a Message) or when the attached Guild is on a separate Shard then the one this is called.</b>
      *
      * @return Guild of this emote or null if it is a fake entity
      */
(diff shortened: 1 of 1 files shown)


JohnnyJayJay:
> Emote#getGuild() also returns null in the case that the Guild is in fact one the bot shares, but is on a different shard than where this method was called.

That is the definition of a fake entity, just like the [documentation for IFakeable] describes it. (https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/core/entities/IFakeable.html).

This change is therefore redundant imo and might even confuse people since other documentation of fake entity behaviour does not explicitely mention what fake means.

src/main/java/net/dv8tion/jda/api/entities/Emote.java
```diff
@@ -54,16 +54,15 @@
...
+     * <p><b>This is null if the emote is fake (retrieved from a Message) or when the attached Guild is on a separate Shard then the one this is called.</b>
      *
      * @return Guild of this emote or null if it is a fake entity
      */
     @Nullable
     Guild getGuild();
 
     /**
-     * Roles this emote is active for
-     * <br><a href="https://discordapp.com/developers/docs/resources/guild#emoji-object" target="_blank">Learn More</a>
```
MinnDevelopment: Replace guild with emoji for the new link

src/main/java/net/dv8tion/jda/api/entities/Emote.java
```diff
@@ -54,16 +54,15 @@
     /**
      * The {@link net.dv8tion.jda.api.entities.Guild Guild} this emote is attached to.
      *
-     * <p><b>This is null if the emote is fake (retrieved from a Message)</b>
+     * <p><b>This is null if the emote is fake (retrieved from a Message) or when the attached Guild is on a separate Shard then the one this is called.</b>
```
MinnDevelopment: This is redundant since fake includes this in its definition

Andre601:
Updated links and removed the redundant message

PR #1024, opened 2019-06-19
Add implementation for teams

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Add support for teams in the application info. This can be useful for bots that have multiple owners.


src/main/java/net/dv8tion/jda/api/entities/TeamMember.java
```diff
@@ -0,0 +1,49 @@
...
+
+import javax.annotation.Nonnull;
+
+public interface TeamMember
+{
+    @Nonnull
+    User getUser();
+
+    @Nonnull
+    MembershipState getMembershipState();
```
DV8FromTheWorld: This should probably be `Status`, not `State`

src/main/java/net/dv8tion/jda/api/entities/ApplicationTeam.java
```diff
@@ -0,0 +1,60 @@
...
+        return iconId == null ? null : String.format(ICON_URL, getId(), iconId);
+    }
+
+    /**
+     * The {@link net.dv8tion.jda.api.entities.TeamMember Team Members}.
+     *
+     * @return Immutable list of team members
+     */
+    @Nonnull
+    List<? extends TeamMember> getMembers();
```
MinnDevelopment: This bounded wildcard makes `list.add(new TeamMember())` a compile-time error but its probably not worth it to break our consistency for it.

PR #1027, opened 2019-06-20
Deprecate ShardManager#setGame

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Somehow this slipped through in the rename.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
+++ b/src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
@@ -15,6 +15,8 @@
  */
 package net.dv8tion.jda.api.sharding;
 
+import net.dv8tion.jda.annotations.DeprecatedSince;
+import net.dv8tion.jda.annotations.ReplaceWith;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.JDA.Status;
 import net.dv8tion.jda.api.OnlineStatus;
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
```diff
@@ -1096,28 +1098,53 @@ default VoiceChannel getVoiceChannelById(@Nonnull final String id)
      *
      * @see    net.dv8tion.jda.api.entities.Activity#playing(String)
      * @see    net.dv8tion.jda.api.entities.Activity#streaming(String, String)
+     *
+     * @deprecated
+     *         Use {@link #setActivity(net.dv8tion.jda.api.entities.Activity)} instead
      */
+    @Deprecated
+    @DeprecatedSince("4.0.0")
```
DV8FromTheWorld: The other PR that we just accepted (#1025) uses `@DeprecatedSince("4.BETA.0")`. Do we want to use that here instead?

MinnDevelopment: 4.BETA.0 is already a released version so saying it was deprecated since then would be a lie. I will probably change it there as well.

DV8FromTheWorld: Sounds good.

PR #1028, opened 2019-06-21
Add GatewayPingEvent

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Introduces a new event to track updates to the gateway ping. This can be very useful for metrics or detecting connection issues.

```java
public class ConnectionListener extends ListenerAdapter {
    @Override
    public void onGatewayPing(GatewayPingEvent event) {
        int prev = event.getOldValue(), next = event.getNewValue();
        if (next >= prev + 100 && prev != -1) {
            System.err.println("Gateway ping has increased by " + (next - prev) + "ms!");
        }
    }
}
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/events/GatewayPingEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/GatewayPingEvent.java
@@ -26,11 +26,11 @@ import javax.annotation.Nonnull;
  *
  * <p>Can be used to detect changes to the gateway ping.
  *
- * <p>Identifier: {@code ping}
+ * <p>Identifier: {@code gateway-ping}
  */
 public class GatewayPingEvent extends Event implements UpdateEvent<JDA, Long>
 {
-    public static final String IDENTIFIER = "ping";
+    public static final String IDENTIFIER = "gateway-ping";
     private final long next, prev;
 
     public GatewayPingEvent(@Nonnull JDA api, long old)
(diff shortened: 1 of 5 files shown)


src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -829,7 +830,11 @@ public void setAudioSendFactory(IAudioSendFactory factory)
 
     public void setPing(long ping)
     {
+        long oldPing = this.ping;
         this.ping = ping;
+        getEventManager().handle(
+            new GatewayPingEvent(
+                this, oldPing));
```
MinnDevelopment: This is technically a side-effect but should be ok.

MinnDevelopment: I have proposed a new proxy to avoid exceptions from handle() calls which could cause issues

src/main/java/net/dv8tion/jda/api/events/GatewayPingEvent.java
```diff
@@ -0,0 +1,96 @@
...
+ * Indicates that the gateway ping has been updated by the heartbeat cycle.
+ * <br>You can always get the last ping update with {@link net.dv8tion.jda.api.JDA#getGatewayPing()}.
+ *
+ * <p>Can be used to detect changes to the gateway ping.
+ *
+ * <p>Identifier: {@code ping}
+ */
+public class GatewayPingEvent extends Event implements UpdateEvent<JDA, Long>
+{
+    public static final String IDENTIFIER = "ping";
```
DV8FromTheWorld: Should this be `gateway-ping` instead?

PR #1030, opened 2019-06-23
Make GenericGuildVoiceUpdateEvent public

averen:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

GenericGuildVoiceUpdateEvent is currently package-private, which obviously is a mistake for an event.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/events/guild/voice/GenericGuildVoiceUpdateEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/guild/voice/GenericGuildVoiceUpdateEvent.java
@@ -23,7 +23,7 @@ import net.dv8tion.jda.api.entities.VoiceChannel;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
Whats wrong with using `GuildVoiceUpdateEvent`? The interface exists for this exact purpose and is also in `ListenerAdapter`.

averen:
I must've overlooked `GuildVoiceUpdateEvent`, but I'm curious as to why it is intended behavior.

MinnDevelopment:
It is intended behavior because you should listen for the `GuildVoiceUpdateEvent` and not the `GenericGuildVoiceUpdateEvent`. The latter isn't even in `ListenerAdapter` and that hasn't been addressed by this PR either. There really is no point to making this public, it only exists to reduce duplicate code.

averen:
Fair enough. I assumed since it was the direct parent of `GuildVoiceJoinEvent` and `GuildVoiceMoveEvent` that it could be used as a cast.

MinnDevelopment:
You can cast to the interface as well, its also a direct parent.

PR #1031, opened 2019-06-23
Add EventManagerProxy

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This is done to avoid exceptions that could disrupt the SocketHandler.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/JDAImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/JDAImpl.java
@@ -684,7 +684,7 @@ public class JDAImpl implements JDA
     @Override
     public IEventManager getEventManager()
     {
-        return eventManager;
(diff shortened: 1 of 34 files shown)


src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -721,7 +722,7 @@ public void removeEventListener(@Nonnull Object... listeners)
     @Override
     public List<Object> getRegisteredListeners()
     {
-        return Collections.unmodifiableList(eventManager.getRegisteredListeners());
```
MinnDevelopment: This should be implementation defined.

PR #1032, opened 2019-06-23
Mark getShardInfo() nonnull

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

The shard info indicates which shard id and shard total the current JDA session uses. If sharding is disabled this is equivalent to [0 / 1]. The identification string used for thread names is unchanged and will not report shard details unless explicitly set through `useSharding` or the shard manager.

- This is not a breaking change in java because it only changes an annotation.
- This is not a breaking change in kotlin because you can also store nonnull types in nullable typed variables. Safe-call operators should just emit a warning similar to unnecessary null checks.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDA.java
+++ b/src/main/java/net/dv8tion/jda/api/JDA.java
@@ -113,6 +113,9 @@ public interface JDA
      */
     class ShardInfo
     {
+        /** Default sharding config with one shard */
+        public static final ShardInfo SINGLE = new ShardInfo(0, 1);
+
         int shardId;
         int shardTotal;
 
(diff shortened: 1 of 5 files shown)


Andre601:
Would there still be a way to check if the JDA instance is actually part of a sharded bot?
I have an API that checks if JDA has ShardInfo to optionally add the shard id and shard count and this change would make it always set those which really isn't useful in case of a single, non-sharded bot.

MinnDevelopment:
A non-sharded bot has the shard total of 1 and shard id 0. That means its effectively not sharded either way. This behavior of `useSharding(0, 1)` and not setting sharding configuration is identical should be treated the same.

PR #1042, opened 2019-07-02
Allow more configuration on data caching

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

With these changes I introduce `XData` and `MutableXData` interfaces which can be used to provide `XUpdateDigestEvent` rather than atomic events for each changed property. These interfaces can be implemented by the library user and provided using a callback.

```java
new JDABuilder(TOKEN)
    .setGuildDataProvider((id) -> new CustomGuildData());
```

We provide 2 implementations by default:

| Name | Description |
|-------|-------------|
| LIGHT | No data is stored, only default values (constants) |
| RICH | Everything is stored |

You could easily implement custom providers by extending the existing `LIGHT` configuration:

```java
public class CustomGuildData extends LightGuildData {
    private String description = "";
    private String bannerId = null;
    private String iconId = null;

    @Override
    public CustomGuildData copy() {
        // ...
    }

    // implement getters and setters
}
```

Example using rich and having digest events enabled:

```java
@Override
public void onGuildUpdateDigest(@Nonnull GuildUpdateDigestEvent event) {
    LOG.info("Received guild update digest for guild {}", event.getGuild().getName());
}

@Override
public void onGenericGuildUpdate(@Nonnull GenericGuildUpdateEvent event) {
    if (!(event instanceof GuildUpdateDigestEvent)) {
        LOG.info("Received atomic guild update ({}) for guild {}", event.getPropertyIdentifier(), event.getGuild().getName());
    }
}
```
Output:
```
13:29:01.931 JDA INFO   Received guild update digest for guild Testing Bacon
13:29:01.932 JDA INFO   Received atomic guild update (max_members) for guild Testing Bacon
13:29:01.933 JDA INFO   Received atomic guild update (region) for guild Testing Bacon
13:29:01.933 JDA INFO   Received atomic guild update (afk_timeout) for guild Testing Bacon
13:29:01.934 JDA INFO   Received atomic guild update (afk_channel) for guild Testing Bacon
13:29:01.934 JDA INFO   Received atomic guild update (system_channel) for guild Testing Bacon
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/JDABuilder.java
@@ -19,6 +19,7 @@ import com.neovisionaries.ws.client.WebSocketFactory;
 import net.dv8tion.jda.annotations.Incubating;
 import net.dv8tion.jda.api.audio.factory.IAudioSendFactory;
 import net.dv8tion.jda.api.entities.Activity;
+import net.dv8tion.jda.api.entities.bean.MutableGuildData;
 import net.dv8tion.jda.api.exceptions.AccountTypeException;
 import net.dv8tion.jda.api.hooks.IEventManager;
 import net.dv8tion.jda.api.hooks.VoiceDispatchInterceptor;
@@ -30,10 +31,7 @@ import net.dv8tion.jda.api.utils.cache.CacheFlag;
 import net.dv8tion.jda.internal.JDAImpl;
 import net.dv8tion.jda.internal.managers.PresenceImpl;
 import net.dv8tion.jda.internal.utils.Checks;
-import net.dv8tion.jda.internal.utils.config.AuthorizationConfig;
-import net.dv8tion.jda.internal.utils.config.MetaConfig;
-import net.dv8tion.jda.internal.utils.config.SessionConfig;
-import net.dv8tion.jda.internal.utils.config.ThreadingConfig;
+import net.dv8tion.jda.internal.utils.config.*;
 import net.dv8tion.jda.internal.utils.config.flags.ConfigFlag;
 import okhttp3.OkHttpClient;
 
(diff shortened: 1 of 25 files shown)


MinnDevelopment:
I also want to add a flag to switch between either digest events or atomic events. That way we move more towards 1:1 mapping of the events rather than 1:N like we do currently.

MinnDevelopment:
This is probably not needed when we allow lazy loading and disabling guild subscriptions.

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
```diff
@@ -85,13 +81,14 @@ public JDA getJDA()
     @Override
     public OffsetDateTime getTimeJoined()
     {
-        return OffsetDateTime.ofInstant(Instant.ofEpochMilli(joinDate), OFFSET);
+        return OffsetDateTime.ofInstant(Instant.ofEpochMilli(data.getTimeBoosted()), OFFSET);
```
caneleex: Should be replaced with `data.getTimeJoined()` instead

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
```diff
@@ -245,42 +239,37 @@ public long getIdLong()
...
     {
-        this.nickname = nickname;
+        this.data.setNickname(nickname);
         return this;
     }
 
     public MemberImpl setJoinDate(long joinDate)
     {
-        this.joinDate = joinDate;
+        this.data.setTimeBoosted(joinDate);
```
caneleex: Should be replaced with `data.setTimeJoined()` instead.

MinnDevelopment:
I want to revisit this for 4.2.0. I also want to make an easy switch to use the LIGHT mode on all entities by default like: `builder.setDefaultCachePolicy(CachePolicy.LIGHT)` which makes it easier to opt-in on specific entities.

PR #1046, opened 2019-07-07
Lazy Loading

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Implementation for lazy loading of guild members. This will disable the member chunking performed when the guild is received and instead load members as they appear. This obviously removes a lot of information but can be very useful for large bots that see a lot of members which are never active.

The only information missing when creating a member from a presence update seems to be the join date, we could in theory just disable this caching during lazy loading. Currently I just set the join date to the guild creation time instead.

You can dynamically load the entire member cache for a guild or single members.

```java
guild.retrieveMembers().thenRun(() -> {
    // use guild with members loaded
});

guild.retrieveMemberById(86699011792191488L).queue((member) -> {
    // use member loaded through either cache or http request
});
```

This also adds support for disabling `guild_subscriptions` which disables all guild subscriptions such as presence and typing events. ~~When guild subscriptions are disabled the user cache becomes inconsistent as discord will not inform us of updates directly. However, we can update the cached users through message and voice state events and REST calls.~~
We disable the user cache completely in this case because there are no member update events so there is no way to make a working system of caching them. The only users that get cached are users who participate in voice channels.

### Example 

```java
new JDABuilder(TOKEN)
    .setGuildSubscriptionsEnabled(false)
    .setChunkingFilter(ChunkingFilter.include(125227483518861312L, 81384788765712384L));
```


src/main/java/net/dv8tion/jda/api/entities/Member.java
```diff
@@ -24,7 +24,7 @@
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.awt.Color;
+import java.awt.*;
```
MinnDevelopment: I need to disable this in intellij again.

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -158,6 +160,16 @@ public boolean isCacheFlagSet(CacheFlag flag)
         return metaConfig.getCacheFlags().contains(flag);
     }
 
+    public boolean chunkGuild(long id)
+    {
+        return chunkingFilter.filter(id);
```
MinnDevelopment: Handle exceptions

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -856,6 +866,14 @@ public Message createMessage(DataObject jsonObject, boolean exceptionOnMissingUs
     }
     public Message createMessage(DataObject jsonObject, MessageChannel chan, boolean exceptionOnMissingUser)
     {
+        if (chan.getType().isGuild() && !jsonObject.isNull("member"))
+        {
+            GuildChannel guildChannel = (GuildChannel) chan;
+            DataObject member = jsonObject.getObject("member");
+            member.put("user", jsonObject.getObject("author"));
+            createMember((GuildImpl) guildChannel.getGuild(), member);
```
MinnDevelopment: This shouldn't be called if the member already exists (fully loaded)

src/main/java/net/dv8tion/jda/internal/handle/GuildMemberAddHandler.java
```diff
@@ -46,6 +46,8 @@ protected Long handleInternally(DataObject content)
         }
 
         Member member = getJDA().getEntityBuilder().createMember(guild, content);
+        // Update memberCount
+        guild.onMemberAdd();
```
MinnDevelopment: This should be called before createMember

src/main/java/net/dv8tion/jda/internal/handle/PresenceUpdateHandler.java
```diff
@@ -47,212 +51,229 @@ public PresenceUpdateHandler(JDAImpl api)
...
-            if (jsonUser.hasKey("username"))
-            {
-                String name = jsonUser.getString("username");
-                String discriminator = jsonUser.get("discriminator").toString();
-                String avatarId = jsonUser.getString("avatar", null);
-                String oldAvatar = user.getAvatarId();
+            if (jsonUser.isNull("username") || content.getString("status", "offline").equals(OnlineStatus.OFFLINE.getKey()))
+                return null;
+            // We should have somewhat enough information to create this member, so lets do it!
+            user = createMember(content, guildId, guild, jsonUser);
```
MinnDevelopment: Think about whether this is needed

src/main/java/net/dv8tion/jda/internal/handle/VoiceStateUpdateHandler.java
```diff
@@ -75,7 +76,7 @@ private void handleGuildVoiceState(DataObject content)
             return;
         }
 
-        MemberImpl member = (MemberImpl) guild.getMemberById(userId);
+        MemberImpl member = getJDA().getEntityBuilder().createMember((GuildImpl) guild, content.getObject("member"));
```
MinnDevelopment: Only call this when the member isn't fully loaded yet

MinnDevelopment:
Disabling guild subscriptions seem to require more changes than expected. Member cache seems to be a bad idea when subscriptions are disabled, so we need to rewrite parts of the library to disable this cache when subscriptions are disabled.

MinnDevelopment:
I need to compile a list of events that require guild subscriptions to be enabled.

MinnDevelopment:
Here are some graphs for my small bot with around 55K users:

### setChunkingFilter(ALL) and setGuildsSubscriptsionsEnabled(true)

image

### setChunkingFilter(NONE) and setGuildSubscriptionsEnabled(true)

image

### setGuildsSubscriptionsEnabled(false)

image

MinnDevelopment:
We should also experiment with the large_threshold for lazy loading.

MinnDevelopment:
This will not be part of 4.0.0 because it is too volatile still.

MinnDevelopment:
This is waiting for member state to be added to reactions. Otherwise we would have to workaround the problem until they change it which is not good for a stable release.

Related issue: discordapp/discord-api-docs#1035

PR #1050, opened 2019-07-16
Use single UDP socket for all audio connections

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

UDP is a connection-less protocol so we can just use a single socket for any amount of audio connections. The socket is created on first use and closed by JDA#shutdown.


MinnDevelopment:
I need to figure out a good way to solve receive with this.

MinnDevelopment:
I have decided not to go through with this because it entails too much restructuring due to voice receive.

PR #1054, opened 2019-07-18
Add specification to the default SplitPolicies of the MessageBuilder

MaximevanderSmissen:
In the JavaDoc, it is now specified whether or not a certain default SplitPolicy removes the characters it splits on.

-  I have read the [contributing guidelines][contributing].

### Changes

- Documentation

## Description
Adds some JavaDoc that states whether NEWLINE, SPACE and ANYWHERE do or don't remove the characters they split on (one previously had to check the source code for that).

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/MessageBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/MessageBuilder.java
@@ -944,17 +944,17 @@ public class MessageBuilder implements Appendable
     public interface SplitPolicy
     {
         /**
-         * Splits on newline chars {@code `\n`}.
+         * Splits on newline chars {@code `\n`} and removes the characters it splits on.
          */
         SplitPolicy NEWLINE = new CharSequenceSplitPolicy("\n", true);
 
         /**
-         * Splits on space chars {@code `\u0020`}.
+         * Splits on space chars {@code `\u0020`} and removes the characters it splits on.
          */
         SplitPolicy SPACE = new CharSequenceSplitPolicy(" ", true);
 
         /**
-         * Splits exactly after 2000 chars.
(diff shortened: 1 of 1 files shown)


kantenkugel:
Is this really some needed addition?

I would say that it doesn't matter for spaces and newlines, since those are trimmed by the Discord API **and** users wouldn't want them present anyway.
And for the `ANYWHERE` SplitPolicy, it should be clear to anyone that nothing is lost which makes the javadoc some *Captain Obvious* level addition.

In that sense, it doesn't add any important or relevant informations IMO.

MaximevanderSmissen:
> I would say that it doesn't matter for spaces and newlines, since those are trimmed by the Discord API **and** users wouldn't want them present anyway.

I previously didn't know this and agree that it'd make the addition quite useless, one could argue that it'd provide extra information, and might be useful for those who don't know about the trimming.

> And for the `ANYWHERE` SplitPolicy, it should be clear to anyone that nothing is lost which makes the javadoc some _Captain Obvious_ level addition.

The `ANYWHERE` SplitPolicy addition was mostly made in the thought of being complete (might as well add the, although obvious, 3rd one, while I'm at it).

> In that sense, it doesn't add any important or relevant informations IMO.

In conclusion, I now share @kantenkugel's opinion. I won't close the PR myself, but only for the purposes of getting a second opinion from another maintainer, similar to what @kantenkugel told me on discord.

Andre601:
I also think that this isn't needed.
I myself didn't look at this SplitPolicy and have the feeling that most other people also don't look at it in normal cases. (I didn't even know that this is an actual thing)

MinnDevelopment:
I agree with @kantenkugel so I'm closing this.

Issue #1055, opened 2019-07-19
Channels mentioned with the new format do no show up in Message#getMentionedChannels


BenjaminUrquhart:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [X] I have checked for similar issues.
- [X] I have updated to the [latest JDA version][download].
- [X] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report
Message#getMentionedChannels does not work when a channel is mentioned using the new syntax.

There is no mention of this behavior in the Discord Docs.
### Expected Behavior
Message#getMentionedChannels should work when any channel is mentioned
### Code Example or Reproduction Steps
1. Mention channel in the form `<#channelID-guildID-name>`
2. run `Message#getMentionedChannels`

![Image](https://chat.is-going-to-rickroll.me/i/kQ3UQ_UcUlB8TQ.png)

[Discord API issue](https://github.com/discordapp/discord-api-docs/issues/1029)

Andre601:
This also seems to only be the case with verified Guilds for now.
At least I only saw this new format in such a Discord while the non-verified still have the old system.

The client/Discord itself already support for this across all Discords meaning you can use the new format and it will format.

BenjaminUrquhart:
See: https://github.com/discordapp/discord-api-docs/pull/1030

PR #1056, opened 2019-07-20
Add support for new channel mentions

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1055 

## Description

This adds support for new channel mention syntax in `getMentionedChannels` and `getContentDisplay`.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
@@ -489,7 +489,7 @@ public class ReceivedMessage extends AbstractMessage
             for (TextChannel mentionedChannel : getMentionedChannels())
             {
                 tmp = tmp.replaceAll(
-                        "<#" + mentionedChannel.getId() + "(?::(\\d+):([^>\\s]+))?>", // possible mention formats
+                        "<#" + mentionedChannel.getId() + "(?::(\\d+):([^>]{1,100}))?>", // possible mention formats
                         '#' + Matcher.quoteReplacement(mentionedChannel.getName()));  // proper look and feel
             }
             for (Emote emote : getEmotes())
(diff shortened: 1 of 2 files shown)


Andre601:
Will this perhaps also be added to JDA 3 in case the new channel mention type is released before JDA 4 is out of beta?

MinnDevelopment:
@Andre601 this is a new discord feature and does not replace existing functionality. This means I will most likely not add it to the older versions of the library. The beta version is completely stable and usable right now.

DV8FromTheWorld:
Approved, however, we should probably wait until the feature is actually released to verify everything works, especially with the discussion around `base64` encoding of the name.

DV8FromTheWorld:
Looks like they just released a comment about name length and a new recommended regex
https://github.com/discordapp/discord-api-docs/pull/1030#issuecomment-513892776

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -486,13 +486,15 @@ public String getContentDisplay()
                     name = user.getName();
                 tmp = tmp.replaceAll("<@!?" + Pattern.quote(user.getId()) + '>', '@' + Matcher.quoteReplacement(name));
             }
-            for (Emote emote : getEmotes())
+            for (TextChannel mentionedChannel : getMentionedChannels())
             {
-                tmp = tmp.replace(emote.getAsMention(), ":" + emote.getName() + ":");
+                tmp = tmp.replaceAll(
+                        "<#" + mentionedChannel.getId() + "(?::(\\d+):([^>\\s]+))?>", // possible mention formats
```
DV8FromTheWorld: Do you think it would be possible to somehow reuse the regex we already have instead of having it duplicated twice as a literal?

MinnDevelopment: Would be possible but probably not worth the effort. Its unlikely this will ever be changed again.

MinnDevelopment:
I'm putting this on hold until the situation becomes less confusing.

MinnDevelopment:
It looks like this feature got nixed, gonna wait for an official response though.

PR #1057, opened 2019-07-20
Update documentation to remove references to internals

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This removes documentation referring to internals in JDA. Additionally this adds a replacement for `SessionController#getGatewayBot` which does not use the internal `Pair` utility.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Member.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Member.java
@@ -18,7 +18,6 @@ package net.dv8tion.jda.api.entities;
 
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.OnlineStatus;
-import net.dv8tion.jda.api.Permission;
 import net.dv8tion.jda.api.requests.restaction.AuditableRestAction;
 
 import javax.annotation.CheckReturnValue;
@@ -26,7 +25,6 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.awt.*;
 import java.time.OffsetDateTime;
-import java.util.EnumSet;
 import java.util.List;
 
 /**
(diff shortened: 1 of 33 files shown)


src/main/java/net/dv8tion/jda/api/entities/Member.java
```diff
@@ -24,7 +24,7 @@
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.awt.Color;
+import java.awt.*;
```
DV8FromTheWorld: Intellij did its dumbness again.

src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManager.java
```diff
@@ -509,16 +509,16 @@ protected JDAImpl buildInstance(final int shardId) throws LoginException, Interr
         {
             try
             {
-                Pair<String, Integer> gateway = jda.getGatewayBot();
-                this.gatewayURL = gateway.getLeft();
+                SessionController.GatewayBot gateway = jda.getGatewayBot();
```
DV8FromTheWorld: Does this actually compile? It looks like `getGatewayBot` is still returning a `Pair`

MinnDevelopment: You're confusing `SessionController#getGatewayBot` and `JDAImpl#getGatewayBot`

src/main/java/net/dv8tion/jda/api/entities/Member.java
MinnDevelopment: I honestly don't care at this point

src/main/java/net/dv8tion/jda/api/utils/SessionController.java
```diff
@@ -144,10 +143,81 @@
...
+    default GatewayBot getShardedGateway(@Nonnull JDA api)
+    {
+        Pair<String, Integer> tuple = getGatewayBot(api);
+        return new GatewayBot(tuple.getLeft(), tuple.getRight());
+    }
+
+    /**
+     * POJO containing the gateway endpoint and recommended shard total for a shard manager.
+     */
+    class GatewayBot
```
DV8FromTheWorld: Should this still be called `GatewayBot` since we're moving to the `ShardedGateway` naming?

PR #1062, opened 2019-07-24
Add quote handling and MarkdownUtil

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

I've added a utility class to apply formatting and added handling for quote markdown in the sanitizer.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/MarkdownUtil.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/MarkdownUtil.java
@@ -165,7 +165,7 @@ public final class MarkdownUtil
     /**
      * Escapes already existing quote regions in the input
      * and applies quote formatting to the entire string.
-     * <br>The resulting string will be {@code "> " + escaped(input).replace(LF, "> ")}.
+     * <br>The resulting string will be {@code "> " + escaped(input).replace("\n", "\n> ")}.
      *
      * @param  input
      *         The input to quote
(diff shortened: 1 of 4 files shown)


Andre601:
Is there already known if quote blocks are still normal text in contentRaw or are they in fact embeds (considering their style is close/simlar to embeds)?

MinnDevelopment:
image

The only similar thing with embeds is the colored column to the left of it. The background is missing though and the column is a bit different too.

MinnDevelopment:
The handling for the sanitizer is a bit broken currently because it ignores line-breaks which are relevant for quotes. Additionally jake said they were going to change in a way that it requires a space such as `> hello` rather than `>hello`.

Andre601:
> The handling for the sanitizer is a bit broken currently because it ignores line-breaks which are relevant for quotes. Additionally jake said they were going to change in a way that it requires a space such as `> hello` rather than `>hello`.

The latest info from Discord-Datamining does support this change.
https://github.com/DJScias/Discord-Datamining/commit/56ff440c09745aec01df087959b62059e1b6cd4f#commitcomment-34437277

src/main/java/net/dv8tion/jda/api/utils/MarkdownUtil.java
```diff
@@ -0,0 +1,198 @@
...
+    public static String strike(@Nonnull String input)
+    {
+        String sanitized = MarkdownSanitizer.escape(input, ~MarkdownSanitizer.STRIKE);
+        return "~~" + sanitized + "~~";
+    }
+
+    /**
+     * Escapes already existing quote regions in the input
+     * and applies quote formatting to the entire string.
+     * <br>The resulting string will be {@code "> " + escaped(input).replace(LF, "> ")}.
```
MinnDevelopment:
```suggestion
     * <br>The resulting string will be {@code "> " + escaped(input).replace("\n", "\n> ")}.
```

Issue #1071, opened 2019-07-30
Message#isImage() is returning true on videos


HeyBanditoz:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report
Message#isImage() is returning true on videos, even though it probably shouldn't.

### Expected Behavior
I would expect it to return false on video files.

### Code Example or Reproduction Steps
```Java
import net.dv8tion.jda.core.JDA;
import net.dv8tion.jda.core.JDABuilder;
import net.dv8tion.jda.core.entities.Message;
import net.dv8tion.jda.core.events.message.MessageReceivedEvent;
import net.dv8tion.jda.core.hooks.ListenerAdapter;

import javax.security.auth.login.LoginException;

public class scratch extends ListenerAdapter {
    public static void main(String[] args) throws InterruptedException, LoginException {
        JDA jda = new JDABuilder("token").build();
        jda.awaitReady();
        jda.addEventListener(new scratch());
    }

    @Override
    public void onMessageReceived(MessageReceivedEvent e) {
        for (Message.Attachment a : e.getMessage().getAttachments()) {
            System.out.println(a.getUrl() + "\n" + a.isImage());
        }
    }
}
```
Outputs:
```
https://cdn.discordapp.com/attachments/570771524697718810/605885966879817760/big-buck-bunny_trailer.webm
true
https://cdn.discordapp.com/attachments/570771524697718810/605890023232897024/Replay_2019-05-31_15-55-30.mp4
true
```
even though it is a video file.

### Exception or Error
N/A

Andre601:
This is most likely due to how JDA *checks* if the attachment is an image.

https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/core/entities/Message.Attachment.html#isImage--
![image](https://user-images.githubusercontent.com/11576465/62201691-335a7080-b388-11e9-9bd1-8b1b1bd5bce3.png)

DV8FromTheWorld:
[...]

PR #1074, opened 2019-08-01
Improve isImage() check

caneleex:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: #1071

## Description

Improve isImage() check by using image pattern.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -1449,13 +1449,14 @@ public interface Message extends ISnowflake, Formattable
 
         /**
          * Whether or not this attachment is an Image.
-         * <br>Based on the values of getHeight and getWidth being larger than zero.
+         * <br>Based on the image pattern.
          *
-         * @return True if width and height are greater than zero.
+         * @return True if the file is an image
          */
         public boolean isImage()
         {
-            return height > 0 && width > 0;
+            private final Pattern image = Pattern.compile("([^\\s]+)(?:\\.(?i)(jpe?g|png|gif|bmp|webp))$");
+            return image.matcher(getFileName()).matches();
         }
     }
 }
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -1117,6 +1117,13 @@ public interface Message extends ISnowflake, Formattable
         private final int size;
         private final int height;
         private final int width;
+        
+        /**
+        * Pattern used to check if the file is an image.
+        *
+        * @see #isImage()
+        */
+        private final Pattern image = Pattern.compile("([^\\s]+)(?:\\.(?i)(jpe?g|png|gif|bmp|webp))$");
 
         private final JDAImpl jda;
 
@@ -1455,7 +1462,6 @@ public interface Message extends ISnowflake, Formattable
          */
         public boolean isImage()
         {
-            private final Pattern image = Pattern.compile("([^\\s]+)(?:\\.(?i)(jpe?g|png|gif|bmp|webp))$");
             return image.matcher(getFileName()).matches();
         }
     }


JohnnyJayJay:
I would make that pattern a constant of the `Message` interface or at least not make it a local variable, since it never changes. That way it wouldn't need to be compiled for every call and users may utilise it, too.

natanbc:
You can make it `static final`, there's no need to create a pattern for each instance.

Devoxin:
This will not work (reliably) if the attachment has had its file extension changed.



A way around this would be to check `width > -1` (or height, I guess. JDA returns -1 if the attachment is not an image, assuming JDA pulls this [information straight from the Message object itself](https://discordapp.com/developers/docs/resources/channel#attachment-object-attachment-structure)), along with a file extension check (or if you're feeling super confident, parse the image contents and detect file type).

Lastly, there are more image types out there than what the regex supports, unless this is only intended to support images that Discord can embed.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1117,6 +1117,13 @@ public Pattern getPattern()
         private final int size;
         private final int height;
         private final int width;
+        
+        /**
+        * Pattern used to check if the file is an image.
+        *
+        * @see #isImage()
+        */
+        private final Pattern image = Pattern.compile("([^\\s]+)(?:\\.(?i)(jpe?g|png|gif|bmp|webp))$");
```
MinnDevelopment: Why use a regex for this? This is trivial to implement without a regex and would probably be less error-prone than this.

```java
int index = name.lastIndexOf('.');
if (index + 1 == name.length())
    return false;
String ext = name.substring(index + 1);
return "png|webp|jpeg|jpg|gif".contains(ext);
```

Also I'm pretty sure bmp is not supported.

caneleex: You're right probably. I'll leave the decision up to you and other maintainers.

jagrosh:
As already stated, a regex would be a really slow solution to this problem. Also, the width and height should still be checked (in addition to whatever is used to separate images from video) to ensure that the image is actually being shown.

DV8FromTheWorld:
1. We should consider adding `Attachment#getFileExtension()` and use that in this PR
2. `Attachment#isImage()`should _first_ check if there are width / height values before doing the additional extension check.
3. The extension check should not use Regex.

caneleex:
Why would you all firstly check for width and height and not just the extension check? Isn't it kinda useless?

Devoxin:
Because as per the discord docs, width and height properties exist for attachments that are images. Checking those properties first would be a "less expensive" operation and then the extension check would serve as an extra step to ensure it's an image and not a video (although, I'm not sure if Discord sends width/height properties for videos so even the extension check might be useless here?)

JohnnyJayJay:
The extension check is not useless because extensions don't determine the file format, just how it's displayed by Discord.
I.e. if you do not check the extension, someone could upload a raw text file with a `.jpg` extension and this method would return true.
From what I understand, even with the extension check it's not 100% safe yet, since someone still could upload a video with an image extension. But it seems like the reasonably closest we can get.

Devoxin:
You can send a txt file disguised as a jpg/png/whatever and Discord will not send width/height properties for it (they will be null, or in JDA's case, `-1`), so this is not a problem.

However, I have just checked to see if the width/height properties exist for videos, and they do. 

So the width/height check just checks if the attachment is a video/image, and the extension check would serve to filter out videos. Though, it is worth mentioning that **videos disguised as images by changing the extension, will still have width/height properties**. End result is that this proposed check will still return a false positive in this case.

caneleex:
This is really confusing and I have no idea of reliable check.

DV8FromTheWorld:
At the end of the day, we can only provide as much security as we are given. We can ensure it is at least something that is visually displayable via the width and height checks, and we can _attempt_ to provider a further restriction by using file extension checking, but at the end of the day, if it really really matters

MinnDevelopment:
@caneleex looks like you deleted your branch, I assume you don't intend to continue working on this PR then? In that case could you close it so someone else can give it a shot.

Thank you for your contributions.

caneleex:
I think that this PR doesn't solve the reliability of the check and someone could've done it better; so yeah, I will close the PR and let someone other try to solve it. I also think this isn't very easy to fix as obviously there is no 100% reliable fix.

PR #1076, opened 2019-08-06
Add Guild#getMaxEmotes and BoostTier#getMaxEmotes

Shea4:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds BoostTier#getMaxEmotes and Guild#getMaxEmotes so you can check the max amount of emotes a guild can have based on their boost tier

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -410,6 +410,18 @@ public interface Guild extends ISnowflake
     }
 
     /**
+     * The maximum amount of emotes a guild can have.
+     *
+     * @return The maximum amount of emotes
+     *
+     * @since 4.0.0
+     */
+    default int getMaxEmotes()
+    {
+        return getBoostTier().getMaxEmotes();
+    }
+
+    /**
      * The maximum amount of members that can join this guild.
      *
      * @return The maximum amount of members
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -409,6 +409,18 @@ default int getMaxBitrate()
...
+    /**
+     * The maximum amount of emotes a guild can have.
+     *
+     * @return The maximum amount of emotes
+     *
+     * @since 4.0.0
+     */
+    default int getMaxEmotes()
+    {
+        return getBoostTier().getMaxEmotes();
```
MinnDevelopment: What about the `MORE_EMOJI` feature flag?

Shea4: I don't know how you would handle that because as far as I'm aware there's no way to tell how many more emojis the guild can use

MinnDevelopment: In that case this should at least be mentioned in the documentation since the maximum returned by this might not be accurate for guilds with that feature flag.

DV8FromTheWorld: is `MORE_EMOJI` documented? Maybe we could request a value for the max emojis it allows.

21Joakim: So I took a look into the client source code and they do it like 
`Math.max(getFeatures().contains("MORE_EMOJI")  ? EMOJI_MAX_SLOTS_MORE : EMOJI_MAX_SLOTS, getBoostTier().getMaxEmotes())` where `EMOJI_MAX_SLOTS_MORE` is 200 and `EMOJI_MAX_SLOTS` is 50

PR #1079, opened 2019-08-08
Release 4.0.0

MinnDevelopment:
This Pull Request is only symbolic since the release won't go through a pull request this time. We will be replacing the existing `master` branch with this one.

If you don't have a high end CPU I recommend not opening the changelist on this PR.


MinnDevelopment:
To get a better diff you can use

```sh
git diff -C1% origin/master origin/development -- \
    src/main/java/net/dv8tion/jda/api src/main/java/net/dv8tion/jda/core
```

> See this [gist output](https://gist.github.com/MinnDevelopment/b9de37b7d6e7a2cc12ef784778d7dbe3)

PR #1081, opened 2019-08-11
Implement X-RateLimit-Precision handling

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This implements the upcoming `X-RateLimit-Precision` header. Unfortunately discord doesn't echo this header so we have to wait for it to be released.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/Requester.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/Requester.java
@@ -260,7 +260,8 @@ public class Requester
     private void applyHeaders(Request<?> apiRequest, okhttp3.Request.Builder builder, boolean authorized)
     {
         builder.header("user-agent", USER_AGENT)
-               .header("accept-encoding", "gzip");
+               .header("accept-encoding", "gzip")
+               .header("x-ratelimit-precision", "milliseconds");
 
         //adding token to all requests to the discord api or cdn pages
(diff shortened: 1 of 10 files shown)


src/main/java/net/dv8tion/jda/internal/requests/Requester.java
```diff
@@ -260,7 +260,8 @@ private void applyBody(Request<?> apiRequest, okhttp3.Request.Builder builder)
     private void applyHeaders(Request<?> apiRequest, okhttp3.Request.Builder builder, boolean authorized)
     {
         builder.header("user-agent", USER_AGENT)
-               .header("accept-encoding", "gzip");
+               .header("accept-encoding", "gzip")
+               .header("x-ratelimit-precision", "milliseconds");
```
jhgg: `s/milliseconds/millisecond`

MinnDevelopment:
Waiting for this to be deployed so reactions don't send at lower rates.

MinnDevelopment:
We should expect users to have a proper ntp synchronized clock on their system.

caneleex:
Cloned the repo with this branch and it gives me instant NPE while starting the bot on the line 156 (parseDouble). Will add more details later tomorrow.

binaryoverload:
I'd guess that's because the particular headers are only in planning right now and are not in production?

caneleex:
Ooh I see. I'm dumb, thanks.

Issue #1086, opened 2019-08-19
MarkdownSanitizer#Escape() not escaping bolds correctly


HeyBanditoz:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

MarkdownSanitizer#Escape() is incorrectly escaping bolded letters. For example, `***Test***` would become `\***Test\***`

### Expected Behavior
`\*\*\*Test\*\*\*`

### Code Example or Reproduction Steps
```Java
public static void main(String[] args) {
        String message = "***Test***";
        System.out.println(MarkdownSanitizer.escape(message));
}
``` 
-> `\***Test\***`
### Exception or Error
N/A

V-ed:
I'll take the issue to try and fix it :

My initial guess is that this would be relatively the same fix as the one for UNDERLINE, but for a mixture of `BOLD | ITALICS_A` :

https://github.com/DV8FromTheWorld/JDA/blob/489ed4e222a937802d7f4341f0ac9ff1e9285f65/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java#L414-L421

You should probably rename the issue for "[...] not escaping bold italics correctly", as `***text***` is bold italic, not only bold.

This issue may also happen when trying to escape codeblocks (such as `` ```code block``` ``) - didn't try it yet, I'll see if my assumption is correct later today.

HeyBanditoz:
They don't get escaped if one of the characters is alone, eg `Test*Test`. Intended?

V-ed:
> They don't get escaped if one of the characters is alone, eg `Test*Test`. Intended?

Discord does not format the single `*`, so I believe it is intended. Italics formats uses the `_` char anyway, so there would be no problem here.

... but there would be no problem with the `*` character, the `_` char brings problems for italics. The following unit test actually fails :
```java
Assertions.assertEquals("_Hello \\_Test World_", italics("Hello _Test World"));
```

There may be other cases which it currently fails, I haven't tested further.

HeyBanditoz:
[...]

PR #1087, opened 2019-08-19
Fix BOLD | ITALICS_A not getting escaped properly in MarkdownSanitizer (#1086)

V-ed:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1086

## Description

Fixes the `MarkdownSanitizer`'s `escape` method not properly escaping BOLD | ITALICS_A mixture of text (ie: `***text***`).

~~I did not test the changes, I will look how can I test later today.~~ 

~~I also plan to fix a very similar issue that I think might happen with code blocks, I will look into that later today too, making this PR a WIP one. I will therefore create a new commit, which I could squash them if preferable on any maintainer's request.~~ 

_Also, there might be a need to change the README's Contributing guidelines, as there is no `v4` branch anymore... :_ 
```
If you want to contribute to JDA, make sure to base your branch off of our v4 branch (or a feature-branch) and create your PR into that same branch. We will be rejecting any PRs between branches or into release branches!
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
@@ -416,6 +416,8 @@ public class MarkdownSanitizer
             throw new IllegalStateException("Found illegal region for strategy ESCAPE '" + region + "' with no known format token!");
         if (region == UNDERLINE)
             token = "_\\_"; // UNDERLINE needs special handling because the client thinks its ITALICS_U if you only escape once
+        else if (region == BOLD)
+            token = "*\\*"; // BOLD needs special handling because the client thinks its ITALICS_A if you only escape once
         else if (region == (BOLD | ITALICS_A))
             token = "*\\*\\*"; // BOLD | ITALICS_A needs special handling because the client thinks its BOLD if you only escape once
         builder.append("\\").append(token)
--- a/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
@@ -416,6 +416,8 @@ public class MarkdownSanitizer
             throw new IllegalStateException("Found illegal region for strategy ESCAPE '" + region + "' with no known format token!");
         if (region == UNDERLINE)
             token = "_\\_"; // UNDERLINE needs special handling because the client thinks its ITALICS_U if you only escape once
+        else if (region == (BOLD | ITALICS_A))
+            token = "*\\*\\*"; // BOLD | ITALICS_A needs special handling because the client thinks its BOLD if you only escape once
         builder.append("\\").append(token)
                .append(seq)
                .append("\\").append(token);
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -101,7 +101,7 @@ val javadoc: Javadoc by tasks
 val jar: Jar by tasks
 val build: Task by tasks
 val clean: Task by tasks
-val test: Task by tasks
+val test: Test by tasks
 val check: Task by tasks
 
 shadowJar.classifier = "withDependencies"
@@ -257,6 +257,11 @@ bintrayUpload.apply {
     onlyIf { getProjectProperty("bintrayUsername").isNotEmpty() }
     onlyIf { getProjectProperty("bintrayApiKey").isNotEmpty() }
     onlyIf { System.getenv("BUILD_NUMBER") != null }
+}
+
+test.apply {
+    useJUnitPlatform()
+    failFast = true
 }
 
 publishing {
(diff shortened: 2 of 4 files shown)


V-ed:
I will also look at adding / fixing tests later today, will update here accordingly.

V-ed:
Hey @MinnDevelopment , I've got a question about your test suite - is your build step includes your unit test? I'm not used to gradle and I can't seem to make the tests fail, even when basically doing a assertTrue(false), and the build says it succeeded but one of the test should actually fail :

https://github.com/DV8FromTheWorld/JDA/blob/489ed4e222a937802d7f4341f0ac9ff1e9285f65/src/test/java/MarkdownTest.java#L283-L289

The first Assertion above should fail with my branch, as I fixed the escape function to be what is intended. I ran a manual test and indeed, this now works :
```java
public static void main(String[] args){
    String message = "***Test***";
    System.out.println(MarkdownSanitizer.escape(message));
}

>>  \*\*\*Test\*\*\*

**Comment by MinnDevelopment** [comment]
@V-ed I currently only run tests from intellij but they can be enabled in gradle with:

```kotlin
val test: Test by tasks

test.apply {
    useJUnitPlatform()
    failFast = true
}
```

Here is a patch you can apply:

```diff
diff --git a/build.gradle.kts b/build.gradle.kts
index bfcb0e85..df58c7aa 100644
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -101,7 +101,7 @@ val javadoc: Javadoc by tasks
 val jar: Jar by tasks
 val build: Task by tasks
 val clean: Task by tasks
-val test: Task by tasks
+val test: Test by tasks
 val check: Task by tasks
 
 shadowJar.classifier = "withDependencies"
@@ -259,6 +259,11 @@ bintrayUpload.apply {
     onlyIf { System.getenv("BUILD_NUMBER") != null }
 }
 
+test.apply {
+    useJUnitPlatform()
+    failFast = true
+}
+
 publishing {
     publications {
         register("BintrayRelease", MavenPublication::class) {
```
You can apply the patch through `git apply` or with intellij **VCS** | **Apply Patch from Clipboard**

V-ed:
After some testing, I concluded that there is no similar issues with code blocks (Discord formats the string `` \```text\``` `` as `` ```text``` `` correctly).

Althought, I knew there would be a problem with escaping every format as Discord's Markdown is a bit more convoluted than this, and indeed, the `monospace` function suffers from the lack of a proper escape mecanism :

image

as the second Assertion here would infer : https://github.com/DV8FromTheWorld/JDA/blob/489ed4e222a937802d7f4341f0ac9ff1e9285f65/src/test/java/MarkdownUtilTest.java#L48-L54

Should I also try to fix the monospace issue in this PR or would this be another scope, therefore another pull request?

MinnDevelopment:
The monospace escape is working as intended, there is no reliable way of escaping it and this is an acceptable margin of error.

V-ed:
> The monospace escape is working as intended, there is no reliable way of escaping it and this is an acceptable margin of error.

I actually beg to differ : there is a somewhat reliable way of escaping it, as I did in my own project, althought it's more of a workaround than a proper fix, and I do not believe that the image I've shown (and added below as well) is an acceptable margin of error, as it seriously degrades the actual format when formatting monospace with backticks in them.

image

As you can see below, that is what I would expect to see, which I believe *should* be the way to go, in expense of making the escaping mechanism a bit more complex :
image

... but I think this may be out of scope for this PR and that while there may be work on this later (at least, I hope), this PR here may be merged if accepted of course.

MinnDevelopment:
> The monospace escape is working as intended

src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
```diff
@@ -416,6 +416,8 @@ private void applyStrategy(int region, @Nonnull String seq, @Nonnull StringBuild
             throw new IllegalStateException("Found illegal region for strategy ESCAPE '" + region + "' with no known format token!");
         if (region == UNDERLINE)
             token = "_\\_"; // UNDERLINE needs special handling because the client thinks its ITALICS_U if you only escape once
+        else if (region == (BOLD | ITALICS_A))
+            token = "*\\*\\*"; // BOLD | ITALICS_A needs special handling because the client thinks its BOLD if you only escape once
```
MinnDevelopment: This probably also needs to be done for simple bold `\*\*` rather than `\**`

V-ed: Good catch, fixed - a few more tests were affected by this change, I tweaked them [in my latest commit](https://github.com/DV8FromTheWorld/JDA/pull/1087/commits/8f245bb7b3ef36f05c67fd67e7e4627ce7124560) to be more accurate on what's supposed to come out.

Issue #1091, opened 2019-08-31
Incorrect nullability expectation


hasan-imam:
## Bug Report

`getJumpUrl()` in `ReceivedMessage` calls `getGuild()` and expects it to return null for direct messages [[code](https://github.com/DV8FromTheWorld/JDA/blob/19acdd57a88c2b93c1c61f76ba59ddfb849a6d45/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java#L177)]. However, `getGuild()` never returns null. It throws when the message isn't a DM.

### Expected Behavior

`getJumpUrl()` should work for direct/private messages. 

### Code Example or Reproduction Steps

Call `getJumpUrl()` on a private message.

### Exception or Error

```
java.lang.IllegalStateException: This message was not sent in a text channel
	at net.dv8tion.jda.internal.entities.ReceivedMessage.getTextChannel(ReceivedMessage.java:570)
	at net.dv8tion.jda.internal.entities.ReceivedMessage.getGuild(ReceivedMessage.java:584)
	at net.dv8tion.jda.internal.entities.ReceivedMessage.getJumpUrl(ReceivedMessage.java:177)
```

PR #1092, opened 2019-08-31
Fix getJumpUrl() in ReceivedMessage for DMs

kantenkugel:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1091

## Description

Fixes a bug with the new Nullability handling and `ReceivedMessage` using `getGuild()` in its `getJumpUrl()` implementation

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
@@ -174,7 +174,7 @@ public class ReceivedMessage extends AbstractMessage
     @Override
     public String getJumpUrl()
     {
-        return String.format("https://discordapp.com/channels/%s/%s/%s", getGuild() == null ? "@me" : getGuild().getId(), getChannel().getId(), getId());
+        return String.format("https://discordapp.com/channels/%s/%s/%s", isFromType(ChannelType.PRIVATE) ? "@me" : getGuild().getId(), getChannel().getId(), getId());
     }
 
     private User matchUser(Matcher matcher)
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -174,7 +174,7 @@ public long getIdLong()
     @Override
     public String getJumpUrl()
     {
-        return String.format("https://discordapp.com/channels/%s/%s/%s", getGuild() == null ? "@me" : getGuild().getId(), getChannel().getId(), getId());
+        return String.format("https://discordapp.com/channels/%s/%s/%s", isFromType(ChannelType.PRIVATE) ? "@me" : getGuild().getId(), getChannel().getId(), getId());
```
hasan-imam: Wouldn't this throw null pointer exception when type is `GROUP`? How about accessing guild if `getChannelType().isGuild()` is `true`?

kantenkugel: groups are not supported in v4 anymore and were probably even removed

kantenkugel: Yea, Groups were removed, just not the ChannelType enum constant.

kantenkugel: I might still go for isGuild() like suggested tho, considering theres also Store (which hopefully doesn't receive messages, but you never know)

MinnDevelopment: Definitely seems like `isFromGuild` is more appropriate/clear

Issue #1094, opened 2019-09-05
MessageReaction.isSelf() does not work correctly


tzann:
## Bug Report

MessageReaction.isSelf() returns `true` only if the reaction was created by the self user, but not if the self user has reacted with the same reaction.

### Expected Behavior

The documentation in `src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java` states that MessageReaction.isSelf() returns true if the self user has reacted to the message with that reaction, not that the reacting user must be the self user.

### Code Example or Reproduction Steps

In an EventListener:
```
if (event instanceof MessageReactionAddEvent) {
    System.out.println("isSelf(): " + ((MessageReactionAddEvent) event).getReaction().isSelf());  
}
```
This will return true if the bot reacts to a message itself, but false if someone else then reacts to the same message with the same reaction.

### Exception or Error

N/A

MinnDevelopment:
We should probably do the same as the `getCount()` method and throw an exception in events.

AtriusX:
I believe this might be the source of the issue. 
https://github.com/DV8FromTheWorld/JDA/blob/1a97dd9367ab5c61ea452134770e1d5a325be891/src/main/java/net/dv8tion/jda/internal/handle/MessageReactionHandler.java#L128

The `EntityBuilder` class references the "me" field in the api when building a reaction; https://github.com/DV8FromTheWorld/JDA/blob/1a97dd9367ab5c61ea452134770e1d5a325be891/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java#L966 
however, no such call is made in the `MessageReactionHandler` class from what I can tell.

MinnDevelopment:
The api doesn't provide this information on events

AtriusX:
Ah ok, there's probably not an easy way to get the data without checking the message for each event then I guess.

PR #1098, opened 2019-09-15
Patch for isSelf returning incorrect values

AtriusX:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1094 

## Description

This is just a quick fix for the issue with the `MessageReaction#isSelf` method. Since the add/remove events don't supply the proper information directly, this is the only workaround I could think of.


MinnDevelopment:
This is unacceptable. You cannot make a REST request for every reaction event.

PR #1108, opened 2019-09-30
Create constant for max slowmode

duncte123:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This pull request adds a field for the maximum slowmode that a text channel can have.
I also added an extra check to the channel action class making it throw if the slowmode was not set on a text channel.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/requests/restaction/ChannelAction.java
+++ b/src/main/java/net/dv8tion/jda/api/requests/restaction/ChannelAction.java
@@ -166,8 +166,8 @@ public interface ChannelAction<T extends GuildChannel> extends AuditableRestActi
      * @param  slowmode
      *         The number of seconds required to wait between sending messages in the channel.
      *
-     * @throws IllegalStateException
-     *         If the selected {@link net.dv8tion.jda.api.entities.GuildChannel GuildChannel}'s type is not {@link net.dv8tion.jda.api.entities.ChannelType#TEXT TEXT}
+     * @throws UnsupportedOperationException
+     *         If this ChannelAction is not for a TextChannel
      * @throws IllegalArgumentException
      *         If the {@code slowmode} is greater than 21600, or less than 0
      *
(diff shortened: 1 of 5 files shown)


caneleex:
Shouldn't constants be `public static final`?

duncte123:
> Shouldn't constants be `public static final`?

TextChannel is an interface so this is not needed

src/main/java/net/dv8tion/jda/internal/requests/restaction/ChannelActionImpl.java
```diff
@@ -146,7 +146,9 @@ public ChannelType getType()
     @CheckReturnValue
     public ChannelActionImpl<T> setSlowmode(int slowmode)
     {
-        Checks.check(slowmode <= 21600 && slowmode >= 0, "Slowmode must be between 0 and 21600 (seconds)!");
+        if (type != ChannelType.TEXT)
+            throw new IllegalStateException("Can only set slowmode on text channels");
```
MinnDevelopment: Is this exception documented anywhere?

src/main/java/net/dv8tion/jda/internal/requests/restaction/ChannelActionImpl.java
```diff
@@ -146,7 +146,9 @@ public ChannelType getType()
...
+        Checks.check(slowmode <= TextChannel.MAX_SLOWMODE && slowmode >= 0, "Slowmode must be between 0 and %s (seconds)!", TextChannel.MAX_SLOWMODE);
```
MinnDevelopment: This should be `%d` not `%s`.

src/main/java/net/dv8tion/jda/internal/requests/restaction/ChannelActionImpl.java
duncte123: Here https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/api/requests/restaction/ChannelAction.html#setSlowmode(int%29

src/main/java/net/dv8tion/jda/api/requests/restaction/ChannelAction.java
```diff
@@ -166,6 +166,8 @@
      * @param  slowmode
      *         The number of seconds required to wait between sending messages in the channel.
      *
+     * @throws UnsupportedOperationException
+     *         If this ChannelAction is not for a TextChannel
      * @throws IllegalArgumentException
      *         If the {@code slowmode} is greater than 21600, or less than 0
```
MinnDevelopment: You should use an `{@link TextChannel.MAX_SLOWMODE}` here.

PR #1112, opened 2019-10-01
Remove Voice channel from JDA cache on deletion

duncte123:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

When a voice channel is deleted on discord the handler only ever deleted it from the guild cache and not from the JDA cache. This caused an exception being thrown when getting the guild from a voice channel via the JDA cache because the guild was garbage collected.
This pull request fixes that by also deleting the voice channel from the JDA cache.


MinnDevelopment:
I fixed this a year ago but it seems to have been re-introduced by v4 somewhere along the way.

caneleex:
This is strange as i forked `lazy-loading` branch before v4 has been released and this bug seems to be fixed in the lazy loading branch instead i guess?

PR #1114, opened 2019-10-01
Fix for removing an emoji reaction in codepoint style

DManstrator:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds a missing ("breaking") check on the codepoint variant in `MessageChannel#removeReactionById` which lead to an `ErrorResponseException: 10014: Unknown Emoji` experienced by MarRue#3658 first. It also makes the code more uniform in terms of `addReactionById` and `removeReactionById`.

#### The Problem
```JAVA
default RestAction<Void> addReactionById(@Nonnull String messageId, @Nonnull String unicode)
{
	Checks.isSnowflake(messageId, "Message ID");
	Checks.notNull(unicode, "Provided Unicode");
	unicode = unicode.trim();
	Checks.notEmpty(unicode, "Provided Unicode");

	String encoded;
	if (unicode.startsWith("U+"))  // checking for codepoint style
		encoded = EncodingUtil.encodeCodepointsUTF8(unicode);
	else
		encoded = EncodingUtil.encodeUTF8(unicode);
	Route.CompiledRoute route = Route.Messages.ADD_REACTION.compile(getId(), messageId, encoded, "@me");
	// ...
}
```
```JAVA
default RestAction<Void> removeReactionById(@Nonnull String messageId, @Nonnull String unicode)
{
	Checks.isSnowflake(messageId, "Message ID");
	Checks.noWhitespace(unicode, "Emoji");

	final String code = EncodingUtil.encodeUTF8(unicode);  // missing check if (unicode.startsWith("U+"))
	final Route.CompiledRoute route = Route.Messages.REMOVE_REACTION.compile(getId(), messageId, code, "@me");
	// ...
}
```

#### Removing Code Redundancy
Before, the `removeReactionById` logic was both in the `MessageChannel` and the `TextChannelImpl` leading to code redundancy. Now the method in the `MessageChannel` calls the method in the `TextChannelImpl`. **This is imo a bad design and I would like to get feedback on that because within 3 hours, I couldn't come up with a better approach.**

#### Making the code more uniform
Before, all four methods of `MessageChannel#removeReactionById` were leading to one implementation. However for adding reactions, there were two (or three) small implementations. For this PR, I also made it that all four methods are leading to one implementation as well.

With removing the code redundancy, there is now **just one implementation for both adding and removing reactions** making it more maintainable. 

**Also a point to discuss would be if `noWhitespace` or `notNull` with `notEmpty` should be used for adding and removing** since `addReaction("\uXXXX ")` will work currently but `removeReaction("\uXXXX ")` won't (approved by Kantenkugel). **I'm unsure what is better (where I am preferring `noWhitespace`) and so is Kantenkugel so I would like to hear your opinions on that.**

(Nearly) all of this has already been discussed in the JDA-Server with Kantenkugel starting [here](https://discordapp.com/channels/125227483518861312/125227483518861312/628315511654055959).

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
@@ -1773,11 +1773,8 @@ public interface MessageChannel extends ISnowflake, Formattable
         unicode = unicode.trim();
         Checks.notEmpty(unicode, "Provided Unicode");
 
-        String encoded;
-        if (unicode.startsWith("U+") || unicode.startsWith("u+"))
-            encoded = EncodingUtil.encodeCodepointsUTF8(unicode);
-        else
-            encoded = EncodingUtil.encodeUTF8(unicode);
+        final String encoded = EncodingUtil.encodeReaction(unicode);
+
         Route.CompiledRoute route = Route.Messages.ADD_REACTION.compile(getId(), messageId, encoded, "@me");
         return new RestActionImpl<>(getJDA(), route);
     }
@@ -2044,11 +2041,7 @@ public interface MessageChannel extends ISnowflake, Formattable
         unicode = unicode.trim();
         Checks.notEmpty(unicode, "Provided Unicode");
 
-        String encoded;
-        if (unicode.startsWith("U+") || unicode.startsWith("u+"))
-            encoded = EncodingUtil.encodeCodepointsUTF8(unicode);
-        else
-            encoded = EncodingUtil.encodeUTF8(unicode);
+        final String encoded = EncodingUtil.encodeReaction(unicode);
 
         final Route.CompiledRoute route = Route.Messages.REMOVE_REACTION.compile(getId(), messageId, encoded, "@me");
         return new RestActionImpl<>(getJDA(), route);
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
@@ -1772,6 +1772,7 @@ public interface MessageChannel extends ISnowflake, Formattable
         Checks.notNull(unicode, "Provided Unicode");
         unicode = unicode.trim();
         Checks.notEmpty(unicode, "Provided Unicode");
+        // TODO Maybe checking for Checks.noWhitespace(unicode, "Unicode emoji") instead (in terms of consistency / making it uniform)
 
         String encoded;
         if (unicode.startsWith("U+"))
@@ -1916,11 +1917,8 @@ public interface MessageChannel extends ISnowflake, Formattable
     @CheckReturnValue
     default RestAction<Void> addReactionById(@Nonnull String messageId, @Nonnull Emote emote)
     {
-        Checks.isSnowflake(messageId, "Message ID");
         Checks.notNull(emote, "Emote");
-
-        Route.CompiledRoute route = Route.Messages.ADD_REACTION.compile(getId(), messageId, String.format("%s:%s", emote.getName(), emote.getId()), "@me");
-        return new RestActionImpl<Void>(getJDA(), route);
+        return addReactionById(messageId, emote.getName() + ":" + emote.getId());
     }
 
     /**
@@ -2042,12 +2040,8 @@ public interface MessageChannel extends ISnowflake, Formattable
     @CheckReturnValue
     default RestAction<Void> removeReactionById(@Nonnull String messageId, @Nonnull String unicode)
     {
-        Checks.isSnowflake(messageId, "Message ID");
-        Checks.noWhitespace(unicode, "Emoji");
-
-        final String code = EncodingUtil.encodeUTF8(unicode);
-        final Route.CompiledRoute route = Route.Messages.REMOVE_REACTION.compile(getId(), messageId, code, "@me");
-        return new RestActionImpl<>(getJDA(), route);
+        // removes code redundancy but kind of bad design
+        return ((TextChannel) this).removeReactionById(messageId, unicode, getJDA().getSelfUser());
     }
 
     /**
@@ -2216,8 +2210,7 @@ public interface MessageChannel extends ISnowflake, Formattable
     @CheckReturnValue
     default RestAction<Void> removeReactionById(long messageId, @Nonnull Emote emote)
     {
-        Checks.notNull(emote, "Emote");
-        return removeReactionById(messageId, emote.getName() + ":" + emote.getId());
+        return removeReactionById(Long.toUnsignedString(messageId), emote);
     }
 
     /**
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
@@ -1772,10 +1772,9 @@ public interface MessageChannel extends ISnowflake, Formattable
         Checks.notNull(unicode, "Provided Unicode");
         unicode = unicode.trim();
         Checks.notEmpty(unicode, "Provided Unicode");
-        // TODO Maybe checking for Checks.noWhitespace(unicode, "Unicode emoji") instead (in terms of consistency / making it uniform)
 
         String encoded;
-        if (unicode.startsWith("U+"))
+        if (unicode.startsWith("U+") || unicode.startsWith("u+"))
             encoded = EncodingUtil.encodeCodepointsUTF8(unicode);
         else
             encoded = EncodingUtil.encodeUTF8(unicode);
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -1772,6 +1772,7 @@ default MessagePaginationAction getIterableHistory()
         Checks.notNull(unicode, "Provided Unicode");
         unicode = unicode.trim();
         Checks.notEmpty(unicode, "Provided Unicode");
+        // TODO Maybe checking for Checks.noWhitespace(unicode, "Unicode emoji") instead (in terms of consistency / making it uniform)
```
MinnDevelopment: No this is intended, to allow things like `addReaction("U+AABB U+CCDD")`

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -2042,12 +2040,8 @@ default MessagePaginationAction getIterableHistory()
...
     default RestAction<Void> removeReactionById(@Nonnull String messageId, @Nonnull String unicode)
     {
-        Checks.isSnowflake(messageId, "Message ID");
-        Checks.noWhitespace(unicode, "Emoji");
-
-        final String code = EncodingUtil.encodeUTF8(unicode);
-        final Route.CompiledRoute route = Route.Messages.REMOVE_REACTION.compile(getId(), messageId, code, "@me");
-        return new RestActionImpl<>(getJDA(), route);
+        // removes code redundancy but kind of bad design
+        return ((TextChannel) this).removeReactionById(messageId, unicode, getJDA().getSelfUser());
```
MinnDevelopment: What if this is not a `TextChannel`?

src/main/java/net/dv8tion/jda/internal/entities/TextChannelImpl.java
```diff
@@ -458,14 +458,23 @@ public MessageAction sendFile(@Nonnull InputStream data, @Nonnull String fileNam
...
         Checks.noWhitespace(unicode, "Unicode emoji");
         Checks.notNull(user, "User");
+
         if (!getJDA().getSelfUser().equals(user))
             checkPermission(Permission.MESSAGE_MANAGE);
-        final String code = EncodingUtil.encodeUTF8(unicode);
-        Route.CompiledRoute route;
+
+        String encoded;
+        if (unicode.startsWith("U+"))
```
MinnDevelopment: It's probably better to also check for lowercase version.

src/main/java/net/dv8tion/jda/internal/entities/TextChannelImpl.java
```diff
@@ -458,14 +458,23 @@ public MessageAction sendFile(@Nonnull InputStream data, @Nonnull String fileNam
         Checks.isSnowflake(messageId, "Message ID");
         Checks.noWhitespace(unicode, "Unicode emoji");
         Checks.notNull(user, "User");
+
```
MinnDevelopment: The `noWhitespace` check should be changed, see other comment above

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: For adding multiple reactions at once? Thought you don't allow that. Or is it like 2 concatenated unicodes `\uD83D\uDCFA` for adding like 📺?

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: Well. I thought it has to be castable since TextChannel implements MessageChannel. Isn't this the case?

I know that it was at least bad design but I couldn't come up with a better approach, do you have any ideas how to get rid of the code redundancy here?

src/main/java/net/dv8tion/jda/internal/entities/TextChannelImpl.java
DManstrator: Good catch, will replace it with `unicode.toLowerCase().startsWith("u+")` so no second if is needed.

src/main/java/net/dv8tion/jda/internal/entities/TextChannelImpl.java
DManstrator: Gotcha.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
MinnDevelopment: You can't cast a `PrivateChannel` to a `TextChannel` they both implement this interface.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
MinnDevelopment: The latter

src/main/java/net/dv8tion/jda/internal/entities/TextChannelImpl.java
MinnDevelopment: It would be more efficient to just use a logical or than to lowercase the entire string

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: So to stay on this concrete example:
Both `addReaction("U+1F4FA")` and `addReaction("U+D83D U+DCFA")` would add the reaction correctly? Because I never saw the codepoints used like that.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
DManstrator: Just tried it myself really quick.
```
Exception in thread "main" In MessageChannel
java.lang.ClassCastException: implementscheck.PrivateChannel cannot be cast to implementscheck.TextChannel
```
You're totally right ofc. Somehow I thought that only happens if you try to cast something like a String to a Double (nothing in common like an abstract class or an interface). Somehow I thought it would be okay since both classes are implementing MessageChannel. :man_facepalming:

The problem is: The method in the TextChannel can also be with another user so simply calling MessageChannel's default method won't work. Also no private helping functions are allowed in an Interface. Guess it has to be implemented twice then, hasn't it?

src/main/java/net/dv8tion/jda/internal/entities/TextChannelImpl.java
DManstrator: Gotcha. Just tried to avoid another check since I thought the string would be at most like 6 chars so lowering it wouldn't matter too much.

DManstrator:
Fixed all the things. I still don't like that there are now like 3 implementations [here](https://github.com/DV8FromTheWorld/JDA/blob/7ac975b876eceec730ab021ddc9abaa53be33715/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java#L1769-L1783), [here](https://github.com/DV8FromTheWorld/JDA/blob/7ac975b876eceec730ab021ddc9abaa53be33715/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java#L2040-L2055) and [here](https://github.com/DV8FromTheWorld/JDA/blob/7ac975b876eceec730ab021ddc9abaa53be33715/src/main/java/net/dv8tion/jda/internal/entities/TextChannelImpl.java#L456-L481) for basically the same thing.

Maybe a helper method (where?) which can take a msgId, a Unicode / String, a boolean (for adding / removing) and a User?
```JAVA
public RestAction<Void> handleReactionById(@Nonnull String messageId, @Nonnull String unicode, boolean add)
{
    return handleReactionById(messageId, unicode, add, getJDA().getSelfUser());
}

public RestAction<Void> handleReactionById(@Nonnull String messageId, @Nonnull String unicode, boolean add, @Nonnull User user)
{
    Checks.isSnowflake(messageId, "Message ID");
    Checks.notNull(unicode, "Provided Unicode");
    unicode = unicode.trim();
    Checks.notEmpty(unicode, "Provided Unicode");
    Checks.notNull(user, "User");

    if (!getJDA().getSelfUser().equals(user))
        checkPermission(Permission.MESSAGE_MANAGE);

    String encoded;
    if (unicode.startsWith("U+") || unicode.startsWith("u+"))
        encoded = EncodingUtil.encodeCodepointsUTF8(unicode);
    else
        encoded = EncodingUtil.encodeUTF8(unicode);

    String targetUser;
    if (user.equals(getJDA().getSelfUser()))
        targetUser = "@me";
    else
        targetUser = user.getId();

    Route wayToGo;
    if (add)  {
        wayToGo = Route.Messages.ADD_REACTION;
    }  else  {
        wayToGo = Route.Messages.REMOVE_REACTION;
    }

    final Route.CompiledRoute route = wayToGo.compile(getId(), messageId, encoded, targetUser);
    return new RestActionImpl<>(getJDA(), route);
}
```

Or something like this. Just want to get rid of the code redundancy. Or do you have a better approach?

src/main/java/net/dv8tion/jda/internal/entities/TextChannelImpl.java
```diff
@@ -456,16 +456,27 @@ public MessageAction sendFile(@Nonnull InputStream data, @Nonnull String fileNam
...
         if (!getJDA().getSelfUser().equals(user))
             checkPermission(Permission.MESSAGE_MANAGE);
-        final String code = EncodingUtil.encodeUTF8(unicode);
-        Route.CompiledRoute route;
+
+        String encoded;
+        if (unicode.startsWith("U+") || unicode.startsWith("u+"))
+            encoded = EncodingUtil.encodeCodepointsUTF8(unicode);
+        else
+            encoded = EncodingUtil.encodeUTF8(unicode);
```
MinnDevelopment: We should move this if/else into `EncodingUtil` as something like `encodeReaction(String)`

DManstrator:
Sorry for the late response, it's done now. But I would still like to hear your opinion on that `handleReactionById` function I provided in the comments above. With that we only have one main implementation making it even more maintainable.

PR #1116, opened 2019-10-02
Add Member#getActiveClients()

kantenkugel:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

This adds the method `Member#getActiveClients() : Set<ClientType>` which returns all ClientTypes the Member is not Offline or with Unknown status with

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Member.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Member.java
@@ -26,6 +26,7 @@ import javax.annotation.Nullable;
 import java.awt.Color;
 import java.time.OffsetDateTime;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Represents a Guild-specific User.
@@ -128,6 +129,21 @@ public interface Member extends IMentionable, IPermissionHolder
      */
     @Nonnull
     OnlineStatus getOnlineStatus(@Nonnull ClientType type);
+
+    /**
+     * A Set of all active {@link net.dv8tion.jda.api.entities.ClientType ClientTypes} of this Member.
+     * Every {@link net.dv8tion.jda.api.OnlineStatus OnlineStatus} other than {@code OFFLINE} and {@code UNKNOWN}
+     * is interpreted as active.
+     * <br>If the Member is currently not active with any Client, this returns an empty Set.
+     * <br>Since a user can be connected from multiple different devices such as web and mobile,
+     * discord specifies a status for each {@link net.dv8tion.jda.api.entities.ClientType}.
+     *
+     * @return Immutable Set of all active {@link net.dv8tion.jda.api.entities.ClientType ClientTypes}
+     *
+     * @since  4.0.0
+     */
+    @Nonnull
+    Set<ClientType> getActiveClients();
 
     /**
      * Returns the current nickname of this Member for the parent Guild.
--- a/src/main/java/net/dv8tion/jda/api/entities/Member.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Member.java
@@ -25,8 +25,8 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.awt.Color;
 import java.time.OffsetDateTime;
+import java.util.EnumSet;
 import java.util.List;
-import java.util.Set;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/entities/Member.java
```diff
@@ -129,6 +130,21 @@
...
+     * <br>If the Member is currently not active with any Client, this returns an empty Set.
+     * <br>Since a user can be connected from multiple different devices such as web and mobile,
+     * discord specifies a status for each {@link net.dv8tion.jda.api.entities.ClientType}.
+     *
+     * @return Immutable Set of all active {@link net.dv8tion.jda.api.entities.ClientType ClientTypes}
+     *
+     * @since  4.0.0
+     */
+    @Nonnull
+    Set<ClientType> getActiveClients();
```
MinnDevelopment: This should return `EnumSet<ClientType>`

src/main/java/net/dv8tion/jda/api/entities/Member.java
```diff
@@ -129,6 +130,21 @@
     @Nonnull
     OnlineStatus getOnlineStatus(@Nonnull ClientType type);
 
+    /**
+     * A Set of all active {@link net.dv8tion.jda.api.entities.ClientType ClientTypes} of this Member.
+     * Every {@link net.dv8tion.jda.api.OnlineStatus OnlineStatus} other than {@code OFFLINE} and {@code UNKNOWN}
+     * is interpreted as active.
+     * <br>If the Member is currently not active with any Client, this returns an empty Set.
```
MinnDevelopment: Mention `INVISIBLE` here as well since some people might think its tracked because its present in the enum

src/main/java/net/dv8tion/jda/api/entities/Member.java
```diff
@@ -129,6 +130,21 @@
...
+     * <br>Since a user can be connected from multiple different devices such as web and mobile,
+     * discord specifies a status for each {@link net.dv8tion.jda.api.entities.ClientType}.
+     *
+     * @return Immutable Set of all active {@link net.dv8tion.jda.api.entities.ClientType ClientTypes}
```
MinnDevelopment: This doesn't need to be immutable (and can't be for an `EnumSet` return type).

src/main/java/net/dv8tion/jda/api/entities/Member.java
MinnDevelopment: Also it should be pointed out that this will always be empty if the `CacheFlag.CLIENT_STATUS` is disabled.

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
```diff
@@ -127,6 +127,21 @@ public OnlineStatus getOnlineStatus(@Nonnull ClientType type)
...
+    public Set<ClientType> getActiveClients()
+    {
+        EnumSet<ClientType> activeClients = EnumSet.noneOf(ClientType.class);
+        this.clientStatus.forEach((client, status) ->
+        {
+            if (status != OnlineStatus.OFFLINE && status != OnlineStatus.UNKNOWN)
+            {
+                activeClients.add(client);
+            }
+        });
```
MinnDevelopment: `OFFLINE` and `UNKNOWN` should technically never be in this map, we could enforce that in the `PresenceUpdateHandler` and simplify this to `EnumSet.copyOf(clientStatus.keySet())`.

MinnDevelopment:
Any update on this?

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
kantenkugel: The `OFFLINE`/`UNKNOWN` check seems to be already handled in `Member#setOnlineStatus(...)`:

https://github.com/DV8FromTheWorld/JDA/blob/feature/member_active_clients/src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java#L292-L301

kantenkugel:
Sorry for the wait. Should be done now

PR #1121, opened 2019-10-09
Promote commons-collections4 to an API dependency (#1120)

### LINKED ISSUE #1120: commons-collections4 should be api dependency
In your build script you define `implementation("org.apache.commons:commons-collections4:4.1")`.
But `commons-collections4` is used in your API (e. g. as return type of several methods in `Message`).
Due to that it should instead be declared as `api("org.apache.commons:commons-collections4:4.1")`.

This is not only a philosophical change, this is how `api` and `implementation` are meant to be used.
And for me it also has a practical impact. (That I of course can work-around easily by including the lib on my compile classpath manually, but I shouldn't need to)

Due to it being defined as `implementation` dependency it lands as `runtime`-scoped dependency in the published POM and thus is not part of the compile classpath of Gradle 5+ consumers.

Now I have a project that is a consumer and have a Spock test that mocks a method call of `Message`.
This causes the test compilation to fail as the Spock Groovy AST transformation tries to get the methods of `Message` and fails with a `NoClassDefFoundError` as `Bag` is not on the compile classpath.

Vampire:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1120

## Description

Promotes the commons-collections4 dependency to API scope to properly provide bags in the message interface.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -83,7 +83,7 @@ dependencies {
     /* Internal dependencies */
 
     //General Utility
-    implementation("org.apache.commons:commons-collections4:4.1")
+    api("org.apache.commons:commons-collections4:4.1")
     implementation("net.sf.trove4j:trove4j:3.0.3")
     implementation("com.fasterxml.jackson.core:jackson-databind:2.9.8")
 


MinnDevelopment:
Refer to [Contributing Guidelines](https://github.com/DV8FromTheWorld/JDA/blob/master/.github/CONTRIBUTING.md) please base your changes on development.

Vampire:
rebase is pushed, sorry

PR #1124, opened 2019-10-16
Sanitize text channel names

JakeJMattson:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Automatically sanitize spaces from text channel names, and replace them with dashes (`-`) to comply with Discord channel naming rules.

- `GuildImpl#createTextChannel`
- `ChannelManagerImpl#setName`

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
@@ -1095,7 +1095,7 @@ public class GuildImpl implements Guild
     {
         checkPermission(Permission.MANAGE_CHANNEL);
         Checks.notBlank(name, "Name");
-        name = name.trim();
+        name = Helpers.sanitizeChannelName(name.trim());
 
         Checks.check(name.length() > 0 && name.length() <= 100, "Provided name must be 1 - 100 characters in length");
         return new ChannelActionImpl<>(TextChannel.class, name, this, ChannelType.TEXT);
--- a/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
@@ -1095,7 +1095,7 @@ public class GuildImpl implements Guild
     {
         checkPermission(Permission.MANAGE_CHANNEL);
         Checks.notBlank(name, "Name");
-        name = Helpers.sanitizeChannelName(name.trim());
+        name = name.trim();
 
         Checks.check(name.length() > 0 && name.length() <= 100, "Provided name must be 1 - 100 characters in length");
         return new ChannelActionImpl<>(TextChannel.class, name, this, ChannelType.TEXT);
--- a/src/main/java/net/dv8tion/jda/internal/managers/ChannelManagerImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/managers/ChannelManagerImpl.java
@@ -240,10 +240,6 @@ public class ChannelManagerImpl extends ManagerBase<ChannelManager> implements C
     {
         Checks.notBlank(name, "Name");
         Checks.check(name.length() > 0 && name.length() <= 100, "Name must be between 1-100 characters long");
-
-        if (getType() == ChannelType.TEXT)
-            name = Helpers.sanitizeChannelName(name.trim());
-
         this.name = name;
         set |= NAME;
         return this;
--- a/src/main/java/net/dv8tion/jda/internal/managers/ChannelManagerImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/managers/ChannelManagerImpl.java
@@ -28,6 +28,7 @@ import net.dv8tion.jda.internal.entities.AbstractChannelImpl;
 import net.dv8tion.jda.internal.requests.Route;
 import net.dv8tion.jda.internal.requests.restaction.PermOverrideData;
 import net.dv8tion.jda.internal.utils.Checks;
+import net.dv8tion.jda.internal.utils.Helpers;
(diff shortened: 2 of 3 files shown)


src/main/java/net/dv8tion/jda/internal/managers/ChannelManagerImpl.java
```diff
@@ -239,8 +240,10 @@ public ChannelManagerImpl setName(@Nonnull String name)
     {
         Checks.notBlank(name, "Name");
         Checks.check(name.length() > 0 && name.length() <= 100, "Name must be between 1-100 characters long");
+
         if (getType() == ChannelType.TEXT)
-            Checks.noWhitespace(name, "Name");
+            name = Helpers.sanitizeChannelName(name.trim());
```
MinnDevelopment: This isn't documented

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1095,7 +1095,7 @@ public long getIdLong()
     {
         checkPermission(Permission.MANAGE_CHANNEL);
         Checks.notBlank(name, "Name");
-        name = name.trim();
+        name = Helpers.sanitizeChannelName(name.trim());
```
MinnDevelopment: This isn't documented

src/main/java/net/dv8tion/jda/internal/utils/Helpers.java
```diff
@@ -72,6 +72,10 @@ public static int countMatches(final CharSequence seq, final char c)
         return count;
     }
 
+    public static String sanitizeChannelName(String input) {
+        return input.replace(' ', '-');
+    }
```
MinnDevelopment: Formatting

src/main/java/net/dv8tion/jda/internal/utils/Helpers.java
```diff
@@ -72,6 +72,10 @@ public static int countMatches(final CharSequence seq, final char c)
         return count;
     }
 
+    public static String sanitizeChannelName(String input) {
+        return input.replace(' ', '-');
```
DManstrator: If the input should be sanitized (instead of leaving it up to Discord / throwing on whitespace), I would recommend `String#replaceAll("\\s+", "-")` instead since then also tabs will be replaced and multiple spaces will just result in one dash and not one for each space.

kantenkugel:
While this might be nice in a few cases, i don't really see, how the extra computation is needed at all. If you hardcode an invalid string into those methods or when you just blindly pass in invalid strings from user input then IMO thats not something JDA should **silently** deal with.

Due to this silent nature, what you get is not what you expect and thats also not really good (while this part could be documented)

DV8FromTheWorld:
I agree with @kantenkugel.
While the webapp might do these manipulations, I don't believe it is JDAs responsibility to take on this work.

DManstrator:
@DV8FromTheWorld It's not only a webapp thing, it's in the Discord Backend. I thought you tested it with the eval?
Creating a TextChannel with spaces in JDA will create a TextChannel having spaces replaced by dashes.

We could let Discord do that instead ofc but throwing on both creating and setting a name via. the ChannelManager shouldn't be the way to go since Discord allows and handles it. But not sure if it is that good to let Discord do that. At least it should be documented then.

kantenkugel:
@DManstrator if I understood that correctly, you are saying that the Discord API already supports (and handles) whitespaces in names?

If so, then this approach to the problem is redundant, as we should just give those "valid" names to the API and let it handle the sanitization.
All that has to be done in JDA would be to remove the noWhitespaces check (and use some nonEmpty one/one for minimum/maximum length) and document the behavior that the Discord API does in the docs

JakeJMattson:
@kantenkugel 
No, that is not correct. Creating a text channel with spaces is handled by the API, whereas *setting* a name with spaces is not handled by the API. It throws an exception, which is inconsistent behavior. This PR adds that consistency/sanity and applies the required Discord convention to the String implicitly. It just mocks the existing behavior provided by Discord itself.

You can test this behavior yourself with the sample snippet below.

```java
public static void main(String[] args) throws LoginException, InterruptedException {
    //Build JDA
    JDA jda = new JDABuilder(args[0]).build();
    jda.awaitReady();

    //Fetch guild
    Guild guild = jda.getGuilds().get(0);
    System.out.println("Guild: " + guild.getName());

    //Create channel with spaces - This is allowed
    guild.createTextChannel("A TextChannel With Spaces").queue(channel ->

        //Set channel name with spaces - This throws an exception
        channel.getManager().setName("channel for studying the blade").queue()
    );
}
```

DManstrator:
@kantenkugel Yes, correct. Here two links to proove that:
DV8 evaled it in the JDA Guild: https://discordapp.com/channels/125227483518861312/125227483518861312/634082089444048931
I evaled it on my server: https://discordapp.com/channels/125227483518861312/125227483518861312/634082384639164428

And yes, removing noWhitespace in ChannelManager#setName and checking nonEmpty etc. would be the best way to go imo.

@JakeJMattson No, the Discord API really handles it. The only reason this code fails is because JDA checks for whitespaces in ChannelManager#setName but not while creating it. And since it works while creating, it means it is handled by the Discord API and JDA is too restrictive.

JakeJMattson:
So should I modify this PR accordingly, or open up a new PR for this one line change?

DManstrator:
I would recommend fixing this PR instead.

kantenkugel:
Could you add some JDoc infos in the corresponding methods about what the Discord API does? So that users know that valid input doesn't mean that the name is created as-given

DManstrator:
And could you remove the unneeded Helpers import in the ChannelMangerImpl class?

PR #1126, opened 2019-10-25
Add readonly support for custom status

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This adds **readonly** support for custom status. This includes the emoji and the new activity type.


src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -550,8 +550,18 @@ public static Activity createActivity(DataObject gameJson)
             }
         }
 
+        RichPresence.Emoji emoji = null;
+        if (!gameJson.isNull("emoji"))
```
kantenkugel: Maybe add "emoji" to the `richGameFields` as well?

https://github.com/DV8FromTheWorld/JDA/blob/aba40e211af889796b3d431b6a84458a14a0cdde/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java#L61-L74

kantenkugel:
Regarding the Emoji inner class. I assume thats just cuz its a combination of Emote + (usual) Emoji?

Could we make this some general Entity and use it in place of eg MessageReaction?

Would just need some isEmote(), getEmote() i guess

MinnDevelopment:
Making a shared entity and using it in `MessageReaction#getReactionEmote()` might be an ABI change if done incorrectly. Worth considering though.

MinnDevelopment:
@kantenkugel we can't have a `getEmote()` on this because activities are not connected to JDA instances.

kantenkugel:
Hm, i guess

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -70,7 +70,6 @@
         tmp.add("session_id");
         tmp.add("state");
         tmp.add("sync_id");
-        tmp.add("emoji");
```
kantenkugel: Why remove this again? Does the Rich presence for Custom status have one of the other fields at least?

MinnDevelopment: We no longer treat it as rich presence

kantenkugel: Hm? So currently, a normal Activity is created, but the activity itself doesn't hold the correct text or the emoji yet, right?

The emoji would be given to the RichPresence Constructor, but that one is only called, when one of the rich-presence flags is set

kantenkugel:
Feel free to request a review from me when this is no longer a draft btw

src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -84,6 +86,14 @@
     @Nullable
     Timestamps getTimestamps();
 
+    /**
+     * The emoji (or custom emoji) attached to a custom status.
+     *
+     * @return {@link Emoji}
```
Andre601: I would suggest changing this to `{@link Emoji} or null if no emoji/emote was set` since it's nullable.

src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -241,7 +251,7 @@ static Activity of(@Nonnull ActivityType type, @Nonnull String name, @Nullable S
      */
     static boolean isValidStreamingUrl(@Nullable String url)
     {
-        return url != null && url.matches("https?://(www\\.)?twitch\\.tv/.+");
+        return url != null && url.matches("https?://(www\\.)?(twitch\\.tv/|youtube.com/watch\\?v=).+");
```
Andre601: Does this really work?
Testing it in Regex101 gives [some errors](https://regex101.com/r/XmfbR0/1) about the / being unescaped.

jagrosh: Which language did you have Regex101 set to? Not all languages implement regex the same way.

Andre601: I did have it on the default one which in this case was PHP.
Javascript does also return the same issue while Golang and Python don't.
It's a bit sad that Regex101 doesn't have the syntax/system for Java, but I guess Python would be close enough in that case.

Andre601: It was just pointed out in the Discord that the url of youtube is missing a `\\` before a dot.

kantenkugel: you can't just copy java -> regex101, cuz in java, every \ is double

DManstrator: Also wanted to point that out in the commit right now. Since `.tv` was also escaped, I'm pretty sure it also has to be escaped. The slash shouldn't matter since it's not a special char like a dot, a question mark, etc.

kantenkugel: This is what you get when you fix the regex for non-java regex engines: https://regex101.com/r/XL5Rel/1 and it looks 100% ok to me

Andre601: The dot after youtube is still not escaped, making it a quantifier according to regex101

kantenkugel: Ok, that one i didn't see. Thanks

src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -251,7 +251,7 @@ static Activity of(@Nonnull ActivityType type, @Nonnull String name, @Nullable S
...
```
caneleex: I think `https://youtu.be/code` is also a valid url for videos/livestreams, isn't it?

kantenkugel: thats a url-shortener... and discord will probably not use it

src/main/java/net/dv8tion/jda/internal/entities/RichPresenceImpl.java
```diff
@@ -181,6 +181,13 @@ public Party getParty()
         return party;
     }
 
+    @Nullable
+    @Override
+    public Emoji getEmoji()
```
kantenkugel: This override isn't needed, or is it? ActivityImpl already implements it exactly the same way + RichPresence just inherits it from there

src/main/java/net/dv8tion/jda/api/entities/Activity.java
MinnDevelopment: I tested and this url format doesn't work.

MinnDevelopment:
This should be ready to go now. I implemented a forward-compatible workaround for the current state of the API.

caneleex:
> I implemented a forward-compatible workaround for the current state of the API.
Didn't you say you don't want to implement workarounds? What if Discord suddenly fixes/changes the json? We all know Discord™.

Andre601:
I agree

PR #1129, opened 2019-11-02
Add ErrorResponseException.ignore(...)

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This allows to ignore specific error responses without a complicated piece of code.

### Example

```java
public static void selfDestruct(MessageChannel channel, String content) {
    channel.sendMessage(content).queue((message) -> 
        message.delete().queueAfter(30, SECONDS, null, ignore(UNKNOWN_MESSAGE))
    );
}
```


Andre601:
That's actually quite handy since I myself have the issue of constantly receiving "RestAction returned failure ....: Unknown Message" even tho I do everything to make sure the message is known (Check if message is null, make sure bot has message history perms before performing actions, etc)
So this would (at least) clean the console a bit.

PR #1139, opened 2019-11-17
compile -> implementation in gradle info

Jonbeckas:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Replaced compile with implementation in gradle info of Readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,7 +399,7 @@ Be sure to replace the **VERSION** key below with the one of the versions shown 
 **Gradle**
 ```gradle
 dependencies {
-    compile 'net.dv8tion:JDA:VERSION'
+    implementation 'net.dv8tion:JDA:VERSION'
 }
 
 repositories {


kantenkugel:
Please rebase your branch to be inline with develop

kantenkugel:
Actually... let me think about it... since this is only in the readme and minn apparently only does those changes to master

gpluscb:
For consistency, changing the `compile ('net.dv8tion:JDA:VERSION') {` a few lines down as well would make sense.

README.md
```diff
@@ -399,7 +399,7 @@ Be sure to replace the **VERSION** key below with the one of the versions shown
 **Gradle**
 ```gradle
 dependencies {
-    compile 'net.dv8tion:JDA:VERSION'
+    implementation 'net.dv8tion:JDA:VERSION'
```
MinnDevelopment: This also needs to be changed below. Also, I think it's reasonable to change this to be `implementation("...")` to be compatible with kotlin-dsl. Additionally, we should add a note that older versions of gradle still require `compile` since the majority is using older versions.

```md
´´´gradle
dependencies {
    ...
}
repositories {
    ....
}
´´´
> Note: Older versions of gradle use `compile` instead of `implementation`!
```

MinnDevelopment:
Closing due to inactivity.

PR #1143, opened 2019-11-22
EmoteManagerImpl#finalizeData resets the role list before serializing

averen:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

When the `EmoteManager` is reset in `EmoteManagerImpl#finalizeData`, the role list is still referenced by the `DataObject`, clearing it before the rest action serializes it. This means that an empty list is always sent to Discord no matter the contents provided in `EmoteManager#setRoles(Set<Role>)`.

State before `reset()` is called:

before-reset

State after `reset()` is called:

after-reset

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/managers/EmoteManagerImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/managers/EmoteManagerImpl.java
@@ -154,9 +154,8 @@ public class EmoteManagerImpl extends ManagerBase<EmoteManager> implements Emote
         withLock(this.roles, (list) ->
         {
             if (shouldUpdate(ROLES))
-                object.put("roles", list);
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/managers/EmoteManagerImpl.java
```diff
@@ -154,9 +154,8 @@ protected RequestBody finalizeData()
         withLock(this.roles, (list) ->
         {
             if (shouldUpdate(ROLES))
-                object.put("roles", list);
+                object.put("roles", new ArrayList<>(list));
```
MinnDevelopment: I think this should be fixed in the implementation of `DataObject#put` instead.

PR #1145, opened 2019-11-30
List instead of Collection in Javadoc of Message#delete

danthe1st:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

The JavaDoc of `net.dv8tion.jda.api.entities.Message#delete()`, referres to `net.dv8tion.jda.api.entities.MessageChannel#purgeMessages`.

While the link is correct, the text says `Collection` and not `List`.

This PR fixes this.


kantenkugel:
Was there a reason that purgeMessages was switched from Collection -> List? cuz it seems more natural to me to change the actual method signature and not the javadocs referencing to it

MinnDevelopment:
Changing the method signature is a breaking change. It would have to be done with an overload.

PR #1146, opened 2019-12-02
Fixed nullpointer exceptions for when doing owner checks

ToxicMushroom:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [x] Other: bugfix

Closes Issue: NaN

## Description
Fixes an npe that gets thrown when checking if a member can interact with another member in a server without an owner (owner banned)

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/utils/PermissionUtil.java
+++ b/src/main/java/net/dv8tion/jda/internal/utils/PermissionUtil.java
@@ -48,9 +48,9 @@ public class PermissionUtil
         Guild guild = issuer.getGuild();
         if (!guild.equals(target.getGuild()))
             throw new IllegalArgumentException("Provided members must both be Member objects of the same Guild!");
-        if(guild.getOwner().equals(issuer))
+        if(issuer.equals(guild.getOwner()))
             return true;
-        if(guild.getOwner().equals(target))
+        if(target.equals(guild.getOwner()))
             return false;
         List<Role> issuerRoles = issuer.getRoles();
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/utils/PermissionUtil.java
```diff
@@ -48,9 +48,9 @@ public static boolean canInteract(Member issuer, Member target)
         Guild guild = issuer.getGuild();
         if (!guild.equals(target.getGuild()))
             throw new IllegalArgumentException("Provided members must both be Member objects of the same Guild!");
-        if(guild.getOwner().equals(issuer))
+        if(issuer.equals(guild.getOwner()))
             return true;
-        if(guild.getOwner().equals(target))
+        if(target.equals(guild.getOwner()))
```
MinnDevelopment: Change this to `member.isOwner()` for both

src/main/java/net/dv8tion/jda/internal/utils/PermissionUtil.java
```diff
@@ -80,7 +80,7 @@ public static boolean canInteract(Member issuer, Role target)
         Guild guild = issuer.getGuild();
         if (!guild.equals(target.getGuild()))
             throw new IllegalArgumentException("Provided Member issuer and Role target must be from the same Guild!");
-        if(guild.getOwner().equals(issuer))
+        if(issuer.equals(guild.getOwner()))
```
MinnDevelopment: See above

PR #1148, opened 2019-12-06
Correcting wrong information for Message#delete();

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

- Corrects a typo where the Javadoc says `The pin was attempted after the Message had been deleted.` for the method `Message#delete();` which should be about the message being already deleted, when the request was performed.
- Added additional mention of `Attachment#getWidth()` and `Attachment#getHeight()` returning -1 if Attachment is neither an image nor video.
- Smaller grammar correction about the description of what Reactions are.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -913,7 +913,7 @@ public interface Message extends ISnowflake, Formattable
      *         or lost {@link net.dv8tion.jda.api.Permission#MESSAGE_MANAGE}.</li>
      *
      *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
-     *         The message got already deleted at the time the request was send.</li>
+     *         The message was already deleted at the time the request was sent.</li>
      * </ul>
      *
      * @throws java.lang.UnsupportedOperationException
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1982,7 +1982,7 @@ public int getSize()
          * The height of the Attachment if this Attachment is an image/video.
          * <br>If this Attachment is neither an image, nor a video, this returns -1.
          *
-         * @return int containing image/video Attachment height.
+         * @return int containing image/video Attachment height, or -1 if attachment is neither image nor video.
```
MinnDevelopment: This is already explained in the description above anyways but sure

Andre601: I added this as there are methods with a similar design.
For example do `Message#getTimeEdited` and `Message#getMember` mention in both description and `@return` that they may return null even if according to your message now, this would only be required in the description itself.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -913,7 +913,7 @@ default boolean isFromGuild()
      *         or lost {@link net.dv8tion.jda.api.Permission#MESSAGE_MANAGE}.</li>
      *
      *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
-     *         The pin was attempted after the Message had been deleted.</li>
+     *         The message got already deleted at the time the request was send.</li>
```
Devoxin:
```suggestion
     *         The message was already deleted at the time the request was sent.</li>
```

PR #1156, opened 2019-12-12
Disable owner caching when guild subs are off

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

The owner cache will be outdated when guild subscriptions are disabled thus we need to disable it as well.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2139,7 +2139,7 @@ public interface Guild extends ISnowflake
     /**
      * Shortcut for {@code guild.retrieveMemberById(guild.getOwnerIdLong())}.
      * <br>This will retrieve the current owner of the guild.
-     * It is possible that the owner of a guild is no longer a registered disocrd user in which case this will fail.
+     * It is possible that the owner of a guild is no longer a registered discord user in which case this will fail.
      *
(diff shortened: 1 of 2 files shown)


DV8FromTheWorld:
Is there a way to explicitly fetch the current owner?

MinnDevelopment:
@DV8FromTheWorld yes there is: 
```java
long ownerId = guild.getOwnerIdLong();
guild.retrieveMemberById(ownerId).queue(...);
```

MinnDevelopment:
I added a convenience method and linked to it from `getOwner()`:

```java
guild.retrieveOwner().queue(
    (member) -> System.out.printf("The owner is %#s\n", member.getUser())
);
```

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -2133,6 +2136,32 @@ default Emote getEmoteById(long id)
     @Nonnull
     RestAction<Member> retrieveMemberById(long id);
 
+    /**
+     * Shortcut for {@code guild.retrieveMemberById(guild.getOwnerIdLong())}.
+     * <br>This will retrieve the current owner of the guild.
+     * It is possible that the owner of a guild is no longer a registered disocrd user in which case this will fail.
```
Devoxin:
```suggestion
     * It is possible that the owner of a guild is no longer a registered discord user in which case this will fail.
```

DV8FromTheWorld:
Good job Github.. Double commits.

PR #1158, opened 2019-12-14
Add DeferredRestAction

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This fixes an issue that might occur with delayed RestActions that are backed by cache state.

**Problem**

```java
jda.retrieveUserById(userId).queueAfter(24, HOURS, (user) -> {
    System.out.printf("%#s\n", user);
});
```

The user here might be directly in cache which will return an `EmptyRestAction` with the user stored immediately. This user is then kept in that rest action instance for 24 hours while JDA might reconnect and invalidate.

**Solution**

With a deferred implementation we try to load the user from cache lazily. This means the moment that `queue` is called will decide whether to load the member from cache or make a request.


DV8FromTheWorld:
I know it's late, but looks good.

Issue #1173, opened 2019-12-26
Guild#kick reason parameter has no effect


Devoxin:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

The reason parameter in `Guild#kick(String userId, String reason)` has no effect -- the action has no associated reason when displayed in the audit log.

Screenshots demonstrating this behavior:
![](https://serux.pro/fd0598cebf.png)
![](https://serux.pro/60a0b1dcb2.png)

It's perhaps worth noting that using `AuditableRestAction#reason` to set an audit log reason **works** as expected.

### Expected Behavior

The audit log displays the reason associated with the kick action.

### Code Example or Reproduction Steps

```java
public void kickMember(Guild guild, String userId, String reason) {
  guild.kick(userId, reason).queue();
}

@Override
public void onGuildMemberJoin(GuildMemberJoinEvent event) {
  kickMember(event.getGuild(), event.getMember().getId(), "Auto-kick.");
}
```

### Exception or Error

N/A

MinnDevelopment:
This should be fixed in 4.1.0_85

PR #1174, opened 2019-12-26
Correctly apply reason to query

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1173 

## Description

The `addQueryParameter` method returns a new route instance that we should be using. I also removed some duplicate code.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
@@ -914,7 +914,7 @@ public class GuildImpl implements Guild
     {
         Route.CompiledRoute route = Route.Guilds.KICK_MEMBER.compile(getId(), userId);
         if (!Helpers.isBlank(reason))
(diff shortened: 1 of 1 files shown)


DManstrator:
I am a tiny bit confused. Why is there this fix listed in the PR and [this commit](https://github.com/DV8FromTheWorld/JDA/commit/a88709833def35d96c6e43a42e344680bb1f90ca) as well (found it in the logs from Discord)? According to the logs it should also be in the `patch-kick-ban-reason` branch but I can't see it. I guess you reverted that commit somehow?

I also prefer the version from this PR, I am just a bit confused where this commit went and if there could be some problems with it later on (2 different versions) although I'm sure you are already aware of that.

MinnDevelopment:
@DManstrator force-push removed it.

DManstrator:
I see, never worked with force-pushes. Thanks!

PR #1175, opened 2019-12-26
Allow setting a maximum decompression buffer size

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This also changes the default behavior to re-allocate the decompression buffer when it exceeds 2048 bytes. That seems to be a reasonable default. To get the previous behavior you can do `setMaxBufferSize(Integer.MAX_VALUE)`.


MinnDevelopment:
@DV8FromTheWorld yes, the buffer becomes rather large after receiving GUILD_CREATE payloads and then is never used to its full extent again until another huge payload is received. 2048 bytes should cover most payloads but this is an engineering constant.

PR #1177, opened 2019-12-29
Allow retrieving member presences at runtime

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This can be used to retrieve activities despite them not being cached.


src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -781,6 +785,57 @@ public boolean isAvailable()
...
 
+    @Nonnull
+    @Override
+    public CompletableFuture<MemberPresence> retrieveMemberPresence(long memberId)
+    {
+        Member member = getMemberById(memberId);
+        if (member != null
+            && api.isCacheFlagSet(CacheFlag.ACTIVITY)
+            && api.isCacheFlagSet(CacheFlag.CLIENT_STATUS))
+            return CompletableFuture.completedFuture(new MemberPresenceImpl(member));
```
DV8FromTheWorld: Why do we need to recreate a MemberPresence here when Member itself implements the interface?

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -781,6 +785,57 @@ public boolean isAvailable()
...
+
+        CompletableFuture<MemberPresence> future = new CompletableFuture<>();
+        if (memberId == 0)
+        {
+            future.completeExceptionally(new IllegalArgumentException("Member with the specified userId does not exist"));
+            return future;
+        }
+
+        api.getClient().requestMembers((response) -> {
```
DV8FromTheWorld: Let's make the callback the last argument to make this more readable

src/main/java/net/dv8tion/jda/api/entities/MemberPresence.java
```diff
@@ -0,0 +1,84 @@
...
+
+package net.dv8tion.jda.api.entities;
+
+import net.dv8tion.jda.api.OnlineStatus;
+
+import javax.annotation.Nonnull;
+import java.util.EnumSet;
+import java.util.List;
+
+public interface MemberPresence
```
DV8FromTheWorld: If possible, let's add class level docs

src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java
```diff
@@ -1106,6 +1122,25 @@ else if (requestStage == ConnectionStage.CONNECT)
         return null;
     }
 
+    public void requestMembers(Consumer<DataObject> callback, long guildId, long userId)
```
DV8FromTheWorld: Should we support requesting multiple in one request so that users don't have to make a ton of requests to to the work that a single could provide?

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -781,6 +785,57 @@ public boolean isAvailable()
...
+            if (notFound)
+            {
+                future.completeExceptionally(new IllegalArgumentException("Member with the specified userId does not exist"));
+                return;
+            }
+
+            DataArray presences = response.getArray("presences");
+            if (presences.isEmpty())
+            {
+                future.complete(MemberPresenceImpl.EMPTY);
```
DV8FromTheWorld: What does it mean if we get no presences back but the user didn't result in 'not found'?

src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java
```diff
@@ -1106,6 +1122,25 @@ else if (requestStage == ConnectionStage.CONNECT)
...
+            Set<Consumer<DataObject>> callbacks = memberRequests.get(guildId);
+            if (callbacks == null)
+                memberRequests.put(guildId, callbacks = new HashSet<>());
+            callbacks.add(callback);
+
+            send(DataObject.empty()
+                .put("op", WebSocketCode.MEMBER_CHUNK_REQUEST)
+                .put("d", DataObject.empty()
+                    .put("guild_id", guildId)
+                    .put("user_ids", Arrays.asList(userId, 0L))
```
DV8FromTheWorld: Why does this array end in 0L?

src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java
```diff
@@ -837,6 +840,19 @@ protected void onDispatch(DataObject raw)
                         jda.setStatus(JDA.Status.LOADING_SUBSYSTEMS);
                     }
                     break;
+                case "GUILD_MEMBERS_CHUNK":
+                    synchronized (memberRequests)
```
DV8FromTheWorld: While this will prevent the modification of `memberRequests` while the payload is being handled, it doesn't handle the fact that the following situation could occur:

1) request presence for Id 123
2) request presence for Id 234
3) payload response received for 123
 3.a) callbacks run.
 3.b) callback for 123 succeeds
 3.c) callback for 234 gets empty presence?
 3.d) all callbacks removed
4) payload response received for 234
 4.a) no callback to handle 234

Am I missing something?

src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java
MinnDevelopment: I use it as a marker to check later that the search is meant as a user lookup. The `not_found` will only be present in the payload if an id was not resolved, passing 0 forces it to be present.

src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java
MinnDevelopment: I'm still working on this and I know its incomplete. I'm trying to think of a good way to implement this.

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
MinnDevelopment: It means the response was either not meant for this request or discord fucked up somehow.

src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java
MinnDevelopment: I'm not sure if its worth supporting that. If you need access to multiple presences its likely better to just enable the cache. The interface only becomes more complicated.

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
MinnDevelopment: I had it this way because the `userId` parameter was meant as a varargs to allow requesting multiple users. I'd rather keep it like this for future proofing.

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
MinnDevelopment: I think its better to pass a snapshot to the future as the member state might change before the future is runing callbacks. The `toString` is also the same for both cases this way.

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
DV8FromTheWorld: This is an internal method, so having it be a single parameter and later changing that to am array should be fine. We needn't build in future proofing for that

src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java
DV8FromTheWorld: Not sure I agree with this all or single mentality. Seems perfectly reasonable to want to request more than one at a timeout without wanting the entire cache.

src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java
DV8FromTheWorld: What of we used a range from 0 - 100 or so to represent unique identifiable member chunk requests instead of just 0?

We could use a different number for each one so that we only run and clear a callback when we get a payload response for the one we requested.

MinnDevelopment:
Closing this in favor of #1292

PR #1178, opened 2019-12-31
Allow using JDA without gateway connection

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This allows using JDA without any gateway connection.

### Example

```kotlin
fun main() {
    val jda = JDABuilder.createRest(token).build()

    val route = Route.Messages.SEND_MESSAGE.compile(CHANNEL_ID)
    val body = DataObject.fromJson("""{"content": "Testing"}""")
    RestAction.makeAction(jda, route, body)
            .map(RestPayload::toString)
            .queue(::println)
}
```


MinnDevelopment:
I believe we close the response body before calling the rest action callback so a handler setup needs to be implemented to properly use this.

MinnDevelopment:
To be considered:

- Remove the RateLimiter abstract class, we only need one implementation
- Add an interface for the RateLimiter to make it easier to analyze usage
- Add a way to track bucket statistics such as queue size

MinnDevelopment:
This PR is getting discontinued for now.

PR #1183, opened 2020-01-09
Add getPermissionHolder method to PermissionOverride

DManstrator:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds a function `PermissionOverride#getPermissionHolder` to get the `IPermissionHolder`. It can be used to e. g. get the general permissions of that PermissionHolder no matter if it is a Member or a Role.

Also adds missing `@Nullable` annotations in the `PermissionOverrideImpl` class for consistency.

### Example Use Case
Check if a GuildChannel has a an override for a Member / Role which itself has a high permission (like Administrator, Manage Messages, etc.).

```java
     for (Permission perm : HIGH_PERMISSIONS)
     {
         // Any member with an override has a high permission?
         boolean highMemberOverride = permissionOverrides.stream()
             .filter(PermissionOverride::isMemberOverride)
             .anyMatch(po -> po.getMember().hasPermission(perm));
         if (highMemberOverride)
             return true;

         // Any role with an override has a high permission?
         boolean highRoleOverride = permissionOverrides.stream()
             .filter(PermissionOverride::isRoleOverride)
             .anyMatch(po -> po.getRole().hasPermission(perm));
         if (highRoleOverride)
           return true;
     }
```
With that new method:
```java
     for (Permission perm : HIGH_PERMISSIONS)
     {
         // Any permission holder with an override has a high permission?
         boolean highOverride = permissionOverrides.stream()
             .anyMatch(po -> po.getPermissionHolder().hasPermission(perm));
         if (highOverride)
             return true;
     }
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
@@ -23,6 +23,7 @@ import net.dv8tion.jda.api.requests.restaction.PermissionOverrideAction;
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+
 import java.util.EnumSet;
 
 /**
@@ -99,6 +100,63 @@ public interface PermissionOverride extends ISnowflake
      */
     @Nonnull
     JDA getJDA();
+
+    /**
+     * This method will return the {@link net.dv8tion.jda.api.entities.IPermissionHolder PermissionHolder} of this {@link net.dv8tion.jda.api.entities.PermissionOverride PermissionOverride}.
+     * It can be used to e. g. get the general permissions of that PermissionHolder no matter if it is a {@link net.dv8tion.jda.api.entities.Member Member} or a {@link net.dv8tion.jda.api.entities.Role Role}.
+     *
+     * <p><b>To get the concrete Member or Role, use {@link PermissionOverride#getMember()} or {@link PermissionOverride#getRole()}!</b>
+     *
+     * <h2>Example</h2>
+     * <pre><code>
+     * private static final List{@literal <}Permission{@literal >} HIGH_PERMISSIONS = Arrays.asList(Permission.ADMINISTRATOR, Permission.BAN_MEMBERS,
+     *      Permission.MANAGE_SERVER, Permission.KICK_MEMBERS);  // list because the order matters
+     *
+     * guild.getTextChannelCache().stream().filter(tc -{@literal >}
+     * {
+     *     List{@literal <}PermissionOverride{@literal >} permissionOverrides = tc.getPermissionOverrides();
+     *
+     *     // OLD
+     *     for (Permission perm : HIGH_PERMISSIONS)
+     *     {
+     *         // Any member with an override has a high permission?
+     *         boolean highMemberOverride = permissionOverrides.stream()
+     *             .filter(PermissionOverride::isMemberOverride)
+     *             .anyMatch(po -{@literal >} po.getMember().hasPermission(perm));
+     *         if (highMemberOverride)
+     *             return true;
+     *
+     *         // Any role with an override has a high permission?
+     *         boolean highRoleOverride = permissionOverrides.stream()
+     *             .filter(PermissionOverride::isRoleOverride)
+     *             .anyMatch(po -{@literal >} po.getRole().hasPermission(perm));
+     *         if (highRoleOverride)
+     *           return true;
+     *     }
+     *
+     *     // NEW
+     *     for (Permission perm : HIGH_PERMISSIONS)
+     *     {
+     *         // Any permission holder with an override has a high permission?
+     *         boolean highOverride = permissionOverrides.stream()
+     *             .anyMatch(po -{@literal >} po.getPermissionHolder().hasPermission(perm));
+     *         if (highOverride)
+     *             return true;
+     *     }
+     *
+     *     return false;
+     * });
+     * </code></pre>
+     *
+     * @return Never-null {@link net.dv8tion.jda.api.entities.IPermissionHolder PermissionHolder} of this {@link net.dv8tion.jda.api.entities.PermissionOverride PermissionOverride}.
+     *
+     * @see    PermissionOverride#getRole()
+     * @see    PermissionOverride#getMember()
+     * @see    net.dv8tion.jda.api.entities.IPermissionHolder#getPermissions() IPermissionHolder#getPermissions
+     * @see    net.dv8tion.jda.api.entities.IPermissionHolder#hasPermission(Permission... permissions) IPermissionHolder#hasPermission
+     */
+    @Nonnull
+    IPermissionHolder getPermissionHolder();
 
     /**
      * If this {@link net.dv8tion.jda.api.entities.PermissionOverride PermissionOverride} is an override dealing with
--- a/src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
@@ -93,8 +93,7 @@ public interface PermissionOverride extends ISnowflake
     EnumSet<Permission> getDenied();
 
     /**
-     * The {@link net.dv8tion.jda.api.JDA JDA} instance that this {@link net.dv8tion.jda.api.entities.PermissionOverride PermissionOverride}
-     * is related to.
+     * The {@link net.dv8tion.jda.api.JDA JDA} instance that this PermissionOverride is related to.
      *
      * @return Never-null {@link net.dv8tion.jda.api.JDA JDA} instance.
      */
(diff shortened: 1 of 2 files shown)


Andre601:
Why do you make like `List{@literal <}Permission{@literal >}` and not `List{@literal <Permission>}` or even `{@literal List<Permission>}`?

I can't remember seeing the docs having this weird syntax with the literal tag.

DManstrator:
Good point, will fix that. I actually just noticed it when trying to copy-paste the code that the generic was missing. Thought you can only use it on a single char.

src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java
```diff
@@ -103,12 +105,21 @@ public JDA getJDA()
...
+    }
+
+    @Nullable
     @Override
     public Member getMember()
     {
         return getGuild().getMemberById(id);
     }
 
+    @Nullable
```
MinnDevelopment: Adding these to implementations is redundant.

src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java
```diff
@@ -103,12 +105,21 @@ public JDA getJDA()
         return api;
     }
 
+    @Nonnull
+    @Override
+    public IPermissionHolder getPermissionHolder()
```
MinnDevelopment: You mark this as `Nonnull` but both methods used for resolution may return null. This is an error, the correct annotation to use is `@Nullable`.

src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
```diff
@@ -100,6 +101,63 @@
...
+     *             .anyMatch(po -{@literal >} po.getPermissionHolder().hasPermission(perm));
+     *         if (highOverride)
+     *             return true;
+     *     }
+     *
+     *     return false;
+     * });
+     * </code></pre>
+     *
+     * @return Never-null {@link net.dv8tion.jda.api.entities.IPermissionHolder PermissionHolder} of this {@link net.dv8tion.jda.api.entities.PermissionOverride PermissionOverride}.
```
MinnDevelopment: This is nullable and linking to the same class in javadoc is redundant.

src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
```diff
@@ -100,6 +101,63 @@
...
+     *
+     * @see    PermissionOverride#getRole()
+     * @see    PermissionOverride#getMember()
+     * @see    net.dv8tion.jda.api.entities.IPermissionHolder#getPermissions() IPermissionHolder#getPermissions
+     * @see    net.dv8tion.jda.api.entities.IPermissionHolder#hasPermission(Permission... permissions) IPermissionHolder#hasPermission
```
MinnDevelopment: I don't see why to link to getPermissions and hasPermissions, they are completely unrelated to this.

src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
```diff
@@ -100,6 +101,63 @@
...
     JDA getJDA();
 
+    /**
+     * This method will return the {@link net.dv8tion.jda.api.entities.IPermissionHolder PermissionHolder} of this {@link net.dv8tion.jda.api.entities.PermissionOverride PermissionOverride}.
+     * It can be used to e. g. get the general permissions of that PermissionHolder no matter if it is a {@link net.dv8tion.jda.api.entities.Member Member} or a {@link net.dv8tion.jda.api.entities.Role Role}.
+     *
+     * <p><b>To get the concrete Member or Role, use {@link PermissionOverride#getMember()} or {@link PermissionOverride#getRole()}!</b>
+     *
+     * <h2>Example</h2>
+     * <pre><code>
```
MinnDevelopment: Why does thie getter need a complex example?

src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
```diff
@@ -100,6 +101,63 @@
     @Nonnull
     JDA getJDA();
 
+    /**
+     * This method will return the {@link net.dv8tion.jda.api.entities.IPermissionHolder PermissionHolder} of this {@link net.dv8tion.jda.api.entities.PermissionOverride PermissionOverride}.
```
MinnDevelopment: You shouldn't link to the same class in javadoc, its redundant.

src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
Andre601: Does it even need a before-after comparison?
All the examples I saw in the Javadoc so far were always just what you can/should do with it, not what you can replace with it.

src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java
DManstrator: I guess I can then remove all Nullable / Nonnull annotations then? Because there were already a lot of Nonnulls but no Nullable ones so I added them.

src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
DManstrator: There were already other JavaDocs with links to itself. I will remove those links as well then.

src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
DManstrator: As an Example which could be used for that. Will remove them.

src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
DManstrator: To show a sample use-case and how it makes that easier. Will remove the old one and simplify it.

src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
DManstrator: Same as above.

src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java
DManstrator: I thought it is Nonnull because the PermissionOverride is for a Member or a Role and those methods return null if the override is for the other entity. So checking both should cover all cases making it Nonnull?

MinnDevelopment: No. It's nullable.

src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java
MinnDevelopment: The nonnull is not redundant, nullable is the default so they are not needed.

src/main/java/net/dv8tion/jda/api/entities/PermissionOverride.java
MinnDevelopment: I don't think the example is necessary, its a getter everyone knows how to use it.

src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java
DManstrator: All methods are marked as Nonnull / Nullable in the interface already, aren't the Nonnulls redundant as well then?

src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java
DManstrator: Will mark it as Nullable. For a better understanding, could you give me an example in which case it could return null? Or is it nullable just by definition (all methods used for it are nullable)?

DV8FromTheWorld: If we change `getPermissionHolder` to not call `getRole/getMember` (which arguably we should), then this _can_ be marked `@Notnull` as the underlying implementation of the PermissionOverride will be providing the explicit entity

DV8FromTheWorld: Ah, actually no. I didn't realize we weren't storing an actual reference to a `IPermissionHolder` object or soft-reference. You'll note that we're only storing the id here: https://github.com/DV8FromTheWorld/JDA/blob/ed2b57b38d357de1df18773178997b88f2a7ac6b/src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java#L59
This means that a lookup of `getRoleById` or `getMemberById` can return null due to timing or cache invalidate.

This is a question here as to whether we should drop using the `id` only and instead have a reference to the IPermissionHolder object instead as there is actual use for it here.

DManstrator: Didn't thought about the cache invalidation, good point!

The IPermissionHolder was stored as a class variable until 2 weeks ago. And because I didn't want to store it again, I made it like this.
[Link to the commit](https://github.com/DV8FromTheWorld/JDA/commit/4426e2e2e8eda2faff049f4503dfe0cd0c478f32#diff-3acd3d97f0f475cc7d2d1fadb219b3f0)

src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java
MinnDevelopment: No, they indicate to the implementation writer that returning null is an error.

src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java
MinnDevelopment: I have plans to make these available even if the member is currently not cached. So returning null is in all situations the behavior I want to convey as correct here.

DManstrator:
Just in case it went under the radar, I just re-requested a review.

Issue #1212, opened 2020-03-01
queued RestActions lost on shutdown()


bantu:
- [X] I have checked for similar issues.
- [X] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

The code snippet below does not send all messages to the channel. This is unexpected because https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/api/JDA.html#shutdown() says that "Already enqueued RestActions are still going to be executed." The code works correctly, when complete() is used instead of queue().

### Expected Behavior

15 messages (Test0 ... Test14) are sent to the channel.

### Code Example or Reproduction Steps

```java
JDA jda = new JDABuilder("MyToken").build().awaitReady();
Guild guild = jda.getGuildsByName("MyGuild", true).get(0);
TextChannel channel = guild.getTextChannelsByName("MyChannel", true).get(0);
for (int i = 0; i < 15; i++) {
    channel.sendMessage("Test" + i).queue();
}
jda.shutdown();
```

### Exception or Error

N/A

MinnDevelopment:
This should be fixed in 4.1.1_110

bantu:
@MinnDevelopment Thanks for the quick patch. Unfortunately the issue persists using 4.1.1_110.

MinnDevelopment:
Are you sure? I tested your code and it works as expected sending messages from 0 to 14

bantu:
I only see "Test0". I verified that I am actually using 4.1.1_110 by printing the version number. I am open to debugging suggestions.

DManstrator:
I can actually reproduce this on 4.1.1_110.

Code:
```
public class TestBot {
    private final static String TOKEN = "hereToken";
    private final static long CHANNEL_ID = hereChannelId;
    
    public static void main(String[] args) throws LoginException, InterruptedException  {
        JDA jda = new JDABuilder(TOKEN).build().awaitReady();
        TextChannel channel = jda.getTextChannelById(CHANNEL_ID);
        channel.sendMessageFormat("I am on Version %s!", JDAInfo.VERSION).queue();
        for (int i = 0; i < 15; i++) {
            channel.sendMessage("Test" + i).queue();
        }
        jda.shutdown();
    }
}
```
Console-Output:
```
[main] INFO net.dv8tion.jda.api.JDA - Login Successful!
[...]

PR #1215, opened 2020-03-02
Change defaults for requester to be non-daemon and have a shorter timeout

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Now the rate limit pool will keep doing requests until all requests are exhausted. I also decided to add an info log about this behavior so people know why the process doesn't stop immediately when calling `JDA#shutdown`.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/JDAImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/JDAImpl.java
@@ -663,8 +663,8 @@ public class JDAImpl implements JDA
     @Override
     public synchronized void shutdownNow()
     {
-        shutdown();
         requester.shutdown(); // stop all requests
+        shutdown();
         threadConfig.shutdownNow();
     }
 
--- a/src/main/java/net/dv8tion/jda/internal/JDAImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/JDAImpl.java
@@ -664,8 +664,8 @@ public class JDAImpl implements JDA
     public synchronized void shutdownNow()
     {
         shutdown();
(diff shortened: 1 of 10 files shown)


src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -664,8 +664,8 @@ public SelfUser getSelfUser()
     public synchronized void shutdownNow()
     {
         shutdown();
-        threadConfig.shutdownNow();
         requester.shutdown(); // stop all requests
+        threadConfig.shutdownNow();
```
DManstrator: Since you have changed the order, couldn't `shutdownRequester()` be called instead now or would it make a difference since it is a synchronized method?

MinnDevelopment: Calling `shutdownRequester` is unnecessary since we call `shutdownNow` on the executor anyway.

DManstrator: As I just realized, you call `threadConfig.shutdownRequester();` in the `shutdownRequester()` method, not `threadConfig.shutdownNow();`. Misread that somehow. My bad.

DManstrator:
Just tested this branch, it fixed #1212 for me.
Got the note `[main] INFO RateLimiter - Waiting for 1 bucket(s) to finish. Average queue size of 16 requests` and the program immediately terminated after the last message was sent.

Great job!

PR #1216, opened 2020-03-04
Adding a check in reaction methods before calling getTextChannel()

DManstrator:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds a check in removing / clearing reaction methods, checking if the message was sent in a TextChannel and else tells the user that he is unable to remove reactions of others or clear them if the message is not from a TextChannel.

Before, `getTextChannel()` was called without a check throwing `This message was not sent in a text channel` if the message was from a PrivateChannel (or group) which could be misleading ([and actually was](https://discordapp.com/channels/81384788765712384/381889648827301889/684520138367827971) which is why I make that PR) in that context.

This also fixes a **bug** where if the SelfUser was passed to the function, it would still have thrown that exception even that would be allowed.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -1378,6 +1378,7 @@ public interface Message extends ISnowflake, Formattable
      *             <li>If the provided {@link net.dv8tion.jda.api.entities.Emote Emote} is fake {@link net.dv8tion.jda.api.entities.Emote#isFake() Emote.isFake()}.</li>
      *             <li>If the provided {@link net.dv8tion.jda.api.entities.Emote Emote} cannot be used in the current channel.
      *                 See {@link Emote#canInteract(User, MessageChannel)} or {@link Emote#canInteract(Member)} for more information.</li>
+     *             <li>If the provided user is null</li>
      *         </ul>
      * @throws java.lang.IllegalStateException
      *         If this message was <b>not</b> sent in a
@@ -1488,7 +1489,7 @@ public interface Message extends ISnowflake, Formattable
      *             <li>{@link net.dv8tion.jda.api.Permission#MESSAGE_MANAGE Permission.MESSAGE_MANAGE}</li>
      *         </ul>
      * @throws java.lang.IllegalArgumentException
-     *         If the provided unicode emoji is null or empty.
+     *         If the provided unicode emoji is null or empty or if the provided user is null.
      * @throws java.lang.IllegalStateException
      *         If this message was <b>not</b> sent in a
      *         {@link net.dv8tion.jda.api.entities.TextChannel TextChannel}
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -1163,7 +1163,7 @@ public interface Message extends ISnowflake, Formattable
      * Removes all reactions from this Message.
      * <br>This is useful for moderator commands that wish to remove all reactions at once from a specific message.
      *
-     * <p>Please note that you <b>can't</b> clear reactions if this is a private message / this message was sent in a {@link net.dv8tion.jda.api.entities.PrivateChannel PrivateChannel}!
+     * <p>Please note that you <b>can't</b> clear reactions if this message was sent in a {@link net.dv8tion.jda.api.entities.PrivateChannel PrivateChannel}!
      *
      * <p><b>Neither success nor failure of this request will affect this Message's {@link #getReactions()} return as Message is immutable.</b>
      *
@@ -1191,7 +1191,7 @@ public interface Message extends ISnowflake, Formattable
      *         in the channel.
      * @throws java.lang.IllegalStateException
      *         If this message was <b>not</b> sent in a
-     *         {@link net.dv8tion.jda.api.entities.TextChannel TextChannel}.
+     *         {@link net.dv8tion.jda.api.entities.Guild Guild}.
      *
      * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link java.lang.Void}
      */
@@ -1202,7 +1202,7 @@ public interface Message extends ISnowflake, Formattable
     /**
      * Removes all reactions for the specified emoji.
      *
-     * <p>Please note that you <b>can't</b> clear reactions if this is a private message / this message was sent in a {@link net.dv8tion.jda.api.entities.PrivateChannel PrivateChannel}!
+     * <p>Please note that you <b>can't</b> clear reactions if this message was sent in a {@link net.dv8tion.jda.api.entities.PrivateChannel PrivateChannel}!
      *
      * <h2>Example</h2>
      * <pre><code>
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -163,22 +163,23 @@ public boolean isPinned()
     @Override
     public RestAction<Void> clearReactions()
     {
-        if (!isFromType(ChannelType.TEXT))
-            throw new IllegalStateException("Cannot clear reactions from a message in a Group or PrivateChannel.");
```
gpluscb: This is likely intentional. Checks.check throws an IllegalArgumentException, which suggests that an illegal argument has been passed to the method. That is not the case here, so throwing an IllegalStateException is more appropriate. The same applies to the methods below.

DManstrator: Yeah, I just recognized this 10min ago and I already started changing that back to throw an IllegalStateException. Before I didn't noticed that I was throwing the wrong exception now.

I will also add a note to the Docs of Message about that case.

Thanks for telling me though!

DManstrator:
Docs and code should be correct now, waiting for requested changes.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -172,13 +172,17 @@ public boolean isPinned()
...
+            throw new IllegalStateException("Cannot clear reactions from a message in a Group or PrivateChannel.");
         return getTextChannel().clearReactionsById(getId(), unicode);
     }
 
     @Nonnull
     @Override
     public RestAction<Void> clearReactions(@Nonnull Emote emote)
     {
+        if (!isFromType(ChannelType.TEXT))
+            throw new IllegalStateException("Cannot clear reactions from a message in a Group or PrivateChannel.");
```
duncte123: You should probably remove the group from here since bots can't be in groups

DManstrator: I am fully aware of this but since this was [the original output](https://github.com/DV8FromTheWorld/JDA/blob/fb0c6281a3eec4c07fc33ca1699d417077ada61b/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java#L166-L167) and the throws section saying ["if was not sent in a TextChannel"](https://github.com/DV8FromTheWorld/JDA/blob/fb0c6281a3eec4c07fc33ca1699d417077ada61b/src/main/java/net/dv8tion/jda/api/entities/Message.java#L1190-L1192), I decided to stay with this.

However in my JavaDoc additions, I explicitly talk about private messages since it's easier to understand that than telling it has to come from a TextChannel imo.

TL;DR: I would wait for Minn's response since this is likely to be unwanted.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1468,7 +1489,11 @@ default boolean isFromGuild()
      *             <li>{@link net.dv8tion.jda.api.Permission#MESSAGE_MANAGE Permission.MESSAGE_MANAGE}</li>
      *         </ul>
      * @throws java.lang.IllegalArgumentException
-     *         If the provided unicode emoji is null or empty.
+     *         If the provided unicode emoji is null or empty or if the provided user is null.
+     * @throws java.lang.IllegalStateException
+     *         If this message was <b>not</b> sent in a
+     *         {@link net.dv8tion.jda.api.entities.TextChannel TextChannel}
```
MinnDevelopment: I'd say not sent in a guild to also account for other channel types such as news channels

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -193,6 +197,14 @@ public boolean isPinned()
     @Override
     public RestAction<Void> removeReaction(@Nonnull Emote emote, @Nonnull User user)
     {
+        Checks.notNull(user, "User");  // to prevent NPEs
+        // check if the passed user is the SelfUser, then the ChannelType doesn't matter
+        if (user.equals(getJDA().getSelfUser()))
+            // we can safely remove that
+            return channel.removeReactionById(getIdLong(), emote);
+
+        if (!isFromType(ChannelType.TEXT))
```
MinnDevelopment: These checks should be changed to isFromGuild checks

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1324,6 +1336,8 @@ default boolean isFromGuild()
     /**
      * Removes a {@link net.dv8tion.jda.api.entities.User User's} reaction from this Message using an {@link net.dv8tion.jda.api.entities.Emote Emote}.
      *
+     * <p>Please note that you <b>can't</b> remove reactions of other users if this is a private message / this message was sent in a {@link net.dv8tion.jda.api.entities.PrivateChannel PrivateChannel}!
```
MinnDevelopment:
```suggestion
     * <p>Please note that you <b>can't</b> remove reactions of other users if this message was sent in a {@link net.dv8tion.jda.api.entities.PrivateChannel PrivateChannel}!
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
MinnDevelopment: Current behavior doesn't mean its permanent, being vague allows us to make changes more easily.

DManstrator: How should I implement it then? As it is now or should I revert that back? Or do I get your comment wrong?

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -682,15 +700,17 @@ public PrivateChannel getPrivateChannel()
     @Override
     public TextChannel getTextChannel()
     {
-        if (!isFromType(ChannelType.TEXT))
+        // this is not really safe but at the moment this is the only GuildChannel with messages
+        // and a MessageChannel can just be Text or Private so this is assumable as of now
+        if (!isFromGuild())
```
MinnDevelopment: This check should be reverted.

DManstrator:
In case it went under the radar, I re-requested a review.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -193,6 +197,14 @@ public boolean isPinned()
     @Override
     public RestAction<Void> removeReaction(@Nonnull Emote emote, @Nonnull User user)
     {
+        Checks.notNull(user, "User");  // to prevent NPEs
+        // check if the passed user is the SelfUser, then the ChannelType doesn't matter
+        if (user.equals(getJDA().getSelfUser()))
+            // we can safely remove that
+            return channel.removeReactionById(getIdLong(), emote);
```
MinnDevelopment: nit: When the if body is more than one line it should use braces, the comment could be moved behind the return or use braces.

DV8FromTheWorld: I'd like to see this change made too.

DV8FromTheWorld:
What is the progress on this PR? I see @MinnDevelopment has approved already.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
DManstrator: Let me change that really quick. :)

PR #1217, opened 2020-03-06
Deprecate GuildUpdateBannerEvent#getNewBannerIdUrl

averen:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

The method `GuildUpdateBannerEvent#getNewBannerIdUrl` is a typo for `getNewBannerUrl` and it is corrected in this PR.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/events/guild/update/GuildUpdateBannerEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/guild/update/GuildUpdateBannerEvent.java
@@ -16,6 +16,8 @@
 
 package net.dv8tion.jda.api.events.guild.update;
 
+import net.dv8tion.jda.annotations.DeprecatedSince;
+import net.dv8tion.jda.annotations.ReplaceWith;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.Guild;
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/events/guild/update/GuildUpdateBannerEvent.java
```diff
@@ -55,11 +57,27 @@ public String getNewBannerId()
...
+     *
+     * @deprecated This will be replaced by {@link #getNewBannerUrl()}
+     */
+    @Nullable
+    @Deprecated
+    @DeprecatedSince("4.1.1")
+    @ReplaceWith("getNewBannerUrl()")
+    public String getNewBannerIdUrl()
+    {
+        return getNewBannerUrl();
```
Andre601: That makes no sense.
If you deprecate the method, then why return the new field?

MinnDevelopment: Backwards compatibility.

Andre601: Was on mobile (which sucked a bit) so I saw all of this in a wrong way (assumed getNewBannerUrl() was actually a field xD)
So issue on my end.

MinnDevelopment:
It should be the next version, which would be `4.2.0` in this case.

src/main/java/net/dv8tion/jda/api/events/guild/update/GuildUpdateBannerEvent.java
```diff
@@ -55,11 +57,27 @@ public String getNewBannerId()
...
+    /**
+     * The new banner url
+     *
+     * @return The new banner url, or null if the banner was removed
+     *
+     * @deprecated This will be replaced by {@link #getNewBannerUrl()}
+     */
+    @Nullable
+    @Deprecated
+    @DeprecatedSince("4.1.1")
```
MinnDevelopment:
```suggestion
    @DeprecatedSince("4.2.0")
```

PR #1220, opened 2020-03-09
Add RestAction#timeout and RestAction#deadline

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds a simple check to timeout requests and avoid backpressure due to network problems.

### Example

image

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/requests/RestAction.java
+++ b/src/main/java/net/dv8tion/jda/api/requests/RestAction.java
@@ -210,6 +210,27 @@ public interface RestAction<T>
     }
 
     /**
+     * Default timeout to apply to every RestAction.
+     * <br>This will use no timeout unless specified otherwise.
+     * <br>If the request doesn't get executed within the timeout it will be cancelled.
+     *
+     * <p>This cancel happens when the request would be executed otherwise, similar to {@link #setCheck(BooleanSupplier)}.
+     * You can use this to avoid backpressure due to global rate limits.
+     *
+     * @param  timeout
+     *         The default timeout to use
+     * @param  unit
+     *         {@link TimeUnit Unit} for the timeout value
+     *
+     * @throws IllegalArgumentException
+     *         If the provided unit is null
+     */
+    static void setDefaultTimeout(long timeout, @Nonnull TimeUnit unit)
+    {
+        RestActionImpl.setDefaultTimeout(timeout, unit);
+    }
+
+    /**
      * The default failure callback used when none is provided in {@link #queue(Consumer, Consumer)}.
      *
      * @return The fallback consumer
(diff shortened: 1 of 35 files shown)


MinnDevelopment:
Should we use a different exception to signal timeout rather than using the same cancel exception? If so, which exception would seem appropriate? The docs for `java.util.concurrent.TimeoutException` explicitly state it's designed for blocking operations like `get(timeout)` and similar but we can probably ignore that and use it anyway.

src/main/java/net/dv8tion/jda/api/requests/RestAction.java
```diff
@@ -252,6 +285,60 @@ static void setDefaultSuccess(@Nullable final Consumer<Object> callback)
...
+     * @param  timestamp
+     *         Millisecond timestamp at which the request will timeout
+     *
+     * @return The same RestAction with the applied deadline
+     *
+     * @see    #timeout(long, TimeUnit)
+     * @see    #setDefaultTimeout(long, TimeUnit)
+     */
+    @Nonnull
+    default RestAction<T> deadline(long timestamp)
```
DV8FromTheWorld: Why is this implemented as an empty default method?

src/main/java/net/dv8tion/jda/api/requests/restaction/ChannelAction.java
```diff
@@ -49,6 +50,14 @@
     @Override
     ChannelAction<T> setCheck(@Nullable BooleanSupplier checks);
 
+    @Nonnull
+    @Override
+    ChannelAction<T> timeout(long timeout, @Nonnull TimeUnit unit);
+
+    @Nonnull
+    @Override
+    ChannelAction<T> deadline(long timestamp);
```
DV8FromTheWorld: Should this `@inheritDoc` ?

src/main/java/net/dv8tion/jda/api/requests/RestAction.java
DV8FromTheWorld: I see that you made a commit exclusively to do this, but I don't understand why

MinnDevelopment: To provide backwards compatibility with any other implementations of this interface. This can be done to implement operators like we did for `flatMap` or to write mock versions of this interface for unit testing.

src/main/java/net/dv8tion/jda/api/requests/restaction/ChannelAction.java
MinnDevelopment: No, that is unnecessary. I'm not sure why its present in AuditableRestAction, might just be a mistake on my end from back when i wrote it.

PR #1222, opened 2020-03-10
Documentation updates

MinnDevelopment:


src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -58,7 +58,8 @@
     private MessageEmbed.ImageInfo image;
 
     /**
-     * Creates an EmbedBuilder to be used to creates an embed to send.
+     * Constructs a new EmbedBuilder instance, which can be used to create {@link net.dv8tion.jda.api.entities.MessageEmbed MessageEmbeds}.
+     * These can then be sent to a channel using {@link net.dv8tion.jda.api.entities.MessageChannel#sendMessage(MessageEmbed)}.
```
DV8FromTheWorld:
```suggestion
     * These can then be sent to a channel using {@link net.dv8tion.jda.api.entities.MessageChannel#sendMessage(MessageEmbed) MessageChannel#sendMessage(MessageEmbed)}.
```

MinnDevelopment: We usually do `Foo.bar(Baz)` for this, that's also what the docs generate.

PR #1223, opened 2020-03-10
Add JDA#cancelRequests

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Allows to cancel all current requests in the queue.


MinnDevelopment:
I'll wait for #1220

src/main/java/net/dv8tion/jda/internal/requests/RateLimiter.java
```diff
@@ -64,6 +64,11 @@ public boolean isRateLimited(Route.CompiledRoute route)
         return getRateLimit(route) != null;
     }
 
+    public int cancelRequests()
```
DV8FromTheWorld: This should be an abstract method shouldn't it? What's the point of making it a default-implemented method?

src/main/java/net/dv8tion/jda/internal/requests/ratelimit/BotRateLimiter.java
```diff
@@ -102,6 +103,33 @@ private ScheduledExecutorService getScheduler()
...
+            buckets.values()
+                .stream()
+                .map(Bucket::getRequests)
+                .flatMap(Collection::stream)
+                .filter(request -> !request.isPriority())
+                .forEach(request -> {
+                    request.cancel();
+                    count.incrementAndGet();
+                });
+            buckets.clear();
```
DV8FromTheWorld: Do we want to clear priority requests? Based on the filter above, we don't cancel priority requests, but this will clear all request out regardless.

src/main/java/net/dv8tion/jda/internal/requests/ratelimit/BotRateLimiter.java
```diff
@@ -102,6 +103,33 @@ private ScheduledExecutorService getScheduler()
...
+            if (count.get() == 1)
+                RateLimiter.log.warn("Cancelled 1 request!");
+            else if (count.get() > 1)
+                RateLimiter.log.warn("Cancelled {} requests!", count.get());
```
DV8FromTheWorld: Why do we bother differentiating between 1 and > 1?
Also, do we really want to make clearing a warn? Seems like it could be an annoying message to have popup. Curious as to the thought process behind the logging level.

DV8FromTheWorld: We also call `count.get()` 4 times here. Can we relegate the result to a variable instead?

MinnDevelopment: It's a warn because it should be the last resort. Doing this regularly in normal operation should be considered a bug. I differentiate because why not.

src/main/java/net/dv8tion/jda/internal/requests/ratelimit/BotRateLimiter.java
MinnDevelopment: Hmm, will need to think how to handle this nicely.

src/main/java/net/dv8tion/jda/internal/requests/ratelimit/BotRateLimiter.java
DV8FromTheWorld: Ah. you differentiate for the `s`.

PR #1226, opened 2020-03-12
Allow to disable channel cache by type

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/JDABuilder.java
@@ -251,7 +251,7 @@ public class JDABuilder
      *     <li>{@link #setEnabledIntents(Collection)} is set to {@link GatewayIntent#DEFAULT}</li>
      *     <li>{@link #setMemberCachePolicy(MemberCachePolicy)} is set to {@link MemberCachePolicy#NONE}</li>
      *     <li>{@link #setChunkingFilter(ChunkingFilter)} is set to {@link ChunkingFilter#NONE}</li>
-     *     <li>{@link #setEnabledCacheFlags(EnumSet)} is set to none</li>
+     *     <li>{@link #setEnabledCacheFlags(EnumSet)} is set to {@link CacheFlag#channels()}</li>
      * </ul>
      *
      * @param  token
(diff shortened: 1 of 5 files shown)


DV8FromTheWorld:
This is going to cause a backup of cached events for channels that are not created due to their cacheflag being disabled
https://github.com/DV8FromTheWorld/JDA/blob/459a19aa2b5fa90ed3eb4e0509a5f6502b5d35da/src/main/java/net/dv8tion/jda/internal/handle/MessageCreateHandler.java#L68

MinnDevelopment:
That isn't an issue because we have a limit to the number of cached events:

```java
public static final long TIMEOUT_AMOUNT = 100;
```

https://github.com/DV8FromTheWorld/JDA/blob/6ba908ca54b1888211ffa754aa73098c3e77c0fd/src/main/java/net/dv8tion/jda/internal/handle/EventCache.java#L40

https://github.com/DV8FromTheWorld/JDA/blob/6ba908ca54b1888211ffa754aa73098c3e77c0fd/src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java#L916

DV8FromTheWorld:
Even with the timeout stuff, isn't this gonna result in messages being printed about "a large amount of events being cached?"

MinnDevelopment:
No there are no warnings about the event cache.

DV8FromTheWorld (approved):
It's been a while since this PR was created. Might need to look at the implications of the `.enableCache` calls in `createLight` and `createDefault`.

Otherwise, looks good. Lets get it in.

caneleex:
> 
> 
> It's been a while since this PR was created. Might need to look at the implications of the `.enableCache` calls in `createLight` and `createDefault`.
> 
> Otherwise, looks good. Lets get it in.

it should probably wait for stage channels support to be added

DV8FromTheWorld:
¯\_(ツ)_/¯
No real strong reason to wait. We can always add support for the stage channels to CacheFlag when we add support for stage channels.

DV8FromTheWorld:
Is this being help off on until StageChannels are done @MinnDevelopment ?

MinnDevelopment:
I need to read myself back in on this PR and make some decisions regarding which events to cache and if we can be smarter about it.

MinnDevelopment:
These cache settings raise the question of how `ThreadChannel#getParentChannel` is currently implemented. Why does it use a cache getter instead of a field?

PR #1227, opened 2020-03-12
Add GuildImpl#invalidate

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This will be very useful for the new auto-sharding.


DV8FromTheWorld:
Is there a reason we didn't go ahead with this change?

MinnDevelopment:
I added this for the purpose of auto-sharding, but I never got to it, so now it's just stale. To merge this we would need to check if any changes need to be addressed since this PR was created.

DV8FromTheWorld:
We should consider resurrecting this as we just encountered a situation where the cache didn't get cleared because invalidation was forgotten. Having this all in one place would be nice. https://github.com/DV8FromTheWorld/JDA/pull/2029

PR #1228, opened 2020-03-12
Document exception in JDA#getSelfUser and expose JDA#hasSelfUser

napstr:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Noticed that getSelfUser() throws an exception. That should be documented at least. It would also be nice for us enduser to be able to check it beforehand to avoid exception-driven control flow.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDA.java
+++ b/src/main/java/net/dv8tion/jda/api/JDA.java
@@ -1678,9 +1678,19 @@ public interface JDA
     IEventManager getEventManager();
 
     /**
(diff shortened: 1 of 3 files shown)


MinnDevelopment:
This actually only happens when using the shard manager, I'm planning to remove this exception entirely in the future.

MinnDevelopment:
#1231 should resolve this

napstr:
Removing thrown exceptions from the API sounds like an even better idea.

PR #1230, opened 2020-03-13
Add ability to get image url with extension

duncte123:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

## Description

This pull request adds the ability to force an extension on the image URLs returned by JDA.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ public static void main(String[] args) {
     JDABuilder builder = JDABuilder.createDefault(args[0]);
     
     // Disable parts of the cache
-    builder.setDisabledCacheFlags(EnumSet.of(CacheFlag.ACTIVITY, CacheFlag.VOICE_STATE));
+    builder.disableCache(CacheFlag.MEMBER_OVERRIDES, CacheFlag.VOICE_STATE);
     // Enable the bulk delete event
     builder.setBulkDeleteSplittingEnabled(false);
     // Disable compression (not recommended)
@@ -101,9 +101,7 @@ Additionally, you can change the handling of member/user cache by setting either
 ```java
 public void configureMemoryUsage(JDABuilder builder) {
     // Disable cache for member activities (streaming/games/spotify)
-    builder.setDisabledCacheFlags(
-        EnumSet.of(CacheFlag.ACTIVITY)
-    );
+    builder.disableCache(CacheFlag.ACTIVITY);
 
     // Only cache members who are either in a voice channel or owner of the guild
     builder.setMemberCachePolicy(MemberCachePolicy.VOICE.or(MemberCachePolicy.OWNER));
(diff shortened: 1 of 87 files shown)


duncte123:
This PR is ready for review

MinnDevelopment:
Why would you want a jpg instead of a png? What if the image is a gif and you pass `png`? I'm not sure what the point of this is. You can simply replace extensions using a `replaceAll("(png|gif)$", "jpg")`. This also adds a lot of new consistency we need to drag through every single image url and I'm not sure if that's worth it. @DV8FromTheWorld what do you think?

DV8FromTheWorld:
Yeah, I have to agree. 

Some of the stuff that sinks this for met:
1) There are realistically only 2 options: PNG and GIF
2) Does discord auto-convert GIF to WEBP? If it doesn't, this would be an invalid url.
3) Definitely shouldn't support both JPG and JPEG. That's bloat for no reason.
4) The is no real reason to prefer a JPEG over a PNG. 

I think the idea holds merit and I see how it can and would be a useful addition, however, JDA is already pretty big. Adding additional code to support a small-scale feature that can easily be replicated in userland code doesn't seem worthwhile, especially considering how we'd need to make sure that it gets applied consistently throughout the codebase going forward.

So long as we expose the animated (GIF) and still (PNG) variants of the image as required, I think that we are fine. Given that the developer wants to grab a different extension, they are free to replace the extension or build the url themselves.

duncte123:
The main reason that I made this PR was to avoid having to replace or build URLs manually.
I feel this or similar functionality should be included in JDA since it is part of the discord API.

About the question whether discord converts gifs to WEBP, it returns a non-animated version for any other extensions than gif.

I don't know why someone would be using JPG either but I've come across people that use the JFIF extension recently so it wouldn't surprise me if people want JPG over PNG.

Andre601:
> About the question whether discord converts gifs to WEBP, it returns a non-animated version for any other extensions than gif.

The fun (or even sad?) fact here is, that even the `.gif` extension will cause static images for user-clients, when the extension is anything else but lowercase `.gif` meaning that even `.GIF` or `.Gif` will cause this inconvenience.
If this is implemented should it be made sure to keep the extension lowercase.

MinnDevelopment:
> I don't know why someone would be using JPG either but I've come across people that use the JFIF extension recently so it wouldn't surprise me if people want JPG over PNG.

There will always be that one person that can make use of it, but adding a ton of methods that almost nobody will use, is a bit bloat I'd say. And for those specific edge-cases, the user can simply use a `replaceAll` as I mentioned before.

duncte123:
Closing this pr since the likelihood of acceptance is low

PR #1234, opened 2020-03-17
Added getGuildChannelById methods to ShardManager

Artuto:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

title

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
+++ b/src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
@@ -796,6 +796,133 @@ public interface ShardManager
     }
 
     /**
+     * Get {@link net.dv8tion.jda.api.entities.GuildChannel GuildChannel} for the provided ID.
+     * <br>This checks if any of the channel types in this guild have the provided ID and returns the first match.
+     *
+     * <br>To get more specific channel types you can use one of the following:
+     * <ul>
+     *     <li>{@link #getTextChannelById(String)}</li>
+     *     <li>{@link #getVoiceChannelById(String)}</li>
+     *     <li>{@link #getStoreChannelById(String)}</li>
+     *     <li>{@link #getCategoryById(String)}</li>
+     * </ul>
+     *
+     * @param  id
+     *         The ID of the channel
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided ID is null
+     * @throws java.lang.NumberFormatException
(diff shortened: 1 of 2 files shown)


Andre601:
I'm wondering, where you would use a GuildChannel over a Text or PrivateChannel, which already have stuff declared about what you can or cannot do.

MinnDevelopment:
I would prefer if we would just call the method on the individual JDA instances for easier maintanance

Artuto:
Done

src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
```diff
@@ -795,6 +795,131 @@ default Role getRoleById(@Nonnull final String id)
...
+     *
+     * @param  type
+     *         The {@link net.dv8tion.jda.api.entities.ChannelType}
+     * @param  id
+     *         The ID of the channel
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided ID is null
+     * @throws java.lang.NumberFormatException
+     *         If the provided ID is not a snowflake
```
MinnDevelopment: What if the type is null?

src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
```diff
@@ -795,6 +795,131 @@ default Role getRoleById(@Nonnull final String id)
...
+     *     <li>{@link #getStoreChannelById(long)}</li>
+     *     <li>{@link #getCategoryById(long)}</li>
+     * </ul>
+     *
+     * @param  type
+     *         The {@link net.dv8tion.jda.api.entities.ChannelType}
+     * @param  id
+     *         The ID of the channel
+     *
+     * @return The GuildChannel or null
```
MinnDevelopment: No throws docs for `notNull(type)`

Artuto: I copied the docs from the original methods https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/api/JDA.java#966-1016

Do you want me to add the missing throws there too?

MinnDevelopment: Yes, please do.

Artuto:
Done

PR #1236, opened 2020-03-19
Add MessageChannel#retrieveReactionUsersById and update reaction related docs

gpluscb:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fix documentation of ErrorResponses and Exceptions on reaction related methods.
~~This includes removing the documentation for InsuficientPermissionException on the retrieveReactionUsers methods because permissions are not checked currently. This should be added back if the check is added.~~ The missing permission check was implemented together with MessageChannel#retrieveReactionUsersById methods.


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1504,9 +1533,6 @@ default boolean isFromGuild()
      *
      * @throws java.lang.UnsupportedOperationException
      *         If this is not a Received Message from {@link net.dv8tion.jda.api.entities.MessageType#DEFAULT MessageType.DEFAULT}
-     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
-     *         If the MessageChannel this message was sent in was a {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} and the
-     *         logged in account does not have {@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY} in the channel.
```
MinnDevelopment: If this endpoint requires this permission the check should be added, instead of removing this from the docs.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1542,9 +1573,6 @@ default boolean isFromGuild()
...
```
MinnDevelopment: See above

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1564,26 +1592,11 @@ default boolean isFromGuild()
...
-     * <ul>
-     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#MISSING_ACCESS MISSING_ACCESS}
-     *     <br>The request was attempted after the account lost access to the {@link net.dv8tion.jda.api.entities.TextChannel TextChannel}
-     *         due to {@link net.dv8tion.jda.api.Permission#MESSAGE_READ Permission.MESSAGE_READ} being revoked
-     *     <br>Also can happen if the account lost the {@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY}</li>
-     *
-     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
-     *         The reaction request was attempted after the Message had been deleted.</li>
-     * </ul>
-     *
```
MinnDevelopment: Why remove this?

src/main/java/net/dv8tion/jda/api/entities/Message.java
gpluscb: Should the check be added to this PR, or separately? Also, should the change include `MessageChannel#retrieveReactionUsersById` methods?

src/main/java/net/dv8tion/jda/api/entities/Message.java
gpluscb: I don't see how `getReactionByUnicode` could produce ErrorResponses.

gpluscb: I don't see how `getReactionByUnicode` could produce ErrorResponses.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1447,6 +1468,11 @@ default boolean isFromGuild()
      *         due to {@link net.dv8tion.jda.api.Permission#MESSAGE_READ Permission.MESSAGE_READ} being revoked
      *     <br>Also can happen if the account lost the {@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY}</li>
      *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_EMOJI UNKNOWN_EMOJI}
+     *     <br>The provided unicode character does not refer to a known emoji unicode character.
+     *     <br>Proper unicode characters for emojis can be found at
+     *         <a href="http://unicode.org/emoji/charts/full-emoji-list.html" target="_blank">http://unicode.org/emoji/charts/full-emoji-list.html</a></li>
```
MinnDevelopment:
```suggestion
     *         <a href="https://unicode.org/emoji/charts/full-emoji-list.html" target="_blank">Emoji Table</a></li>
```

It would probably be worth using HTTPS where possible. (You can also change the existing link to https.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1346,6 +1359,9 @@ default boolean isFromGuild()
      *     <br>The reaction request was attempted after the account lost {@link net.dv8tion.jda.api.Permission#MESSAGE_MANAGE Permission.MESSAGE_MANAGE}
      *         in the {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} when removing the reaction.</li>
      *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_EMOJI UNKNOWN_EMOJI}
+     *     <br>The provided emote was deleted or doesn't exist.</li>
```
MinnDevelopment: I think this code can also mean that the emote is not available in the channel:

https://github.com/discord/discord-api-docs/issues/1350

src/main/java/net/dv8tion/jda/api/entities/Message.java
MinnDevelopment: If the check isn't already there, it should be added here.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -210,13 +210,16 @@ public boolean isPinned()
...
     {
         Checks.notNull(emote, "Emote");
 
+        if (getChannelType() == ChannelType.TEXT) {
+            ((TextChannelImpl) channel).checkPermission(Permission.MESSAGE_HISTORY);
+        }
+
         MessageReaction reaction = this.reactions.stream()
             .filter(r -> r.getReactionEmote().isEmote() && r.getReactionEmote().getEmote().equals(emote))
             .findFirst().orElse(null);
```
gpluscb: The docs of [`ReactionPaginationAction#getReaction()`](https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/api/requests/restaction/pagination/ReactionPaginationAction.html#getReaction()) suggest that it will always throw if the ReactionPaginationAction is created via `Message#retrieveReactionUsers`.
Do we care enough about having it sometimes not throw (it won't throw if this stream finds something) to not just use
```java
return channel.retrieveReactionUsersById(id, emote);
```
here?

gpluscb:
With the addition of sizeable code changes, should this be targeted towards the development branch?

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1067,6 +1067,9 @@ default boolean isFromGuild()
      *         or {@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY}
      *         in the {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} when adding the reaction.</li>
      *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_EMOJI UNKNOWN_EMOJI}
+     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently locked in account in this channel.</li>
```
DV8FromTheWorld, line 1071:
```suggestion
     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently logged-in account in this channel.</li>
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1253,7 +1263,7 @@ default boolean isFromGuild()
      *         or losing the {@link net.dv8tion.jda.api.Permission#VIEW_CHANNEL VIEW_CHANNEL} permission</li>
      *
      *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_EMOJI UNKNOWN_EMOJI}
-     *     <br>The provided emote was deleted or doesn't exist.</li>
+     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently locked in account in this channel.</li>
```
DV8FromTheWorld, line 1266:
```suggestion
     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently logged-in account in this channel.</li>
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1296,6 +1306,9 @@ default boolean isFromGuild()
      *         due to {@link net.dv8tion.jda.api.Permission#MESSAGE_READ Permission.MESSAGE_READ} being revoked
      *     <br>Also can happen if the account lost the {@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY}</li>
      *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_EMOJI UNKNOWN_EMOJI}
+     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently locked in account in this channel.</li>
```
DV8FromTheWorld, line 1310:
```suggestion
     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently logged-in account in this channel.</li>
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1346,6 +1359,9 @@ default boolean isFromGuild()
      *     <br>The reaction request was attempted after the account lost {@link net.dv8tion.jda.api.Permission#MESSAGE_MANAGE Permission.MESSAGE_MANAGE}
      *         in the {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} when removing the reaction.</li>
      *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_EMOJI UNKNOWN_EMOJI}
+     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently locked in account in this channel.</li>
```
DV8FromTheWorld, line 1363:
```suggestion
     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently logged-in account in this channel.</li>
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1495,6 +1525,9 @@ default boolean isFromGuild()
      *         due to {@link net.dv8tion.jda.api.Permission#MESSAGE_READ Permission.MESSAGE_READ} being revoked
      *     <br>Also can happen if the account lost the {@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY}</li>
      *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_EMOJI UNKNOWN_EMOJI}
+     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently locked in account in this channel.</li>
```
DV8FromTheWorld, line 1529:
```suggestion
     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently logged-in account in this channel.</li>
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1601,29 +1624,16 @@ default boolean isFromGuild()
...
+     *         The string id of the reaction emote.
      *
      * @throws java.lang.UnsupportedOperationException
      *         If this is not a Received Message from {@link net.dv8tion.jda.api.entities.MessageType#DEFAULT MessageType.DEFAULT}
-     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
-     *         If the MessageChannel this message was sent in was a {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} and the
-     *         logged in account does not have {@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY} in the channel.
      * @throws java.lang.IllegalArgumentException
+     *         If the provided id is null.
+     * @throws java.lang.NumberFormatException
```
DV8FromTheWorld, line 1636: This isn't true. `Checks.isSnowflake` throws `IllegalArgumentException`
Additionally, an id being `null` already makes it "not a snowflake", thus illegal.

This should be reverted.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -2065,10 +2061,7 @@ default MessagePaginationAction getIterableHistory()
      *         The unicode characters of the emoji
      *
      * @throws java.lang.IllegalArgumentException
-     *         <ul>
-     *             <li>If provided {@code messageId} is {@code null} or empty.</li>
-     *             <li>If provided {@code unicode} is {@code null} or empty.</li>
-     *         </ul>
+     *         If provided {@code unicode} is {@code null} or empty.
```
DV8FromTheWorld, line 2064: Why did this lose the bullet point about `messageId` ?

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -2155,10 +2146,8 @@ default MessagePaginationAction getIterableHistory()
      *         {@link net.dv8tion.jda.api.Permission#MESSAGE_ADD_REACTION Permission.MESSAGE_ADD_REACTION} in the
      *         {@link net.dv8tion.jda.api.entities.TextChannel TextChannel}.</li>
      *
-     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_EMOJI}
-     *     <br>The provided unicode character does not refer to a known emoji unicode character.
-     *     <br>Proper unicode characters for emojis can be found at
-     *         <a href="http://unicode.org/emoji/charts/full-emoji-list.html" target="_blank">http://unicode.org/emoji/charts/full-emoji-list.html</a></li>
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_EMOJI UNKNOWN_EMOJI}
+     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently locked in account in this channel.</li>
```
DV8FromTheWorld, line 2150:
```suggestion
     *     <br>The provided emote was deleted, doesn't exist, or is not available to the currently logged-in account in this channel.</li>
```

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -2174,10 +2163,7 @@ default MessagePaginationAction getIterableHistory()
      *         The emote to remove
      *
      * @throws java.lang.IllegalArgumentException
-     *         <ul>
-     *             <li>If provided {@code messageId} is {@code null} or empty.</li>
-     *             <li>If provided {@code emote} is {@code null}.</li>
-     *         </ul>
+     *         If provided {@code emote} is {@code null}.
```
DV8FromTheWorld, line 2166: Why did this lose the bullet point about the `messageId`?

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -2261,8 +2446,6 @@ default MessagePaginationAction getIterableHistory()
      * @param  messageId
      *         The message to pin.
      *
-     * @throws IllegalArgumentException
-     *         if the provided messageId is not positive.
```
DV8FromTheWorld, line 2265: Why was this removed?
The description "positive" is not correct, however, we definitely can still throw an exception if the provided id isn't a snowflake.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -210,13 +210,16 @@ public boolean isPinned()
...
 
-
     @Nonnull
     @Override
     public ReactionPaginationAction retrieveReactionUsers(@Nonnull Emote emote)
     {
         Checks.notNull(emote, "Emote");
 
+        if (getChannelType() == ChannelType.TEXT) {
+            ((TextChannelImpl) channel).checkPermission(Permission.MESSAGE_HISTORY);
```
DV8FromTheWorld, line 220: Do we also need to check for the READ permissions? The docs say that we can get a `MISSING_ACCESS` error if we don't have it.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -230,8 +233,13 @@ public ReactionPaginationAction retrieveReactionUsers(@Nonnull Emote emote)
     @Override
     public ReactionPaginationAction retrieveReactionUsers(@Nonnull String unicode)
     {
+        Checks.notEmpty(unicode, "Emoji");
         Checks.noWhitespace(unicode, "Emoji");
 
+        if (getChannelType() == ChannelType.TEXT) {
+            ((TextChannelImpl) channel).checkPermission(Permission.MESSAGE_HISTORY);
+        }
```
DV8FromTheWorld, line 241: Same question as above

src/main/java/net/dv8tion/jda/internal/entities/TextChannelImpl.java
```diff
@@ -500,6 +502,31 @@ public MessageAction sendFile(@Nonnull InputStream data, @Nonnull String fileNam
...
 
+    @Nonnull
+    @Override
+    public ReactionPaginationAction retrieveReactionUsersById(@Nonnull String messageId, @Nonnull String unicode)
+    {
+        Checks.isSnowflake(messageId, "Message ID");
+        Checks.notEmpty(unicode, "Emoji");
+        Checks.noWhitespace(unicode, "Emoji");
+
+        checkPermission(Permission.MESSAGE_HISTORY);
```
DV8FromTheWorld, line 513: Same question as above

src/main/java/net/dv8tion/jda/internal/entities/TextChannelImpl.java
```diff
@@ -500,6 +502,31 @@ public MessageAction sendFile(@Nonnull InputStream data, @Nonnull String fileNam
...
+    }
+
+    @Nonnull
+    @Override
+    public ReactionPaginationAction retrieveReactionUsersById(@Nonnull String messageId, @Nonnull Emote emote)
+    {
+        Checks.isSnowflake(messageId, "Message ID");
+        Checks.notNull(emote, "Emote");
+
+        checkPermission(Permission.MESSAGE_HISTORY);
```
DV8FromTheWorld, line 525: Same question as above

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
MinnDevelopment: You can't have the message history permissions without read, its an implicit permission.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
MinnDevelopment: I think we should use `channel.retrieveReactionUsersById` regardless since it also does the permission checks. The throw happens if the action is created without a `MessageReaction` instance. If we find one here, it won't throw but the docs say it anyway because users shouldn't rely on this.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
MinnDevelopment: I feel like this cast shouldn't be here, this checkPermission method shouldn't be used outside the `TextChannelImpl` class. You should avoid doing that. The `channel.retrieveReactionUsersById` method should be doing this check anyway, I think switching to that instead would be a better solution that doesn't involve using internal helper methods of the TextChannelImpl class like that. This cast makes unit testing very complicated.

src/main/java/net/dv8tion/jda/api/entities/Message.java
gpluscb: In this case, `MiscUtil.parseSnowflake` was called without an `isSnowflake` check. I added the check and reverted the documentation change.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
gpluscb: Since this is the `long` overload, null or something empty cannot be passed as `messageId`. However, I misunderstood the `Checks.isSnowflake` method and this can still throw if `messageId` is not a valid snowflake. I added a bullet point about that situation now.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
gpluscb: Same as above.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
gpluscb: I misunderstood the `isSnowflake` check. The throws documentation was added back for the case that `messageId` is not a valid snowflake.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -264,6 +239,7 @@ public ReactionPaginationAction retrieveReactionUsers(@Nonnull String unicode)
     @Override
     public MessageReaction.ReactionEmote getReactionById(@Nonnull String id)
     {
+        Checks.isSnowflake(id);
         return getReactionById(MiscUtil.parseSnowflake(id));
```
MinnDevelopment, line 243: This check is unnecessary cine `parseSnowflake` will throw if the input is an invalid snowflake anyway.

gpluscb: I added the check as an easy way to consistently throw `IllegalArgumentExceptions` instead of throwing `NumberFormatExceptions` in some cases. https://github.com/DV8FromTheWorld/JDA/pull/1236#discussion_r402305343 is a relevant review comment to this.

MinnDevelopment: We consistenly do just parseSnowflake and rely on that throw. A `NumberFormatException` is a `IllegalArgumentException`.

https://docs.oracle.com/javase/8/docs/api/java/lang/NumberFormatException.html
> public class NumberFormatException
extends IllegalArgumentException

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -2188,6 +2186,211 @@ default MessagePaginationAction getIterableHistory()
...
+     */
+    @Nonnull
+    @CheckReturnValue
+    default ReactionPaginationAction retrieveReactionUsersById(@Nonnull String messageId, @Nonnull String unicode)
+    {
+        Checks.isSnowflake(messageId, "Message ID");
+        Checks.notEmpty(unicode, "Emoji");
+        Checks.noWhitespace(unicode, "Emoji");
+
+        return new ReactionPaginationActionImpl(this, messageId, EncodingUtil.encodeUTF8(unicode));
```
MinnDevelopment, line 2241:
```suggestion
        return new ReactionPaginationActionImpl(this, messageId, EncodingUtil.encodeReaction(unicode));
```

PR #1240, opened 2020-03-23
Add better configuration methods to the builders

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This allows disabling flags without needing to specify an entire config.
This is very useful with the new factory methods.

~~We should also consider adding `enableIntents` and `disableIntents` for the same reason. I've seen users do `GatewayIntent.getIntents(DEFAULT & ~intents)` which would be a lot cleaner using `disableIntents(intents)`.~~ I decided to include this in the PR.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/JDABuilder.java
@@ -170,6 +170,9 @@ public class JDABuilder
      *         The bot token to use
      *
      * @return The new JDABuilder
+     *
+     * @see    #disableIntents(GatewayIntent, GatewayIntent...)
+     * @see    #enableIntents(GatewayIntent, GatewayIntent...)
      */
     @Nonnull
     @CheckReturnValue
@@ -267,6 +270,9 @@ public class JDABuilder
      *         The bot token to use
      *
      * @return The new JDABuilder
+     *
+     * @see    #disableIntents(GatewayIntent, GatewayIntent...)
+     * @see    #enableIntents(GatewayIntent, GatewayIntent...)
      */
     @Nonnull
     @CheckReturnValue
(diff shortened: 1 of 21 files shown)


src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -477,14 +515,69 @@ public JDABuilder setRelativeRateLimit(boolean enable)
...
+     *
+     * @return The JDABuilder instance. Useful for chaining.
+     *
+     * @see    #disableCache(Collection)
+     */
+    @Nonnull
+    public JDABuilder enableCache(@Nonnull Collection<CacheFlag> flags)
+    {
+        Checks.noneNull(flags, "CacheFlags");
+        cacheFlags.removeAll(flags);
```
DV8FromTheWorld:
```suggestion
        cacheFlags.addAll(flags);
```

src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -477,14 +515,69 @@ public JDABuilder setRelativeRateLimit(boolean enable)
...
+     *
+     * @throws IllegalArgumentException
+     *         If provided with null
+     *
+     * @return The JDABuilder instance. Useful for chaining.
+     *
+     * @see    #disableCache(CacheFlag, CacheFlag...)
+     */
+    @Nonnull
+    public JDABuilder enableCache(@Nonnull CacheFlag flag, @Nonnull CacheFlag... flags)
```
DV8FromTheWorld: Would it be worth it to point this at the `Collections` variant method?

src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
```diff
@@ -452,14 +490,69 @@ public DefaultShardManagerBuilder setRelativeRateLimit(boolean enable)
...
+     *
+     * @return The DefaultShardManagerBuilder instance. Useful for chaining.
+     *
+     * @see    #disableCache(Collection)
+     */
+    @Nonnull
+    public DefaultShardManagerBuilder enableCache(@Nonnull Collection<CacheFlag> flags)
+    {
+        Checks.noneNull(flags, "CacheFlags");
+        cacheFlags.removeAll(flags);
```
DV8FromTheWorld:
```suggestion
        cacheFlags.addAll(flags);
```

src/main/java/net/dv8tion/jda/api/JDABuilder.java
MinnDevelopment: You mean an `@see`?

PR #1251, opened 2020-04-04
Add section to bug report template

MinnDevelopment:
This adds a new section to the bug report template. With the new changes to 4.2.0 we will need to always know how the user started the JDA session because of how things rely on intents and cache configuration.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -49,6 +49,15 @@ If you have a reproducible behavior, please provide some details for us.
 Otherwise just put N/A
 -->
 
+### Code for JDABuilder or DefaultShardManagerBuilder Used
+
+<!--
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -49,6 +49,15 @@ If you have a reproducible behavior, please provide some details for us.
 Otherwise just put N/A
 -->
 
+### Code for JDABuilder or DefaultShardManagerBuilder Used
+
+<!--
+Please provide the code you used to start your bot.
+This greatly impacts behavior due to the way cache and events are configured.
+
+Otherwise just put N/A
```
Andre601, line 58: In what cases would you actually put N/A?

MinnDevelopment: Compiler errors or similar

PR #1260, opened 2020-04-11
Fix incorrect size check for MessageChannel#sendFile

gpluscb:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fixes an issue where the `byte[]` overload of `MessageChannel#sendFile` 
does not check permissions and ignores guild boosts in the max allowed size check.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
@@ -770,8 +770,7 @@ public interface MessageChannel extends ISnowflake, Formattable
     {
         Checks.notNull(data, "data");
         Checks.notNull(fileName, "fileName");
-        final long maxSize = getJDA().getSelfUser().getAllowedFileSize();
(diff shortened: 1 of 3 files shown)


MinnDevelopment:
For your info, it does check permissions since the `sendFile(String, byte[])` just calls `sendFile(String, InputStream)` which does have an override.

gpluscb:
Should this behaviour be changed?
The recently added `TextChannelImpl/PrivateChannelImpl#sendFile(File, String, AttachmentOption...)` methods have the same behaviour, so it would make sense to change these as well if this is going to be changed.

MinnDevelopment:
Well, checking permissions twice is kind of redundant. However, the code flow here is confusing, as you can probably tell. I'm fine either way, your call.

gpluscb:
In that case I am going to leave it in for clarity sake.

PR #1263, opened 2020-04-16
Add ReactionEmote#getAsReactionCode method

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description
Adds a `getAsReactionCode()` method to `MessageReaction.ReactionEmote` which would either return the UTF-8 code of the emoji, or the name and id of the Emote in the format `name:id` depending on what type it is.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
@@ -399,7 +399,7 @@ public class MessageReaction
 
     private String getReactionCode()
     {
-        return emote.isEmote()
+        return emote != null
                 ? emote.getName() + ":" + emote.getId()
                 : EncodingUtil.encodeUTF8(emote.getName());
     }
@@ -539,7 +539,7 @@ public class MessageReaction
         @Nonnull
         public String getAsReactionCode()
         {
-            return emote != null
+            return isEmote()
                     ? name + ":" + id
                     : name;
         }
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
@@ -528,6 +528,20 @@ public class MessageReaction
                 throw new IllegalStateException("Cannot get id for emoji reaction");
             return id;
         }
+        
+        /**
+         * The code for this Reaction.
+         * <br>For unicode emojis this will be the unicode of said emoji rather than an alias like {@code :smiley:}.
+         * <br>For custom emotes this will be the name and id of said emote in the format {@code <name>:<id>}.
+         *
+         * @return The unicode if it is an emoji, or the name and id in the format {@code <name>:<id>}
+         */
+        public String getAsReactionCode()
+        {
+            return isEmote()
+                    ? name + ":" + id
+                    : EncodingUtil.encodeUTF8(name);
+        }
 
         /**
          * The unicode representing the emoji used for reacting.
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
@@ -539,7 +539,7 @@ public class MessageReaction
         @Nonnull
         public String getAsReactionCode()
         {
-            return isEmote()
+            return emote != null
                     ? name + ":" + id
                     : name;
         }
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
@@ -536,6 +536,7 @@ public class MessageReaction
          *
          * @return The unicode if it is an emoji, or the name and id in the format {@code <name>:<id>}
          */
+        @Nonnull
         public String getAsReactionCode()
         {
             return isEmote()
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
```diff
@@ -528,6 +528,20 @@ public long getIdLong()
...
+         * <br>For unicode emojis this will be the unicode of said emoji rather than an alias like {@code :smiley:}.
+         * <br>For custom emotes this will be the name and id of said emote in the format {@code <name>:<id>}.
+         *
+         * @return The unicode if it is an emoji, or the name and id in the format {@code <name>:<id>}
+         */
+        public String getAsReactionCode()
+        {
+            return isEmote()
+                    ? name + ":" + id
+                    : EncodingUtil.encodeUTF8(name);
```
MinnDevelopment, line 543: How do you expect this to be used? If you encode it here then it will encoding it twice if used with `addReaction`.

src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
```diff
@@ -528,6 +528,20 @@ public long getIdLong()
                 throw new IllegalStateException("Cannot get id for emoji reaction");
             return id;
         }
+        
+        /**
+         * The code for this Reaction.
+         * <br>For unicode emojis this will be the unicode of said emoji rather than an alias like {@code :smiley:}.
+         * <br>For custom emotes this will be the name and id of said emote in the format {@code <name>:<id>}.
+         *
+         * @return The unicode if it is an emoji, or the name and id in the format {@code <name>:<id>}
```
MinnDevelopment, line 537: If you encode it then you don't return the unicode anymore, you return an encoding of the characters. In this case percent-encoded.

DManstrator:
https://github.com/DV8FromTheWorld/JDA/blob/ed831c7579f2528fcc83c4ee5cbdcefb55f973f1/src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java#L400-L405
I would have used that method instead since it's already there and doing the exact same stuff. But I would use your `isEmote` check since `emote` is nullable.

src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
Andre601: I based it off the `getReactionCode` from the MessageReaction class, since this one had the EncodingUtil.encodeUTF8 method.
I'll remove it I suppose.

src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
```diff
@@ -528,6 +528,20 @@ public long getIdLong()
...
+         */
+        public String getAsReactionCode()
```
MinnDevelopment, line 539: Missing `@Nonnull` annotation here

Andre601: Added.

Andre601:
> Since you don't encode anymore, this obs cannot be used anymore. But would you be so kind and replace that emote.isEmote() with a null check, else this can throw an NPE here as explained above. @Andre601

What should I return then when emote would be null? An empty String like it is the case with getName()

DManstrator:
> What should I return then when emote would be null?

The same as before, just the check shouldn't trigger a possible NPE. But your fix is exactly what I meant, thanks. :)

Andre601:
Changed it.
Also, please just send a separate comment next time as I don't receive notifications for edits (obviously) and don't check my PRs on a regular basis.

DManstrator:
That's interesting. I also tagged MarRue in an edit and he indeed got a notification. Guess that's something not generalized. Will do that. lgtm now.

src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
```diff
@@ -399,7 +399,7 @@ public ReactionPaginationAction retrieveUsers()
 
     private String getReactionCode()
     {
-        return emote.isEmote()
+        return emote != null
```
MinnDevelopment, line 402: When would this ever be null?

Andre601: I don't know either.
`isEmote()` essentially returns the result from `emote != null` and @DManstrator suggested this alteration.

MinnDevelopment: No, that makes no sense. You are calling `isEmote` on an object, that object has an `emote` field which might be null.

DManstrator: @MinnDevelopment @Andre601 Yeah, it was my bad, I got confused here. I saw that in the `ReactionEmote` class `emote` actually can be null in case of an emoji.
https://github.com/DV8FromTheWorld/JDA/blob/5234d2d943342550a9fe6605e78463f25e4589ff/src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java#L437-L443
The variables just had the same name and the method was close to the `ReactionEmote` class, that's why I thought that it is the nullable `emote`. Again sorry for that.

Andre601: So, should I revert it back to emote.isEmote()?

PR #1267, opened 2020-04-18
Json Parsing Optimizations

napstr:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

I finally had some time to deep dive into some performance optimizations. The wins are not huge here, but still around a solid 5 % on a bot that has most intents turned off. I can imagine that number to look better the more events are received, as well as bot that have high REST requests (I did not do any).

This PR has 3 areas of improvement:
A. Uses the binary representation of websocket text messages
B. Avoid intermediate string representation of JSON payloads received on the websocket
C. Avoid intermediate string representation of JSON request bodies for rest actions

These are based on the official recommendations from Jackson, see https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance#basics-things-you-should-do-anyway

### Test Environment
- 100 shards fully loaded
- Enabled intents: GUILD_MESSAGES, DIRECT_MESSAGES, GUILD_MESSAGE_REACTIONS, GUILD_EMOJIS, GUILD_MEMBERS, GUILD_VOICE_STATES
- No commands being executed

Unfortunately the measurements are not as clean as they could be, as I was in parallel working on performance improvements of my metrics. Nevertheless, I have a direct comparison of the most important changes (change B) of this PR to vanilla JDA v4.1.1_133:

Vanilla 4.1.1_133:

string

Including change A and change B:
bytearrays

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/data/DataArray.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/data/DataArray.java
@@ -27,6 +27,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
@@ -600,6 +601,25 @@ public class DataArray implements Iterable<Object>
     {
         data.remove(value);
         return this;
+    }
+
+    /**
+     * Serialize this object as JSON.
+     *
+     * @return a byte array containing the JSON representation of this object.
+     */
+    public byte[] toJson()
+    {
+        try
+        {
+            ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+            mapper.writeValue(outputStream, data);
+            return outputStream.toByteArray();
+        }
+        catch (IOException e)
+        {
+            throw new IllegalStateException(e);
+        }
     }
 
     @Override
--- a/src/main/java/net/dv8tion/jda/api/utils/data/DataArray.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/data/DataArray.java
@@ -31,6 +31,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
+import java.io.UncheckedIOException;
 import java.util.*;
 import java.util.function.Function;
 import java.util.function.UnaryOperator;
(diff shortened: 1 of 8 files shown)


napstr:
As mentioned, I was working on some improvements in my own code as well. In the end, I'm very happy with the result, but it is not that comparable to the other two measurements as I fixed a lot of other stuff as well.

Changes A, B and C + some other optimizations in my own code
all

src/main/java/net/dv8tion/jda/api/utils/data/DataArray.java
```diff
@@ -602,6 +603,25 @@ public DataArray remove(@Nullable Object value)
...
+    {
+        try
+        {
+            ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+            mapper.writeValue(outputStream, data);
+            return outputStream.toByteArray();
+        }
+        catch (IOException e)
+        {
+            throw new IllegalStateException(e);
```
MinnDevelopment, line 621: I'd use `UncheckedIOException` instead even if it can't be thrown.

src/main/java/net/dv8tion/jda/api/utils/data/DataObject.java
```diff
@@ -620,6 +646,25 @@ else if (value instanceof DataArray)
...
+    {
+        try
+        {
+            ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+            mapper.writeValue(outputStream, data);
+            return outputStream.toByteArray();
+        }
+        catch (IOException e)
+        {
+            throw new ParsingException(e);
```
MinnDevelopment, line 664: Why did you use a different exception here?

napstr:
> I'm a bit concerned about encoding. Do these methods properly use UTF-8 encoding regardless of runtime properties?

Should be fine, Jackson appears to be doing some autodetection with UTF-8 being the default. See issues like https://github.com/FasterXML/jackson-core/issues/222

src/main/java/net/dv8tion/jda/api/utils/data/DataArray.java
napstr: Just copied the `toString()` method from right below, no big thoughts went into this. I don't understand the structure and failure handling of JDA good enough to make an educated decision anyways, feel free to edit at will.

src/main/java/net/dv8tion/jda/api/utils/data/DataObject.java
napstr: Just copied the `toString()` method from right below, no big thoughts went into this. I don't understand the structure and failure handling of JDA good enough to make an educated decision anyways, feel free to edit at will.

PR #1270, opened 2020-04-21
Fix invalid link in Permission enum

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Removes a link in the `Permission` enum, which is no longer valid.
If there is a replacement for this information, let me know and I implement it.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/Permission.java
+++ b/src/main/java/net/dv8tion/jda/api/Permission.java
@@ -137,8 +137,6 @@ public enum Permission
 
     /**
      * The binary offset of the permission.
-     * <br>For more information about Discord's offset system refer to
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/Permission.java
```diff
@@ -137,8 +137,6 @@ public String getName()
 
     /**
      * The binary offset of the permission.
-     * <br>For more information about Discord's offset system refer to
-     * <a href="https://discordapi.readthedocs.org/en/latest/reference/channels/permissions.html#permissions-number">Discord Permission Numbers</a>.
```
MinnDevelopment, line 141: https://discordapp.com/developers/docs/topics/permissions

Andre601: Added

PR #1271, opened 2020-04-21
Add User#getFlags

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This implements the `public_flags` value for the User object, which essentially contains all [user flags](https://discord.com/developers/docs/resources/user#user-object-user-flags) with some exceptions like what type of nitro (or if they even have nitro) the user has and some other hidden flags.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/Permission.java
+++ b/src/main/java/net/dv8tion/jda/api/Permission.java
@@ -223,7 +223,7 @@ public enum Permission
 
     /**
      * A set of all {@link net.dv8tion.jda.api.Permission Permissions} that are specified by this raw long representation of
-     * permissions. This is best used with the getRaw methods in {@link net.dv8tion.jda.api.entities.Role Role} or
+     * permissions. The is best used with the getRaw methods in {@link net.dv8tion.jda.api.entities.Role Role} or
      * {@link net.dv8tion.jda.api.entities.PermissionOverride PermissionOverride}.
      *
      * <p>Example: {@link net.dv8tion.jda.api.entities.Role#getPermissionsRaw() Role.getPermissionsRaw()}
--- a/src/main/java/net/dv8tion/jda/api/Permission.java
+++ b/src/main/java/net/dv8tion/jda/api/Permission.java
@@ -223,7 +223,7 @@ public enum Permission
 
     /**
      * A set of all {@link net.dv8tion.jda.api.Permission Permissions} that are specified by this raw long representation of
-     * permissions. The is best used with the getRaw methods in {@link net.dv8tion.jda.api.entities.Role Role} or
+     * permissions. This is best used with the getRaw methods in {@link net.dv8tion.jda.api.entities.Role Role} or
      * {@link net.dv8tion.jda.api.entities.PermissionOverride PermissionOverride}.
      *
      * <p>Example: {@link net.dv8tion.jda.api.entities.Role#getPermissionsRaw() Role.getPermissionsRaw()}
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -225,4 +225,24 @@ public interface User extends IMentionable, IFakeable
      */
     @Nonnull
     JDA getJDA();
+    
+    public enum Flags{
+        NONE(0, "None"),
+        EMPLOYEE(1, "Discord Employee"),
+        PARTNER(2, "Discord Partner"),
+        HYPESQUAD(4, "HypeSquad Events"),
+        BUG_HUNTER_1(8, "Bug Hunter Level 1"),
+        BRAVERY(64, "HypeSquad House Bravery"),
+        BRILLIANCE(128, "HypeSquad House Brilliance"),
+        BALANCE(256, "HypeSquad House Balance"),
+        SYSTEM(4096, "System"),
+        BUG_HUNTER_2(16384, "Bug Hunter Level 2");
+        
+        private final int key;
+        private final String name;
+        
+        Flags(int key, String name){
+            this.key = key;
+            this.name = name;
+    }
 }
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -225,24 +225,63 @@ public interface User extends IMentionable, IFakeable
      */
     @Nonnull
     JDA getJDA();
-    
-    public enum Flags{
-        NONE(0, "None"),
-        EMPLOYEE(1, "Discord Employee"),
-        PARTNER(2, "Discord Partner"),
-        HYPESQUAD(4, "HypeSquad Events"),
-        BUG_HUNTER_1(8, "Bug Hunter Level 1"),
-        BRAVERY(64, "HypeSquad House Bravery"),
-        BRILLIANCE(128, "HypeSquad House Brilliance"),
-        BALANCE(256, "HypeSquad House Balance"),
-        SYSTEM(4096, "System"),
+
+    /**
+     * Represents the bit offsets used by Discord for public flags
+     */
+    enum Flags{
+        NONE(            0, "None"),
+        EMPLOYEE(        1, "Discord Employee"),
+        PARTNER(         2, "Discord Partner"),
+        HYPESQUAD(       4, "HypeSquad Events"),
+        BUG_HUNTER_1(    8, "Bug Hunter Level 1"),
+        BRAVERY(        64, "HypeSquad House Bravery"),
+        BRILLIANCE(    128, "HypeSquad House Brilliance"),
+        BALANCE(       256, "HypeSquad House Balance"),
+        SYSTEM(       4096, "System"),
         BUG_HUNTER_2(16384, "Bug Hunter Level 2");
         
-        private final int key;
+        private final int offset;
+        private final long raw;
         private final String name;
         
-        Flags(int key, String name){
-            this.key = key;
+        Flags(int offset, @Nonnull String name)
+        {
+            this.offset = offset;
+            this.raw = 1 << offset;
             this.name = name;
+        }
+
+        /**
+         * The readable name as used in the Discord Client.
+         * 
+         * @return The readable name of this {@link net.dv8tion.jda.api.entities.User.Flags Flag}.
+         */
+        @Nonnull
+        public String getName()
+        {
+            return this.name;
+        }
+
+        /**
+         * The binary offset of the flag.
+         * 
+         * @return The offset that represents this {@link net.dv8tion.jda.api.entities.User.Flags Flag}.
+         */
+        public int getOffset()
+        {
+            return offset;
+        }
+
+        /**
+         * The value of this flag when viewed as raw value.
+         * <br>This is equivalent to: <code>1 {@literal <<} {@link #getOffset()}</code>
+         * 
+         * @return The raw value of this specific flag.
+         */
+        public long getRawValue()
+        {
+            return raw;
+        }
     }
 }
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -230,16 +230,23 @@ public interface User extends IMentionable, IFakeable
      * Represents the bit offsets used by Discord for public flags
      */
     enum Flags{
-        NONE(            0, "None"),
-        EMPLOYEE(        1, "Discord Employee"),
-        PARTNER(         2, "Discord Partner"),
-        HYPESQUAD(       4, "HypeSquad Events"),
-        BUG_HUNTER_1(    8, "Bug Hunter Level 1"),
-        BRAVERY(        64, "HypeSquad House Bravery"),
-        BRILLIANCE(    128, "HypeSquad House Brilliance"),
-        BALANCE(       256, "HypeSquad House Balance"),
-        SYSTEM(       4096, "System"),
-        BUG_HUNTER_2(16384, "Bug Hunter Level 2");
+        NONE(                      0, "None"),
+        STAFF(                     1, "Discord Employee"),
+        PARTNER(                   2, "Discord Partner"),
+        HYPESQUAD(                 4, "HypeSquad Events"),
+        BUG_HUNTER_LEVEL_1(        8, "Bug Hunter Level 1"),
+        // Bravery
+        HYPESQUAD_ONLINE_HOUSE_1( 64, "HypeSquad Online House Bravery"),
+        // Brilliance
+        HYPESQUAD_ONLINE_HOUSE_2(128, "HypeSquad Online House Brilliance"),
+        // Balance
+        HYPESQUAD_ONLINE_HOUSE_3(256, "HypeSquad Online House Balance"),
+        PREMIUM_EARLY_SUPPORTER( 512, "Early Supporter"),
+        TEAM_USER(              1024, "Team User"),
+        SYSTEM(                 4096, "System User"),
+        BUG_HUNTER_LEVEL_2(    16384, "Bug Hunter Level 2"),
+        VERIFIED_BOT(          65536, "Verified Bot"),
+        VERIFIED_DEVELOPER(   131072, "Verified Developer");
         
         private final int offset;
         private final long raw;
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -230,7 +230,7 @@ public interface User extends IMentionable, IFakeable
      * Represents the bit offsets used by Discord for public flags
      */
     enum Flags {
-        UNKOWN(                   -1, "Unknown"),
+        UNKNOWN(                   -1, "Unknown"),
         STAFF(                     0, "Discord Employee"),
         PARTNER(                   1, "Discord Partner"),
         HYPESQUAD(                 2, "HypeSquad Events"),
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -18,6 +18,7 @@ package net.dv8tion.jda.api.entities;
 
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.requests.RestAction;
+import net.dv8tion.jda.internal.utils.Checks;
 
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
@@ -305,8 +306,39 @@ public interface User extends IMentionable, IFakeable
         {
             return raw;
         }
+
+        /**
+         * Gets the first {@link net.dv8tion.jda.api.entities.User.Flag Flag} relating to the provided offset.
+         * <br>If there is no {@link net.dv8tion.jda.api.entities.User.Flag Flag} that matches the provided offset,
+         * {@link net.dv8tion.jda.api.entities.User.Flag#UNKNOWN Flag.UNKNOWN} is returned.
+         * 
+         * @param  offset
+         *         The offset to match a {@link net.dv8tion.jda.api.entities.User.Flag Flag} to.
+         *         
+         * @return {@link net.dv8tion.jda.api.entities.User.Flag Flag} relating to the provided offset.
+         */
+        @Nonnull
+        public static Flag getFromOffset(int offset)
+        {
+            for (Flag flag : values())
+            {
+                if(flag.offset == offset)
+                    return flag;
+            }
+            return UNKNOWN;
+        }
         
-        public static EnumSet<Flag> getFlags(int flags)
+        /**
+         * A set of all {@link net.dv8tion.jda.api.entities.User.Flag Flags} that are specified by this raw long representation of
+         * flags.
+         * 
+         * @param  flags
+         *         The raw {@code long} representation if flags.
+         *         
+         * @return Possibly-empty EnumSet of {@link net.dv8tion.jda.api.entities.User.Flag Flags}.
+         */
+        @Nonnull
+        public static EnumSet<Flag> getFlags(long flags)
         {
             if(flags == 0)
                 return EnumSet.noneOf(Flag.class);
@@ -317,6 +349,45 @@ public interface User extends IMentionable, IFakeable
                     flagsSet.add(flag);
             }
             return flagsSet;
+        }
+
+        /**
+         * This is effectively the oposite of {@link #getFlags(long)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.Flag Flags}
+         * and returns the raw offset {@code long} representation of the flags.
+         * 
+         * @param  flags
+         *         The array of flags of which to form into the raw long representation.
+         *         
+         * @return Unsigned long representing the provided flags.
+         */
+        public static long getRaw(@Nonnull Flag... flags){
+            long raw = 0;
+            for (Flag flag : flags)
+            {
+                if (flag != null && flag != UNKNOWN)
+                    raw |= flag.raw;
+            }
+            
+            return raw;
+        }
+
+        /**
+         * This is effectively the oposite of {@link #getFlags(long)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.Flag Flags}
+         * and returns the raw offset {@code long} representation of the flags.
+         * <br>Example: {@code getRaw(EnumSet.of(Flag.STAFF, Flag.HYPESQUAD))}
+         *
+         * @param  flags
+         *         The Collection of flags of which to form into the raw long representation.
+         *
+         * @return Unsigned long representing the provided flags.
+         * 
+         * @see java.util.EnumSet EnumSet
+         */
+        public static long getRaw(@Nonnull Collection<Flag> flags)
+        {
+            Checks.notNull(flags, "Flag Collection");
+            
+            return getRaw(flags.toArray(EMPTY_FLAGS));
         }
     }
 }
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -331,7 +331,7 @@ public interface User extends IMentionable, IFakeable
         }
         
         /**
-         * A set of all {@link net.dv8tion.jda.api.entities.User.UserFlag Flags} that are specified by this raw long representation of
+         * A set of all {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} that are specified by this raw long representation of
          * flags.
          * 
          * @param  flags
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -262,12 +262,12 @@ public interface User extends IMentionable, IFakeable
         UNKNOWN(                   -1, "Unknown");
 
         /**
-         * Empty array of Flag enum, useful for optimized use in {@link java.util.Collection#toArray(Object[])}.
+         * Empty array of UserFlag enum, useful for optimized use in {@link java.util.Collection#toArray(Object[])}.
          */
         public static final UserFlag[] EMPTY_FLAGS = new UserFlag[0];
         
         private final int offset;
-        private final long raw;
+        private final int raw;
         private final String name;
 
         UserFlag(int offset, @Nonnull String name)
@@ -304,7 +304,7 @@ public interface User extends IMentionable, IFakeable
          * 
          * @return The raw value of this specific flag.
          */
-        public long getRawValue()
+        public int getRawValue()
         {
             return raw;
         }
@@ -331,7 +331,7 @@ public interface User extends IMentionable, IFakeable
         }
         
         /**
-         * A set of all {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} that are specified by this raw long representation of
+         * A set of all {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} that are specified by this raw int representation of
          * flags.
          * 
          * @param  flags
@@ -356,7 +356,7 @@ public interface User extends IMentionable, IFakeable
 
         /**
          * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
-         * and returns the raw offset {@code long} representation of the flags.
+         * and returns the raw offset {@code int} representation of the flags.
          * 
          * @param  flags
          *         The array of flags of which to form into the raw long representation.
@@ -376,7 +376,7 @@ public interface User extends IMentionable, IFakeable
 
         /**
          * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
-         * and returns the raw offset {@code long} representation of the flags.
+         * and returns the raw offset {@code int} representation of the flags.
          * <br>Example: {@code getRaw(EnumSet.of(UserFlag.STAFF, UserFlag.HYPESQUAD))}
          *
          * @param  flags
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -342,15 +342,16 @@ public interface User extends IMentionable, IFakeable
         @Nonnull
         public static EnumSet<UserFlag> getFlags(int flags)
         {
-            if(flags == 0)
-                return EnumSet.noneOf(UserFlag.class);
-            EnumSet<UserFlag> userFlags = EnumSet.noneOf(UserFlag.class);
-            for (UserFlag flag : UserFlag.values())
-            {
+            final EnumSet<UserFlag> foundFlags = EnumSet.noneOf(UserFlag.class);
+            
+            if (flags == 0)
+                return foundFlags; //empty
+            
+            for (UserFlag flag : values())
                 if (flag != UNKNOWN && (flags & flag.raw) == flag.raw)
-                    userFlags.add(flag);
-            }
-            return userFlags;
+                    foundFlags.add(flag);
+                    
+            return foundFlags;
         }
 
         /**
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -291,7 +291,7 @@ public interface User extends IMentionable, IFakeable
         /**
          * The binary offset of the flag.
          * 
-         * @return The offset that represents this {@link net.dv8tion.jda.api.entities.User.UserFlag Flag}.
+         * @return The offset that represents this {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag}.
          */
         public int getOffset()
         {
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -238,9 +238,9 @@ public interface User extends IMentionable, IFakeable
     EnumSet<UserFlag> getFlags();
 
     /**
-     * Returns the raw {@code int} value of the {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} of this user.
+     * Returns the bitmask representation of the {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} of this user.
      * 
-     * @return Integer representing the raw value of the user's flags.
+     * @return bitmask representation of the user's flags.
      */
     int getFlagsRaw();
 
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -230,7 +230,7 @@ public interface User extends IMentionable, IFakeable
     JDA getJDA();
 
     /**
-     * Returns the {@link net.dv8tion.jda.api.entities.User.UserFlag Flags} of this user.
+     * Returns the {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} of this user.
      * 
      * @return EnumSet containing the flags of the user.
      */
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -280,7 +280,7 @@ public interface User extends IMentionable, IFakeable
         /**
          * The readable name as used in the Discord Client.
          * 
-         * @return The readable name of this {@link net.dv8tion.jda.api.entities.User.UserFlag Flag}.
+         * @return The readable name of this {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag}.
          */
         @Nonnull
         public String getName()
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -238,6 +238,13 @@ public interface User extends IMentionable, IFakeable
     EnumSet<UserFlag> getFlags();
 
     /**
+     * Returns the raw {@code int} value of the {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} of this user.
+     * 
+     * @return Integer representing the raw value of the user's flags.
+     */
+    int getFlagsRaw();
+
+    /**
      * Represents the bit offsets used by Discord for public flags
      */
     enum UserFlag
@@ -280,7 +287,7 @@ public interface User extends IMentionable, IFakeable
         /**
          * The readable name as used in the Discord Client.
          * 
-         * @return The readable name of this {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag}.
+         * @return The readable name of this UserFlag.
          */
         @Nonnull
         public String getName()
@@ -291,7 +298,7 @@ public interface User extends IMentionable, IFakeable
         /**
          * The binary offset of the flag.
          * 
-         * @return The offset that represents this {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag}.
+         * @return The offset that represents this UserFlag.
          */
         public int getOffset()
         {
@@ -310,14 +317,14 @@ public interface User extends IMentionable, IFakeable
         }
 
         /**
-         * Gets the first {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} relating to the provided offset.
-         * <br>If there is no {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} that matches the provided offset,
+         * Gets the first UserFlag relating to the provided offset.
+         * <br>If there is no UserFlag that matches the provided offset,
          * {@link net.dv8tion.jda.api.entities.User.UserFlag#UNKNOWN UserFlag.UNKNOWN} is returned.
          * 
          * @param  offset
-         *         The offset to match a {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} to.
+         *         The offset to match a UserFlag to.
          *         
-         * @return {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} relating to the provided offset.
+         * @return UserFlag relating to the provided offset.
          */
         @Nonnull
         public static UserFlag getFromOffset(int offset)
@@ -331,13 +338,13 @@ public interface User extends IMentionable, IFakeable
         }
         
         /**
-         * A set of all {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} that are specified by this raw int representation of
+         * A set of all UserFlags that are specified by this raw int representation of
          * flags.
          * 
          * @param  flags
          *         The raw {@code int} representation if flags.
          *         
-         * @return Possibly-empty EnumSet of {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}.
+         * @return Possibly-empty EnumSet of UserFlags.
          */
         @Nonnull
         public static EnumSet<UserFlag> getFlags(int flags)
@@ -357,15 +364,20 @@ public interface User extends IMentionable, IFakeable
         }
 
         /**
-         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
+         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more UserFlags
          * and returns the raw offset {@code int} representation of the flags.
          * 
          * @param  flags
          *         The array of flags of which to form into the raw int representation.
          *         
-         * @return Unsigned int representing the provided flags.
+         * @return Unsigned bitmask representing the provided flags.
+         * 
+         * @throws java.lang.IllegalArgumentException
+         *         When the provided UsefLags are null.
          */
         public static int getRaw(@Nonnull UserFlag... flags){
+            Checks.noneNull(flags, "UserFlags");
+            
             int raw = 0;
             for (UserFlag flag : flags)
             {
@@ -384,7 +396,7 @@ public interface User extends IMentionable, IFakeable
          * @param  flags
          *         The Collection of flags of which to form into the raw int representation.
          *
-         * @return Unsigned int representing the provided flags.
+         * @return Unsigned bitmask representing the provided flags.
          * 
          * @see java.util.EnumSet EnumSet
          */
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -324,7 +324,7 @@ public interface User extends IMentionable, IFakeable
         {
             for (UserFlag flag : values())
             {
-                if(flag.offset == offset)
+                if (flag.offset == offset)
                     return flag;
             }
             return UNKNOWN;
@@ -348,8 +348,10 @@ public interface User extends IMentionable, IFakeable
                 return foundFlags; //empty
             
             for (UserFlag flag : values())
+            {
                 if (flag != UNKNOWN && (flags & flag.raw) == flag.raw)
                     foundFlags.add(flag);
+            }
                     
             return foundFlags;
         }
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -230,17 +230,17 @@ public interface User extends IMentionable, IFakeable
     JDA getJDA();
 
     /**
-     * Returns the {@link net.dv8tion.jda.api.entities.User.Flag Flags} of this user.
+     * Returns the {@link net.dv8tion.jda.api.entities.User.UserFlag Flags} of this user.
      * 
      * @return EnumSet containing the flags of the user.
      */
     @Nonnull
-    EnumSet<Flag> getFlags();
+    EnumSet<UserFlag> getFlags();
 
     /**
      * Represents the bit offsets used by Discord for public flags
      */
-    enum Flag
+    enum UserFlag
     {
         STAFF(                     0, "Discord Employee"),
         PARTNER(                   1, "Discord Partner"),
@@ -262,15 +262,15 @@ public interface User extends IMentionable, IFakeable
         UNKNOWN(                   -1, "Unknown");
 
         /**
-         * Empty array of Flag enum, useful for optimized use in {@link Collection#toArray(Object[])}.
+         * Empty array of Flag enum, useful for optimized use in {@link java.util.Collection#toArray(Object[])}.
          */
-        public static final Flag[] EMPTY_FLAGS = new Flag[0];
+        public static final UserFlag[] EMPTY_FLAGS = new UserFlag[0];
         
         private final int offset;
         private final long raw;
         private final String name;
 
-        Flag(int offset, @Nonnull String name)
+        UserFlag(int offset, @Nonnull String name)
         {
             this.offset = offset;
             this.raw = 1 << offset;
@@ -280,7 +280,7 @@ public interface User extends IMentionable, IFakeable
         /**
          * The readable name as used in the Discord Client.
          * 
-         * @return The readable name of this {@link Flag Flag}.
+         * @return The readable name of this {@link net.dv8tion.jda.api.entities.User.UserFlag Flag}.
          */
         @Nonnull
         public String getName()
@@ -291,7 +291,7 @@ public interface User extends IMentionable, IFakeable
         /**
          * The binary offset of the flag.
          * 
-         * @return The offset that represents this {@link Flag Flag}.
+         * @return The offset that represents this {@link net.dv8tion.jda.api.entities.User.UserFlag Flag}.
          */
         public int getOffset()
         {
@@ -310,19 +310,19 @@ public interface User extends IMentionable, IFakeable
         }
 
         /**
-         * Gets the first {@link net.dv8tion.jda.api.entities.User.Flag Flag} relating to the provided offset.
-         * <br>If there is no {@link net.dv8tion.jda.api.entities.User.Flag Flag} that matches the provided offset,
-         * {@link net.dv8tion.jda.api.entities.User.Flag#UNKNOWN Flag.UNKNOWN} is returned.
+         * Gets the first {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} relating to the provided offset.
+         * <br>If there is no {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} that matches the provided offset,
+         * {@link net.dv8tion.jda.api.entities.User.UserFlag#UNKNOWN Flag.UNKNOWN} is returned.
          * 
          * @param  offset
-         *         The offset to match a {@link net.dv8tion.jda.api.entities.User.Flag Flag} to.
+         *         The offset to match a {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} to.
          *         
-         * @return {@link net.dv8tion.jda.api.entities.User.Flag Flag} relating to the provided offset.
+         * @return {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} relating to the provided offset.
          */
         @Nonnull
-        public static Flag getFromOffset(int offset)
+        public static UserFlag getFromOffset(int offset)
         {
-            for (Flag flag : values())
+            for (UserFlag flag : values())
             {
                 if(flag.offset == offset)
                     return flag;
@@ -331,30 +331,30 @@ public interface User extends IMentionable, IFakeable
         }
         
         /**
-         * A set of all {@link net.dv8tion.jda.api.entities.User.Flag Flags} that are specified by this raw long representation of
+         * A set of all {@link net.dv8tion.jda.api.entities.User.UserFlag Flags} that are specified by this raw long representation of
          * flags.
          * 
          * @param  flags
          *         The raw {@code long} representation if flags.
          *         
-         * @return Possibly-empty EnumSet of {@link net.dv8tion.jda.api.entities.User.Flag Flags}.
+         * @return Possibly-empty EnumSet of {@link UserFlag Flags}.
          */
         @Nonnull
-        public static EnumSet<Flag> getFlags(int flags)
+        public static EnumSet<UserFlag> getFlags(int flags)
         {
             if(flags == 0)
-                return EnumSet.noneOf(Flag.class);
-            EnumSet<Flag> flagsSet = EnumSet.noneOf(Flag.class);
-            for (Flag flag : Flag.values())
+                return EnumSet.noneOf(UserFlag.class);
+            EnumSet<UserFlag> userFlags = EnumSet.noneOf(UserFlag.class);
+            for (UserFlag flag : UserFlag.values())
             {
                 if (flag != UNKNOWN && (flags & flag.raw) == flag.raw)
-                    flagsSet.add(flag);
+                    userFlags.add(flag);
             }
-            return flagsSet;
+            return userFlags;
         }
 
         /**
-         * This is effectively the oposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.Flag Flags}
+         * This is effectively the oposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}
          * and returns the raw offset {@code long} representation of the flags.
          * 
          * @param  flags
@@ -362,9 +362,9 @@ public interface User extends IMentionable, IFakeable
          *         
          * @return Unsigned long representing the provided flags.
          */
-        public static int getRaw(@Nonnull Flag... flags){
+        public static int getRaw(@Nonnull UserFlag... flags){
             int raw = 0;
-            for (Flag flag : flags)
+            for (UserFlag flag : flags)
             {
                 if (flag != null && flag != UNKNOWN)
                     raw |= flag.raw;
@@ -374,7 +374,7 @@ public interface User extends IMentionable, IFakeable
         }
 
         /**
-         * This is effectively the oposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.Flag Flags}
+         * This is effectively the oposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}
          * and returns the raw offset {@code long} representation of the flags.
          * <br>Example: {@code getRaw(EnumSet.of(Flag.STAFF, Flag.HYPESQUAD))}
          *
@@ -385,7 +385,7 @@ public interface User extends IMentionable, IFakeable
          * 
          * @see java.util.EnumSet EnumSet
          */
-        public static int getRaw(@Nonnull Collection<Flag> flags)
+        public static int getRaw(@Nonnull Collection<UserFlag> flags)
         {
             Checks.notNull(flags, "Flag Collection");
             
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -365,13 +365,13 @@ public interface User extends IMentionable, IFakeable
 
         /**
          * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more UserFlags
-         * and returns the raw offset {@code int} representation of the flags.
+         * and returns the bitmask representation of the flags.
          * 
          * @param  flags
          *         The array of flags of which to form into the raw int representation.
          *
          * @throws java.lang.IllegalArgumentException
-         *         When the provided UserFLags are null.
+         *         When the provided UserFlags are null.
          *         
          * @return bitmask representing the provided flags.
          */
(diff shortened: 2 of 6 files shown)


duncte123:
Why didn't you add the early supporter one?

Andre601:
> Why didn't you add the early supporter one?

I'm not sure if all flags are available, because people said that some aren't insider the public_flags, and my assumption is, that any kind of nitro indicator (including ea) are not in this...
There's sadly no proper documentation (yet) about which flags aren't included, because if public_flags = flags, then why not just use flags?

Andre601:
@duncte123 I now added all flags that are considered "public" (based on [this file](https://github.com/vDelite/DiscordLists/blob/master/flags.md))

I also renamed some flags to make them match the internal names Discord uses, although I think having the names of the HypeSquad houses would be better.

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +225,70 @@ default String getEffectiveAvatarUrl()
...
+        VERIFIED_DEVELOPER(   131072, "Verified Developer");
+        
+        private final int offset;
+        private final long raw;
+        private final String name;
+        
+        Flags(int offset, @Nonnull String name)
+        {
+            this.offset = offset;
+            this.raw = 1 << offset;
```
boomboompower, line 258: This value is incorrect, you are already passing the raw offset as the first paramater;
Example: the "raw" varaible for NONE will be 1, 2 for STAFF, 4 fpr HYPESQUAD.

To fix this issue you should change each Enums value to the offset instead of the raw bit shifted value

boomboompower, line 258: This value here is incorrect, you are already passing the raw offset as the first paramater;
Example: the "raw" varaible for NONE will be 1, 2 for STAFF, 4 fpr HYPESQUAD.

To fix this issue you should change each Enums value to the offset instead of the raw bit shifted value

boomboompower: I've fixed it in a PR [here](https://github.com/Andre601/JDA/pull/3)

The "Flags" enum should probably be renamed to UserFlags as well

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -230,16 +230,23 @@ default String getEffectiveAvatarUrl()
...
+        STAFF(                     1, "Discord Employee"),
+        PARTNER(                   2, "Discord Partner"),
+        HYPESQUAD(                 4, "HypeSquad Events"),
+        BUG_HUNTER_LEVEL_1(        8, "Bug Hunter Level 1"),
+        // Bravery
+        HYPESQUAD_ONLINE_HOUSE_1( 64, "HypeSquad Online House Bravery"),
+        // Brilliance
+        HYPESQUAD_ONLINE_HOUSE_2(128, "HypeSquad Online House Brilliance"),
+        // Balance
+        HYPESQUAD_ONLINE_HOUSE_3(256, "HypeSquad Online House Balance"),
```
MinnDevelopment, line 243: Let's call them `HYPESQUAD_BRAVERY` and such.

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -230,16 +230,23 @@ default String getEffectiveAvatarUrl()
...
-        EMPLOYEE(        1, "Discord Employee"),
-        PARTNER(         2, "Discord Partner"),
-        HYPESQUAD(       4, "HypeSquad Events"),
-        BUG_HUNTER_1(    8, "Bug Hunter Level 1"),
-        BRAVERY(        64, "HypeSquad House Bravery"),
-        BRILLIANCE(    128, "HypeSquad House Brilliance"),
-        BALANCE(       256, "HypeSquad House Balance"),
-        SYSTEM(       4096, "System"),
-        BUG_HUNTER_2(16384, "Bug Hunter Level 2");
+        NONE(                      0, "None"),
```
MinnDevelopment, line 233: How would this be used?

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -230,16 +230,23 @@ default String getEffectiveAvatarUrl()
...
+        // Brilliance
+        HYPESQUAD_ONLINE_HOUSE_2(128, "HypeSquad Online House Brilliance"),
+        // Balance
+        HYPESQUAD_ONLINE_HOUSE_3(256, "HypeSquad Online House Balance"),
+        PREMIUM_EARLY_SUPPORTER( 512, "Early Supporter"),
+        TEAM_USER(              1024, "Team User"),
+        SYSTEM(                 4096, "System User"),
+        BUG_HUNTER_LEVEL_2(    16384, "Bug Hunter Level 2"),
+        VERIFIED_BOT(          65536, "Verified Bot"),
+        VERIFIED_DEVELOPER(   131072, "Verified Developer");
```
MinnDevelopment, line 249: Instead of using raw ints, please use the offsets and calculate the raw value in the constructor.

Andre601:
@MinnDevelopment How should we treat `NONE` then when using the offset?
Because `NONE` is a flag in Discord, using the value 0 and I think using `-1` should be reserved for a `UNKNOWN` flag.

I suggested [here](https://github.com/Andre601/JDA/pull/3) that we perhaps should have both the offset and the raw value? Would kind of make sense considering Discord provides the flags as their raw (total) value.

MinnDevelopment:
> @MinnDevelopment How should we treat NONE then when using the offset?
Because NONE is a flag in Discord, using the value 0 and I think using -1 should be reserved for a UNKNOWN flag.

NONE isn't a flag, it's a bitmask representing no flags being set. If we have something like `getFlags(): EnumSet<UserFlag>` then it makes no sense to return a `EnumSet.of(NONE)` rather than just an empty set. There is no reasonable way to use this. `UNKNOWN` is a placeholder constant used when we can't resolve the flag to an enum constant.

> I suggested here that we perhaps should have both the offset and the raw value? Would kind of make sense considering Discord provides the flags as their raw (total) value.

The offset gives a good and understandable code. Using raw values is confusing and can lead to bugs. You should simply use the offset to do `1 << offset` in a `getRaw()`. See [Permission](https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/api/Permission.java#L120). The [documentation](https://discordapp.com/developers/docs/resources/user#user-object-user-flags) also uses offsets for the same reason.

src/main/java/net/dv8tion/jda/api/entities/User.java
Andre601: Renamed it to `UNKNOWN` with -1 as value (Placeholder).

src/main/java/net/dv8tion/jda/api/entities/User.java
Andre601: Was renamed.
https://github.com/DV8FromTheWorld/JDA/pull/3

src/main/java/net/dv8tion/jda/api/entities/User.java
Andre601: Was changed int https://github.com/DV8FromTheWorld/JDA/pull/3

Andre601:
I merged the suggested changes from @boomboompower in https://github.com/Andre601/JDA/pull/3 which also covered most, if not all requested changes by Minn

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -229,29 +229,29 @@ default String getEffectiveAvatarUrl()
...
-        HYPESQUAD_ONLINE_HOUSE_3(256, "HypeSquad Online House Balance"),
-        PREMIUM_EARLY_SUPPORTER( 512, "Early Supporter"),
-        TEAM_USER(              1024, "Team User"),
-        SYSTEM(                 4096, "System User"),
-        BUG_HUNTER_LEVEL_2(    16384, "Bug Hunter Level 2"),
-        VERIFIED_BOT(          65536, "Verified Bot"),
-        VERIFIED_DEVELOPER(   131072, "Verified Developer");
-        
+    enum Flags {
+        UNKOWN(                   -1, "Unknown"),
```
MinnDevelopment, line 233:
```suggestion
        UNKNOWN(                   -1, "Unknown"),
```

Andre601:
I think now should I implement this into the User object somehow?
I never done this before, so I'm unsure where I should alter stuff to implement this. A helping hand on this (i.e. guiding me into the right direction) would be appreciated.

MinnDevelopment:
The user objects are created in the `EntityBuilder` and the flags should be stored as an `int` in the `UserImpl` class. The getter is declared in the `User` interface and should return an `EnumSet`. The implementation will need the same getter plus one for the `int` itself, used to compare against in the `updateUser` method in `EntityBuilder` to fire an update event.

Andre601:
Would the Permission enum be a good example for the Getter?

Andre601:
I just noticed something.

In the Permission enum do you have this code:
```java
    @Nonnull
    public static EnumSet<Permission> getPermissions(long permissions)
    {
        if (permissions == 0)
            return EnumSet.noneOf(Permission.class);
        EnumSet<Permission> perms = EnumSet.noneOf(Permission.class);
        for (Permission perm : Permission.values())
        {
            if (perm != UNKNOWN && (permissions & perm.raw) == perm.raw)
                perms.add(perm);
        }
        return perms;
    }
```

Wouldn't it be somewhat better to have it like this?
```java
    @Nonnull
    public static EnumSet<Permission> getPermissions(long permissions)
    {
        EnumSet<Permission> perms = EnumSet.noneOf(Permission.class);
        if (permissions == 0)
            return perms;
        for (Permission perm : Permission.values())
        {
            if (perm != UNKNOWN && (permissions & perm.raw) == perm.raw)
                perms.add(perm);
        }
        return perms;
    }
```

MinnDevelopment:
Permissions are a good example for flags, yes.

Andre601:
I now implemented a `getFlags()` of type `EnumSet<Flag>` to the User Object and updated the UserImpl accordingly.

Right now am I stuck with the following things:
- Not sure how I should set the flags (int) and enum set of it for UserImpl
- Not sure how to best return the flags (int)
- Best way to compare values in the EntityBuilder... Similar as the one used in `updateUser`? (`!Objects.equasls(Object, Object)`)?

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
```diff
@@ -216,6 +231,12 @@ public UserImpl setFake(boolean fake)
         this.fake = fake;
         return this;
     }
+    
+    public UserImpl setFlags(int flags){
```
MinnDevelopment, line 235:
```suggestion
    public UserImpl setFlags(int flags)
    {
```

Andre601:
I've added a `UserUpdateFlagsEvent` that would get triggered when the user's flags get updated.

Not sure if I made everything alright tho.

src/main/java/net/dv8tion/jda/api/Permission.java
```diff
@@ -223,7 +223,7 @@ public static Permission getFromOffset(int offset)
 
     /**
      * A set of all {@link net.dv8tion.jda.api.Permission Permissions} that are specified by this raw long representation of
-     * permissions. The is best used with the getRaw methods in {@link net.dv8tion.jda.api.entities.Role Role} or
+     * permissions. This is best used with the getRaw methods in {@link net.dv8tion.jda.api.entities.Role Role} or
```
Andre601, line 226: I wasn't sure if this was a typo (Probably was?) so I corrected this...
Or should it be `The method is ...`?

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,168 @@ default String getEffectiveAvatarUrl()
      */
     @Nonnull
     JDA getJDA();
+
+    /**
+     * Returns the {@link net.dv8tion.jda.api.entities.User.Flag Flags} of this user.
```
HydroPage90, line 233: Why don't you call it UserFlag? Could be a bit more readable

src/main/java/net/dv8tion/jda/api/entities/User.java
HydroPage90: @boomboompower Agreed. UserFlag would be more readable

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,168 @@ default String getEffectiveAvatarUrl()
...
+        @Nonnull
+        public String getName()
+        {
+            return this.name;
+        }
+
+        /**
+         * The binary offset of the flag.
+         * 
+         * @return The offset that represents this {@link Flag Flag}.
```
HydroPage90, line 294: The [JDA Structure Guide](https://github.com/DV8FromTheWorld/JDA/wiki/6)-JDA-Structure-Guide) advises against documenting with imported names. I'm not sure if it doesn't apply here for some reason, but I'm mentioning it just in case

Andre601: I blame IntelliJ for auto-cutting the {@link} parameter as it had the full path before.

HydroPage90: Fair enough

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,168 @@ default String getEffectiveAvatarUrl()
      */
     @Nonnull
     JDA getJDA();
+
+    /**
+     * Returns the {@link net.dv8tion.jda.api.entities.User.UserFlag Flags} of this user.
```
HydroPage90, line 233:
```suggestion
     * Returns the {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} of this user.
```

Probably change these to accommodate the recent commit

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,168 @@ default String getEffectiveAvatarUrl()
...
+         * <br>If there is no {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} that matches the provided offset,
+         * {@link net.dv8tion.jda.api.entities.User.UserFlag#UNKNOWN Flag.UNKNOWN} is returned.
+         * 
+         * @param  offset
+         *         The offset to match a {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} to.
+         *         
+         * @return {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} relating to the provided offset.
+         */
+        @Nonnull
+        public static UserFlag getFromOffset(int offset)
```
HydroPage90, line 323: I wonder what a valid use-case for this would be. When would you work with the raw offsets in client-code? Isn't that exposing the implementation a bit too much?

What I'm saying is: if you're implementing a `UserFlag` enum to encapsulate the `public_flag` numbers, why would you allow the client code to still work with the numbers? Seems a bit off to me, I'd like to hear your thoughts

Andre601: Most of the code is based on the `Permission` enum, which also has a `getFromOffset` method for this.

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,168 @@ default String getEffectiveAvatarUrl()
...
+        {
+            if(flags == 0)
+                return EnumSet.noneOf(UserFlag.class);
+            EnumSet<UserFlag> userFlags = EnumSet.noneOf(UserFlag.class);
+            for (UserFlag flag : UserFlag.values())
+            {
+                if (flag != UNKNOWN && (flags & flag.raw) == flag.raw)
+                    userFlags.add(flag);
+            }
+            return userFlags;
```
HydroPage90, line 353:
```suggestion
            final EnumSet<UserFlag> foundFlags = EnumSet.noneOf(UserFlag.class);
            
            if (flags == 0)
                return foundFlags; //empty
            
            for (UserFlag flag : values())
                if (flag != UNKNOWN && (flags & flag.raw) == flag.raw)
                    foundFlags.add(flag);
                    
            return foundFlags;
```

Not sure if this would help a bit, or if you want to keep the look of the method the exact same as in `Permissions`

Andre601: I also asked Minn why it is like that and not different (See [here](https://github.com/DV8FromTheWorld/JDA/pull/1271#issuecomment-619544061))
No idea why this comment was marked as off-topic.

I myself can't see a reason why `EnumSet.noneOf(Class)` is used twice as it could just be used once and modified later on...
I'll merge this suggestion and probably update the Permission class accordingly, unless Minn is against this.

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+            
+            for (UserFlag flag : values())
+                if (flag != UNKNOWN && (flags & flag.raw) == flag.raw)
+                    foundFlags.add(flag);
+                    
+            return foundFlags;
+        }
+
+        /**
+         * This is effectively the oposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}
```
HydroPage90, line 358:
```suggestion
         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}
```

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,77 @@
...
+    @Nullable
+    public Integer getNewFlags(){
+        return getNewValue();
+    }
+
+    /**
+     * The new Set of {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}.
+     *
+     * @return The previous EnumSet of {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}.
+     */
```
HydroPage90, line 72:
```suggestion
    /**
     * @return The new EnumSet of {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}.
     */
```

You said "previous" before, even though this is the method for the new ones

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,77 @@
...
+    @Nullable
+    public EnumSet<User.UserFlag> getOlfFlagSet(){
+        return previous == null ? null : User.UserFlag.getFlags(previous);
+    }
+
+    /**
+     * The new raw {@code Integer} representation of the set {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}.
+     * 
+     * @return The new {@code Integer} representation of the {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}.
+     */
```
HydroPage90, line 62:
```suggestion
    /**
     * @return The new {@code Integer} representation of the {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}.
     */
```

src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
```diff
@@ -171,6 +171,7 @@ public void onUserUpdateActivityOrder(@Nonnull UserUpdateActivityOrderEvent even
     public void onUserTyping(@Nonnull UserTypingEvent event) {}
     public void onUserActivityStart(@Nonnull UserActivityStartEvent event) {}
     public void onUserActivityEnd(@Nonnull UserActivityEndEvent event) {}
+    public void onUserUpdateFlagsEvent(@Nonnull UserUpdateFlagsEvent event) {}
```
HydroPage90, line 174:
```suggestion
    public void onUserUpdateFlags(@Nonnull UserUpdateFlagsEvent event) {}
```

Follow the naming convention

src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
```diff
@@ -459,6 +460,8 @@ else if (event instanceof UserActivityStartEvent)
             onUserActivityStart((UserActivityStartEvent) event);
         else if (event instanceof UserActivityEndEvent)
             onUserActivityEnd((UserActivityEndEvent) event);
+        else if (event instanceof UserUpdateFlagsEvent)
+            onUserUpdateFlagsEvent((UserUpdateFlagsEvent) event);
```
HydroPage90, line 464:
```suggestion
            onUserUpdateFlags((UserUpdateFlagsEvent) event);
```

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -367,6 +371,15 @@ public void updateUser(UserImpl userObj, DataObject user)
                     jda, responseNumber,
                     userObj, oldAvatar));
         }
+        
+        if (!Objects.equals(oldFlags, newFlags))
```
HydroPage90, line 375: Why not just simple `!=`? This is less readable

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+        public static int getRaw(@Nonnull UserFlag... flags){
+            int raw = 0;
+            for (UserFlag flag : flags)
+            {
+                if (flag != null && flag != UNKNOWN)
+                    raw |= flag.raw;
+            }
+            
+            return raw;
+        }
```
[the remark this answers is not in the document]
Andre601: Long is from when this was a long and not an integer.
Also, this is - like many other things - based on the Permission enum, so I'll keep it like that for consistency.

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+         */
+        @Nonnull
+        public static UserFlag getFromOffset(int offset)
+        {
+            for (UserFlag flag : values())
+            {
+                if(flag.offset == offset)
+                    return flag;
+            }
+            return UNKNOWN;
```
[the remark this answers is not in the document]
Andre601: See my previous post

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+         * A set of all {@link net.dv8tion.jda.api.entities.User.UserFlag Flags} that are specified by this raw long representation of
+         * flags.
+         * 
+         * @param  flags
+         *         The raw {@code long} representation if flags.
+         *         
+         * @return Possibly-empty EnumSet of {@link UserFlag Flags}.
+         */
+        @Nonnull
+        public static EnumSet<UserFlag> getFlags(int flags)
```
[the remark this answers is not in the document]
Andre601: It's int.
It was previously long, but I changed it, because Discord seems to send the flags as Integer and not Long.

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,77 @@
...
+    
+    public UserUpdateFlagsEvent(@Nonnull JDA api, long responseNumber, @Nonnull User user, int oldFlags)
+    {
+        super(api, responseNumber, user, oldFlags, User.UserFlag.getRaw(user.getFlags()), IDENTIFIER);
+    }
+
+    /**
+     * The previous raw {@code Integer} representation of the set {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}.
+     * 
+     * @return The previous {@code Integer} representation of the {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}.
```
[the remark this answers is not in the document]
Andre601: I don't see where this is wrong.
I could rephrase it but don't see a reason as to why I should leave the comment itself away.

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * The previous Set of {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}.
+     * 
+     * @return The previous EnumSet of {@link net.dv8tion.jda.api.entities.User.UserFlag Flags}.
+     */
+    @Nullable
+    public EnumSet<User.UserFlag> getOlfFlagSet(){
+        return previous == null ? null : User.UserFlag.getFlags(previous);
+    }
```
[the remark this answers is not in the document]
Andre601: Shouldn't it be UserFlags instead of Flags? 😉

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
Andre601: Because the above one also uses Objects.equals, so I do this here too for consistency.

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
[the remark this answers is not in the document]
HydroPage90: The comment basically says "This returns the integer" and then you have a `@return` tag that literally says "This returns the integer". The comment itself feels redundant if the purpose can be more literally specified by other means, no?

src/main/java/net/dv8tion/jda/api/entities/User.java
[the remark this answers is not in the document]
HydroPage90: The documentation still says it's a long, that's what I was talking about

src/main/java/net/dv8tion/jda/api/entities/User.java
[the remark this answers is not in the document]
HydroPage90: Understood

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+         * <br>Example: {@code getRaw(EnumSet.of(Flag.STAFF, Flag.HYPESQUAD))}
+         *
+         * @param  flags
+         *         The Collection of flags of which to form into the raw long representation.
+         *
+         * @return Unsigned long representing the provided flags.
+         * 
+         * @see java.util.EnumSet EnumSet
+         */
+        public static int getRaw(@Nonnull Collection<UserFlag> flags)
```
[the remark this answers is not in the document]
HydroPage90: Again, the `@return` still claims `long`

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,69 @@
...
+    public Integer getOldFlags()
+    {
+        return getOldValue();
+    }
+
+    /**
+     * @return Possibly-null EnumSet of previous {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}.
+     */
+    @Nullable
+    public EnumSet<User.UserFlag> getOlfFlagSet(){
```
HydroPage90, line 50:
```suggestion
    public EnumSet<User.UserFlag> getOldFlagSet(){
```

Still misspelled

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+
+        /**
+         * Gets the first {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} relating to the provided offset.
+         * <br>If there is no {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} that matches the provided offset,
+         * {@link net.dv8tion.jda.api.entities.User.UserFlag#UNKNOWN Flag.UNKNOWN} is returned.
+         * 
+         * @param  offset
+         *         The offset to match a {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} to.
+         *         
+         * @return {@link net.dv8tion.jda.api.entities.User.UserFlag Flag} relating to the provided offset.
```
HydroPage90, line 320:
```suggestion
         * Gets the first {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} relating to the provided offset.
         * <br>If there is no {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} that matches the provided offset,
         * {@link net.dv8tion.jda.api.entities.User.UserFlag#UNKNOWN UserFlag.UNKNOWN} is returned.
         * 
         * @param  offset
         *         The offset to match a {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} to.
         *         
         * @return {@link net.dv8tion.jda.api.entities.User.UserFlag FlagFlag} relating to the provided offset.
```

This is tedious for me. Can you do it in your IDE?

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+        }
+
+        /**
+         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
+         * and returns the raw offset {@code int} representation of the flags.
+         * 
+         * @param  flags
+         *         The array of flags of which to form into the raw long representation.
+         *         
+         * @return Unsigned long representing the provided flags.
```
HydroPage90, line 364:
```suggestion
         * @return Unsigned {@code int} representing the provided flags.
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+
+        /**
+         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
+         * and returns the raw offset {@code int} representation of the flags.
+         * <br>Example: {@code getRaw(EnumSet.of(UserFlag.STAFF, UserFlag.HYPESQUAD))}
+         *
+         * @param  flags
+         *         The Collection of flags of which to form into the raw long representation.
+         *
+         * @return Unsigned long representing the provided flags.
```
HydroPage90, line 385:
```suggestion
         * @return Unsigned {@code int} representing the provided flags.
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+            }
+            return UNKNOWN;
+        }
+        
+        /**
+         * A set of all {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} that are specified by this raw int representation of
+         * flags.
+         * 
+         * @param  flags
+         *         The raw {@code long} representation if flags.
```
HydroPage90, line 338:
```suggestion
         *         The raw {@code int} representation if flags.
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+                    
+            return foundFlags;
+        }
+
+        /**
+         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
+         * and returns the raw offset {@code int} representation of the flags.
+         * 
+         * @param  flags
+         *         The array of flags of which to form into the raw long representation.
```
HydroPage90, line 362:
```suggestion
         *         The array of flags of which to form into the raw {@code int} representation.
```

HydroPage90:
Why is this still a draft? Do you have anything else in mind?

Andre601:
It's a draft, because I'm not sure if everything was added, that has to be added.

HydroPage90:
How do we make sure, then? Shall we review request Minn?

Andre601:
That is probably the best option here.

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+         *         The offset to match a {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} to.
+         *         
+         * @return {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} relating to the provided offset.
+         */
+        @Nonnull
+        public static UserFlag getFromOffset(int offset)
+        {
+            for (UserFlag flag : values())
+            {
+                if(flag.offset == offset)
```
HydroPage90, line 327:
```suggestion
                if (flag.offset == offset)
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,169 @@ default String getEffectiveAvatarUrl()
...
+        public static EnumSet<UserFlag> getFlags(int flags)
+        {
+            final EnumSet<UserFlag> foundFlags = EnumSet.noneOf(UserFlag.class);
+            
+            if (flags == 0)
+                return foundFlags; //empty
+            
+            for (UserFlag flag : values())
+                if (flag != UNKNOWN && (flags & flag.raw) == flag.raw)
+                    foundFlags.add(flag);
```
HydroPage90, line 352:
```suggestion
            for (UserFlag flag : values())
            {
                if (flag != UNKNOWN && (flags & flag.raw) == flag.raw)
                    foundFlags.add(flag);
            }
```

Didn't realize the convention for JDA was having braces on the loops

boomboompower:
Unmark the PR as a draft to recieve feedback ^

Andre601:
Just because it's a draft doesn't mean it can't receive feedback. Your point is completely pointless.

DManstrator:
You can post feedback but when I see a MR on WiP / draft, I wouldn't give feedback since I don't want to say stuff on things which are not final yet. Maybe the author wants to do exactly what I have in my mind but they had no time to do that yet. So I can kinda understand it.

HydroPage90:
@DManstrator I'd still comment, man. What you have on your mind could be a new idea we haven't discussed yet. And if we're working on it, there's no problem telling you. This is what PRs are for

DManstrator:
@HydroPage90 I'm literally too lazy for that. Worst case: You suggest things which are going to be implemented nevertheless -> waste of time.
Most people just want to see the final results and improve that. At least this is my attitude.

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,171 @@ default String getEffectiveAvatarUrl()
...
+        {
+            this.offset = offset;
+            this.raw = 1 << offset;
+            this.name = name;
+        }
+
+        /**
+         * The readable name as used in the Discord Client.
+         * 
+         * @return The readable name of this {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag}.
```
MinnDevelopment, line 283: Linking to the same class is redundant

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,171 @@ default String getEffectiveAvatarUrl()
...
+         * and returns the raw offset {@code int} representation of the flags.
+         * 
+         * @param  flags
+         *         The array of flags of which to form into the raw int representation.
+         *         
+         * @return Unsigned int representing the provided flags.
+         */
+        public static int getRaw(@Nonnull UserFlag... flags){
+            int raw = 0;
+            for (UserFlag flag : flags)
```
MinnDevelopment, line 370: Add a `Checks.noneNull` for this, we don't want to throw NPE's if we can avoid it. Also add docs for the `IllegalArgumentException`.

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,171 @@ default String getEffectiveAvatarUrl()
...
+
+        /**
+         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
+         * and returns the raw offset {@code int} representation of the flags.
+         * <br>Example: {@code getRaw(EnumSet.of(UserFlag.STAFF, UserFlag.HYPESQUAD))}
+         *
+         * @param  flags
+         *         The Collection of flags of which to form into the raw int representation.
+         *
+         * @return Unsigned int representing the provided flags.
```
MinnDevelopment, line 387: Java doesn't support unsigned ints. Call it `bitmask`

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,171 @@ default String getEffectiveAvatarUrl()
...
+            {
+                if (flag != null && flag != UNKNOWN)
+                    raw |= flag.raw;
+            }
+            
+            return raw;
+        }
+
+        /**
+         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
```
MinnDevelopment, line 380:
```suggestion
         * This is effectively the opposite of {@link #getFlags(int)}. This takes a collection of UserFlags
```

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,69 @@
+package net.dv8tion.jda.api.events.user.update;
```
MinnDevelopment, line 1: Missing copyright

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,69 @@
...
+ * <p>This event requires the {@link net.dv8tion.jda.api.requests.GatewayIntent#GUILD_MEMBERS GUILD_MEMBERS} intent to be enabled.
+ * <br>{@link net.dv8tion.jda.api.JDABuilder#createDefault(String) createDefault(String)} and
+ * {@link net.dv8tion.jda.api.JDABuilder#createLight(String) createLight(String)} disable this by default!
+ *
+ * <p>Additionally, this event also requires the {@link net.dv8tion.jda.api.utils.MemberCachePolicy MemberCachePolicy}
+ * to cache the updated members. Discord does not specifically tell us about the updates, but merely tells us the
+ * member was updated and gives us the updated member object. In order to fire a specific event like this we
+ * need to have the old member cached to compare against.
+ */
+public class UserUpdateFlagsEvent extends GenericUserUpdateEvent<Integer>
```
MinnDevelopment, line 28: I think this should be using the `EnumSet` instead of the int representation. Then you can add stuff like `getNewFlagsRaw()` here, etc.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
MinnDevelopment: We use `Objects.equals` for things that can be null, int cannot be null so you should use `!=`.

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
```diff
@@ -155,6 +158,18 @@ public boolean isFake()
...
     }
 
+    @Nonnull
+    @Override
+    public EnumSet<UserFlag> getFlags()
+    {
+        return flagsSet;
+    }
+    
+    public int getFlagsRaw()
```
MinnDevelopment, line 168: Add this to the interface. This can be useful for better compatibilty when discord adds new flags.

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
```diff
@@ -43,6 +44,8 @@
     protected long privateChannel = 0L;
     protected boolean bot;
     protected boolean fake = false;
+    protected EnumSet<UserFlag> flagsSet;
```
MinnDevelopment, line 47: This shouldn't be stored, just compute it when its needed.

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
```diff
@@ -155,6 +158,18 @@ public boolean isFake()
         return fake;
     }
 
+    @Nonnull
+    @Override
+    public EnumSet<UserFlag> getFlags()
+    {
+        return flagsSet;
```
MinnDevelopment, line 165: You're returning a modifiable set here. This shouldn't be a problem if you would compute it inplace rather than ahead of time.

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
```diff
@@ -216,6 +231,13 @@ public UserImpl setFake(boolean fake)
         this.fake = fake;
         return this;
     }
+    
+    public UserImpl setFlags(int flags)
+    {
+        this.flagsRaw = flags;
+        this.flagsSet = UserFlag.getFlags(flags);
```
MinnDevelopment, line 238: Always converting these to an EnumSet is a bad idea, most user instances won't need this representation.

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
```diff
@@ -43,6 +44,8 @@
     protected long privateChannel = 0L;
     protected boolean bot;
     protected boolean fake = false;
+    protected EnumSet<Flag> flagsSet;
+    protected int flags;
```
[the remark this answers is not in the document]
Andre601: Then how should I handle the Override of `getFlags()` which is a EnumSet?
Maybe return `UserFlag.getFlags(flags)`? Or shouldn't this be a thing in the User interface?
That would be against what you said (https://github.com/DV8FromTheWorld/JDA/pull/1271#issuecomment-619538928)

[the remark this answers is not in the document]
MinnDevelopment:
```java
EnumSet<UserFlag> getFlags()
{
  return UserFlag.getFlags(flags);
}
```

src/main/java/net/dv8tion/jda/api/entities/User.java
HydroPage90, line 283:
```suggestion
```

HydroPage90, line 283:
```suggestion
         * @return The readable name of this UserFlag
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,171 @@ default String getEffectiveAvatarUrl()
...
+
+        /**
+         * Gets the first {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} relating to the provided offset.
+         * <br>If there is no {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} that matches the provided offset,
+         * {@link net.dv8tion.jda.api.entities.User.UserFlag#UNKNOWN UserFlag.UNKNOWN} is returned.
+         * 
+         * @param  offset
+         *         The offset to match a {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} to.
+         *         
+         * @return {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag} relating to the provided offset.
```
HydroPage90, line 320:
```suggestion
         * Gets the first UserFlag relating to the provided offset.
         * <br>If there is no UserFlag that matches the provided offset,
         * {@link net.dv8tion.jda.api.entities.User.UserFlag#UNKNOWN UNKNOWN} is returned.
         * 
         * @param  offset
         *         The offset to match a UserFlag to.
         *         
         * @return UserFlag relating to the provided offset.
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,171 @@ default String getEffectiveAvatarUrl()
...
+        }
+        
+        /**
+         * A set of all {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags} that are specified by this raw int representation of
+         * flags.
+         * 
+         * @param  flags
+         *         The raw {@code int} representation if flags.
+         *         
+         * @return Possibly-empty EnumSet of {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}.
```
HydroPage90, line 340:
```suggestion
         * A set of all UserFlags that are specified by this raw int representation of
         * flags.
         * 
         * @param  flags
         *         The raw {@code int} representation if flags.
         *         
         * @return Possibly-empty EnumSet of UserFlag.
```

src/main/java/net/dv8tion/jda/api/entities/User.java
Andre601: This should then be corrected in the Permission enum too.
https://github.com/DV8FromTheWorld/JDA/blob/cc11f87957c2512ce33f84664df9df7ad43e2908/src/main/java/net/dv8tion/jda/api/Permission.java#L129
https://github.com/DV8FromTheWorld/JDA/blob/cc11f87957c2512ce33f84664df9df7ad43e2908/src/main/java/net/dv8tion/jda/api/Permission.java#L142
https://github.com/DV8FromTheWorld/JDA/blob/cc11f87957c2512ce33f84664df9df7ad43e2908/src/main/java/net/dv8tion/jda/api/Permission.java#L203-L204
https://github.com/DV8FromTheWorld/JDA/blob/cc11f87957c2512ce33f84664df9df7ad43e2908/src/main/java/net/dv8tion/jda/api/Permission.java#L208
https://github.com/DV8FromTheWorld/JDA/blob/cc11f87957c2512ce33f84664df9df7ad43e2908/src/main/java/net/dv8tion/jda/api/Permission.java#L224
https://github.com/DV8FromTheWorld/JDA/blob/cc11f87957c2512ce33f84664df9df7ad43e2908/src/main/java/net/dv8tion/jda/api/Permission.java#L233
https://github.com/DV8FromTheWorld/JDA/blob/cc11f87957c2512ce33f84664df9df7ad43e2908/src/main/java/net/dv8tion/jda/api/Permission.java#L251
https://github.com/DV8FromTheWorld/JDA/blob/cc11f87957c2512ce33f84664df9df7ad43e2908/src/main/java/net/dv8tion/jda/api/Permission.java#L272

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,171 @@ default String getEffectiveAvatarUrl()
...
+            {
+                if (flag != UNKNOWN && (flags & flag.raw) == flag.raw)
+                    foundFlags.add(flag);
+            }
+                    
+            return foundFlags;
+        }
+
+        /**
+         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
```
HydroPage90, line 360:
```suggestion
         * This is effectively the opposite of {@link #getFlags(int)}. This takes a collection of UserFlags
```

src/main/java/net/dv8tion/jda/api/entities/User.java
HydroPage90, line 370:
```suggestion
            int raw = 0;
            
            Checks.noneNull(flags, "No flags should be null");
            for (UserFlag flag : flags)
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,171 @@ default String getEffectiveAvatarUrl()
...
+         * and returns the raw offset {@code int} representation of the flags.
+         * 
+         * @param  flags
+         *         The array of flags of which to form into the raw int representation.
+         *         
+         * @return Unsigned int representing the provided flags.
```
HydroPage90, line 366:
```suggestion
         * @return bitmask representing the provided flags.
```

src/main/java/net/dv8tion/jda/api/entities/User.java
Andre601: So `Unsigned int` to `Bitmask`, or `Unsigned bitmask`?

MinnDevelopment: Just `bitmask` and yes best to refer to it as `bitmask` everywhere

src/main/java/net/dv8tion/jda/api/entities/User.java
MinnDevelopment: The permission enum is not the topic of this PR.

Andre601: I did remove any links to the same class in the UserFlag enum

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
Andre601: Added

src/main/java/net/dv8tion/jda/api/entities/User.java
Andre601: Added the check and the `@throws` about the IllegalArgumentException

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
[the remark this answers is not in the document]
Andre601: Did it implement like that now.

src/main/java/net/dv8tion/jda/api/entities/User.java
Andre601: changed

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
Andre601: Added

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
Andre601: I changed it.
Please take a look and let me know if the changes are alright or need tweaking

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
Andre601: Was changed and is now computed in `getFlags()`

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
Andre601: Was changed to `UserFlag.getFlags(flags)`

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -25,45 +40,45 @@
...
     }
 
     /**
-     * @return Possibly-null EnumSet of previous {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}.
+     * @return The old raw {@code Integer} value of the {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}.
      */
     @Nullable
-    public EnumSet<User.UserFlag> getOldFlagSet(){
-        return previous == null ? null : User.UserFlag.getFlags(previous);
+    public Integer getOldFlagsRaw(){
```
MinnDevelopment, line 65: Change this to `int` and remove the `Nullable`

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
Andre601: removed the flagSet

src/main/java/net/dv8tion/jda/api/entities/User.java
Andre601: The second arg is the name of the argument provided and not a message.
I implemented it tho. ;)

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
Andre601: I assume for both methods?
I changed it

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,183 @@ default String getEffectiveAvatarUrl()
...
+         */
+        public int getRawValue()
+        {
+            return raw;
+        }
+
+        /**
+         * Gets the first UserFlag relating to the provided offset.
+         * <br>If there is no UserFlag that matches the provided offset,
+         * {@link net.dv8tion.jda.api.entities.User.UserFlag#UNKNOWN UserFlag.UNKNOWN} is returned.
```
MinnDevelopment, line 322:
```suggestion
         * {@link #UNKNOWN} is returned.
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,183 @@ default String getEffectiveAvatarUrl()
...
+         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more UserFlags
+         * and returns the raw offset {@code int} representation of the flags.
+         * 
+         * @param  flags
+         *         The array of flags of which to form into the raw int representation.
+         *         
+         * @return bitmask representing the provided flags.
+         * 
+         * @throws java.lang.IllegalArgumentException
+         *         When the provided UsefLags are null.
```
MinnDevelopment, line 376: We usually put the `@throws` before the `@return`

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,183 @@ default String getEffectiveAvatarUrl()
...
+         * @param  flags
+         *         The Collection of flags of which to form into the raw int representation.
+         *
+         * @return bitmask representing the provided flags.
+         * 
+         * @see java.util.EnumSet EnumSet
+         */
+        public static int getRaw(@Nonnull Collection<UserFlag> flags)
+        {
+            Checks.notNull(flags, "Flag Collection");
```
MinnDevelopment, line 405: Missing `@throws` for this exception

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,82 @@
...
+ * <p>Additionally, this event also requires the {@link net.dv8tion.jda.api.utils.MemberCachePolicy MemberCachePolicy}
+ * to cache the updated members. Discord does not specifically tell us about the updates, but merely tells us the
+ * member was updated and gives us the updated member object. In order to fire a specific event like this we
+ * need to have the old member cached to compare against.
+ */
+public class UserUpdateFlagsEvent extends GenericUserUpdateEvent<EnumSet<User.UserFlag>>
+{
+    public static final String IDENTIFIER = "public_flags";
+    
+    public UserUpdateFlagsEvent(@Nonnull JDA api, long responseNumber, @Nonnull User user, EnumSet<User.UserFlag> oldFlags)
```
MinnDevelopment, line 47:
```suggestion
    public UserUpdateFlagsEvent(@Nonnull JDA api, long responseNumber, @Nonnull User user, @Nonnull EnumSet<User.UserFlag> oldFlags)
```

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,82 @@
...
+    {
+        super(api, responseNumber, user, oldFlags, user.getFlags(), IDENTIFIER);
+    }
+
+    /**
+     * @return The old {@code EnumSet<{@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag}>} representation of the User's flags.
+     */
+    @Nullable
```
MinnDevelopment, line 55:
```suggestion
    @Nonnull
```

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,82 @@
...
+     * @return The old raw {@code Integer} value of the {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}.
+     */
+    public int getOldFlagsRaw(){
+        return User.UserFlag.getRaw(previous);
+    }
+
+    /**
+     * @return The new {@code EnumSet<{@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag}>} representation of the User's flags.
+     */
+    @Nullable
```
MinnDevelopment, line 71:
```suggestion
    @Nonnull
```

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,82 @@
...
+    public EnumSet<User.UserFlag> getOldFlags()
+    {
+        return getOldValue();
+    }
+
+    /**
+     * @return The old raw {@code Integer} value of the {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}.
+     */
+    public int getOldFlagsRaw(){
```
MinnDevelopment, line 64:
```suggestion
    public int getOldFlagsRaw()
    {
```

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,82 @@
...
+        return User.UserFlag.getRaw(previous);
+    }
+
+    /**
+     * @return The new {@code EnumSet<{@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag}>} representation of the User's flags.
+     */
+    @Nullable
+    public EnumSet<User.UserFlag> getNewFlags(){
```
MinnDevelopment, line 72:
```suggestion
    public EnumSet<User.UserFlag> getNewFlags()
    {
```

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,82 @@
...
+        return getNewValue();
+    }
+
+    /**
+     * @return The old raw {@code Integer} value of the {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}.
+     */
+    public int getNewFlagsRaw(){
```
MinnDevelopment, line 79:
```suggestion
    public int getNewFlagsRaw()
    {
```

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -367,6 +371,15 @@ public void updateUser(UserImpl userObj, DataObject user)
...
                     userObj, oldAvatar));
         }
+        
+        if (oldFlags != newFlags)
+        {
+            userObj.setFlags(newFlags);
+            jda.handleEvent(
+                    new UserUpdateFlagsEvent(
+                        jda, responseNumber,
+                        userObj, oldFlags));
```
MinnDevelopment, line 381: Doesn't this accept an `EnumSet` now?

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
Andre601: The change of this results in `getOldValue()` returning an error that it might evaluate as null while it is annotated as Nonnull

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
Andre601: The change of this results in `getNewFlags()` returning an error that it might evaluate as null while it is annotated as Nonnull

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
Andre601: Should I change it to `User.UserFlag.getFlags(oldFlags)` or should I use a different approach?

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
MinnDevelopment: Ignore it.

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,186 @@ default String getEffectiveAvatarUrl()
...
+
+        /**
+         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more UserFlags
+         * and returns the raw offset {@code int} representation of the flags.
+         * 
+         * @param  flags
+         *         The array of flags of which to form into the raw int representation.
+         *
+         * @throws java.lang.IllegalArgumentException
+         *         When the provided UserFLags are null.
```
MinnDevelopment, line 374:
```suggestion
         *         When the provided UserFlags are null.
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,186 @@ default String getEffectiveAvatarUrl()
...
+                if (flag != UNKNOWN && (flags & flag.raw) == flag.raw)
+                    foundFlags.add(flag);
+            }
+                    
+            return foundFlags;
+        }
+
+        /**
+         * This is effectively the opposite of {@link #getFlags(int)}, this takes 1 or more UserFlags
+         * and returns the raw offset {@code int} representation of the flags.
```
MinnDevelopment, line 368:
```suggestion
         * and returns the bitmask representation of the flags.
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,186 @@ default String getEffectiveAvatarUrl()
...
+                if (flag != null && flag != UNKNOWN)
+                    raw |= flag.raw;
+            }
+            
+            return raw;
+        }
+
+        /**
+         * This is effectively the opposite of {@link #getFlags(int)}. This takes a collection of UserFlags
+         * and returns the raw offset {@code int} representation of the flags.
```
MinnDevelopment, line 393:
```suggestion
         * and returns the bitmask representation of the flags.
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -225,4 +228,186 @@ default String getEffectiveAvatarUrl()
...
+         * <br>Example: {@code getRaw(EnumSet.of(UserFlag.STAFF, UserFlag.HYPESQUAD))}
+         *
+         * @param  flags
+         *         The Collection of flags of which to form into the raw int representation.
```
MinnDevelopment, line 397:
```suggestion
         *         The flags to convert
```

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,85 @@
...
+{
+    public static final String IDENTIFIER = "public_flags";
+    
+    public UserUpdateFlagsEvent(@Nonnull JDA api, long responseNumber, @Nonnull User user, @Nonnull EnumSet<User.UserFlag> oldFlags)
+    {
+        super(api, responseNumber, user, oldFlags, user.getFlags(), IDENTIFIER);
+    }
+
+    /**
+     * @return The old {@code EnumSet<{@link net.dv8tion.jda.api.entities.User.UserFlag UserFlag}>} representation of the User's flags.
```
MinnDevelopment, line 53: Missing description (same for the other methods).

Also:
```suggestion
     * @return {@link EnumSet} of the old {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
```

src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
```diff
@@ -171,6 +171,7 @@ public void onUserUpdateActivityOrder(@Nonnull UserUpdateActivityOrderEvent even
     public void onUserTyping(@Nonnull UserTypingEvent event) {}
     public void onUserActivityStart(@Nonnull UserActivityStartEvent event) {}
     public void onUserActivityEnd(@Nonnull UserActivityEndEvent event) {}
+    public void onUserUpdateFlags(@Nonnull UserUpdateFlagsEvent event) {}
```
MinnDevelopment, line 174: This should be grouped with the other update events above

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
Andre601: Should the Link for EnumSet not be the full path, rather than just EnumSet?

src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
Andre601: I don't quite understand what you mean exactly with that.

MinnDevelopment: All the other `onUserUpdate` methods are grouped together a few lines above, this one should be in that group.

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
MinnDevelopment: This way you don't need to make an alias for it. It's already imported so javadoc knows what to reference.

src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
Andre601: Updated

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateFlagsEvent.java
```diff
@@ -0,0 +1,85 @@
...
+{
+    public static final String IDENTIFIER = "public_flags";
+    
+    public UserUpdateFlagsEvent(@Nonnull JDA api, long responseNumber, @Nonnull User user, @Nonnull EnumSet<User.UserFlag> oldFlags)
+    {
+        super(api, responseNumber, user, oldFlags, user.getFlags(), IDENTIFIER);
+    }
+
+    /**
+     * @return {@link EnumSet} of the old {@link net.dv8tion.jda.api.entities.User.UserFlag UserFlags}
```
MinnDevelopment, line 53: I think you missed my comment in the last review, methods need descriptions.

Andre601: Yeah, I missed it.
I did  had descriptions but iirc did @boomboompower (Or was it @HydroPage90?) suggest removing them to essentially prevent mentioning the same stuff twice.

MinnDevelopment: That's a bad suggestion because the description is also used to show a short summary of the method in the list of methods. Redundancy is always good in the documentation. Better to say it twice than have people confused.

Andre601: I now added it.

Artuto:
update on this?

Andre601:
> update on this?

Uhm..... It's ready?
I mean considering that Minn approved it, and that it has a set Milestone can we consider this done and the merge is the only thing waiting. So no need to hurry now.

HydroPage90:
This is probably the largest conversation I'll have ever contributed to, lmao. Glad this seems to be done, and glad to have helped out a bit

Andre601:
I would say it's kind of small if you remove all the review stuff.

PR #1276, opened 2020-04-24
Add retrieveMembersByPrefix

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This implements the new get by prefix chunking features. This can be useful for bots that need to fetch members by names for various reasons.

### Example

```java
guild.retrieveMembersByPrefix("Minn", 100).onSuccess(list -> {
         list.stream()
             .filter(it -> it.getUser().getAsTag().equals("Minn#6688"))
             .findFirst()
             .ifPresent(member -> {
                 System.out.println("Minn#6688 is a member of this server");
             });
});
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/concurrent/Task.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/concurrent/Task.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2015-2020 Austin Keener, Michael Ritter, Florian Spieß, and the JDA contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.api.utils.concurrent;
+
+import javax.annotation.Nonnull;
+import java.util.function.Consumer;
+
+/**
+ * Represents an asynchronous task.
+ * <br>Note: The underlying task may already be started.
+ *
+ * @param <T>
+ *        The result type
+ */
+public interface Task<T>
+{
+    /**
+     * Whether this task has started.
+     *
+     * @return True, if this task has already started.
+     */
+    boolean isStarted();
+
+    /**
+     * Provide a callback for exception handling.
+     * <br>This is an asynchronous operation.
+     *
+     * @param  callback
+     *         The error callback
+     *
+     * @throws IllegalArgumentException
+     *         If null is provided
+     *
+     * @return The current Task instance for chaining
+     */
+    Task<T> onError(@Nonnull Consumer<? super Throwable> callback);
+
+    /**
+     * Provide a callback for success handling.
+     * <br>This is an asynchronous operation.
+     *
+     * @param  callback
+     *         The success callback
+     *
+     * @throws IllegalArgumentException
+     *         If null is provided
+     *
+     * @return The current Task instance for chaining
+     */
+    Task<T> onSuccess(@Nonnull Consumer<? super T> callback);
+
+    /**
+     * Blocks the current thread until the result is ready.
+     * <br>This will not work on the default JDA event thread because it might depend on other events to be processed,
+     * which could lead to a deadlock.
+     *
+     * @throws UnsupportedOperationException
+     *         If this is called on the default JDA event thread
+     * @throws java.util.concurrent.CompletionException
+     *         If some exception occurred (such as {@link java.util.concurrent.TimeoutException}).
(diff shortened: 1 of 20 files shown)


Andre601:
This is now just a replacement of `getUsersByName`?

MinnDevelopment:
> This is now just a replacement of getUsersByName?

No, because it doesn't check exact matches but name prefixes (including nicknames i believe). This also only works per guild not globally. This method is also limited to a maximum of 100 matches due to API limitations.

src/main/java/net/dv8tion/jda/internal/requests/MemberChunkManager.java
```diff
@@ -18,143 +18,156 @@
...
-import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.locks.ReentrantLock;
 
 public class MemberChunkManager
 {
+    private static final long maxAge = 10 * 1000; // 10 seconds
```
DV8FromTheWorld, line 32:
```suggestion
    private static final long MAX_CHUNK_AGE = 10 * 1000; // 10 seconds
```

src/main/java/net/dv8tion/jda/internal/requests/MemberChunkManager.java
```diff
@@ -18,143 +18,156 @@
...
+
+    private class TimeoutHandler implements Runnable
+    {
+        @Override
+        public void run()
+        {
+            MiscUtil.locked(lock, () ->
+            {
+                requests.forEachValue(request -> {
+                    if (request.getAge() > maxAge)
```
DV8FromTheWorld, line 165:
```suggestion
                    if (request.getAge() > MAX_CHUNK_AGE)
```

src/main/java/net/dv8tion/jda/internal/requests/WebSocketSendingThread.java
```diff
@@ -148,10 +148,17 @@ else if (!attemptedToSend)
...
 
-    private void handleChunkSync(String chunkOrSyncRequest)
+    private void handleChunkSync(DataObject chunkOrSyncRequest)
     {
         LOG.debug("Sending chunk/sync request {}", chunkOrSyncRequest);
-        if (send(chunkOrSyncRequest))
+        boolean success = send(
+            DataObject.empty()
+                .put("d", chunkOrSyncRequest)
+                .put("op", WebSocketCode.MEMBER_CHUNK_REQUEST)
```
DV8FromTheWorld, line 157:
```suggestion
                .put("op", WebSocketCode.MEMBER_CHUNK_REQUEST)
                .put("d", chunkOrSyncRequest)
```

src/main/java/net/dv8tion/jda/internal/utils/concurrent/task/GatewayTask.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public boolean isStarted()
+    {
+        return true;
+    }
+
+    @Override
+    public Task<T> onError(@Nonnull Consumer<? super Throwable> callback)
```
DV8FromTheWorld, line 47: What happens if someone doesn't register an `onError` callback? Is there messaging that gets logged?

MinnDevelopment: I think, the error just gets swallowed by default. I'm also not a fan of that, might have to implement a default behavior.

src/main/java/net/dv8tion/jda/internal/requests/WebSocketSendingThread.java
```diff
@@ -116,9 +116,9 @@ public void run()
             queueLock.lockInterruptibly();
 
             audioRequest = client.getNextAudioConnectRequest();
-            chunkOrSyncRequest = chunkSyncQueue.peek();
-            if (chunkOrSyncRequest != null)
-                handleChunkSync(chunkOrSyncRequest);
+            chunkRequest = chunkSyncQueue.peek();
```
DV8FromTheWorld, line 119: Should we rename `chunkSyncQueue` to `chunkQueue`?

MinnDevelopment: Yes, I already did in the WebSocketClient but forgot it here.

src/main/java/net/dv8tion/jda/api/utils/concurrent/Task.java
```diff
@@ -39,6 +39,9 @@
      * Provide a callback for exception handling.
      * <br>This is an asynchronous operation.
      *
+     * <p>The error will be logged regardless of your callback, this only exists to handle
```
DV8FromTheWorld, line 42: This isn't technically guaranteed because it depends on whether the Future has had an `.exceptionally` registered to it _before_ it gets passed to the Task.

MinnDevelopment: We *are* doing that though. The exception handler is already active before we create the `GatewayTask` instance. If some other implementation of the interface doesn't do that, then it violates the contract of this method.

PR #1288, opened 2020-05-04
Change domain to discord.com

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Change to the new domain which will be required by 7th November 2020.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageActivity.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageActivity.java
@@ -73,7 +73,7 @@ public class MessageActivity
      * Represents the {@link net.dv8tion.jda.api.entities.MessageActivity.Application Application} of a
      * {@link net.dv8tion.jda.api.entities.MessageActivity MessageActivity} if it's set.
      *
-     * @see <a href="https://discordapp.com/api/v7/games" target="_blank">https://discordapp.com/api/v7/games</a>
+     * @see <a href="https://discord.com/api/v7/games" target="_blank">https://discord.com/api/v7/games</a>
      */
     public static class Application implements ISnowflake
     {
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageActivity.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageActivity.java
@@ -39,7 +39,7 @@ public class MessageActivity
     /**
      * The current {@link net.dv8tion.jda.api.entities.MessageActivity.ActivityType ActivityType}
      *
-     * @return the type of the activity.
+     * @return the type of the activity, or {@link ActivityType#UNKNOWN UNKNOWN}
      */
     @Nonnull
     public ActivityType getType()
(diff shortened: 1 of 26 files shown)


DManstrator:
I'm not sure if this is also in the scope of this PR, but ReceivedMessage#getJumpLink would also need to be adjusted I guess. I verified that this already [works in the client](https://discordapp.com/channels/125227483518861312/169484473333710848/706974300191719446).

Andre601:
There are also various places (mainly links to the API) that use discordapp.com and I think those should also be updated accordingly.

src/main/java/net/dv8tion/jda/api/entities/Webhook.java
```diff
@@ -93,7 +93,7 @@
      * @return A fake {@link net.dv8tion.jda.api.entities.User User} instance
      *         representing the default webhook user.
      *
-     * @see    <a href="https://discordapp.com/developers/docs/resources/webhook#execute-webhook">Execute Webhook Docs</a>
+     * @see    <a href="https://v.com/developers/docs/resources/webhook#execute-webhook">Execute Webhook Docs</a>
```
MinnDevelopment, line 96:
```suggestion
     * @see    <a href="https://discord.com/developers/docs/resources/webhook#execute-webhook">Execute Webhook Docs</a>
```

src/main/java/net/dv8tion/jda/api/entities/MessageActivity.java
```diff
@@ -73,7 +73,7 @@ public String getPartyId()
      * Represents the {@link net.dv8tion.jda.api.entities.MessageActivity.Application Application} of a
      * {@link net.dv8tion.jda.api.entities.MessageActivity MessageActivity} if it's set.
      *
-     * @see <a href="https://discordapp.com/api/v7/games" target="_blank">https://discordapp.com/api/v7/games</a>
+     * @see <a href="https://discord.com/api/v7/games" target="_blank">https://discord.com/api/v7/games</a>
```
MinnDevelopment, line 76: This endpoint is gone, is there a documented replacement?

Vankka: Appears like the replacement is `https://discordapp.com/api/v6/applications/detectable`, but not documented

Andre601: That would be kinda counter-productive considering that A) The domain will change to discord.com and B) the API will update to v7.

So it would actually be https://discord.com/api/v7/applications/detectable

Vankka: Accidentally put it as discordapp.com & v6, but the endpoint is still the correct one.

MinnDevelopment: Where is this endpoint documented? If it's not documented I'd rather just remove it entirely.

Andre601: > Where is this endpoint documented? If it's not documented I'd rather just remove it entirely.

> Appears like the replacement is `https://discordapp.com/api/v6/applications/detectable`, **but not documented**

MinnDevelopment: If this isn't documented, I'll just remove it.

PR #1291, opened 2020-05-06
Deprecate IFakeable

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This gets rid of the outdated "fake" terminology and interfaces. Almost every user can be considered "fake" and it just doesn't make sense to keep this design anymore.

We will not cache "fake" users for private channels. Instead, we simply keep the user reference in the private channel instance. That way we get a better cache structure that is easier to maintain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,27 +319,16 @@ public static void main(String[] args) throws Exception
 
 ## Entity Lifetimes
 
-An **Entity** is the term used to describe types such as **GuildChannel**/**Message**/**User** and other entities
-that Discord provides.
-Instances of these entities are created and deleted by JDA when Discord instructs it. This means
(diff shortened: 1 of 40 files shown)


Andre601:
Will this have any major impact on the code of bots or is it mainly an internal change?
And what exactly is the reason for the Deprecation? Is there any replacement or is that no longer needed at all?

Andre601:
I assume the Readme entry here would also be removed/updated accordingly?
<https://github.com/DV8FromTheWorld/JDA/#fake-entities>

PR #1292, opened 2020-05-09
Add methods to bulk retrieve members

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This adds a method to retrieve a list of members, filtered by user IDs.

Note: We can't add a `retrieveMembersByIds(Collection<String>)` due to java type erasure.

### Example

```java
guild.retrieveMembersByIds(86699011792191488L, 107562988810027008L)
    .onSuccess((list) -> System.out.println("Found members: " + list));
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2484,8 +2484,9 @@ public interface Guild extends ISnowflake
      * <br>If the user does not resolve to a member of this guild, then it will not appear in the resulting list.
      * It is possible that none of the users resolve to a member, in which case an empty list will be the result.
      *
-     * <p>This will not load the {@link net.dv8tion.jda.api.OnlineStatus OnlineStatus} and {@link Activity Activities}
-     * of the members. You can use {@link #retrieveMembers(boolean, Collection)} to load presences.
+     * <p>If the {@link GatewayIntent#GUILD_PRESENCES GUILD_PRESENCES} intent is enabled,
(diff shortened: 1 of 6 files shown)


src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -837,6 +838,50 @@ public boolean isAvailable()
                 });
     }
 
+    @Nonnull
+    @Override
+    public Task<List<Member>> retrieveMembersByIds(boolean includePresence, @Nonnull long... ids)
+    {
+        Checks.notNull(ids, "ID Array");
+        Checks.check(!includePresence || api.isIntent(GatewayIntent.GUILD_PRESENCES),
+                "Cannot retrieve persences of members without GUILD_PRESENCES intent!");
```
caneleex, line 847:
```suggestion
                "Cannot retrieve presences of members without GUILD_PRESENCES intent!");
```

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -2478,6 +2479,301 @@ default Emote getEmoteById(long id)
...
+     *             <li>If the input is more than 100 IDs</li>
+     *         </ul>
+     *
+     * @return {@link Task} handle for the request
+     */
+    @Nonnull
+    @CheckReturnValue
+    default Task<List<Member>> retrieveMembersByIds(@Nonnull long... ids)
+    {
+        return retrieveMembersByIds(false, ids);
```
MinnDevelopment, line 2624: Should we just set this to true if `GUILD_PRESENCES` intent is enabled?

PR #1293, opened 2020-05-10
Implement max_concurrency in SessionController

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This implements the `max_concurrency` buckets in a new `ConcurrentSessionController` class (new default). The concurrency means that multiple shards can connect simultaneously, without hitting the identify rate limit. This is very important for larger bots which can connect 16 or 64 shards at once to reduce startup time.

The buckets work by using a modulo over the `shard_id`:

```
bucket = shard_id % concurrency
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/ConcurrentSessionController.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/ConcurrentSessionController.java
@@ -62,7 +62,7 @@ public class ConcurrentSessionController extends SessionControllerAdapter implem
         return worker;
     }
 
-    private final class Worker extends Thread
+    private static class Worker extends Thread
     {
         private final Queue<SessionConnectNode> queue = new ConcurrentLinkedQueue<>();
 
--- a/src/main/java/net/dv8tion/jda/api/utils/ConcurrentSessionController.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/ConcurrentSessionController.java
@@ -121,15 +121,22 @@ public class ConcurrentSessionController extends SessionControllerAdapter implem
                 node.run(false); // we don't use isLast anymore because it can be a problem with many reconnecting shards
             }
             catch (NoSuchElementException ignored) {/* This means the node was removed before we started it */}
+            catch (InterruptedException e)
+            {
+                queue.add(node);
+                throw e;
+            }
             catch (IllegalStateException e)
             {
                 Throwable t = e.getCause();
                 if (t instanceof OpeningHandshakeException)
                     log.error("Failed opening handshake, appending to queue. Message: {}", e.getMessage());
-                else if (!JDA.Status.RECONNECT_QUEUED.name().equals(t.getMessage()))
+                else if (t != null && !JDA.Status.RECONNECT_QUEUED.name().equals(t.getMessage()))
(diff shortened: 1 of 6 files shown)


src/main/java/net/dv8tion/jda/api/utils/ConcurrentSessionController.java
```diff
@@ -0,0 +1,143 @@
...
+import com.neovisionaries.ws.client.OpeningHandshakeException;
+import net.dv8tion.jda.api.JDA;
+
+import javax.annotation.Nonnull;
+import java.util.NoSuchElementException;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.TimeUnit;
+
+public class ConcurrentSessionController extends SessionControllerAdapter implements SessionController
```
MinnDevelopment, line 28: Missing docs here

src/main/java/net/dv8tion/jda/api/utils/ConcurrentSessionController.java
```diff
@@ -0,0 +1,159 @@
...
+            try
+            {
+                node = queue.remove();
+                log.debug("Running connect node for shard {}", node.getShardInfo());
+                node.run(false); // we don't use isLast anymore because it can be a problem with many reconnecting shards
+            }
+            catch (NoSuchElementException ignored) {/* This means the node was removed before we started it */}
+            catch (InterruptedException e)
+            {
+                queue.add(node);
```
DV8FromTheWorld, line 142: We are readding the node to the queue but still throwing the exception. This exception will kill the thread.
Are we just relying on the next call to `enqueue` to restart the thread?

MinnDevelopment: Yes. The thread can only be interrupted by a third-party so it must be either intentional or a bug. Ideally, the third-party will also handle starting a new one in that case.

DV8FromTheWorld: Okay, just wanted to make sure we were on the same page. 👍

src/main/java/net/dv8tion/jda/api/utils/SessionControllerAdapter.java
```diff
@@ -104,8 +141,10 @@ public void handleResponse(Response response, Request<ShardedGateway> request)
 
                         String url = object.getString("url");
                         int shards = object.getInt("shards");
+                        int concurrency = object.getObject("session_start_limit")
+                                                .getInt("max_concurrency", 1);
```
GoldRenard: This is missing in master after release merge. Is that ok?

GoldRenard: So it means that actual `max_concurrency` is never set to SessionController now in master branch and always equals 1.

MinnDevelopment: Fixed in bdcece1

PR #1297, opened 2020-05-12
Automatically adjust settings based on intents

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This seems to be a common segfault for users so we
should adjust the settings automatically instead.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
@@ -2243,7 +2243,7 @@ public class  DefaultShardManagerBuilder
                 .forEach(JDAImpl.LOG::warn);
 
             // Tell user how to disable this warning
-            JDAImpl.LOG.warn("You can manually disable these flags to remove this warning by using disableCache({}) on your JDABuilder",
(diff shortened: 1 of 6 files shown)


src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
```diff
@@ -436,7 +478,23 @@ public static DefaultShardManagerBuilder create(@Nullable String token, @Nonnull
...
+
+    private DefaultShardManagerBuilder applyIntents()
+    {
+        EnumSet<CacheFlag> disabledCache = EnumSet.allOf(CacheFlag.class);
+        for (CacheFlag flag : CacheFlag.values())
+        {
+            GatewayIntent requiredIntent = flag.getRequiredIntent();
+            if (requiredIntent == null || (requiredIntent.getRawValue() & intents) != 0)
+                disabledCache.remove(flag);
+        }
```
DV8FromTheWorld, line 492: I really do think we need a warning here letting users know that the flag was automatically marked disabled. It concerns me that users won't understand why certain functions don't work if they don't know what cache flags are when they use those functions.

A function might say it doesn't work if X cache flag is disabled, but if a new person doesn't know what a cache flag is but _does_ know that they didn't disable any of them, then they'll be confused

PR #1302, opened 2020-05-19
Improve handling of VOICE_SERVER_UPDATE

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This adjusts the internals to properly handle disconnect -> voice region change. Discord seems to sometimes first disconnect our socket and then send the new voice region. This change now means that we will always establish a new connection when we get a VOICE_SERVER_UPDATE. According to testing, these are only sent to the session that requested the connection so that additional check is no longer present.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/managers/AudioManagerImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/managers/AudioManagerImpl.java
@@ -44,7 +44,6 @@ public class AudioManagerImpl implements AudioManager
 
     protected final ListenerProxy connectionListener = new ListenerProxy();
     protected final GuildImpl guild;
-    protected VoiceChannel queuedAudioConnection = null;
     protected AudioConnection audioConnection = null;
     protected EnumSet<SpeakingMode> speakingModes = EnumSet.of(SpeakingMode.VOICE);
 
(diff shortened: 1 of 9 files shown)


src/main/java/net/dv8tion/jda/internal/managers/AudioManagerImpl.java
```diff
@@ -328,30 +317,23 @@ public ConnectionListener getListenerProxy()
...
         if (audioConnection == null)
+        {
+            this.audioConnection = null;
             return;
+        }
 
-        this.queuedAudioConnection = null;
+        // This will set the audioConnection to null, which we then immediately override with the new connection
+        if (this.audioConnection != null)
+            closeAudioConnection(ConnectionStatus.AUDIO_REGION_CHANGE);
```
DV8FromTheWorld, line 328: So we are just assuming that if we had an audio connection already and we're changing to a new one that it is a region change?

So long as `setAudioConnection` is called with a non-null value _only_ from `VOICE_SERVER_UPDATE`, I suppose that is fine.

MinnDevelopment: Yeah, every time we call `setAudioConnection` it's because of a region/server change. "Region" in this case basically just means we switched voice server. One guild can now have multiple voice servers for individual channels, moving channel would then "change region". It is an unfortunate naming of this status but we'll keep using it.

DV8FromTheWorld: > One guild can now have multiple voice servers for individual channels, moving channel would then "change region". It is an unfortunate naming of this status but we'll keep using it.

Wow really? Interesting.

nikammerlaan:
I've been running this is prod for a couple days and it seems perfect to me 👍

PR #1303, opened 2020-05-21
Add Emote#isAvailable() method

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description
Adds `Emote#isAvailable()` method which tells if the emote is available or not.
The [Emoji structure](https://discord.com/developers/docs/resources/emoji#emoji-object-emoji-structure) has a `available` field that indicates if the emote is available for the users, depending on the current boost tier of the Guild.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Emote.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Emote.java
@@ -133,8 +133,16 @@ public interface Emote extends IMentionable, IFakeable
     boolean isManaged();
 
     /**
-     * Whether this emote is available. A Emote is unavailable when the {@link net.dv8tion.jda.api.entities.Guild.BoostTier BoostTier}
-     * of the Guild is lower than required for this Emote.
+     * Whether this emote is available. When an emote becomes unavailable, it cannot be used in messages. An emote becomes
+     * unavailable when the {@link net.dv8tion.jda.api.entities.Guild.BoostTier BoostTier} of the guild drops such that
+     * the maximum allowed emotes is lower than the total amount of emotes added to the guild.
+     * 
+     * <p>If an emote is added to the guild when the boost tier allows for more than 50 normal and 50 animated emotes
+     * (BoostTier is at least {@link net.dv8tion.jda.api.entities.Guild.BoostTier#TIER_1 TIER_1}) and the emote is at least
+     * the 51st one added, then the emote becomes unavaiable when the BoostTier drops below a level that allows those emotes
+     * to be used.
+     * <br>Emotes that where added as part of a lower BoostTier (i.e. the 51st emote on BoostTier 2) will remain available,
+     * as long as the BoostTier stays about the required level.
      * 
      * @return True, if this emote is available
      */
--- a/src/main/java/net/dv8tion/jda/api/entities/Emote.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Emote.java
@@ -142,7 +142,7 @@ public interface Emote extends IMentionable, IFakeable
      * the 51st one added, then the emote becomes unavaiable when the BoostTier drops below a level that allows those emotes
      * to be used.
      * <br>Emotes that where added as part of a lower BoostTier (i.e. the 51st emote on BoostTier 2) will remain available,
-     * as long as the BoostTier stays about the required level.
+     * as long as the BoostTier stays above the required level.
      * 
      * @return True, if this emote is available
      */
--- a/src/main/java/net/dv8tion/jda/api/entities/Emote.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Emote.java
@@ -133,6 +133,14 @@ public interface Emote extends IMentionable, IFakeable
     boolean isManaged();
 
     /**
+     * Whether this emote is available. A Emote is unavailable when the {@link net.dv8tion.jda.api.entities.Guild.BoostTier BoostTier}
+     * of the Guild is lower than required for this Emote.
+     * 
+     * @return True, if this emote is available
+     */
+    boolean isAvailable();
+
+    /**
      * The {@link net.dv8tion.jda.api.JDA JDA} instance of this Emote
      *
      * @return The JDA instance of this Emote
--- a/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
@@ -806,7 +806,7 @@ public class EntityBuilder
                 .setName(json.getString("name", ""))
                 .setAnimated(json.getBoolean("animated"))
                 .setManaged(json.getBoolean("managed"))
-                .setAvailable(json.getBoolean("available"));
+                .setAvailable(json.getBoolean("available", true));
     }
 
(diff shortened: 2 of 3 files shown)


src/main/java/net/dv8tion/jda/api/entities/Emote.java
```diff
@@ -132,6 +132,14 @@ default boolean hasRoles()
      */
     boolean isManaged();
 
+    /**
+     * Whether this emote is available. A Emote is unavailable when the {@link net.dv8tion.jda.api.entities.Guild.BoostTier BoostTier}
+     * of the Guild is lower than required for this Emote.
```
DV8FromTheWorld, line 137: > lower than required
What defines this? We need to explain what "lower than required" means so that if a developer finds that an emote is unavailable they understand how to make it available. From the current docs, I don't understand how to do that.

Andre601: I'm not quite sure how to explain this.
The way it works is, that when a emote was added while the Discord is boosted (f.e. is the 51st emote) would it become unavailable once the Discord loses enough Boost levels to no longer have the additional emotes.

I would be glad if you had a suggestion on how I could phrase the above system in a simple sentence.

DV8FromTheWorld:
```
Whether this emote is available for use. When an emote becomes unavailable, it cannot be used in messages. An emote becomes unavailable when the boost tier of the guild drops such that the maximum allowed emotes is lower than the total amount of emotes added to the guild. If an emote is added to the guild when the boost tier allows for more than 50 (and thus the emote is 51st emote added), then if the boost tier drops to only allow a maximum of 50, the 51st emote would be marked as unavailable.
```

Often going for the most detailed explanation is the best route.

Andre601: I've added a description which I think should be good enough to explain stuff.
I changed some stuff like trying to explain that emotes below the limit of a higher tier will remain available when the tier doesn't go lower than what they require (i.e. going from 2 to 1)

caneleex:
Will this also get updated once the emote becomes available again or becomes unavailable? Shouldn't there be an event for that or discord doesn't send it?

Andre601:
I'm not sure about the details of this.
The only thing I do know is, that this is part of the emoji/emote object that Discord sends.

My guess is, that the emotes will update when the Boost level of the Discord changes either in positive or negative direction.

Andre601:
Would it be reasonable to add a `isAvailable()` method to Emote related Events?
Considering that on an EmoteAddEvent, this would always be true, since you can't add emotes when you lack the required boost level.

DV8FromTheWorld:
For this PR to continue, we need to find out if the `GUILD_EMOJIS_UPDATE` event is fired when boost level changes such that some emojis are no longer available.

Andre601:
From what I could do the docs not mention anything like that(?)

Should I perhaps ask for clarification on the API-docs GitHub?

DV8FromTheWorld:
You could ask in the docs repo, you could also test it out yourself.

Andre601:
https://github.com/discord/discord-api-docs/issues/1725#issuecomment-640947828

Looks like to trigger an event, but won't be guaranteed.

Andre601:
I currently think about how this could be added properly.
My thought was about having an EmoteUpdateX event of some sort that would update its availability, but this would probably require a updateEmote method in the EntityBuilder for what I know.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -805,7 +805,8 @@ public EmoteImpl createEmote(GuildImpl guildObj, DataObject json, boolean fake)
         return emoteObj
                 .setName(json.getString("name", ""))
                 .setAnimated(json.getBoolean("animated"))
-                .setManaged(json.getBoolean("managed"));
+                .setManaged(json.getBoolean("managed"))
+                .setAvailable(json.getBoolean("available"));
```
MinnDevelopment, line 809: `getBoolean` returns `false` if there is no value for the key. Is it correct, that this would be false if `available` is not present in the emote json?

Andre601: I think it's.... weird. The docs don't give a clear indication about what the default value here would be. My guess is, that it would be `true` by default as Emotes added below 50 should always be available, no matter the Boost level and the available flag is probably only used in emotes that go beyond 50.
I'll quickly ask for clarification on that.

MinnDevelopment: You can pass a default value as second parameter like `json.getBoolean("available", true)`

Andre601: I'm aware of this, but before I do this would I first have a clear answer about how the missing field should be treated. Because knowing Discord could this also mean that the Emote isn't available...

Furthermore, would I need to look into adding a new event that would trigger when the Emote's availability changes, which from what I understand is tied to the Boost tier changing, but is also not reliable according to [this response here](https://github.com/discord/discord-api-docs/issues/1725#issuecomment-640947828)

MinnDevelopment: If it isn't reliable we can probably just not add an event for it. It's better to not add one now and add one later if necessary because then it doesn't involve a breaking change.

Andre601: I just pass true as default because I believe that's what Discord would say is the default for the missing field.

Andre601:
Since update events seem unreliable according to night, should this now be pretty much everything to add.
I perhaps should add a notice that the `isAvailable()` boolean might not be updated due to the EmoteUbdateEvent being not reliable on this and that Discord might not send an event in the first place?

src/main/java/net/dv8tion/jda/api/entities/Emote.java
```diff
@@ -132,6 +132,22 @@ default boolean hasRoles()
...
+     * Whether this emote is available. When an emote becomes unavailable, it cannot be used in messages. An emote becomes
+     * unavailable when the {@link net.dv8tion.jda.api.entities.Guild.BoostTier BoostTier} of the guild drops such that
+     * the maximum allowed emotes is lower than the total amount of emotes added to the guild.
+     * 
+     * <p>If an emote is added to the guild when the boost tier allows for more than 50 normal and 50 animated emotes
+     * (BoostTier is at least {@link net.dv8tion.jda.api.entities.Guild.BoostTier#TIER_1 TIER_1}) and the emote is at least
+     * the 51st one added, then the emote becomes unavaiable when the BoostTier drops below a level that allows those emotes
+     * to be used.
+     * <br>Emotes that where added as part of a lower BoostTier (i.e. the 51st emote on BoostTier 2) will remain available,
+     * as long as the BoostTier stays about the required level.
```
MinnDevelopment, line 145:
```suggestion
     * as long as the BoostTier stays above the required level.
```

PR #1307, opened 2020-05-31
Add some convenience methods to reaction events

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This adds a few convenience methods to reaction events. You can now retrieve the target message and reaction issuer directly from the event.

### Example

```java
@Override
public void onMessageReactionRemove(MessageReactionRemoveEvent event) {
    event.retrieveMessage()
         .map(Message::getContentDisplay)
         .queue(System.out::println);
}
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/events/message/react/GenericMessageReactionEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/message/react/GenericMessageReactionEvent.java
@@ -143,7 +143,7 @@ public class GenericMessageReactionEvent extends GenericMessageEvent
 
     /**
      * Retrieves the {@link Member} who added or removed the reaction.
-     * <br>If a member is known, this will return {@link #getUser()}.
+     * <br>If a member is known, this will return {@link #getMember()}.
      *
      * <p>Note that banning a member will also fire {@link MessageReactionRemoveEvent} and no member will be available
      * in those cases. An {@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MEMBER UNKNOWN_MEMBER} error response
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/events/message/react/GenericMessageReactionEvent.java
```diff
@@ -121,4 +125,53 @@ public MessageReaction getReaction()
...
+        if (member != null)
+            return new CompletedRestAction<>(getJDA(), member);
+        return getGuild().retrieveMemberById(getUserIdLong());
+    }
+
+    /**
+     * Retrieves the message for this reaction event.
+     * <br>Simple shortcut for {@code getChannel().retrieveMessageById(getMessageId())}.
+     *
+     * <p>The {@link Message#getMember() getMember()} method will always return null for the resulting message.
```
DV8FromTheWorld, line 167:
```suggestion
     * <p>The {@link Message#getMember() Message#getMember()} method will always return null for the resulting message.
     *  You can, instead, retrieve the member via {@link #getMember()} or {@link #retrieveMember()}.
```

MinnDevelopment: We usually do `Type.method()` in the docs and I think from context you can tell its a method on Message anyway.

DV8FromTheWorld: The only reason I was adding a separation here (and it should be `.`, you're right) is because in the very next line I reference a different `#getMember()`

Updated the suggestion to use a `.`

src/main/java/net/dv8tion/jda/api/events/message/react/GenericMessageReactionEvent.java
```diff
@@ -121,4 +125,53 @@ public MessageReaction getReaction()
...
+    public RestAction<User> retrieveUser()
+    {
+        if (issuer != null)
+            return new CompletedRestAction<>(getJDA(), issuer);
+        return getJDA().retrieveUserById(getUserIdLong());
+    }
+
+    /**
+     * Retrieves the {@link Member} who added or removed the reaction.
+     * <br>If a member is known, this will return {@link #getUser()}.
```
MinnDevelopment, line 146:
```suggestion
     * <br>If a member is known, this will return {@link #getMember()}.
```

PR #1309, opened 2020-06-04
Add ETF support

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This adds a custom implementation to parse and serialize ETF payloads for the discord gateway connection.

You can enable this by using `JDABuilder#setGatewayEncoding(GatewayEncoding.ETF)`, this will be an opt-in.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/JDABuilder.java
@@ -537,11 +537,15 @@ public class JDABuilder
      * @param  encoding
      *         The {@link GatewayEncoding} (default: JSON)
      *
+     * @throws IllegalArgumentException
+     *         If null is provided
+     *
      * @return The JDABuilder instance. Useful for chaining.
      */
     @Nonnull
     public JDABuilder setGatewayEncoding(@Nonnull GatewayEncoding encoding)
     {
+        Checks.notNull(encoding, "GatewayEncoding");
         this.encoding = encoding;
         return this;
     }
--- a/src/main/java/net/dv8tion/jda/api/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/JDABuilder.java
@@ -531,6 +531,14 @@ public class JDABuilder
         return this;
     }
 
+    /**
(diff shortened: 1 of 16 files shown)


src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -530,6 +531,13 @@ private JDABuilder setDisabledCache(EnumSet<CacheFlag> flags)
         return this;
     }
 
+    @Nonnull
+    public JDABuilder setGatewayEncoding(@Nonnull GatewayEncoding encoding)
```
MinnDevelopment, line 535: This also needs to be added to the shard manager probably

src/main/java/net/dv8tion/jda/api/utils/data/etf/ExTermEncoder.java
```diff
@@ -0,0 +1,147 @@
...
+    {
+        buffer = realloc(buffer, 5);
+        buffer.put(INT);
+        buffer.putInt(value);
+        return buffer;
+    }
+
+    private static ByteBuffer packLong(ByteBuffer buffer, long value)
+    {
+        byte sign = (byte) (value < 0 ? 1 : 0);
```
MinnDevelopment, line 116: I'm not sure if we should just treat all longs here as unsigned, discord doesn't really need signed longs for anything.

MinnDevelopment:
@napstr gave me some performance insights. Using ETF instead of JSON seems to have almost no performance impact. This is most likely due to the decompression time which should far outweigh the parsing time. Regardless, we can add support for it as an option since the code works fine and having fallbacks is always good.

napstr:
Adding some more details: Without doing in depth profiling, general JVM metrics like CPU & memory usage and GC activity look fairly similar. Unsure if the slight decrease in CPU compared to -7d can be attributed to the ETF change, or it just being a fresh start. A snapshot of the data can be found here: https://grafana.npstr.space/dashboard/snapshot/p6Myi0ARKjW4uTBk9mBWggga0focaNnT

PR #1311, opened 2020-06-09
Use FormBody for prune post request

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

We are supposed to use the body for sending parameters in POST requests.

Reference: https://github.com/discord/discord-api-docs/issues/1720#issuecomment-640948326

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
@@ -61,6 +61,7 @@ import net.dv8tion.jda.internal.utils.cache.MemberCacheViewImpl;
 import net.dv8tion.jda.internal.utils.cache.SnowflakeCacheViewImpl;
 import net.dv8tion.jda.internal.utils.cache.SortedSnowflakeCacheViewImpl;
 import net.dv8tion.jda.internal.utils.concurrent.task.GatewayTask;
+import okhttp3.FormBody;
 
 import javax.annotation.CheckReturnValue;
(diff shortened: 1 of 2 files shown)


MinnDevelopment:
Adding the varargs parameter to the method changes the signature in a breaking way. However, due to the niche use-case of this method I'm willing to accept this small ABI change.

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1024,16 +1025,25 @@ public long getIdLong()
...
+        form.add("days", Integer.toString(days));
         if (!wait)
-            route = route.withQueryParams("compute_prune_count", "false");
-        return new AuditableRestActionImpl<>(getJDA(), route, (response, request) -> response.getObject().getInt("pruned", 0));
+            form.add("compute_prune_count", "false");
+        for (Role role : roles)
+        {
+            Checks.notNull(role, "Role");
+            Checks.check(role.getGuild().equals(this), "Role is not from the same guild!");
+            form.add("include_roles", role.getId());
```
MinnDevelopment, line 1044: The way this works is basically building a form-body in this format:

```
days=30&compute_prune_count=false&include_roles=123&include_roles=456&include_roles=789
``` 

This is the correct way to pass an array through a query format, in the case of the discord API (which uses Flask).

Issue #1313, opened 2020-06-11
Throw a nice exception if there are duplicated roles when using "modifyMemberRoles"


MrPowerGamerBR:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [X] I have checked for similar issues.
- [X] I have updated to the [latest JDA version][download].
- [X] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request

Discord's error message when you provide a list with duplicate roles in the `modifyMemberRoles` is... pretty bad.

`RestAction queue returned failure: [ErrorResponseException] 400: {"roles":["2"]}`

It would be nice if JDA checked that there are duplicate roles provided and throw an exception, this would be more "user friendly" than trying to figure out why Discord is returning a bad request.

### Example Use-Case

```kotlin
val roles = listOf(guild.getRoleById(12345L), guild.getRoleById(12345L))

// Currently, this code throws an RestAction queue returned failure: [ErrorResponseException] 400: {"roles":["2"]}, which isn't very user friendly
guild.modifyMemberRoles(member, roles) // Exception! Can't give duplicated roles to the member
```

MinnDevelopment:
I think it would be better to simply filter out duplicates since its technically not an issue.

PR #1314, opened 2020-06-11
Remove duplicate roles when using "Guild.modifyMemberRoles(...)"

MrPowerGamerBR:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1313

## Description

This avoids Discord returning "{roles:[2]}" if your list contains duplicate roles... which isn't a very helpful thing when you are trying to debug why that is happening.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
@@ -1261,7 +1261,7 @@ public class GuildImpl implements Guild
         }
 
         DataObject body = DataObject.empty()
-            .put("roles", roles.stream().map(Role::getId).collect(Collectors.toList()));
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1261,7 +1261,7 @@ public long getIdLong()
         }
 
         DataObject body = DataObject.empty()
-            .put("roles", roles.stream().map(Role::getId).collect(Collectors.toList()));
+            .put("roles", roles.stream().distinct().map(Role::getId).collect(Collectors.toList()));
```
MinnDevelopment, line 1264:
```suggestion
            .put("roles", roles.stream().map(Role::getId).collect(Collectors.toSet()));
```

PR #1325, opened 2020-06-22
added missing "!" in getAsMention()

skoliveira:
"!" is necessary if you need matches or replace User.getAsMention() in Message.getContentRaw()

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
@@ -138,7 +138,7 @@ public class UserImpl implements User
     @Override
     public String getAsMention()
     {
-        return "<@" + getId() + '>';
+        return "<@!" + getId() + '>';
     }
 
     @Override


caneleex:
Exclamation mark is not always present in mentions if i remember correctly. It signals that a member has a nickname.

MinnDevelopment:
https://discord.com/developers/docs/reference#message-formatting

https://github.com/DV8FromTheWorld/JDA/blob/459a19aa2b5fa90ed3eb4e0509a5f6502b5d35da/src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java#L362-L365

MinnDevelopment:
You shouldn't use `getAsMention()` to compare something that can have 2 different values. Both with and without the "!" it is a valid mention and you should handle both of those cases. We provide a mention regex in [Message.MentionType#USER](https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/api/entities/Message.MentionType.html#USER).

Example:

```java
String message = "<@86699011792191488> Hello <@!107562988810027008> World";

Pattern pattern = Message.MentionType.USER.getPattern();
Matcher matcher = pattern.matcher(message);

while (matcher.find()) {
  String id = matcher.group(1);
  System.out.println(id);
}

Andre601:
> Exclamation mark is not always present in mentions if i remember correctly. It signals that a member has a nickname.

To answer this: When the Member has a nickname does the mention often contain a ! to mark a mention displaying a nickname (I think)
From what I know is this however no longer needed and mostly a relict of Discord's old system (not sure tho)

One thing that I am curious about tho is, if User#getAsMention() will always just return the mention without ! and Member#getAsMention() will return it, when a nickname is present.

skoliveira:
all right guy,
i realized that linux and android clients behave differently
linux client always uses "!"
and android client never
regardless of nickname
why? I don't know!
i will deal with both cases.
tks

PR #1327, opened 2020-06-26
Introduce better methods to load members

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This introduces better methods for loading members on a guild.

### Example

**Current:**
```java
guild.retrieveMembers().thenRun(() -> {
  List<Member> members = guild.getMembers();
  members.forEach(System.out::println);
  guild.pruneMemberCache(); // manually clear the cache again
});

guild.retrieveMembers().thenRun(() -> {
  List<Member> members = guild.getMembers();
  System.out.println(members.size());
  guild.pruneMemberCache(); // manually clear the cache again
});
```
**New:**
```java
guild.loadMembers(System.out::println); // only adds members to cache which the policy accepts
guild.loadMembers().onSuccess((members) -> {
  System.out.println(members.size());
});
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -54,6 +54,7 @@ import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Future;
 import java.util.function.Consumer;
+import java.util.function.Predicate;
 
 /**
  * Represents a Discord {@link net.dv8tion.jda.api.entities.Guild Guild}.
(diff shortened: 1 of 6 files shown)


MinnDevelopment:
Unfortunately, the overloads seem to result in "ambiguous calls" which makes `guild.loadMembers((x) -> System.out.println(x));` a compiler error. This could be fixed by using different method names, open for suggestions.

Andre601:
I could come up with a name idea, if you can quickly list what the loadMembers method will actually do next to getting the members obviously.

MinnDevelopment:
I'm thinking of renaming `loadMembers(Predicate)` to `findMembers(Predicate)` which seems more accurate and solves the name conflict. The methods would work like this:

`loadMembers(): Task<List<Member>>` load all members into a list 
`findMembers(Predicate): Task<List<Member>>` load all members but only retain the ones that match the filter
`loadMembers(Consumer): Task<Void>` load all members but don't put them into a list, instead use a callback for each member like a forEach method

Andre601:
Seems like a good idea to make.

PR #1334, opened 2020-07-04
Add UserById wrapper

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This implements a wrapper object which only contains a user's ID. Useful to avoid tons of overloads that do the same thing. This requires a bit of testing before merging.

### Example

```java
User user = User.fromId(86699011792191488L);
channel.removeReaction(unicode, user).queue();
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/UserById.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/UserById.java
@@ -43,6 +43,35 @@ public class UserById implements User
         return this.id;
     }
 
+    @Nonnull
+    @Override
+    public String getAsMention()
+    {
+        return "<@" + getId() + ">";
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Long.hashCode(id);
+    }
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        if (obj == this)
+            return true;
+        if (!(obj instanceof User))
+            return false;
+        return ((User) obj).getIdLong() == this.id;
+    }
+
+    @Override
+    public String toString()
+    {
+        return "U:(" + getId() + ')';
+    }
+
     @Contract("->fail")
     private void unsupported()
     {
@@ -147,13 +176,5 @@ public class UserById implements User
     {
         unsupported();
         return false;
-    }
-
-    @Nonnull
-    @Override
-    public String getAsMention()
-    {
-        unsupported();
-        return null;
     }
 }
--- a/src/main/java/net/dv8tion/jda/internal/entities/UserById.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/UserById.java
@@ -30,7 +30,7 @@ import java.util.List;
 
 public class UserById implements User
 {
-    private final long id;
+    protected final long id;
 
     public UserById(long id)
     {
--- a/src/main/java/net/dv8tion/jda/internal/entities/UserById.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/UserById.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2015-2020 Austin Keener, Michael Ritter, Florian Spieß, and the JDA contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/internal/entities/UserById.java
```diff
@@ -0,0 +1,159 @@
...
+        unsupported();
+        return false;
+    }
+
+    @Nonnull
+    @Override
+    public String getAsMention()
+    {
+        unsupported();
+        return null;
```
MinnDevelopment, line 157: We could probably support this one

src/main/java/net/dv8tion/jda/internal/entities/UserById.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    @Override
+    public boolean hasPrivateChannel()
+    {
+        unsupported();
+        return false;
```
MinnDevelopment, line 96: Should we just return false?

src/main/java/net/dv8tion/jda/internal/entities/UserById.java
```diff
@@ -0,0 +1,159 @@
...
+    {
+        unsupported();
+        return 0;
+    }
+
+    @Override
+    public boolean isFake()
+    {
+        unsupported();
+        return false;
```
duncte123, line 149: Shouldn't you return true here?

MinnDevelopment: No, because isFake is not gonna stick around.

src/main/java/net/dv8tion/jda/internal/entities/UserById.java
DV8FromTheWorld: I'd keep it consistent as the point is that this entity exists exclusively to provide access to the entity. Returning a value, even if it is false, would break that consistency and is technically not true as well.

We don't actually have the info needed to inform true or false, so let's just throw

DV8FromTheWorld:
Looks good. I'm interesting in seeing people try this out during draft stage to see how viable this is.

MinnDevelopment:
I want to merge #1291 before this.

src/main/java/net/dv8tion/jda/internal/entities/UserById.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public RestAction<PrivateChannel> openPrivateChannel()
+    {
+        unsupported();
```
Sanduhr32, line 132: I feel like, that this method should be available, because this request only requires the already provided ID (and a JDA instance).
I am aware that this breaks the pattern of JDA, that JDA you would need to manually pass a JDA reference for being able to execute this request.

I don't see why an extra request for retrieving the user by id should be required if you have user / member cache disabled. I know that this is a temporarily created user object that gets collected by the gc fairly quickly, but it still feels unnecessary, if it's possible to achieve this without an extra footprint, if this low memory profile wrapper exists in the future.

MinnDevelopment: You can just use JDA#openPrivateChannelById instead

Sanduhr32: Let's pretend this didn't happen.

napstr:
I could imagine a different approach to this where you don't wrap the user, but only the user id.

```java
public class UserId {

    public static UserId of(long snowflake) {
        return new UserId(snowflake);
    }

    public static UserId of(String snowflake) {
        return new UserId(Long.parseLong(snowflake);
    }

    private final long snowflake;

    private UserId(long snowflake) {
        this.snowflake = snowflake;
    }

    public long getSnowflake() {
        return this.snowflake;
    }
}
```

One of the advantages here would be that you avoid all the footguns of so many unsupported methods.

MinnDevelopment:
@napstr that would require making breaking changes to every method signature that accepts a user.

napstr:
That is unfortunately correct. Would make the interfaces even more ugly until it can be removed in the next major. However the transition for users would also be super simple (either `User#getId` or `UserId.of`).

MinnDevelopment:
@napstr that is not entirely correct. This doesn't only introduce compiler errors for the user but it also introduces breaking changes to the *binary interface* which would result in much more difficult transitions due to every extension breaking.

napstr:
Wouldn't there be plenty of time for tools to migrate if such a change was introduced in 4.3?

napstr:
Oh I see it would take 2 majors to actually end up in a happy place again.

¯\_(ツ)_/¯

PR #1342, opened 2020-07-10
Add support for Role Tags

Chew:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

Discord recently added "tags" object to Roles. 
Tags appear for any of the following reasons:
1) To indicate a Server Booster role, example

```json
"tags": {
    "premium_subscriber": null
}
```

"premium_subscriber" will always be null, we only care about the key.

2) To indicate a Bot integration, example

```json
"tags": {
    "bot": "489034866206310410"
 }
```

3) To indicate a Twitch/YouTube member role, example

```json
"tags": {
    "integration": "717460314068746300"
}
```

If not any of the above 3, tags will not appear.

I've only implemented the first one (partially) by adding `Role#isBoosterRole()`. I plan to move forward and possibly add Role#isBotRole() or Role#isIntegrationRole() or similar. It would also be nice to have a Guild#getBoosterRole() as well.

This has not been tested so as of now it's a draft, it's really just here to get feedback as I work on it and to make sure everything is good along the way :)


Andre601:
Iirc is there a `Role#isManaged()` function that is afaik for Bot and Twitch/YouTube roles.
Maybe we can implement those tags for those?

Chew:
> Iirc is there a `Role#isManaged()` function that is afaik for Bot and Twitch/YouTube roles.
> Maybe we can implement those tags for those?

"Is Managed" applies to all 3 of these roles. 
Managed = 
image

RedstonecraftHD:
Maybe we could add Role#getManagedType() that returns a Enum with the Type

Chew:
Yeah, perhaps

RoleType
NORMAL - A normal everyday role
BOT - A role for bot integrations
BOOSTER - The booster role
INTEGRATION - Role from integration e.g. twitch or youtube

Andre601:
Woldn't it be better and simpler to have the RoleType as an enum inside the Role?
I don't see it as a separate entity.

Chew:
Good call, considering how tiny it is. Forgot about classes in classes

averen:
It also might be better to rename NORMAL as NONE instead. Since there is no field present, it feels more grammatically correct and follows the patterns set by other enums.

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -97,6 +97,20 @@
      */
     boolean isMentionable();
 
+    /**
+     * Whether or not this Role is the server's Server Booster role.
+     *
+     * @ return True, if Role is Server Booster role.
+     */
+    boolean isBoosterRole();
```
kantenkugel, line 105: This is probably not necessary anymore, since there is a getRoleType method now.
It would be nice to add more details to both the getRoleType and the isManaged() javadocs

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -271,4 +285,33 @@ default RoleAction createCopy()
      */
     @Nonnull
     JDA getJDA();
+
+
+    /**
+     * Enum used to differentiate between the different types of Guild Roles.
+     */
+    enum Type
```
kantenkugel, line 293: Should either be named RoleType (what current naming + getter suggest; imo preferred), or ManagementType

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -271,4 +285,33 @@ default RoleAction createCopy()
...
+    {
+        /**
+         * A basic, user-created role.
+         */
+        NONE,
```
kantenkugel, line 298: In case of this enum showing Types of Roles, NONE is not applicable. its not a NONE Role. NORMAL like in initial draft would fit better

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1059,6 +1059,29 @@ public Role createRole(GuildImpl guild, DataObject roleJson, long guildId)
             .setHoisted(roleJson.getBoolean("hoist"))
             .setColor(color == 0 ? Role.DEFAULT_COLOR_RAW : color)
             .setMentionable(roleJson.getBoolean("mentionable"));
+        if(roleJson.hasKey("tags"))
+        {
+            if(roleJson.hasKey("premium_subscriber"))
```
kantenkugel, line 1064: these statements don't look like they would work, since they check for the key on the roleJson, instead of the sub-object "tags".
All the tag-specific role types could also be solved via RoleType#getTag() property and RoleType.fromTag(), and the default being NORMAL

src/main/java/net/dv8tion/jda/api/entities/Role.java
Chew: Removed, didn't do the javadocs yet

src/main/java/net/dv8tion/jda/api/entities/Role.java
Chew: Changed back to RoleType

src/main/java/net/dv8tion/jda/api/entities/Role.java
Chew: Back to NORMAL. I thought normal felt better too

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
Chew: Replace the giant if statement and moved it to the RoleType#fromTags(), where the DataObject is passed and parsed from there. If has(), parse, if not, then just go with NORMAL.

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -77,7 +78,10 @@
 
     /**
      * Whether this {@link net.dv8tion.jda.api.entities.Role Role} is managed by an integration
+     * This is any role where the "This role is managed by an integration..." appears in the role list.
```
MinnDevelopment, line 81:
```suggestion
     * <br>This is any role where the "This role is managed by an integration..." appears in the role list.
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -77,7 +78,10 @@
...
+     * Integration can be: Bot role, Booster role, or a connected partner account role.
      *
+     * @see Role#getRoleType()
      * @return True, if this {@link net.dv8tion.jda.api.entities.Role Role} is managed.
```
MinnDevelopment, line 85:
```suggestion
     * @return True, if this {@link net.dv8tion.jda.api.entities.Role Role} is managed.
     *
     * @see    Role#getRoleType()
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -97,6 +101,14 @@
      */
     boolean isMentionable();
 
+    /**
+     * Return the type of role this is
+     * Typically used for finding what kind of managed/integration role this is.
+     *
+     * @return this role's type
+     */
+    RoleType getRoleType();
```
MinnDevelopment, line 110:
```suggestion
    @Nonnull
    RoleType getRoleType();
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -97,6 +101,14 @@
      */
     boolean isMentionable();
 
+    /**
+     * Return the type of role this is
+     * Typically used for finding what kind of managed/integration role this is.
```
MinnDevelopment, line 106:
```suggestion
     * typically used to determine what kind of managed role this is.
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -271,4 +283,51 @@ default RoleAction createCopy()
      */
     @Nonnull
     JDA getJDA();
+
+
```
MinnDevelopment, line 287:
```suggestion
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -271,4 +283,51 @@ default RoleAction createCopy()
...
+     * Enum used to differentiate between the different types of Guild Roles.
+     */
+    enum RoleType
+    {
+        /**
+         * A basic, user-created role.
+         */
+        NORMAL,
+        /**
+         * A role created from a bot on join.
```
MinnDevelopment, line 298:
```suggestion
         * A role created for a bot on join.
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -271,4 +283,51 @@ default RoleAction createCopy()
...
+         */
+        BOT,
+        /**
+         * A role created from a YouTube Member or Twitch Subscriber integration..\
```
MinnDevelopment, line 302:
```suggestion
         * A role created for an integration. For example, twitch subscribers
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -271,4 +283,51 @@ default RoleAction createCopy()
...
+         */
+        INTEGRATION,
+        /**
+         * The Server Booster role.
```
MinnDevelopment, line 306:
```suggestion
         * The Guild Booster role.
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -271,4 +283,51 @@ default RoleAction createCopy()
...
+                return RoleType.BOT;
+            }
+            else if(tags.hasKey("integration_id"))
+            {
+                return RoleType.INTEGRATION;
+            } else
+            {
+                return RoleType.UNKNOWN;
+            }
+        }
```
MinnDevelopment, line 330:
```suggestion
        public static RoleType fromTags(DataObject tags)
        {
            if (tags.hasKey("premium_subscriber"))
                return RoleType.BOOSTER;
            else if (tags.hasKey("bot_id"))
                return RoleType.BOT;
            else if (tags.hasKey("integration_id"))
                return RoleType.INTEGRATION;
            else
                return RoleType.UNKNOWN;
        }
```

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1059,6 +1059,14 @@ public Role createRole(GuildImpl guild, DataObject roleJson, long guildId)
...
             .setColor(color == 0 ? Role.DEFAULT_COLOR_RAW : color)
             .setMentionable(roleJson.getBoolean("mentionable"));
+        if(roleJson.hasKey("tags"))
+        {
+            role.setType(Role.RoleType.fromTags(roleJson.getObject("tags")));
+        }
+        else
+        {
+            role.setType(Role.RoleType.NORMAL);
+        }
```
MinnDevelopment, line 1069: This is not a good way to handle tags. You should be storing the tags in a map so that you can also retrieve the associated values. For example, which bot a role belongs to.
```suggestion
        if (roleJson.hasKey("tags"))
            role.setType(Role.RoleType.fromTags(roleJson.getObject("tags")));
        else
            role.setType(Role.RoleType.NORMAL);
```

src/main/java/net/dv8tion/jda/internal/entities/RoleImpl.java
```diff
@@ -56,6 +56,7 @@
     private boolean managed;
     private boolean hoisted;
     private boolean mentionable;
+    private RoleType roleType;
```
MinnDevelopment, line 59: This is a bad way to handle tags.

src/main/java/net/dv8tion/jda/api/entities/Role.java
Chew: The role is explicitly called "Server Booster" when it's created. So this should be "The Guild's Server Booster" role

MinnDevelopment: In API terminology were always use the term "guild".

MinnDevelopment:
I'm gonna implement this myself in #1343. I appreciate your efforts, but I have already planned how I would like to implement this a while ago.

PR #1344, opened 2020-07-12
Replace SnowflakeReference

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This removes the usage of `SnowflakeReference` which was used to avoid memory leaks from continuously storing JDA entities such as members. Memory leaks are now solved in a different way, by replacing the upstream reference when the cache has updated. This does not handle cases where the entire JDA session is replaced, however that is a specific use-case that we shouldn't need to handle explicitly for the users.

### Motivation

The `SnowflakeReference` class adds a severe overhead in both memory usage and GC usage. Some GC implementations seem to struggle a lot with `WeakReference`, which is a crucial part of this system.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
@@ -24,7 +24,6 @@ import net.dv8tion.jda.api.utils.cache.CacheFlag;
 import net.dv8tion.jda.internal.JDAImpl;
 import net.dv8tion.jda.internal.utils.Checks;
 import net.dv8tion.jda.internal.utils.PermissionUtil;
-import net.dv8tion.jda.internal.utils.cache.SnowflakeReference;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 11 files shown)


src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
```diff
@@ -63,27 +62,25 @@ public MemberImpl(GuildImpl guild, User user)
...
-        updateUser();
         return user;
     }
 
     @Nonnull
     @Override
     public GuildImpl getGuild()
     {
-        return (GuildImpl) guild.resolve();
+        GuildImpl realGuild = (GuildImpl) getJDA().getGuildById(guild.getIdLong());
```
MinnDevelopment, line 80:
```suggestion
        GuildImpl realGuild = (GuildImpl) api.getGuildById(guild.getIdLong());
```

src/main/java/net/dv8tion/jda/internal/managers/RoleManagerImpl.java
```diff
@@ -64,7 +63,10 @@ public RoleManagerImpl(Role role)
     @Override
     public Role getRole()
     {
-        return role.resolve();
+        Role realRole = getGuild().getRoleById(role.getIdLong());
```
MinnDevelopment, line 66:
```suggestion
        Role realRole = role.getGuild().getRoleById(role.getIdLong());
```

src/main/java/net/dv8tion/jda/internal/managers/PermOverrideManagerImpl.java
```diff
@@ -92,7 +69,7 @@ private void setupValues()
     @Override
     public PermissionOverride getPermissionOverride()
     {
-        return override.resolve();
+        return override;
```
DV8FromTheWorld, line 72: This getter isn't following the "lookup `realEntity`" system that the other getters are following.
Is that intentional?

MinnDevelopment: Yes this is intentional because looking up the override is pretty complicated and I'd like to avoid that in this edge case.

PR #1349, opened 2020-07-22
Add findMembersWithRoles

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds a replacement for `getMembersWithRoles` which performs chunking.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2334,7 +2334,10 @@ public interface Guild extends ISnowflake
     default Task<List<Member>> findMembersWithRoles(@Nonnull Collection<Role> roles)
     {
         Checks.noneNull(roles, "Roles");
-        if (isLoaded())
+        for (Role role : roles)
+            Checks.check(this.equals(role.getGuild()), "All roles must be from the same guild!");
+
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -2297,6 +2301,68 @@ default Emote getEmoteById(long id)
...
+     * @throws IllegalStateException
+     *         If the {@link GatewayIntent#GUILD_MEMBERS GatewayIntent.GUILD_MEMBERS} is not enabled
+     *
+     * @return {@link Task} - Type: {@link List} of {@link Member}
+     */
+    @Nonnull
+    @CheckReturnValue
+    default Task<List<Member>> findMembersWithRoles(@Nonnull Collection<Role> roles)
+    {
+        Checks.noneNull(roles, "Roles");
```
DV8FromTheWorld, line 2328: We should check to make sure that the roles given to us are from this guild as a developer sanity check.

We should also consider ensuring that atleast one role was provided.

MinnDevelopment:
Waiting for response on https://github.com/discord/discord-api-docs/issues/2003

PR #1351, opened 2020-07-27
Add AutoListener

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This replaces the `ListenerAdapter` class which has become an unmaintainable mess due to the number of events it needs to check.

### Pros

- Easier to maintain
- Not as much of a footgun since you don't need to worry about method name spelling anymore
- You can have multiple methods listening to the same event
- Works with custom events as well

### Cons

- You can no longer use auto-complete for method overrides
- Requires reflections
- Methods appear as unused by default
- No annotation which can be used as a compile-time verification

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ public void configureMemoryUsage(JDABuilder builder) {
 The event system in JDA is configured through a hierarchy of classes/interfaces.
 We offer two implementations for the `IEventManager`:
 
-- **InterfacedEventManager** which uses an `EventListener` interface and the `ListenerAdapter` abstract class
+- **InterfacedEventManager** which uses an `EventListener` interface and the `AutoListener` abstract class
 - **AnnotatedEventManager** which uses the `@SubscribeEvent` annotation that can be applied to methods
 
 By default the **InterfacedEventManager** is used.
@@ -159,10 +159,10 @@ public class ReadyListener implements EventListener
 }
 ```
 
-**Using ListenerAdapter**:
+**Using AutoListener**:
 
 ```java
-public class MessageListener extends ListenerAdapter
+public class MessageListener extends AutoListener
 {
     public static void main(String[] args)
             throws LoginException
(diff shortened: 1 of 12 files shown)


Andre601:
If I see this correctly, would this now allow to have any void name you like, as long as the requirements are met?

MinnDevelopment:
> If I see this correctly, would this now allow to have any void name you like, as long as the requirements are met?

Yes.

danthe1st:
Idea for performance improvements:

Do not iterate over all listener methods but iterate over (the class and) implemented interfaces. Use `.getClass()` on the event and walk through all the implementing interfaces. Then, use `Map#get` in every iteration with the interface `Class<?>` type and execute the returned listeners.

This may be faster than just testing every method, especially if the class does not implement a lot of interfaces but has a lot of methods.

MinnDevelopment:
The common listener does not have many event listener methods while the common event does have a lot of inheritance. I don't think it makes sense to optimize things with only guesses, especially when it's just micro-optimization.

thebrightspark:
How does this new `InterfacedEventManager` now differ from the existing `AnnotatedEventManager`? They're both functionally rather similar, only difference being one explicitly marks event handler methods while the other finds them on classes implementing the interface. Would the use of either just simply be user preference?

Personally I would prefer using the `AnnotatedEventManager` so I can be explicit with event handler methods, so I don't see where the new `InterfacedEventManager` fits in.

Apologies for this being a rather late comment, just curiously looking through open PRs.

Andre601:
> How does this new `InterfacedEventManager` now differ from the existing `AnnotatedEventManager`? They're both functionally rather similar, only difference being one explicitly marks event handler methods while the other finds them on classes implementing the interface. Would the use of either just simply be user preference?
> 
> Personally I would prefer using the `AnnotatedEventManager` so I can be explicit with event handler methods, so I don't see where the new `InterfacedEventManager` fits in.
> 
> Apologies for this being a rather late comment, just curiously looking through open PRs.

You're quite right. Both are extremely similar to one another, with the difference being pretty much a missing annotation.
It was mentioned in the OPs Comment as a "Con" that there isn't an annotation to verify on compile-time, so I think this is another change.

PR #1354, opened 2020-07-28
Switch to new permission key

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Reference: https://github.com/discord/discord-api-docs/pull/1843

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
@@ -1033,7 +1033,7 @@ public class EntityBuilder
         final int color = roleJson.getInt("color");
         role.setName(roleJson.getString("name"))
             .setRawPosition(roleJson.getInt("position"))
-            .setRawPermissions(roleJson.getLong("permissions"))
+            .setRawPermissions(roleJson.getLong("permissions_new"))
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1033,7 +1033,7 @@ public Role createRole(GuildImpl guild, DataObject roleJson, long guildId)
         final int color = roleJson.getInt("color");
         role.setName(roleJson.getString("name"))
             .setRawPosition(roleJson.getInt("position"))
-            .setRawPermissions(roleJson.getLong("permissions"))
+            .setRawPermissions(roleJson.getLong("permissions_new"))
```
tpcstld, line 1036: Wow you're fast.

FYI you might need to do `Long.parseLong(roleJson.getString("permissions_new"))` here, some libraries don't automatically cast the string to a long.

MinnDevelopment: Don't worry this is a higher-level abstraction which does the type coalescing for me already.

PR #1355, opened 2020-07-30
Handle cloudflare rate limits

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

When we get rate-limited by cloudflare the retry-after is in seconds. We can detect cloudflare rate-limits by checking whether the [via header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Via) is present.


DV8FromTheWorld:
Side note, this is applying the 429 ratelimit to the specific bucket, but should it instead be applying it to the global ratelimit since cloudflare (to my knowledge) doesn't care about specific endpoints?

MinnDevelopment:
> but should it instead be applying it to the global ratelimit 

That's a good point, maybe we should add a special logging for this as well.

PR #1356, opened 2020-08-01
Remove wrong Nullable annotation

Biospheere:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This Pull Requests removes a wrong @Nullable Annotation in the EmbedBuilder#addField method. The previous Annotation stated that both paramaters could be null. Instead the method throws an IllegalArgumentException.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -769,7 +769,7 @@ public class EmbedBuilder
      * @return the builder after the field has been added
      */
     @Nonnull
-    public EmbedBuilder addField(@Nullable String name, @Nullable String value, boolean inline)
+    public EmbedBuilder addField(String name, String value, boolean inline)
     {
         if (name == null && value == null)
             return this;


Andre601:
Sounds more like a bug?
There is `if(name == null && value == null) return this;` so I think it *should* be nullable and is therefore not wrong annotations.

Biospheere:
But its && and not ||. So the method just returns the current embed without an exception if **both** are null. But the exception is currently thrown if one of the two paramters is null.

MinnDevelopment:
No, these annotations are correct. You can set both values to null.

MinnDevelopment:
This behavior is documented:

> @throws IllecalArgumentException [...] If only {@code name} or {@code value} is set. Both must be set.

niklas2810:
> This behavior is documented:
> 
> > @throws IllecalArgumentException [...] If only {@code name} or {@code value} is set. Both must be set.

@MinnDevelopment 

The documentation does not reflect the actual API behavior: When both parameters are `null`, the method will simply return the current `EmbedBuilder` object without throwing an IllegalArgumentException. Link: https://github.com/DV8FromTheWorld/JDA/blob/fc59416fd9b024e1ac98da229f7726a28d1900e6/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java#L774

Furthermore, I personally would find it more precise to document that both parameters "must not be null" instead of "must be set" in order to highlight that an empty String is a valid parameter.

Issue #1357, opened 2020-08-01
[Suggestion] Add support for publishing messages


GMatrixGames:
## General Troubleshooting
  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request
I'd like to suggest the integration of publishing messages since the opening of news channels to every discord. And messages from bots don't auto-publish

### Example Use-Case
A use case would be if a server has a bot that posts a tweet in a news channel every time a specific account tweets. People wouldn't get the message if they follow that channel since it doesn't publish on post.

RedstoneDaedalus:
It's still entirely possible Discord will block this endpoint for bots, it is better to wait for them to add it to the developer documentation firwt.

GMatrixGames:
I mean, discord.py added it already, so I'm fairly certain it's fine.

GMatrixGames:
@MinnDevelopment If you don't mind me asking, what does the freezer status label mean?

averen:
> It's still entirely possible Discord will block this endpoint for bots, it is better to wait for them to add it to the developer documentation first.

As per discord/discord-api-docs#1692 it is quite clear they are enthusiastic about exposing these endpoints. The freezer label are for issues that are waiting on another issue to be resolved, in this case it is waiting for the documentation to be merged. Until then, implementation details may change as they clean up the feature on their end.

GMatrixGames:
Ah alright. I'll be awaiting the PR to get pulled on discord's end.

Andre601:
> @MinnDevelopment If you don't mind me asking, what does the freezer status label mean?

You can just hover over the label to get a tooltip with its description:
`this is currently put on hold`, so essentially freezer means, the PR/issue won't be worked on for reasons.

GMatrixGames:
> You can just hover over the label to get a tooltip with its description:
> this is currently put on hold, so essentially freezer means, the PR/issue won't be worked on for reasons.

I was on mobile as it was 11PM, and that feature doesn't exist on there. Otherwise I wouldn't have asked.

PR #1362, opened 2020-08-05
Implement support for news channels

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1357 

## Description

We don't handle news channels by using the NEWS type. We will instead use a boolean which makes the code a lot easier to understand. Using a new channel type only for news channels makes no sense since the only difference to text channels is the possibility to crosspost messages.

You can have news channels in a guild without having the NEWS feature. However, those channels do not support crossposting so it makes sense to simply handle them like they are not news channels.

```java
@Override
public boolean isNews()
{
    return news && getGuild().getFeatures().contains("NEWS");
}
```

This also simplifies the checks we need to perform in our code for crossposting.

Gotcha's:

- Trying to crosspost a message that is already crossposted will result in the error response "ALREADY_CROSSPOSTED"
- You can crosspost your own messages if you can send them
- You can crosspost other people's messages with MESSAGE_MANAGE permission
- You cannot crosspost messages if the guild does not have the NEWS feature enabled

### Example

```java
channel.sendMessage("PSA: I like trains")
       .flatMap(Message::crosspost)
       .queue();
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/events/channel/text/update/TextChannelUpdateNewsEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/channel/text/update/TextChannelUpdateNewsEvent.java
@@ -21,6 +21,14 @@ import net.dv8tion.jda.api.entities.TextChannel;
 
 import javax.annotation.Nonnull;
 
+/**
+ * Indicates that a {@link net.dv8tion.jda.api.entities.TextChannel TextChannel}'s has been converted into a news channel
+ * or reverted into a normal channel.
+ *
+ * <p>Can be used to detect when a TextChannel becomes a news channel.
+ *
+ * <p>Identifier: {@code news}
+ */
 @SuppressWarnings("ConstantConditions")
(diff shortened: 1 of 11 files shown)


src/main/java/net/dv8tion/jda/api/events/channel/text/update/TextChannelUpdateNewsEvent.java
```diff
@@ -0,0 +1,47 @@
...
+
+package net.dv8tion.jda.api.events.channel.text.update;
+
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.entities.TextChannel;
+
+import javax.annotation.Nonnull;
+
+@SuppressWarnings("ConstantConditions")
+public class TextChannelUpdateNewsEvent extends GenericTextChannelUpdateEvent<Boolean>
```
MinnDevelopment, line 25: Missing docs

Issue #1360, opened 2020-08-03
MarkdownSanitizer fails to sanitize characters when used inside of quotes 


Owen1212055:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [X ] I have checked for similar issues.
- [X] I have updated to the [latest JDA version][download].
- [X] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

It seems that when you try to escape a string that has quotes and newlines characters inside of the quotes are ignored. For example,
```java
String test = "> _\n" + "> test_";
```

is sanitized as 
```
\> _
\> test_
```
![image](https://user-images.githubusercontent.com/23108066/89228199-28df4200-d5ad-11ea-8b9d-17d1ea7c07e7.png)

### Expected Behavior

I expect it to correctly sanitize characters that are inside of quotes. 
```
\> \_
\> test\_
```

### Code Example or Reproduction Steps

```java
String test = "> _\n" + "> test_";
System.out.println(MarkdownSanitizer.escape(test));
```

### Code for JDABuilder or DefaultShardManagerBuilder Used
N/A

### Exception or Error
N/A

PR #1363, opened 2020-08-05
Fix sanitization issues within quotes

Owen1212055:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1360 

## Description

This fixes the issue causing characters not to sanitize when put into multiple multiline quotes.

Basically, instead of breaking the quotes by their ending we break them up at the start.


Owen1212055:
I apologize for that mess, I forgot to test some of the other methods as well and then ran into other problems!

PR #1372, opened 2020-08-14
Update README.md

GreemDev:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

 Remove unnecessary hyphens, change wording, etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,13 +24,13 @@
 
 # JDA (Java Discord API)
 
-JDA strives to provide a clean and full wrapping of the Discord REST api and its Websocket-Events for Java.
-This library is a helpful tool that provides the functionality to create a discord bot in java.
+JDA strives to provide a clean and feature-complete implementation of the Discord REST API and its WebSocket-based events for Java.
+This library is a helpful tool that provides the functionality to create a Discord bot in Java.
 
 ## Summary
 
 Due to official statements made by the Discord developers we will no longer support unofficial features. These features
-are undocumented API endpoints or protocols that are not available to bot-accounts.
+are undocumented API endpoints or protocols that are not available to bot accounts.
 
 _Please see the [Discord docs](https://discord.com/developers/docs/reference) for more information about bot accounts._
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -279,13 +279,13 @@ Discord allows Bot-accounts to share load across sessions by limiting them to a
...
 
-To use sharding in JDA you will need to use `JDABuilder.useSharding(int shardId, int shardTotal)`. The **shardId** is 0-based which means the first shard
+To use sharding in JDA you will need to use `JDABuilder#useSharding(int shardId, int shardTotal)`. The **shardId** is 0-based which means the first shard
 has the ID 0. The **shardTotal** is the total amount of shards (not 0-based) which can be seen similar to the length of an array, the last shard has the ID of
 `shardTotal - 1`.
 
 The [`SessionController`](https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/api/utils/SessionController.html) is a tool of the JDABuilder
 that allows to control state and behaviour between shards (sessions). When using multiple builders to build shards you have to create one instance
-of this controller and add the same instance to each builder: `builder.setSessionController(controller)`
+of this controller and add the same instance to each builder: `JDABuilder#setSessionController(controller)`
```
MinnDevelopment, line 288: setSessionController isn't specific to JDABuilder so this change is not useful

README.md
```diff
@@ -470,8 +470,8 @@ SLF4J: Defaulting to no-operation (NOP) logger implementation
 SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
 ```
 
-JDA currently provides a fallback Logger in case that no SLF4J implementation is present.
-We strongly recommend to use one though, as that can improve speed and allows you to customize the Logger as well as log to files
+JDA provides a fallback Logger in the case that no SLF4J implementation is present.
+We strongly recommend to use one though, as that can improve speed and allows you to customize the Logger as well as route your Logger's output to a file.
```
MinnDevelopment, line 474: This seems like a more confusing wording, I don't get why this was changed

README.md
```diff
@@ -550,7 +550,7 @@ Features include:
 TODO: Ensure this is compatible with version 4
 ### [JDAction](https://github.com/sedmelluq/jdaction)
 
-Created and maintained by [sedmelluq](https://github.com/sedmelluq)
+Created and maintained by [sedmelluq](https://github.com/sedmelluq),
```
MinnDevelopment, line 553: This comma is useless

README.md
GreemDev: Changed because the word currently was redundant as it's likely not gonna get removed, and the final bit was to make it clearer (though it wasn't necessary)

README.md
GreemDev: It's not? I changed it to `JDABuilder#` because earlier in that section is says that `SessionController` is a tool on the JDABuilder class.

MinnDevelopment: We also never use this notation (Type#method) in the documentation. It's also part of the shard manager.

README.md
MinnDevelopment: Doesn't seem like a useful change to me and the last part makes it more confusing than anything

README.md
GreemDev: Isn't that proper Java documentation notation though? Considering it's part of the Shard Manager as well I'll revert it

MinnDevelopment: Not everyone is familiar with this notation

GreemDev: But it is the standard, right?

MinnDevelopment: No there is no standard

MinnDevelopment:
Closing for inactivity

PR #1374, opened 2020-08-14
new method to check if an attachment is tagged as spoiler

TopiSenpai:
added a method to determine if a attachement is a spoiler

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

added a new method to determine if an attachment is tagged as spoiler

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -2347,5 +2347,16 @@ public interface Message extends ISnowflake, Formattable
             String extension = getFileExtension();
             return extension != null && VIDEO_EXTENSIONS.contains(extension.toLowerCase());
         }
+
+        /**
+         * Whether or not this attachment is tagged as spoiler,
+         * based on {@link #getFileName()}.
+         *
+         * @return True if this attachment is tagged as spoiler
+         */
+        public boolean isSpoiler()
+        {
+            return getFileName().startsWith("SPOILER_");
+        }
     }
 }


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2347,5 +2347,16 @@ public boolean isVideo()
             String extension = getFileExtension();
             return extension != null && VIDEO_EXTENSIONS.contains(extension.toLowerCase());
         }
+
+        /**
+         * Whether or not this attachment is tagged as spoiler,
+         * based on {@link #getFileName()}.
+         *
+         * @return True if this attachment is tagged as spoiler
```
DManstrator, line 2355: I would call it `marked as a spoiler` instead, same above.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2347,5 +2347,16 @@ public boolean isVideo()
...
+         */
+        public boolean isSpoiler()
```
DManstrator, line 2357: I would call it `isSpoiled` or `isMarkedAsSpoiler` to be more precise, preferring the first.

DManstrator:
Maybe I would even add a method `isSpoiledImage` since this is most likely the most common case and will be searched for most times. Impl could be just checking both methods.

TopiSenpai:
yeah sounds not too bad

src/main/java/net/dv8tion/jda/api/entities/Message.java
Andre601: That would sound really weird. When I myself hear "spoiled" do I think about sentences like "Such a spoiled brat" as spoiled essentially means ruined, corrupted or bad.
So `isSpoiled` would imo translate to `isBad` and therefore wouldn't make sense. I vote for keeping the method name as is, as it's more than clear enough.

MinnDevelopment: Yeah, isSpoiler is better.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2347,5 +2347,16 @@ public boolean isVideo()
...
+        {
+            return getFileName().startsWith("SPOILER_");
```
MinnDevelopment, line 2359: Can we check if this is case sensitive or not?

src/main/java/net/dv8tion/jda/api/entities/Message.java
Andre601: > Yeah, isSpoiler is better.

It is case sensitive.
Only prefixed with SPOILER_ works.

src/main/java/net/dv8tion/jda/api/entities/Message.java
caneleex: it's case sensitive

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2347,5 +2347,27 @@ public boolean isVideo()
...
+        }
+        
+        /**
+         * Whether or not this attachment is an image and marked as spoiler,
+         * based on {@link #isSpoiler()}, and {@link #isImage()}.
+         *
+         * @return True if this attachment is and image and marked as spoiler
+         */
+        public boolean isImageSpoiler()
```
MinnDevelopment, line 2368: I don't get the point of this method. We don't need a helper method for this and this will just result in people asking "why is there no isVideoSpoiler and isTextSpoiler and isNotVideoSpoiler" which I just really don't wanna deal with that.

PR #1379, opened 2020-08-24
Fix inconsistency with MarkdownSanitizer 

Owen1212055:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

A couple of weeks ago, I submitted a pull request that attempted to resolve the issue causing symbols to not be sanitized correctly inside of quotes. Unfortunately, I found that putting a symbol in front of a quote causes my previous fix to not work correctly.

Example: ``> _Hello \n> _World`` ->``\\> \\_Hello \n> \\_World``

This fixes the issue by changing the regex to allow for quotes that contain just spaces. While Discord removes these extra spaces itself, they are still able to generate quotes (a bit of a weird fix).

BEFORE

``> ​``
(As stated before, these are still valid quotes)
image

AFTER

``\> ​``
image

If there are any more issues related to this in the future, it might be better just to redo how quotes are handled entirely.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
@@ -71,7 +71,7 @@ public class MarkdownSanitizer
     private static final int ESCAPED_QUOTE_BLOCK = Integer.MIN_VALUE | QUOTE_BLOCK;
 
     private static final Pattern codeLanguage = Pattern.compile("^\\w+\n.*", Pattern.MULTILINE | Pattern.DOTALL);
-    private static final Pattern quote = Pattern.compile("> +\\S.*", Pattern.DOTALL | Pattern.MULTILINE);
+    private static final Pattern quote = Pattern.compile("> +(\\S.*)*", Pattern.DOTALL | Pattern.MULTILINE);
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
The client doesn't render this as a quote:

image


The WYSIWYG editor does this because you might enter more content to indicate that it will be in a quote.

src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
```diff
@@ -71,7 +71,7 @@
     private static final int ESCAPED_QUOTE_BLOCK = Integer.MIN_VALUE | QUOTE_BLOCK;
 
     private static final Pattern codeLanguage = Pattern.compile("^\\w+\n.*", Pattern.MULTILINE | Pattern.DOTALL);
-    private static final Pattern quote = Pattern.compile("> +\\S.*", Pattern.DOTALL | Pattern.MULTILINE);
+    private static final Pattern quote = Pattern.compile("> +(\\S.*)*", Pattern.DOTALL | Pattern.MULTILINE);
```
MinnDevelopment, line 74: What's the point of this group? Just do `> +.*`

Owen1212055: No point, just my not so good experience with regex.

PR #1398, opened 2020-09-19
Make audio thread pool configurable by users

napstr:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

Allow the users to provide the thread pool used for the audio stuff.
Why? This helps to get rid of 1 mostly idling thread per shard. In my largest use case, thats 300+ threads gone.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/JDAImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/JDAImpl.java
@@ -81,7 +81,6 @@ public class JDAImpl implements JDA
     public static final Logger LOG = JDALogger.getLog(JDA.class);
 
     protected final Object audioLifeCycleLock = new Object();
-    protected ScheduledThreadPoolExecutor audioLifeCyclePool;
 
     protected final SnowflakeCacheViewImpl<User> userCache = new SnowflakeCacheViewImpl<>(User.class, User::getName);
     protected final SnowflakeCacheViewImpl<Guild> guildCache = new SnowflakeCacheViewImpl<>(Guild.class, Guild::getName);
@@ -715,8 +714,6 @@ public class JDAImpl implements JDA
         // stop accepting new requests
         if (requester.stop()) // returns true if no more requests will be executed
             shutdownRequester(); // in that case shutdown entirely
-        if (audioLifeCyclePool != null)
-            audioLifeCyclePool.shutdownNow();
         threadConfig.shutdown();
 
         if (shutdownHook != null)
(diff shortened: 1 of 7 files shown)


src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -1020,16 +1017,19 @@ public void resetGatewayUrl()
...
             synchronized (audioLifeCycleLock)
             {
-                pool = audioLifeCyclePool;
+                pool = threadConfig.getAudioPool();
                 if (pool == null)
-                    pool = audioLifeCyclePool = ThreadingConfig.newScheduler(1, this::getIdentifierString, "AudioLifeCycle");
+                {
+                    pool = ThreadingConfig.newScheduler(1, this::getIdentifierString, "AudioLifeCycle");
+                    threadConfig.setAudioPool(pool, threadConfig.isShutdownAudioPool());
+                }
```
MinnDevelopment, line 1032: This is not a thread-safe solution

MinnDevelopment: I think it would be better to just make this a lazy initialization in `ThreadConfig#getAudioPool`.

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -715,8 +714,6 @@ public synchronized void shutdownInternals()
         // stop accepting new requests
         if (requester.stop()) // returns true if no more requests will be executed
             shutdownRequester(); // in that case shutdown entirely
-        if (audioLifeCyclePool != null)
-            audioLifeCyclePool.shutdownNow();
```
MinnDevelopment, line 719: Shouldn't this still shutdown the pool at some point when its default? Thats what `ThreadConfig#isShutdownAudioPool` is used for.

napstr: I'm not 100% sure what you mean here, can you elaborate?

I think this change, removing these two lines, is safe, because the next line, `threadConfig.shutdown()`, takes care of shutting down the audio pool now. There is only a small difference of behaviour, in that `threadConfig.shutdown()` triggers `shutdown()` instead of `shutdownNow()`, but I don't think that makes a difference.

What do you think?

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
napstr: Good point. I'm just realizing now, that it already initializes other pools as well, so moving the audio pool init there should be fine.

src/main/java/net/dv8tion/jda/internal/utils/config/ThreadingConfig.java
```diff
@@ -141,6 +156,24 @@ public ExecutorService getEventPool()
...
+        ScheduledExecutorService pool = audioPool;
+        if (pool == null)
+        {
+            synchronized (audioLock)
+            {
+                pool = audioPool;
+                if (pool == null)
+                {
+                    pool = audioPool = ThreadingConfig.newScheduler(1, identifier, "AudioLifeCycle");
+                }
```
MinnDevelopment, line 171:
```suggestion
                if (pool == null)
                    pool = audioPool = ThreadingConfig.newScheduler(1, identifier, "AudioLifeCycle");
```

MinnDevelopment (commented):
Only some nits

PR #1404, opened 2020-09-27
Add length check to Activity#playing and similar methods

Simon-official:
## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____

Closes Issue: NaN

## Description

Add check to Activity#playing and similar methods to test if the provided name is longer than 128 characters.
Why? Discord won't show activities longer than that.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Activity.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Activity.java
@@ -231,7 +231,7 @@ public interface Activity
      *         The streaming url to use, required to display as "streaming".
      *
      * @throws IllegalArgumentException
-     *         If the specified name is null or empty
+     *         If the specified name is null, empty or longer than 128 characters
      *
      * @return A valid Activity instance with the provided name and url
      *
--- a/src/main/java/net/dv8tion/jda/api/entities/Activity.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Activity.java
@@ -19,6 +19,7 @@ import net.dv8tion.jda.annotations.Incubating;
 import net.dv8tion.jda.internal.entities.EntityBuilder;
 import net.dv8tion.jda.internal.utils.Checks;
 import net.dv8tion.jda.internal.utils.EncodingUtil;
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -114,14 +114,15 @@
...
-     *         if the specified name is null, empty or blank
+     *         if the specified name is null, empty, blank or longer than 128 characters
      *
      * @return A valid Activity instance with the provided name with {@link net.dv8tion.jda.api.entities.Activity.ActivityType#DEFAULT}
      */
     @Nonnull
     static Activity playing(@Nonnull String name)
     {
         Checks.notBlank(name, "Name");
+        Checks.check(name.length() <= 128, "Name must not be greater than 128 characters in length");
```
MinnDevelopment, line 125: Maybe we should trim the name first since discord will not care about leading or tailing spaces.

src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -204,7 +208,7 @@ static Activity watching(@Nonnull String name)
      *         The not-null name of the newly created game
      *
      * @throws IllegalArgumentException
-     *         If the specified name is null or empty
+     *         If the specified name is null, empty or longer than 128 characters
```
MinnDevelopment, line 211: You forgot the other `of` static method

src/main/java/net/dv8tion/jda/api/entities/Activity.java
Simon-official: Seems reasonable, maybe we should also move those checks into [this method](https://github.com/DV8FromTheWorld/JDA/blob/development/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java#L124)  to avoid code duplication

MinnDevelopment: That would be problematic if we use that method later on for creating activities we receive. I prefer having this in the factory methods here instead.

PR #1408, opened 2020-10-02
Implement inline replies feature

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

https://github.com/discord/discord-api-docs/pull/2118

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -23,6 +23,7 @@ import net.dv8tion.jda.api.requests.RestAction;
 import net.dv8tion.jda.api.requests.restaction.AuditableRestAction;
 import net.dv8tion.jda.api.requests.restaction.MessageAction;
 import net.dv8tion.jda.api.requests.restaction.pagination.ReactionPaginationAction;
+import net.dv8tion.jda.api.utils.AttachmentOption;
 import net.dv8tion.jda.internal.JDAImpl;
 import net.dv8tion.jda.internal.requests.FunctionalCallback;
 import net.dv8tion.jda.internal.requests.Requester;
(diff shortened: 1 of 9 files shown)


src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageActionImpl.java
```diff
@@ -62,6 +62,7 @@
     protected EnumSet<Message.MentionType> allowedMentions;
     protected Set<String> mentionableUsers = new HashSet<>();
     protected Set<String> mentionableRoles = new HashSet<>();
+    protected long messageReference;
```
HydroPage90, line 65: Why not stick with `referencedMessage`?

HydroPage90: Nvm, I looked down, and the raw JSON is called `message_reference`, I guess that's why. But then why is everything else referencedMessage?

RedDaedalus:
Support for the new allowed mention type `replied_user` would be nice here.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -948,6 +961,174 @@ default boolean isFromGuild()
...
+    {
+        return getChannel().sendFile(data, name, options).reference(this);
+    }
+
+    /**
+     * Replies and references this message.
+     * <br>This is identical to {@code message.getChannel().sendFile(data, name, options).reference(message)}.
+     * You can use {@link MessageAction#mentionRepliedUser(boolean) mentionRepliedUser(false)} to not mention the author of the message.
+     *
+     * <p>For further info, see {@link MessageChannel#sendFile(File, net.dv8tion.jda.api.utils.AttachmentOption...)} and {@link MessageAction#reference(Message)}.
```
MinnDevelopment, line 1091: Fix this link

caneleex:
would be adding a static setter(/getter), such as `setDefaultMentionRepliedUser` to MessageAction a bad idea? just like with default mentions

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageActionImpl.java
```diff
@@ -480,11 +509,15 @@ protected DataObject getJSON()
             if (nonce != null)
                 obj.put("nonce", nonce);
         }
-        obj.put("tts", tts);
-        if (allowedMentions != null || !mentionableUsers.isEmpty() || !mentionableRoles.isEmpty())
+        if (messageReference != 0)
         {
-            obj.put("allowed_mentions", getAllowedMentionsObj());
+            obj.put("message_reference", DataObject.empty()
```
GoldRenard, line 514: Just tested this. Not sure is that an issue or not, but replying is not working in case of just sending a file without any other message content.

In this case rest action ignores this `payload_json` as it thinks it's empty so `message_reference` never sent to the Discord with a file.

MinnDevelopment: This has been fixed

PR #1410, opened 2020-10-04
improve some permission check related methods' docs

caneleex:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this pr removes dots from params of some methods and clarifies hasAccess docs


MinnDevelopment:
These changes seem superficial, it should be obvious that VOICE_CONNECT only matters in voice channels.

caneleex:
> 
> 
> These changes seem superficial, it should be obvious that VOICE_CONNECT only matters in voice channels.

yes, it's obvious, but from the wording it sounds like it will check for VOICE_CONNECT in every case, even if the provided GuildChannel is not a VoiceChannel

PR #1418, opened 2020-10-16
Implement TextChannel#follow and add source references

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This allows bots to follow announcement channels.

See: https://github.com/discord/discord-api-docs/pull/1837

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
@@ -1437,10 +1437,23 @@ public class EntityBuilder
             }
         }
         
-        return new WebhookImpl(channel, id, type)
+        WebhookImpl webhook = new WebhookImpl(channel, id, type)
                 .setToken(token)
                 .setOwner(owner == null ? null : channel.getGuild().getMember(owner))
                 .setUser(defaultUser);
+
+        if (!object.isNull("source_channel"))
+        {
+            DataObject source = object.getObject("source_channel");
+            webhook.setSourceChannel(new Webhook.ChannelReference(source.getUnsignedLong("id"), source.getString("name")));
+        }
+        if (!object.isNull("source_guild"))
+        {
+            DataObject source = object.getObject("source_guild");
+            webhook.setSourceGuild(new Webhook.GuildReference(source.getUnsignedLong("id"), source.getString("name")));
+        }
+
+        return webhook;
     }
 
     public Invite createInvite(DataObject object)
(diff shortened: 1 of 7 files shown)


caneleex:
i see you added an ErrorResponse for max webhooks, what about adding it to createWebhook docs too if it's in this pr's scope?

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1437,10 +1442,23 @@ public WebhookImpl createWebhook(DataObject object)
             }
         }
         
-        return new WebhookImpl(channel, id, type)
+        WebhookImpl webhook = new WebhookImpl(channel, getJDA(), id, type)
                 .setToken(token)
-                .setOwner(owner == null ? null : channel.getGuild().getMember(owner))
+                .setOwner(owner == null || channel == null ? null : channel.getGuild().getMember(owner))
```
DV8FromTheWorld, line 1447: Is it possible that `owner` wouldn't be null but `channel` would be in this situation and thus this `||` would actually erroneously set `owner` to `null` even though we activelly have it?

src/main/java/net/dv8tion/jda/internal/entities/WebhookImpl.java
```diff
@@ -68,20 +77,22 @@ public WebhookType getType()
...
-        return channel.getGuild();
+        return getChannel().getGuild();
     }
 
     @Nonnull
     @Override
     public TextChannel getChannel()
     {
+        if (channel == null)
+            throw new IllegalStateException("Cannot provide channel for this Webhook instance");
```
DV8FromTheWorld, line 95: Should specify _why_ it can't be provided.

src/main/java/net/dv8tion/jda/internal/entities/WebhookImpl.java
```diff
@@ -68,20 +77,22 @@ public WebhookType getType()
...
-        return channel.getJDA();
+        return api;
     }
 
     @Nonnull
     @Override
     public Guild getGuild()
     {
-        return channel.getGuild();
+        return getChannel().getGuild();
```
DV8FromTheWorld, line 87: Since `getChannel()` can now throw `IllegalStateException`, this should probably manually check `channel == null` and have a Guild-specific `IllegalStateException` instead of the channel-specific one bubbling up instead.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
MinnDevelopment: The problem is we can't get the member instance since there is no guild for it in cache

PR #1469, opened 2020-12-08
Implement a timeout for unavailable guilds

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This timeout handler will automatically mark all guilds unavailable after ~~10~~ 75 seconds when hitting <=60 finished guilds. ~~I'm not entirely sure if 10 seconds is an appropriate timeout here, considering that the websocket rate limit is 60 seconds long. Maybe we should use 70 seconds instead? This would theoretically only be an issue for bots with a lot of guilds that perform chunking for every guild.~~ This is now set to 75 seconds which is the rate limit period plus 15 additional seconds for them to respond.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/handle/GuildSetupController.java
+++ b/src/main/java/net/dv8tion/jda/internal/handle/GuildSetupController.java
@@ -44,7 +44,7 @@ public class GuildSetupController
 {
     protected static final Logger log = JDALogger.getLog(GuildSetupController.class);
 
-    private static final long timeoutDuration = 10; // seconds
+    private static final long timeoutDuration = 75; // seconds
     private static final int timeoutThreshold = 60; // Half of 120 rate limit
 
     private final JDAImpl api;
--- a/src/main/java/net/dv8tion/jda/internal/handle/GuildSetupController.java
+++ b/src/main/java/net/dv8tion/jda/internal/handle/GuildSetupController.java
@@ -16,11 +16,8 @@
 
 package net.dv8tion.jda.internal.handle;
 
-import gnu.trove.iterator.TLongLongIterator;
 import gnu.trove.iterator.TLongObjectIterator;
-import gnu.trove.map.TLongLongMap;
 import gnu.trove.map.TLongObjectMap;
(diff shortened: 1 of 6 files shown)


src/main/java/net/dv8tion/jda/internal/handle/GuildSetupController.java
```diff
@@ -35,21 +32,24 @@
...
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 @SuppressWarnings("WeakerAccess")
 public class GuildSetupController
 {
-    protected static final int CHUNK_TIMEOUT = 10000;
     protected static final Logger log = JDALogger.getLog(GuildSetupController.class);
 
+    private static final long timeoutDuration = 10; // seconds
```
MinnDevelopment, line 47:
```suggestion
    private static final long timeoutDuration = 75; // seconds
```

src/main/java/net/dv8tion/jda/internal/handle/GuildSetupController.java
```diff
@@ -399,6 +383,23 @@ public void onUnavailable(long id)
         log.debug("Guild with id {} is now marked unavailable. Total: {}", id, unavailableGuilds.size());
     }
 
+    public void onTimeout()
```
DV8FromTheWorld, line 386: Should this method clear the `timeoutHandle` so that it can be set again via `startTimeout` if new guilds attempt to be loaded?

DV8FromTheWorld: Oh, I guess we are relying on the implicit clearing of `timeoutHandle` in `checkReady` due to us setting `incompleteCount` to `0` just before calling it.

MinnDevelopment: Yeah checkReady will clear it

DV8FromTheWorld (approved):

PR #1471, opened 2020-12-17
Add ActivityType.COMPETING

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds `ActivityType.COMPETING` (ID 5) and `Activity.competing(String)` to the Activity Interface.

This activity type is usable by bots, as shown below by my personal Bot that I quickly used to test this with my fork:
image

For safety did I add the `@Incubating` annotation and `@incubating` javadoc-tag to the `competing` method. I didn't add those to the ActivityType tho as the reason they are there for wating seems to be, that it is undocumented, which is (partially) wrong? Docs do contain info on the activity types, just not directly related to bots.

Will add this, if required.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Activity.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Activity.java
@@ -205,6 +205,30 @@ public interface Activity
     }
 
     /**
+     * Creates a new Activity instance with the specified name.
+     * <br>This will display as {@code Competing in name} in the official client
+     * 
+     * @param  name
+     *         The not-null name of the newly created game
+     * 
+     * @throws IllegalArgumentException
+     *         If the specified name is null, empty, blank or longer than 128 characters
+     * 
+     * @return A valid Activity instance with the provided name with {@link net.dv8tion.jda.api.entities.Activity.ActivityType#COMPETING}
+     *
+     * @incubating This feature is not yet confirmed for the official bot API
+     */
+    @Nonnull
+    @Incubating
+    static Activity competing(@Nonnull String name)
+    {
+        Checks.notBlank(name, "Name");
+        name = name.trim();
+        Checks.check(name.length() <= 128, "Name must not be greater than 128 characters in lentgh");
+        return EntityBuilder.createActivity(name, null, ActivityType.COMPETING);
+    }
+
+    /**
      * Creates a new Activity instance with the specified name and url.
      *
      * @param  type
@@ -256,6 +280,8 @@ public interface Activity
                 return listening(name);
             case WATCHING:
                 return watching(name);
+            case COMPETING:
+                return competing(name);
             default:
                 throw new IllegalArgumentException("ActivityType " + type + " is not supported!");
         }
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -204,6 +204,30 @@ static Activity watching(@Nonnull String name)
...
+     * 
+     * @param  name
+     *         The not-null name of the newly created game
+     * 
+     * @throws IllegalArgumentException
+     *         If the specified name is null, empty, blank or longer than 128 characters
+     * 
+     * @return A valid Activity instance with the provided name with {@link net.dv8tion.jda.api.entities.Activity.ActivityType#COMPETING}
+     *
+     * @incubating This feature is not yet confirmed for the official bot API
```
MinnDevelopment, line 219: Are you sure about that? Seems to be documented and working

Andre601: I essentially took this from the `watching` method which has pretty much the same description and also mentions that it isn't confirmed for the official api, despite it having the same amount of documentation as the compete one does.

src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -204,6 +204,30 @@ static Activity watching(@Nonnull String name)
...
+     */
+    @Nonnull
+    @Incubating
+    static Activity competing(@Nonnull String name)
+    {
+        Checks.notBlank(name, "Name");
+        name = name.trim();
+        Checks.check(name.length() <= 128, "Name must not be greater than 128 characters in lentgh");
```
Sheldan, line 227: Typo at the very end. It should be 'length'.

src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -308,7 +334,13 @@ static boolean isValidStreamingUrl(@Nullable String url)
...
          */
         @Incubating
-        CUSTOM_STATUS(4);
+        CUSTOM_STATUS(4),
+
+        /**
+         * Used to indicate that the {@link Activity Activity} should display
+         * as {@code Competing in...} in the official client.
+         */
+        COMPETING(5);
```
Sheldan, line 343: Why is this not marked with @Incubating as well, if the builder method is? Should be added to both for consistency.

src/main/java/net/dv8tion/jda/api/entities/Activity.java
Andre601: corrected. Thanks

src/main/java/net/dv8tion/jda/api/entities/Activity.java
Andre601: It is generally confusing as the other ones that have this annotation have the reason, that it isn't documented, despite it being documented... Just not fully.
I explained this in my initial PR comment.

src/main/java/net/dv8tion/jda/api/entities/Activity.java
MinnDevelopment: https://discord.com/developers/docs/topics/gateway#activity-object-activity-types doesn't mention watching.

Andre601: Must have overlooked it then. Will remove that then.

PR #1476, opened 2020-12-19
Improve tracking of voice members without MemberCachePolicy.VOICE

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

By tracking the connected channel independently from the member, we can persist the information regardless of the member cache policy. VoiceChannels track the connected members internally, for optimization purposes and bypasses the cache policy in that regard. This means `GuildVoiceState#getChannel` is now depending on the voice state cache, rather than being atomic. We could in theory try to optimize this by using some kinda lazy caching system, but that isn't really worth it I presume.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
@@ -406,7 +406,6 @@ public class EntityBuilder
                 VoiceChannelImpl connectedChannel = (VoiceChannelImpl) voiceState.getChannel();
                 if (connectedChannel != null)
                     connectedChannel.getConnectedMembersMap().remove(member.getIdLong());
-                voiceState.setConnectedChannel(null);
             }
 
             return false;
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/api/entities/GuildChannel.java
```diff
@@ -80,7 +80,7 @@
      * <br>For {@link net.dv8tion.jda.api.entities.TextChannel TextChannels},
      * this returns all Members with the {@link net.dv8tion.jda.api.Permission#MESSAGE_READ} Permission.
      * <br>For {@link net.dv8tion.jda.api.entities.VoiceChannel VoiceChannels},
-     * this returns all Members that joined that VoiceChannel.
+     * this returns all Members that joined that VoiceChannel. (Requires at least {@link net.dv8tion.jda.api.utils.MemberCachePolicy#VOICE MemberCachePolcy.VOICE} and {@link net.dv8tion.jda.api.utils.cache.CacheFlag#VOICE_STATE CacheFlag.VOICE_STATE})
```
caneleex, line 83:
```suggestion
     * this returns all Members that joined that VoiceChannel. (Requires at least {@link net.dv8tion.jda.api.utils.MemberCachePolicy#VOICE MemberCachePolicy.VOICE} and {@link net.dv8tion.jda.api.utils.cache.CacheFlag#VOICE_STATE CacheFlag.VOICE_STATE})
```

DV8FromTheWorld:
Bump.
Why is this in the freezer? 
Is this effected by the weirdness introduced by Public Instances of StageChannels?

MinnDevelopment:
I put this on freezer because of stage channels and need to re-evaluate the impact of lurking members. Potentially, its unaffected by that but at the time, and currently still, the stage channel voice states are not documented well.

V-Play-Games:
this might need to be reopened

PR #1478, opened 2020-12-21
add Member#isPending

caneleex:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this PR adds Member#isPending and GuildMemberUpdatePendingEvent. i'm not really sure about the event name as it's a pretty much one-way event but.. yeah

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdatePendingEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdatePendingEvent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 Austin Keener, Michael Ritter, Florian Spieß, and the JDA contributors
+ * Copyright 2015 Austin Keener, Michael Ritter, Florian Spieß, and the JDA contributors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdatePendingEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdatePendingEvent.java
@@ -16,6 +16,7 @@
 
 package net.dv8tion.jda.api.events.guild.member.update;
 
+import net.dv8tion.jda.annotations.Incubating;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.Member;
 
@@ -38,7 +39,10 @@ import javax.annotation.Nonnull;
  * to cache the updated members. Discord does not specifically tell us about the updates, but merely tells us the
  * member was updated and gives us the updated member object. In order to fire a specific event like this we
  * need to have the old member cached to compare against.
+ *
+ * @incubating Discord is still trying to figure this out
  */
+@Incubating
 public class GuildMemberUpdatePendingEvent extends GenericGuildMemberUpdateEvent<Boolean>
 {
     public static final String IDENTIFIER = "pending";
--- a/src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdatePendingEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdatePendingEvent.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2015-2020 Austin Keener, Michael Ritter, Florian Spieß, and the JDA contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.api.events.guild.member.update;
+
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.entities.Member;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Indicates that a {@link Member Member} has agreed to Membership Screening requirements.
+ *
+ * <p>Can be used to retrieve members who have agreed to Membership Screening requirements.
+ *
+ * <p>Identifier: {@code pending}
+ *
+ * <h2>Requirements</h2>
+ *
+ * <p>This event requires the {@link net.dv8tion.jda.api.requests.GatewayIntent#GUILD_MEMBERS GUILD_MEMBERS} intent to be enabled.
+ * <br>{@link net.dv8tion.jda.api.JDABuilder#createDefault(String) createDefault(String)} and
+ * {@link net.dv8tion.jda.api.JDABuilder#createLight(String) createLight(String)} disable this by default!
+ *
+ * <p>Additionally, this event also requires the {@link net.dv8tion.jda.api.utils.MemberCachePolicy MemberCachePolicy}
+ * to cache the updated members. Discord does not specifically tell us about the updates, but merely tells us the
+ * member was updated and gives us the updated member object. In order to fire a specific event like this we
+ * need to have the old member cached to compare against.
+ */
+public class GuildMemberUpdatePendingEvent extends GenericGuildMemberUpdateEvent<Boolean>
+{
+    public static final String IDENTIFIER = "pending";
+
+    public GuildMemberUpdatePendingEvent(@Nonnull JDA api, long responseNumber, @Nonnull Member member, boolean previous)
+    {
+        super(api, responseNumber, member, previous, member.isPending(), IDENTIFIER);
+    }
+
+    /**
+     * The old pending status
+     *
+     * @return The old pending status
+     */
+    public boolean getOldPending()
+    {
+        return getOldValue();
+    }
+
+    /**
+     * The new pending status
+     *
+     * @return The new pending status
+     */
+    public boolean getNewPending()
+    {
+        return getNewValue();
+    }
+}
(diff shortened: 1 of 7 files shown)


kantenkugel:
Does the Discord API also allow for checking if a Guild has rule screening enabled and for fetching/updating said rules?
Cuz that would fit into this PR as well

RedDaedalus:
Yes, it's a guild feature: `MEMBER_VERIFICATION_GATE_ENABLED`

src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdatePendingEvent.java
```diff
@@ -0,0 +1,70 @@
...
+ * <p>This event requires the {@link net.dv8tion.jda.api.requests.GatewayIntent#GUILD_MEMBERS GUILD_MEMBERS} intent to be enabled.
+ * <br>{@link net.dv8tion.jda.api.JDABuilder#createDefault(String) createDefault(String)} and
+ * {@link net.dv8tion.jda.api.JDABuilder#createLight(String) createLight(String)} disable this by default!
+ *
+ * <p>Additionally, this event also requires the {@link net.dv8tion.jda.api.utils.MemberCachePolicy MemberCachePolicy}
+ * to cache the updated members. Discord does not specifically tell us about the updates, but merely tells us the
+ * member was updated and gives us the updated member object. In order to fire a specific event like this we
+ * need to have the old member cached to compare against.
+ */
+public class GuildMemberUpdatePendingEvent extends GenericGuildMemberUpdateEvent<Boolean>
```
duncte123, line 42: This does not seem to be documented https://discord.com/developers/docs/topics/gateway#guild-member-update-guild-member-update-event-fields

duncte123 (commented):
My main concern is that it is not documented if there's an event when this gets changed

RedDaedalus:
> My main concern is that it is not documented if there's an event when this gets changed
There's a pull request to document this, but Discord doesn't make breaking changes regardless of documentation status.

duncte123:
> > My main concern is that it is not documented if there's an event when this gets changed
> 
> There's a pull request to document this, but Discord doesn't make breaking changes regardless of documentation status.

I see https://github.com/discord/discord-api-docs/pull/2396#issuecomment-749360966

duncte123 (approved):
LGTM

caneleex:
i thought about adding membership screening details too, but as those are unfinished, and more types are going to be added in the future, i think it will be better to wait until it's completely finished

kantenkugel:
Regarding latest commit: is PUBLIC gone now?

caneleex:
> 
> 
> Regarding latest commit: is PUBLIC gone now?

yes, at least it's gone from the official docs

kantenkugel:
> > Regarding latest commit: is PUBLIC gone now?
> 
> yes, at least it's gone from the official docs

And is there a replacement? cuz public basically meant discoverable/community

caneleex:
> 
> 
> > > Regarding latest commit: is PUBLIC gone now?
> > 
> > 
> > yes, at least it's gone from the official docs
> 
> And is there a replacement? cuz public basically meant discoverable/community

yep, that's literally `DISCOVERABLE` feature

kantenkugel:
Looking through the diffs of doc state of this branch and the discord docs, there are a few differences tho:
JDA docs are missing:
- COMMUNITY
- FEATURABLE

JDA Docs have extra (apart from the 2 new ones):
- MORE_EMOJI

I wouldn't touch the MORE_EMOJI tho, since i am not sure about whether it is still used for some special servers (and iirc it was always kinda hidden).

The other 2 might be wise to add here, as this MR already touches that part of the docs anyways

caneleex:
> 
> 
> Looking through the diffs of doc state of this branch and the discord docs, there are a few differences tho:
> JDA docs are missing:
> 
>     * COMMUNITY
> 
>     * FEATURABLE
> 
> 
> JDA Docs have extra (apart from the 2 new ones):
> 
>     * MORE_EMOJI
> 
> 
> I wouldn't touch the MORE_EMOJI tho, since i am not sure about whether it is still used for some special servers (and iirc it was always kinda hidden).
> 
> The other 2 might be wise to add here, as this MR already touches that part of the docs anyways

i've added community and featurable features in my community updates/rules channel pr. more_emoji is more about your decision as i don't see any issue with removing it

duncte123:
I searched about 4k guilds and did not find the feature https://canary.discord.com/channels/125227483518861312/169484473333710848/790942477443923968

I suggest that if it was to be removed it should be its own PR

Andre601:
Something I noticed, but why isn't there a `on` method for the event?
Does Discord simply not send one when this happens or what is the main reason behind this not being present? How are bots supposed to listen for this even if there isn't a `on` for the ListenerAdabter?

caneleex:
> 
> 
> Something I noticed, but why isn't there a `on` method for the event?
> Does Discord simply not send one when this happens or what is the main reason behind this not being present? How are bots supposed to listen for this even if there isn't a `on` for the ListenerAdabter?

i just forgot to add it

Andre601:
> > Something I noticed, but why isn't there a `on` method for the event?
> > Does Discord simply not send one when this happens or what is the main reason behind this not being present? How are bots supposed to listen for this even if there isn't a `on` for the ListenerAdabter?
> 
> ~i just forgot to add it~ actually no, all methods have been replaced by ClassWalker

Still needs a `on` method to my knowledge... Not sure tho.

kantenkugel:
> > > Something I noticed, but why isn't there a `on` method for the event?
> > > Does Discord simply not send one when this happens or what is the main reason behind this not being present? How are bots supposed to listen for this even if there isn't a `on` for the ListenerAdabter?
> > 
> > 
> > ~i just forgot to add it~ actually no, all methods have been replaced by ClassWalker
> 
> Still needs a `on` method to my knowledge... Not sure tho.

Yes it does. the classwalker still needs reference methods to call. if the method is not defined, it won't be called

MinnDevelopment:
Member update events only work when the member is cached.

Andre601:
> Member update events only work when the member is cached.

That's the odd thing here. Since I only use the bot on one Guild do I have the Member caching set to all.

GoldRenard:
@MinnDevelopment 
> Member update events only work when the member is cached.

Why? It's not like `GuildMemberRoleRemoveEvent`/`GuildMemberRoleAddEvent` update events where we HAVE to know member's previous data.

In case of pending it doesn't make any sense, it's very important thing that we need the fact that member accepted community rules (current new value) even if it don't cached and we can't get the previous value (which will be false most likely).

It must be fired always IMO. The old pending value could be just nullable if it is not available in cache.

caneleex:
> 
> 
> @MinnDevelopment
> 
> > Member update events only work when the member is cached.
> 
> Why? It's not like `GuildMemberRoleRemoveEvent`/`GuildMemberRoleAddEvent` update events where we HAVE to know member's previous data.
> 
> In case of pending it doesn't make any sense, it's very important thing that we need the fact that member accepted community rules (current new value) even if it doesn't cached and we can't get the previous value (which will be false most likely).
> 
> It must be always fired to get the **current** value IMO. The old pending value could be just nullable in case it isn't available in cache.

it's not like the "pending event" is an event by itself. it's a guild member update event and in order to fire the pending event, jda has to compare the previous data against the new data

MinnDevelopment:
@GoldRenard https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/api/events/guild/member/GuildMemberUpdateEvent.html

Discord only sends an update with the new data, regardless of what changes. The library cannot figure out what changed without a cache of the previous state. See also https://github.com/discord/discord-api-docs/issues/2137

GoldRenard:
> it's not like the "pending event" is an event by itself. it's a guild member update event and in order to fire the pending event, jda has to compare the previous pending boolean against the "new" pending boolean. if the previous state "isn't cached", jda cannot assume it's the pending flag being updated

Yeah I figured it out by myself after reading discord docs just before your comment. Huff...

Andre601:
Something I'm curious to know about is, what exact settings, such as Cache Flags, Chunking filter, etc, are needed for this to work.
As I mentioned before does my bot use the ALL MemberCachePolicy and the ChunkingFilter is also set to ALL, yet the event won't get fired here at all.

I currently think that this could perhaps require the CLIENT_STATUS flag, which is disabled in createDefault, to work? I really see no other reason as to why the even either doesn't get fired or doesn't give back updated values like the new boolean.

MinnDevelopment:
Maybe it makes sense to add a `MemberCachePolicy.PENDING` to make handling of pending members easier?

Example:

```java
static MemberCachePolicy PENDING = Member::isPending;
```

In theory, this should allow bots to handle pending without caching all members.

Andre601:
I think I found my issue.

I used the `getNewPending` wrong. I actually thought `true` means the member is passing the screen, while it actually is false... Didn't test it yet, but I'm sure that's the problem I had.

src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdatePendingEvent.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @return The old pending status
+     */
+    public boolean getOldPending()
+    {
+        return getOldValue();
+    }
+
+    /**
+     * The new pending status
```
Andre601, line 62: I think a little bit more explanation for how the boolean should be understood wouldn't be wrong to prevent confusion, as most people would probably assume that `true` equals passed check, which is wrong.

caneleex: well, considering the flag is called `pending` and not `notPending` and the "event" is pretty much one-way-event, it's obvious it will only fire when the member passed the screening, but sure, why not.

MinnDevelopment:
Since it seems like discord is *still* trying to figure this out we might just want to mark all these things as `@Incubating` so we can put this temporary API design into 4.2.1.

src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdatePendingEvent.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2015-2020 Austin Keener, Michael Ritter, Florian Spieß, and the JDA contributors
```
MinnDevelopment, line 2:
```suggestion
 * Copyright 2015 Austin Keener, Michael Ritter, Florian Spieß, and the JDA contributors
```

MinnDevelopment (changes requested):

MinnDevelopment (approved):

PR #1481, opened 2020-12-25
[Docs] Mention that createDefault and createLight may override GatewayIntent#DEFAULT

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

The methods `createDefault(String, GatewayIntent, GatewayIntent...)`, `createLight(String, GatewayIntent, GatewayIntent...)` as well as their counter-parts with Collection do not mention that using those essentially overrides the `GatewayIntent#DEFAULT` as you basically have to provide your own ones, instead of adding them to the defaults.
This can cause people to believe that those methods are used to define the default intents and also add additional ones you need (i.e. GUILD_MEMBERS),

The only real mention is, when you compare it to `createDefault(String)`/`createLight(String)` where one extra List entry mentions the intents to be set to `GatewayIntent#DEFAULT`

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
@@ -175,6 +175,9 @@ public class  DefaultShardManagerBuilder
      *     <li>This disables {@link CacheFlag#ACTIVITY} and {@link CacheFlag#CLIENT_STATUS}</li>
      * </ul>
      *
+     * <p>You can omit intents in this method to use {@link GatewayIntent#DEFAULT} and enable additional intents with
+     * {@link #enableIntents(Collection)}.
+     *
      * <p>If you don't enable certain intents, the cache will be disabled.
      * For instance, if the {@link GatewayIntent#GUILD_MEMBERS GUILD_MEMBERS} intent is disabled, then members will only
      * be cached when a voice state is available.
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
```diff
@@ -187,9 +187,9 @@ public static DefaultShardManagerBuilder createDefault(@Nullable String token)
      * @param  token
      *         The bot token to use
      * @param  intent
-     *         The intent to enable
+     *         The intent to enable. This will override the {@link GatewayIntent#DEFAULT default Intents} used by this method
```
MinnDevelopment, line 190: This should be in the description and a better wording is:

> You can omit intents in this method to use GatewayIntent.DEFAULT and enable additional intents with enableIntents(...)

Andre601: Where would the comment fit the best?
As a `<br>` under the top one or better as its own paragraph?

MinnDevelopment: Doesn't matter to me

Andre601: I decided to make it a own paragraph underneath the list.

MinnDevelopment:
I fixed your commit history. You can use the following commands to update your local repository:

```sh
git fetch --all
git reset --hard origin/patch-2
```

PR #1482, opened 2020-12-27
[Wiki] updated setup guide for Eclipse

Horstexplorer:
## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: Wiki

Closes Issue: NaN

## Description

This pr adds various images for the project setup wiki page on eclipse (2020-12). 
As I'm unable to properly pr the changes for the wiki page I just placed it on [not hastebin](https://gist.github.com/MinnDevelopment/13b8493f2bedc0859b6305ccaf48ad02)
Not a lot has changed for the setup process in general so I mostly changed the config snippets to match what JDA-Butler returns  and added some new images to make following along easier.


DV8FromTheWorld:
Unfortunately, it appears that the hastebin is empty

Andre601:
> Unfortunately, it appears that the hastebin is empty

It worked yesterday. hastebin.com is extremely wanky and sometimes like to die or not load stuff, which is why people stick with either hasteb.in or any other selfhosted option.

arynxd:
> Unfortunately, it appears that the hastebin is empty

see the updated link

Horstexplorer:
Yea using hastebin wasnt a great idea but I did not expect it to disappear in less than 24 hours.
Thanks for the edit Minn :)

Issue #1485, opened 2020-12-31
GuildChannel#isSynced doesn't respect permissions


MinnDevelopment:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

The check in [GuildChannel#isSynced](https://github.com/DV8FromTheWorld/JDA/blob/2b364c84b61825ea00ca29123cbb5516f5a89838/src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java#L164-L171) only checks if the list of overrides is equal. However, this equality check doesn't include the permissions configured in those overrides. See [PermissionOverrideImpl#equals](https://github.com/DV8FromTheWorld/JDA/blob/2b364c84b61825ea00ca29123cbb5516f5a89838/src/main/java/net/dv8tion/jda/internal/entities/PermissionOverrideImpl.java#L220).

The fix for this is easy, simply check the individual permissions rather than doing an equality check.

### Expected Behavior

The method should also check permissions of the individual overrides.

### Code Example or Reproduction Steps

Make overrides with different permissions but the same ids.

### Code for JDABuilder or DefaultShardManagerBuilder Used

N/A

### Exception or Error

N/A

Taucher2003:
I propably take a look on this

Andre601:
I currently think of how this could be done...
Maybe first check if the size of POs is equal and if not directly return false and then when it is equal check each perm individually?

Taucher2003:
I'm currently working on this and are ready to test if I'm correct

Andre601:
Already cooked up a possible solution used in the AbstractChannelImpl, not sure if it would be useful or good tho:
```java
    @Override
    public boolean isSynced()
    {
        if (getParent() == null)
            return true; // Channels without a parent category are always considered synced. Also the case for categories.
        
        if (getParent().getPermissionOverrides().size() != getPermissionOverrides().size())
            return false; // Amount of Overrides doesn't match the ones of the parent.
        
        for(int i = 0; i < getParent().getPermissionOverrides().size(); i++){
[...]

PR #1486, opened 2020-12-31
Fixed isSynced() in AbstractChannelImpl

Taucher2003:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: https://github.com/DV8FromTheWorld/JDA/issues/1485

## Description

This PR fixes, that the isSynced() method in AbstractChannelImpl doesn't check if the permissions of the PermissionOverrides are equal. It just checked, if the category had the PermissionOverrides with the same id.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
@@ -168,15 +168,15 @@ public abstract class AbstractChannelImpl<T extends GuildChannel, M extends Abst
         Stream<PermissionOverride> thisStream = getRolePermissionOverrides().stream().filter(
                 po -> po.getAllowedRaw() != 0 || po.getDeniedRaw() != 0
         );
-        Stream<PermissionOverride> parentStream = getParent().getRolePermissionOverrides().stream().filter(
+        HashMap<Long, PermissionOverride> parentCheck = new HashMap<>();
+        getParent().getRolePermissionOverrides().stream().filter(
                 po -> po.getAllowedRaw() != 0 || po.getDeniedRaw() != 0
-        );
+        ).forEach(po -> parentCheck.put(po.getIdLong(), po));
         Collection<PermissionOverride> thisCheck = thisStream.collect(Collectors.toList());
-        Collection<PermissionOverride> parentCheck = parentStream.collect(Collectors.toList());
 
         for(PermissionOverride check : thisCheck)
         {
-            PermissionOverride matching = parentCheck.stream().filter(po -> check.getIdLong() == po.getIdLong()).findFirst().orElse(null);
+            PermissionOverride matching = parentCheck.getOrDefault(check.getIdLong(), null);
             if(matching == null)
                 return false;
 
--- a/src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
@@ -38,13 +38,14 @@ import net.dv8tion.jda.internal.requests.restaction.AuditableRestActionImpl;
 import net.dv8tion.jda.internal.requests.restaction.InviteActionImpl;
 import net.dv8tion.jda.internal.requests.restaction.PermissionOverrideActionImpl;
 import net.dv8tion.jda.internal.utils.Checks;
-import net.dv8tion.jda.internal.utils.Helpers;
 
 import javax.annotation.Nonnull;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 public abstract class AbstractChannelImpl<T extends GuildChannel, M extends AbstractChannelImpl<T, M>> implements GuildChannel
 {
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
```diff
@@ -166,8 +164,26 @@ public boolean isSynced()
...
+        );
+        Stream<PermissionOverride> parentStream = getParent().getRolePermissionOverrides().stream().filter(
+                po -> po.getAllowedRaw() != 0 || po.getDeniedRaw() != 0
+        );
+        Collection<PermissionOverride> thisCheck = thisStream.collect(Collectors.toList());
+        Collection<PermissionOverride> parentCheck = parentStream.collect(Collectors.toList());
+
+        for(PermissionOverride check : thisCheck)
+        {
+            PermissionOverride matching = parentCheck.stream().filter(po -> check.getIdLong() == po.getIdLong()).findFirst().orElse(null);
```
MinnDevelopment, line 179: This seems rather inefficient when you have a map of id -> override at your disposal.

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
```diff
@@ -166,8 +164,26 @@ public boolean isSynced()
     {
         if (getParent() == null)
             return true; // Channels without a parent category are always considered synced. Also the case for categories.
-        
-        return Helpers.deepEqualsUnordered(getParent().getPermissionOverrides(), getPermissionOverrides());
+
+        Stream<PermissionOverride> thisStream = getRolePermissionOverrides().stream().filter(
+                po -> po.getAllowedRaw() != 0 || po.getDeniedRaw() != 0
+        );
+        Stream<PermissionOverride> parentStream = getParent().getRolePermissionOverrides().stream().filter(
```
MinnDevelopment, line 171: What about members?

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
```diff
@@ -166,8 +164,26 @@ public boolean isSynced()
...
+                po -> po.getAllowedRaw() != 0 || po.getDeniedRaw() != 0
```
MinnDevelopment, line 172: The client also says its not synced for empty overrides

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
Taucher2003: didn't noticed, that I took the one for the roles, will change that

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
Taucher2003: In my testing, in the JDA there was an empty one in the category and it was not present in the channel. My client showed it as synced. That is why I filtered them out

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
Taucher2003: I will look for an better option

Andre601:
Why even make a separate stream and then turn it into a collection? Couldn't you just stream the overrides, filter and then collect to a Collection or List in one go? Why creating this separate stream instance in the first place?

Taucher2003:
> Why even make a separate stream and then turn it into a collection? Couldn't you just stream the overrides, filter and then collect to a Collection or List in one go? Why creating this separate stream instance in the first place?

Yes, that would be possible, didn't thougt about that I could do it directly

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
```diff
@@ -166,8 +164,26 @@ public boolean isSynced()
...
+        ).forEach(po -> parentCheck.put(po.getIdLong(), po));
+
+        for(PermissionOverride check : thisCheck)
+        {
+            PermissionOverride matching = parentCheck.getOrDefault(check.getIdLong(), null);
+            if(matching == null)
+                return false;
+
+            if(matching.getAllowedRaw() != check.getAllowedRaw() || matching.getDeniedRaw() != check.getDeniedRaw())
```
MinnDevelopment, line 183: I think the client only considers permissions that are valid for channels here. In some circumstances bots mess up permissions on channels and add stuff like `MANAGE_GUILD` or `ADMINISTRATOR` to channel overrides, we should try to filter that out.  A useful constant for this is [`Permission.ALL_CHANNEL_PERMISSIONS`](https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/api/Permission.html#ALL_CHANNEL_PERMISSIONS)

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
```diff
@@ -166,8 +164,26 @@ public boolean isSynced()
...
         if (getParent() == null)
             return true; // Channels without a parent category are always considered synced. Also the case for categories.
-        
-        return Helpers.deepEqualsUnordered(getParent().getPermissionOverrides(), getPermissionOverrides());
+
+        Collection<PermissionOverride> thisCheck = getPermissionOverrides().stream().filter(
+                po -> po.getAllowedRaw() != 0 || po.getDeniedRaw() != 0
+        ).collect(Collectors.toList());
+
+        HashMap<Long, PermissionOverride> parentCheck = new HashMap<>();
```
MinnDevelopment, line 172: Why not `this.overrides`?

PR #1493, opened 2021-01-08
clarify Guild#getLocale for guilds without the COMMUNITY feature

caneleex:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this should hopefully clarify what this method returns if the guild doesn't have the community feature as i think it's not clear at all from current docs

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -417,7 +417,9 @@ public interface Guild extends ISnowflake
 
     /**
      * The preferred locale for this guild.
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -417,6 +417,7 @@ default String getVanityUrl()
 
     /**
      * The preferred locale for this guild.
+     * <br>If the guild doesn't have the COMMUNITY feature, this returns {@link Locale#US Locale.US}.
```
MinnDevelopment, line 420: This is also US when its not set, since it is the default

Issue #1499, opened 2021-01-23
Allow modification of existing permissions in ChannelAction


ScoreUnder:
Inspired by discord/discord-api-docs#2520

## General Troubleshooting
  
- [X] I have checked for similar issues.
- [X] I have updated to the [latest JDA version][download].
- [X] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request

I would like a way to view and update existing permission overrides from a `ChannelAction` object.

### Example Use-Case

Because of discord/discord-api-docs#2520, when using JDA's `Category#createVoiceChannel` you must now sanitise out all `MANAGE_ROLES` permissions unless the bot is given admin permissions. Currently I'm just using this as a quick hack, which is not ideal:

```scala
private object JDAHack {
  lazy val field: Field = {
    val f = classOf[ChannelActionImpl[_]].getDeclaredField("overrides")
    f.setAccessible(true)
    f
  }

  def removeManageRolesPerms(channelReq: ChannelAction[_]): Unit = {
    try {
      val overrides = field.get(channelReq).asInstanceOf[TLongObjectHashMap[PermOverrideData]]
      overrides.transformValues { v =>
        val manageRoles = Permission.MANAGE_ROLES.getRawValue
        new PermOverrideData(v.`type`, v.id, v.allow & ~manageRoles, v.deny & ~manageRoles)
      }
    } catch {
      case NonFatal(_) =>
    }
  }
}
```

Andre601:
I'm not an expert, but when you create the channel should you be able to set allow and deny using setAllow/setDeny.

Alternatively can you do `VoiceChannel#getManager().upsertPermission(...)` in the success callback of `queue()` from the create action.

Tl;dr: there are already ways to change a channel's permission individually and your "hack" is redundant.

ScoreUnder:
> but when you create the channel should you be able to set allow and deny using setAllow/setDeny.

But I still have no control over permissions applied automatically by `Category#createVoiceChannel`

> in the success callback of `queue()` from the create action.

It can never succeed on a non-admin user when this permission is in the action

MinnDevelopment:
You can use [`Guild#createVoiceChannel(name, parent)`](https://github.com/DV8FromTheWorld/JDA/blob/ac6fc8d4c93593cc134825d6cfb7b4ec7453b1b1/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java#L1343) which starts off with no overrides at all.

[...]

PR #1500, opened 2021-01-23
Fix permission sync with manage roles/permissions

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: #1499 

## Description

This is a fix for the breaking change introduced last night by night. See https://github.com/discord/discord-api-docs/issues/2522

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/CategoryImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/CategoryImpl.java
@@ -159,8 +159,7 @@ public class CategoryImpl extends AbstractChannelImpl<Category, CategoryImpl> im
     public ChannelAction<TextChannel> createTextChannel(@Nonnull String name)
     {
         ChannelAction<TextChannel> action = getGuild().createTextChannel(name, this);
-        applyPermission(action);
-        return action;
+        return action.syncPermissionOverrides();
     }
 
     @Nonnull
@@ -168,8 +167,7 @@ public class CategoryImpl extends AbstractChannelImpl<Category, CategoryImpl> im
     public ChannelAction<VoiceChannel> createVoiceChannel(@Nonnull String name)
     {
         ChannelAction<VoiceChannel> action = getGuild().createVoiceChannel(name, this);
-        applyPermission(action);
-        return action;
+        return action.syncPermissionOverrides();
     }
 
     @Nonnull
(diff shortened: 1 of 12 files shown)


MinnDevelopment:
I also have to do the permission changes in `ChannelManager` but I wanna wait for a reply on something first.

src/main/java/net/dv8tion/jda/internal/managers/ChannelManagerImpl.java
```diff
@@ -171,8 +173,23 @@ public ChannelManagerImpl putPermissionOverride(@Nonnull IPermissionHolder permH
     {
         Checks.notNull(permHolder, "PermissionHolder");
         Checks.check(permHolder.getGuild().equals(getGuild()), "PermissionHolder is not from the same Guild!");
-        if (isPermissionChecksEnabled() && !getGuild().getSelfMember().hasPermission(getChannel(), Permission.MANAGE_PERMISSIONS))
-            throw new InsufficientPermissionException(getChannel(), Permission.MANAGE_PERMISSIONS);
+        Member selfMember = getGuild().getSelfMember();
+        if (isPermissionChecksEnabled() && !selfMember.hasPermission(Permission.ADMINISTRATOR))
```
MinnDevelopment, line 177: You also can only set `MANAGE_PERMISSIONS` on a channel if you have it as an override on the channel or `ADMINISTRATOR` on a role.

src/main/java/net/dv8tion/jda/internal/managers/ChannelManagerImpl.java
```diff
@@ -211,11 +230,36 @@ public ChannelManagerImpl sync(@Nonnull GuildChannel syncSource)
...
+                    || (PermissionUtil.getExplicitPermission(getChannel(), selfMember, false) & Permission.MANAGE_PERMISSIONS.getRawValue()) != 0;
+
+            if (!canSetRoles)
+            {
+                // Requires either administrator or MANAGE_PERMISSIONS explicitly on the channel itself
+                long botPerms = PermissionUtil.getEffectivePermission(getChannel(), selfMember) & ~Permission.MANAGE_PERMISSIONS.getRawValue();
+                boolean wouldRequireManageRoles = syncSource
+                        .getPermissionOverrides()
+                        .stream()
+                        .mapToLong(p -> p.getAllowedRaw() | p.getDeniedRaw())
```
MinnDevelopment, line 252: This should check if an existing override exists, and if thats the case, what the changed bits are.

src/main/java/net/dv8tion/jda/internal/entities/CategoryImpl.java
```diff
@@ -159,17 +159,15 @@ public InviteAction createInvite()
...
     }
 
     @Nonnull
     @Override
     public ChannelAction<VoiceChannel> createVoiceChannel(@Nonnull String name)
     {
         ChannelAction<VoiceChannel> action = getGuild().createVoiceChannel(name, this);
-        applyPermission(action);
-        return action;
+        return action.syncPermissionOverrides();
```
MinnDevelopment, line 170: Maybe we should just only sync if we actually have perms to do it?

MinnDevelopment:
We could also add a `IPermissionHolder#canSync(channel, parent)` to simplify the checks and allow common users to also check these?

Issue #1385, opened 2020-09-02
JavaDocs


Horstexplorer:
## Bug Report

I found a small issue in the javadocs for the GuildMessageReactionAddEvent.

The methods getUser() and getMember() are annotated having a non null return but the description copied from the super class GenericGuildMessageReactionEvent states that this might be null.

MinnDevelopment:
This is not a bug. Specifications can override the nullability of the return type.

PR #1516, opened 2021-02-12
Docs: Nullability in GuildMessageReactionAddEvent

MCOfficer:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: #1385

## Description

As of now, the docs for GuildMessageReactionAddEvent are copied from GenericGuildMessageReactionEvent, which states that getMember and getUser can be null. However, GuildMessageReactionAddEvent is guaranteed to have a user and member, and there are annotations in place to indicate this.

This PR adds 2 javadoc sections to GuildMessageReactionAddEvent, so it doesn't wrongly use those of its superclass.

The topic was already brought up in #1385, and deemed not a bug. I'd argue that more documentation and thus less confusion is always good, but if you disagree, it's your call to close this.


MinnDevelopment:
This needs to be rebased on `docs/development`.

DV8FromTheWorld (approved):

PR #1519, opened 2021-02-16
Allow building DefaultShardManager without starting it

nikammerlaan:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

Allows a DefaultShardManager to be built without immediately starting it. This is useful in dependency injection situations, where the ShardManager might need to be injected into several places before it's started.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
@@ -2301,7 +2301,7 @@ public class  DefaultShardManagerBuilder
      *         to whether or not loading has finished when this returns.
      */
     @Nonnull
-    public ShardManager build() throws LoginException, IllegalArgumentException
+    public DefaultShardManager build() throws LoginException, IllegalArgumentException
     {
         return build(true);
     }
@@ -2327,7 +2327,7 @@ public class  DefaultShardManagerBuilder
      *         to whether or not loading has finished when this returns.
      */
     @Nonnull
-    public ShardManager build(boolean login) throws LoginException, IllegalArgumentException
+    public DefaultShardManager build(boolean login) throws LoginException, IllegalArgumentException
     {
         checkIntents();
         boolean useShutdownNow = shardingFlags.contains(ShardingConfigFlag.SHUTDOWN_NOW);
--- a/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
@@ -2316,15 +2316,17 @@ public class  DefaultShardManagerBuilder
      * <p>Note that this method is async and as such will <b>not</b> block until all shards are started.
      *
      * @param  login
-     *         Whether the login process will be started
+     *         Whether the login process will be started. If this is false, then you will need to manually call
+     *         {@link net.dv8tion.jda.api.sharding.ShardManager#login()} to start it.
      *
      * @throws  LoginException
      *          If the provided token is invalid.
      * @throws  IllegalArgumentException
      *          If the provided token is empty or null. Or the provided intents/cache configuration is not possible.
      *
-     * @return A {@link net.dv8tion.jda.api.sharding.ShardManager ShardManager} instance that has started the login process. It is unknown as
-     *         to whether or not loading has finished when this returns.
+     * @return A {@link net.dv8tion.jda.api.sharding.ShardManager ShardManager} instance. If {@code login} is set to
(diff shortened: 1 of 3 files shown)


Andre601:
What exact use-case can be imagined with this?
Like you say it's for stuff like dependency injection and such, but what exactly do you mean?

I myself use Dependency injection a lot in my bot but never encountered any real issues with the shardmanager in that regard.

nikammerlaan:
> What exact use-case can be imagined with this?
> Like you say it's for stuff like dependency injection and such, but what exactly do you mean?
> 
> I myself use Dependency injection a lot in my bot but never encountered any real issues with the shardmanager in that regard.

It's to ensure that everything can properly get injected and created before the ShardManager starts logging in and publishing events. Many of the event listeners I have are indirectly dependent on a ShardManager, so they cannot be registered prior to the ShardManager starting. It's also nice to be able to verify that everything else was properly initialized before starting the ShardManager so that identifies are not wasted in the event of a bad setup.

src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
```diff
@@ -2292,7 +2318,9 @@ public ShardManager build() throws LoginException, IllegalArgumentException
         final ShardingMetaConfig metaConfig = new ShardingMetaConfig(maxBufferSize, contextProvider, cacheFlags, flags, compression, encoding);
         final DefaultShardManager manager = new DefaultShardManager(this.token, this.shards, shardingConfig, eventConfig, presenceConfig, threadingConfig, sessionConfig, metaConfig, chunkingFilter);
 
-        manager.login();
+        if(login) {
+            manager.login();
+        }
```
MinnDevelopment, line 2323:
```suggestion
        if (login)
             manager.login();
```

duncte123:
How will you start the ShardManager after building it? The login method is not public

nikammerlaan:
> How will you start the ShardManager after building it? The login method is not public

It is public

https://github.com/DV8FromTheWorld/JDA/blob/3a9fcd0ecc88f988ac81eb2c18527283a83f12d1/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManager.java#L259

duncte123:
> > How will you start the ShardManager after building it? The login method is not public
> 
> It is public
> 
> https://github.com/DV8FromTheWorld/JDA/blob/3a9fcd0ecc88f988ac81eb2c18527283a83f12d1/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManager.java#L259

That would only work when casting it to the default shard manager, it should probably be added to the interface

src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
```diff
@@ -2301,7 +2301,7 @@ public DefaultShardManagerBuilder setMaxBufferSize(int bufferSize)
      *         to whether or not loading has finished when this returns.
      */
     @Nonnull
-    public ShardManager build() throws LoginException, IllegalArgumentException
+    public DefaultShardManager build() throws LoginException, IllegalArgumentException
```
nikammerlaan, line 2304: Is this a breaking change?

MinnDevelopment: Yes a changed return type is breaking the ABI

nikammerlaan: Would it better to add `login()` to the ShardManager interface?

MinnDevelopment: Yeah, I think that would be better.

nikammerlaan:
I'm not sure how I managed to close this PR

src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
```diff
@@ -2302,6 +2302,34 @@ public DefaultShardManagerBuilder setMaxBufferSize(int bufferSize)
...
+     * being performed like database connection or local resource loading.
+     *
+     * <p>Note that this method is async and as such will <b>not</b> block until all shards are started.
+     *
+     * @param  login
+     *         Whether the login process will be started. If this is false, then you will need to manually call
+     *         {@link net.dv8tion.jda.api.sharding.ShardManager#login()} to start it.
+     *
+     * @throws  LoginException
+     *          If the provided token is invalid.
```
MinnDevelopment, line 2323: This is only thrown if login is true

PR #1533, opened 2021-03-08
Improve some logging for large rate limits

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

We should always log long rate limits on warn level, unless its also a global rate limit (which is already logged elsewhere).

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/ratelimit/BotRateLimiter.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/ratelimit/BotRateLimiter.java
@@ -266,11 +266,12 @@ public class BotRateLimiter extends RateLimiter
                     // Handle hard rate limit, pretty much just log that it happened
                     else
                     {
-                        boolean firstHit = hitRatelimit.add(baseRoute);
+                        boolean firstHit = retryAfter >= 6000 || hitRatelimit.add(baseRoute);
                         // Update the bucket to the new information
                         bucket.remaining = 0;
                         bucket.reset = getNow() + retryAfter;
                         // don't log warning if we hit the rate limit for the first time, likely due to initialization of the bucket
+                        // unless its a long retry-after delay (more than a minute)
                         if (firstHit)
                             log.debug("Encountered 429 on route {} with bucket {} Retry-After: {} ms", baseRoute, bucket.bucketId, retryAfter);
                         else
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/requests/ratelimit/BotRateLimiter.java
```diff
@@ -266,11 +266,12 @@ else if (cloudflare)
                     // Handle hard rate limit, pretty much just log that it happened
                     else
                     {
-                        boolean firstHit = hitRatelimit.add(baseRoute);
+                        boolean firstHit = retryAfter >= 6000 || hitRatelimit.add(baseRoute);
```
gpluscb, line 269: Should this be `<`? We want to log it as debug only if it's *not* long or we haven't hit it before if I understand correctly.

MinnDevelopment: Good catch, my mistake.
```suggestion
                        boolean firstHit = retryAfter < 6000 || hitRatelimit.add(baseRoute);
```

src/main/java/net/dv8tion/jda/internal/requests/ratelimit/BotRateLimiter.java
```diff
@@ -266,11 +266,12 @@ else if (cloudflare)
                     // Handle hard rate limit, pretty much just log that it happened
                     else
                     {
-                        boolean firstHit = hitRatelimit.add(baseRoute);
+                        boolean firstHit = retryAfter < 6000 || hitRatelimit.add(baseRoute);
```
MinnDevelopment, line 269:
```suggestion
                        boolean firstHit = retryAfter < 60000 || hitRatelimit.add(baseRoute);
```

src/main/java/net/dv8tion/jda/internal/requests/ratelimit/BotRateLimiter.java
```diff
@@ -266,11 +266,12 @@ else if (cloudflare)
                     // Handle hard rate limit, pretty much just log that it happened
                     else
                     {
-                        boolean firstHit = hitRatelimit.add(baseRoute);
+                        boolean firstHit = retryAfter < 60000 || hitRatelimit.add(baseRoute);
```
MinnDevelopment, line 269:
```suggestion
                        boolean firstHit = hitRatelimit.add(baseRoute) && retryAfter < 60000;
```

PR #1543, opened 2021-03-13
Mention that CUSTOM_STATUS is read only

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds a line to the `CUSTOM_STATUS` entry mentioning, that it is read-only for bots, to prevent any possible confusion.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Activity.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Activity.java
@@ -334,9 +334,8 @@ public interface Activity
         /**
          * Used to indicate that the {@link Activity Activity} should display as a custom status
          * in the official client.
-         * <br>This Activity type is <b>read only</b> for bots.
          *
-         * @incubating This feature is currently not officially documented and might change
+         * @incubating This Activity type is <b>read-only</b> for bots
          */
         @Incubating
         CUSTOM_STATUS(4),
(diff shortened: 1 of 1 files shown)


Andre601:
Corrected some typos where `of(ActivityType, String)` mentions url and also added case where null or wrong ActivityType would cause an IllegalArgumentException

src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -328,6 +334,7 @@ static boolean isValidStreamingUrl(@Nullable String url)
         /**
          * Used to indicate that the {@link Activity Activity} should display as a custom status
          * in the official client.
+         * <br>This Activity type is <b>read only</b> for bots.
          *
          * @incubating This feature is currently not officially documented and might change
```
caneleex, line 339: now that custom status is documented and i'd say "stable", would removing the incubating annotation be okay?

Andre601: Do you have a Link to the documentation?

caneleex: https://discord.com/developers/docs/topics/gateway#activity-object-activity-types

MinnDevelopment: It is incubating because bots still can't use it.

PR #1547, opened 2021-03-16
Set socket timeout before connecting websockets

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

By default we use a timeout of 10 seconds, just to prevent it from deadlocking. You can configure a lower timeout (min. 1 second) with `WebSocketFactory#setSocketTimeout` and passing that factory to one of the JDA builders.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/audio/AudioWebSocket.java
+++ b/src/main/java/net/dv8tion/jda/internal/audio/AudioWebSocket.java
@@ -122,6 +122,8 @@ class AudioWebSocket extends WebSocketAdapter
         {
             WebSocketFactory socketFactory = new WebSocketFactory(getJDA().getWebSocketFactory());
             IOUtil.setServerName(socketFactory, wssEndpoint);
+            if (socketFactory.getSocketTimeout() < 1000)
+                socketFactory.setSocketTimeout(10000);
             socket = socketFactory.createSocket(wssEndpoint);
             socket.setDirectTextMessage(true);
             socket.addListener(this);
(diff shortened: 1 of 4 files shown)


DV8FromTheWorld:
Based on the comment of "minimum is 1000", shouldn't we be doing a `< 1000` and setting to `1000` in that case?

src/main/java/net/dv8tion/jda/internal/audio/AudioWebSocket.java
```diff
@@ -122,6 +122,8 @@ protected void startConnection()
         {
             WebSocketFactory socketFactory = new WebSocketFactory(getJDA().getWebSocketFactory());
             IOUtil.setServerName(socketFactory, wssEndpoint);
+            if (socketFactory.getSocketTimeout() < 1000)
+                socketFactory.setSocketTimeout(10000);
```
MinnDevelopment, line 126:
```suggestion
            if (socketFactory.getSocketTimeout() > 0)
                socketFactory.setSocketTimeout(Math.max(1000, socketFactory.getSocketTimeout()));
            else
                socketFactory.setSocketTimeout(10000);
```

src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java
```diff
@@ -366,6 +366,8 @@ protected synchronized void connect()
         {
             WebSocketFactory socketFactory = new WebSocketFactory(api.getWebSocketFactory());
             IOUtil.setServerName(socketFactory, url);
+            if (socketFactory.getSocketTimeout() < 1000)
+                socketFactory.setSocketTimeout(10000);
```
MinnDevelopment, line 370:
```suggestion
            if (socketFactory.getSocketTimeout() > 0)
                socketFactory.setSocketTimeout(Math.max(1000, socketFactory.getSocketTimeout()));
            else
                socketFactory.setSocketTimeout(10000);
```

DV8FromTheWorld (approved):

PR #1548, opened 2021-03-20
fix npe's and wrong length in EmbedBuilder#length()

ToxicMushroom:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fixes 2 possible npe's and not cumulatively adding the fields' length


ToxicMushroom:
I noticed that the map { ... }.reduce for the fields also has getters that are nullable. Should I add a commit to validate them as well ?

ToxicMushroom:
Okay the reduce method apparently takes the first parameter and then adds the sum of all elements to it so it was correct.
I guess I can close this pr.

Issue #1550, opened 2021-03-21
Increased CPU usage since the new ListenerAdapter implementation (version 224)


21Joakim:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report
A friend of mine noticed a large increase in CPU usage (on a bot with 30k servers with presence events) after updating to the latest version so I decided to investigate it a bit. Turns out this was due to the new ListenerAdapter implementation that now uses reflection.

I wrote up a simple benchmark using the previous and current ListenerAdapter
```Java
public class Test {
	
	public static final int WARMUP_EVENTS = 10_000;
	public static final int EVENTS = 1_000_000;
	
	public static final RawGatewayEvent DUMMY_EVENT = new RawGatewayEvent(null, 0, null);
	
	public static void main(String[] args) {
		runTest(new ListenerAdapterOld() {}, "old ListenerAdapter");
		runTest(new ListenerAdapter() {}, "current ListenerAdapter");
	}
	
	public static void runTest(EventListener listener, String description) {
		for(int i = 0; i < WARMUP_EVENTS; i++) {
			listener.onEvent(DUMMY_EVENT);
		}
		
		long start = System.nanoTime();
		for(int i = 0; i < EVENTS; i++) {
			listener.onEvent(DUMMY_EVENT);
		}
		long time = System.nanoTime() - start;
		
		System.out.println(String.format("Took a total of: %sns (an average of %.2fns) for %s", time, (double) time/EVENTS, description));
	}
}
```
and the results were as follows
```
Took a total of: 54969800ns (an average of 54,97ns) for old ListenerAdapter
Took a total of: 3878588500ns (an average of 3878,59ns) for current ListenerAdapter
```
to my surprise it was a whole 70x slower than the old method and I could even hear my fans spinning up.

**Note: not only is this 70x slower overall but 70x slower for every ListenerAdapter you have registered!**

My proposed solution would be to cache all of the methods that will be called for each event, example:
```Java
private Map<Class<?>, List<Method>> eventMethods = new HashMap<>();

@Override
[...]

PR #1551, opened 2021-03-21
Use MethodHandle for ListenerAdapter

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1550

## Description

This slightly improves performance by caching the method handles over time. This trades memory for CPU usage.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
+++ b/src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
@@ -89,6 +89,8 @@ import javax.annotation.Nonnull;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
+import java.util.Collections;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
@@ -390,8 +392,19 @@ public abstract class ListenerAdapter implements EventListener
     public void onGenericEmoteUpdate(@Nonnull GenericEmoteUpdateEvent event) {}
     public void onGenericPermissionOverride(@Nonnull GenericPermissionOverrideEvent event) {}
 
-    private static final MethodHandles.Lookup lookup = MethodHandles.publicLookup();
+    private static final MethodHandles.Lookup lookup = MethodHandles.lookup();
     private static final ConcurrentMap<Class<?>, MethodHandle> methods = new ConcurrentHashMap<>();
+    private static final Set<Class<?>> unresolved;
+    static
+    {
+        unresolved = ConcurrentHashMap.newKeySet();
+        Collections.addAll(unresolved,
+            Object.class, // Objects aren't events
+            Event.class, // onEvent is final and would never be found
+            UpdateEvent.class, // onGenericUpdate has already been called
+            GenericEvent.class // onGenericEvent has already been called
+        );
+    }
 
     @Override
     public final void onEvent(@Nonnull GenericEvent event)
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
```diff
@@ -400,26 +406,40 @@ public final void onEvent(@Nonnull GenericEvent event)
...
-            catch (InvocationTargetException e)
-            {
-                Throwable cause = e.getCause();
-                if (cause instanceof RuntimeException)
-                    throw (RuntimeException) cause;
-                if (cause instanceof Error)
-                    throw (Error) cause;
-                throw new RuntimeException(cause);
-            }
+            MethodHandle mh = methods.computeIfAbsent(clazz, ListenerAdapter::findMethod);
```
21Joakim, line 411: ConcurrentHashMap won't enter the value if it is null
> If the specified key is not already associated with a value, attempts to compute its value using the given mapping function and **enters it into this map unless null**.

which means it won't cache the result when it can't find the method (most notably that happened for Event and Object) removing a lot of the optimization (from my benchmark this actually ended up being over 4x slower than the current method)

DV8FromTheWorld:
This needs to be run against https://gist.github.com/DV8FromTheWorld/531712ade36fb0525938c3522fdaeb51 to ensure it generates the same output as the current impl.

Based on the changes I imagine it will, but we should confirm.

src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
```diff
@@ -400,26 +419,42 @@ public final void onEvent(@Nonnull GenericEvent event)
...
+                    throw (Error) throwable;
+                throw new IllegalStateException(throwable);
             }
         }
     }
+
+    private static MethodHandle findMethod(Class<?> clazz)
+    {
+        String name = clazz.getSimpleName();
+        MethodType type = MethodType.methodType(Void.TYPE, clazz);
```
DV8FromTheWorld, line 451: This is a change, isn't it? This now enforces that we only handle methods that return `void`. Before, any return could be used.

Note, I don't think this is a big deal, I just wanted to call it out.

MinnDevelopment: It actually isn't a change since we only checked the methods in ListenerAdapter anyways.

DV8FromTheWorld: Oh that's right. I forgot that it didn't look at the subclass itself.

MinnDevelopment:
I ran your test class: [output.txt](https://gist.github.com/MinnDevelopment/c8e172376abcf1c5cf6b0f3837510767)
Let me know if you spot any problems.

DV8FromTheWorld:
Did a compare of current vs your output. They match, so we're good.

DV8FromTheWorld (approved):

21Joakim:
For completeness sake here's the benchmark with this one added (for anyone looking at this unaware: #1550)
```
Took a total of: 58836000ns (an average of 58,84ns) for old ListenerAdapter
Took a total of: 3433348200ns (an average of 3433,35ns) for current ListenerAdapter
Took a total of: 53795600ns (an average of 53,80ns) for new ListenerAdapter (mine)
Took a total of: 183665700ns (an average of 183,67ns) for JDA's new ListenerAdapter
```
Which puts it at a much more respectable ~3x worse. Worth mentioning that the only reason it's 3x worse compared to about the same with my implementation is that you only cache the methods and not the entire path from the start of the event (skipping the ClassWalker) but I think this is completely fine. 👍

PR #1575, opened 2021-04-02
First pass on stage channels

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/ChannelType.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/ChannelType.java
@@ -48,6 +48,10 @@ public enum ChannelType
      */
     STORE(6, 0, true),
     /**
+     * A {@link StageChannel StageChannel}, Guild-Only.
+     */
+    STAGE(13, 1, true),
+    /**
      * Unknown Discord channel type. Should never happen and would only possibly happen if Discord implemented a new
      * channel type and JDA had yet to implement support for it.
      */
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2253,6 +2253,10 @@ public interface Guild extends ISnowflake
     @Nonnull
     AudioManager getAudioManager();
 
+    @Nonnull
+    @CheckReturnValue
+    RestAction<Void> requestToSpeak();
+
     /**
      * Returns the {@link net.dv8tion.jda.api.JDA JDA} instance of this Guild
      *
--- a/src/main/java/net/dv8tion/jda/api/entities/StageChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/StageChannel.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2015 Austin Keener, Michael Ritter, Florian Spieß, and the JDA contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.api.entities;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Represents a Stage Channel, also known as Radio Channel.
+ *
+ * <p>This is a more advanced version of a {@link VoiceChannel}
+ * that can be used to host events with speakers and listeners.
+ */
+public interface StageChannel extends VoiceChannel
+{
+    /**
+     * The topic of a stage channel.
+     *
+     * @return The topic
+     */
+    @Nonnull
+    String getTopic();
+}
--- a/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
@@ -284,6 +284,7 @@ public class EntityBuilder
         case TEXT:
             createTextChannel(guildObj, channelData, guildObj.getIdLong());
             break;
+        case STAGE:
         case VOICE:
             createVoiceChannel(guildObj, channelData, guildObj.getIdLong());
             break;
@@ -947,7 +948,10 @@ public class EntityBuilder
                 UnlockHook vlock = guildVoiceView.writeLock();
                 UnlockHook jlock = voiceView.writeLock())
             {
-                channel = new VoiceChannelImpl(id, guild);
+                if (json.getInt("type") == ChannelType.STAGE.getId())
+                    channel = new StageChannelImpl(id, guild);
+                else
+                    channel = new VoiceChannelImpl(id, guild);
                 guildVoiceView.getMap().put(id, channel);
                 playbackCache = voiceView.getMap().put(id, channel) == null;
             }
--- a/src/main/java/net/dv8tion/jda/internal/requests/Route.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/Route.java
@@ -115,6 +115,7 @@ public class Route
         public static final Route GET_GUILD_EMOTES =   new Route(GET,    "guilds/{guild_id}/emojis");
         public static final Route GET_AUDIT_LOGS =     new Route(GET,    "guilds/{guild_id}/audit-logs");
         public static final Route GET_VOICE_REGIONS =  new Route(GET,    "guilds/{guild_id}/regions");
+        public static final Route REQUEST_TO_SPEAK =   new Route(PATCH,  "/guilds/{guild_id}/voice-states/@me");
 
         public static final Route GET_INTEGRATIONS =   new Route(GET,    "guilds/{guild_id}/integrations");
         public static final Route CREATE_INTEGRATION = new Route(POST,   "guilds/{guild_id}/integrations");
--- a/src/main/java/net/dv8tion/jda/api/Permission.java
+++ b/src/main/java/net/dv8tion/jda/api/Permission.java
@@ -71,6 +71,8 @@ public enum Permission
     MANAGE_WEBHOOKS(   29, true, true, "Manage Webhooks"),
     MANAGE_EMOTES(     30, true, false, "Manage Emojis"),
 
+    REQUEST_TO_SPEAK(  32, true, true, "Request to Speak"),
+
     UNKNOWN(-1, false, false, "Unknown");
 
     /**
@@ -109,7 +111,7 @@ public enum Permission
      */
     public static final long ALL_VOICE_PERMISSIONS
             = Permission.getRaw(VOICE_STREAM, VOICE_CONNECT, VOICE_SPEAK, VOICE_MUTE_OTHERS,
-                                VOICE_DEAF_OTHERS, VOICE_MOVE_OTHERS, VOICE_USE_VAD);
+                                VOICE_DEAF_OTHERS, VOICE_MOVE_OTHERS, VOICE_USE_VAD, REQUEST_TO_SPEAK);
 
     private final int offset;
     private final long raw;
--- a/src/main/java/net/dv8tion/jda/api/events/guild/voice/GuildVoiceUpdateEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/guild/voice/GuildVoiceUpdateEvent.java
@@ -16,10 +16,12 @@
 
 package net.dv8tion.jda.api.events.guild.voice;
 
+import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.entities.Member;
 import net.dv8tion.jda.api.entities.VoiceChannel;
 import net.dv8tion.jda.api.events.UpdateEvent;
 
+import javax.annotation.Nonnull;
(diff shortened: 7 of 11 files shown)


src/main/java/net/dv8tion/jda/api/entities/ChannelType.java
```diff
@@ -47,6 +47,10 @@
      * A {@link net.dv8tion.jda.api.entities.StoreChannel StoreChannel}, Guild-Only.
      */
     STORE(6, 0, true),
+    /**
+     * A {@link StageChannel StageChannel}, Guild-Only.
+     */
+    STAGE(13, 1, true),
```
MinnDevelopment, line 53: Reminder: we have to update all usages of channel type in the library for this

MinnDevelopment:
Freezing this for now due to discord/discord-api-docs#2751

> Please note that this feature is still under active development (even though it is launched), and thus there is higher than normal risk of things possibly changing.

src/main/java/net/dv8tion/jda/api/entities/Category.java
```diff
@@ -159,6 +159,43 @@
     @CheckReturnValue
     ChannelAction<VoiceChannel> createVoiceChannel(@Nonnull String name);
 
+    /**
+     * Creates a new {@link net.dv8tion.jda.api.entities.StageChannel VoiceChannel} with this Category as parent.
```
caneleex, line 163:
```suggestion
     * Creates a new {@link net.dv8tion.jda.api.entities.StageChannel StageChannel} with this Category as parent.
```

src/main/java/net/dv8tion/jda/internal/requests/Route.java
```diff
@@ -115,6 +115,7 @@
         public static final Route GET_GUILD_EMOTES =   new Route(GET,    "guilds/{guild_id}/emojis");
         public static final Route GET_AUDIT_LOGS =     new Route(GET,    "guilds/{guild_id}/audit-logs");
         public static final Route GET_VOICE_REGIONS =  new Route(GET,    "guilds/{guild_id}/regions");
+        public static final Route REQUEST_TO_SPEAK =   new Route(PATCH,  "/guilds/{guild_id}/voice-states/@me");
```
caneleex, line 118:
```suggestion
        public static final Route REQUEST_TO_SPEAK =   new Route(PATCH,  "guilds/{guild_id}/voice-states/@me");
```

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -4609,6 +4639,70 @@ default Emote getEmoteById(long id)
     @CheckReturnValue
     ChannelAction<VoiceChannel> createVoiceChannel(@Nonnull String name, @Nullable Category parent);
 
+    /**
+     * Creates a new {@link net.dv8tion.jda.api.entities.StageChannel VoiceChannel} in this Guild.
```
caneleex, line 4643:
```suggestion
     * Creates a new {@link net.dv8tion.jda.api.entities.StageChannel StageChannel} in this Guild.
```

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -4609,6 +4639,70 @@ default Emote getEmoteById(long id)
...
+     */
+    @Nonnull
+    @CheckReturnValue
+    default ChannelAction<StageChannel> createStageChannel(@Nonnull String name)
+    {
+        return createStageChannel(name, null);
+    }
+
+    /**
+     * Creates a new {@link net.dv8tion.jda.api.entities.StageChannel VoiceChannel} in this Guild.
```
caneleex, line 4675:
```suggestion
     * Creates a new {@link net.dv8tion.jda.api.entities.StageChannel StageChannel} in this Guild.
```

src/main/java/net/dv8tion/jda/api/entities/Category.java
```diff
@@ -159,6 +159,43 @@
     @CheckReturnValue
     ChannelAction<VoiceChannel> createVoiceChannel(@Nonnull String name);
 
+    /**
+     * Creates a new {@link net.dv8tion.jda.api.entities.StageChannel StageChannel} with this Category as parent.
+     * For this to be successful, the logged in account has to have the
+     * {@link net.dv8tion.jda.api.Permission#MANAGE_CHANNEL MANAGE_CHANNEL} Permission in the {@link net.dv8tion.jda.api.entities.Guild Guild}.
```
caneleex, line 165: is this really true considering you can also allow this permission for permission holders for categories?

MinnDevelopment:
The API for stage channels is kinda weird. I did some digging and found out a few issues that will be annoying to deal with:

- The request to speak endpoint seems to be REST-only (so you can't join and immediately speak)
- The endpoint also requires you to specify which channel you are currently in (which requires caching, not a problem for us)
- They expect an ISO timestamp for the request to speak (i can't understand why but ok)
- You can suppress yourself? I have no idea why this is a thing
- Apparently, bots can also approve requests to speak, this is limited through `MUTE_MEMBERS` permission
- The request to speak timestamp is also part of voice states, so we can derive an event there

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -947,7 +954,10 @@ public VoiceChannel createVoiceChannel(GuildImpl guild, DataObject json, long gu
                 UnlockHook vlock = guildVoiceView.writeLock();
                 UnlockHook jlock = voiceView.writeLock())
             {
-                channel = new VoiceChannelImpl(id, guild);
+                if (json.getInt("type") == ChannelType.STAGE.getId())
+                    channel = new StageChannelImpl(id, guild);
```
MinnDevelopment, line 958: We need to call setTopic here

TheChilliPL:
Can't see any methods to invite someone to speak, I think that's possible in stage channels (works like speaking requests but backward: the user has to accept the invite from a moderator to speak)

jzvi12:
Any updates?

> Can't see any methods to invite someone to speak, I think that's possible in stage channels (works like speaking requests but backward: the user has to accept the invite from a moderator to speak)

The bot I use (which uses this JDA lib) can have a moderator role and bypass any invites??

DV8FromTheWorld:
It will be ready when it is ready. Stage channels are still in flux behind the scenes in the Discord API.

MinnDevelopment:
Should we move the speaker methods such as `approveSpeaker` and `inviteSpeaker` into the `Member` interface instead?

MinnDevelopment:
**Feature TODO**

- [ ] Stage Instance events
- [ ] Stage Instance moderation (set topic etc)
- [ ] Voice states for lurkers

**Overall TODO**

- [ ] Documentation
- [ ] Checks
- [ ] Testing

src/main/java/net/dv8tion/jda/internal/requests/Route.java
```diff
@@ -225,6 +226,14 @@
         public static final Route STOP_CALL =        new Route(POST,   "channels/{channel_id}/call/stop_ringing"); // aka deny or end call
     }
 
+    public static class StageInstances
+    {
+        public static final Route GET_INSTANCE =    new Route(GET,    "stage-instances/{channel_id}");
+        public static final Route DELETE_INSTANCE = new Route(DELETE, "stage-instances/{channel_id}");
+        public static final Route UPDATE_INSTANCE =   new Route(PATCH,  "stage-instances/{channel_id}");
+        public static final Route CREATE_INSTANCE = new Route(POST,   "stage-instances");
```
DV8FromTheWorld, line 234:
```suggestion
        public static final Route GET_INSTANCE =    new Route(GET,    "stage-instances/{channel_id}");
        public static final Route DELETE_INSTANCE = new Route(DELETE, "stage-instances/{channel_id}");
        public static final Route UPDATE_INSTANCE = new Route(PATCH,  "stage-instances/{channel_id}");
        public static final Route CREATE_INSTANCE = new Route(POST,   "stage-instances");
```

src/main/java/net/dv8tion/jda/internal/requests/restaction/ChannelActionImpl.java
```diff
@@ -328,14 +328,15 @@ protected RequestBody finalizeData()
...
             case TEXT:
-                if (topic != null && !topic.isEmpty())
-                    object.put("topic", topic);
                 if (nsfw != null)
                     object.put("nsfw", nsfw);
                 if (slowmode != null)
                     object.put("rate_limit_per_user", slowmode);
                 if (news != null)
                     object.put("type", news ? 5 : 0);
+            case STAGE:
```
DV8FromTheWorld, line 337: This should let you set the bitrate for stage too

src/main/java/net/dv8tion/jda/internal/requests/restaction/StageInstanceActionImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        return this;
+    }
+
+    @Override
+    protected RequestBody finalizeData()
+    {
+        DataObject body = DataObject.empty();
+        body.put("channel_id", channel.getId());
+        body.put("topic", topic);
+        body.put("privacy_level", level.getKey());
```
DV8FromTheWorld, line 94: should all of these always be set? Don't we need to have checks to see if they've been actually set?

src/main/java/net/dv8tion/jda/internal/handle/VoiceStateUpdateHandler.java
```diff
@@ -60,6 +70,14 @@ private void handleGuildVoiceState(DataObject content)
...
         boolean suppressed = content.getBoolean("suppress");
         boolean stream = content.getBoolean("self_stream");
+        String requestToSpeak = content.getString("request_to_speak_timestamp", null);
+        OffsetDateTime requestToSpeakTime = null;
+        long requestToSpeakTimestamp = 0L;
+        if (requestToSpeak != null)
+        {
+            requestToSpeakTime = OffsetDateTime.parse(requestToSpeak);
+            requestToSpeakTimestamp = requestToSpeakTime.toInstant().toEpochMilli();
+        }
```
DV8FromTheWorld, line 80: We should have a Helpers method that encompasses this functionality.
We already have a `long -> OffsetDateTime`. We should have one that does `Timestamp(String) -> long`

src/main/java/net/dv8tion/jda/internal/entities/StageInstanceImpl.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Nonnull
+    @Override
+    public RestAction<Void> requestToSpeak()
+    {
+        Guild guild = getGuild();
+        Route.CompiledRoute route = Route.Guilds.UPDATE_VOICE_STATE.compile(guild.getId(), "@me");
+        DataObject body = DataObject.empty().put("channel_id", channel.getId());
+        // Stage moderators can bypass the request queue by just unsuppressing
+        if (guild.getSelfMember().hasPermission(getChannel(), Permission.VOICE_MUTE_OTHERS))
```
DV8FromTheWorld, line 111: Should this be `this.getChannel().isModerator(guild.getSelfMember())` ?

src/main/java/net/dv8tion/jda/api/entities/StageChannel.java
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * @return The {@link StageInstance} or {@code null} if this stage is not live
+     */
+    @Nullable
+    StageInstance getStageInstance();
+
+    /**
+     * Create a new {@link StageInstance} for this stage channel.
+     *
+     * <p>If there already is an active {@link StageInstance} for this channel, this will fail with {@link net.dv8tion.jda.api.requests.ErrorResponse#STAGE_ALREADY_OPEN ErrorResponse.STAGE_ALREADY_OPEN}.
```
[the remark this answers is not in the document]
MinnDevelopment: I wanted to avoid doing pre-checks to avoid race-conditions when recreating a stage instance.

src/main/java/net/dv8tion/jda/internal/handle/VoiceStateUpdateHandler.java
```diff
@@ -44,6 +46,14 @@ protected Long handleInternally(DataObject content)
             return null; //unhandled for calls
         if (getJDA().getGuildSetupController().isLocked(guildId))
             return guildId;
+
+        // TODO: Handle these voice states properly
+        if (content.isNull("member"))
```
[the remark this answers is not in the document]
MinnDevelopment: Discord bug with no explanation after i complained 10 times already

src/main/java/net/dv8tion/jda/internal/requests/restaction/StageInstanceActionImpl.java
MinnDevelopment: The only one thats optional is `privacy_level` which is defaulted to `GUILD_ONLY` regardless.

src/main/java/net/dv8tion/jda/internal/entities/StageInstanceImpl.java
MinnDevelopment: No. According to the [documentation](https://discord.com/developers/docs/resources/guild#modify-current-user-voice-state):

> You must have the MUTE_MEMBERS permission to unsuppress yourself. You can always suppress yourself.

src/main/java/net/dv8tion/jda/internal/handle/VoiceStateUpdateHandler.java
MinnDevelopment: This is not possible here since we want to extract both the OffsetDateTime and a long. Java doesn't support tuple returns so this will have to do.

PR #1576, opened 2021-04-03
Add Message.Attachment#getContentType

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

See https://github.com/discord/discord-api-docs/pull/2762

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -2200,18 +2200,20 @@ public interface Message extends ISnowflake, Formattable
         private final String url;
         private final String proxyUrl;
         private final String fileName;
+        private final String contentType;
         private final int size;
         private final int height;
         private final int width;
 
         private final JDAImpl jda;
 
-        public Attachment(long id, String url, String proxyUrl, String fileName, int size, int height, int width, JDAImpl jda)
+        public Attachment(long id, String url, String proxyUrl, String fileName, String contentType, int size, int height, int width, JDAImpl jda)
         {
             this.id = id;
             this.url = url;
             this.proxyUrl = proxyUrl;
             this.fileName = fileName;
+            this.contentType = contentType;
             this.size = size;
             this.height = height;
             this.width = width;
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1283,8 +1283,9 @@ public MessageReaction createMessageReaction(MessageChannel chan, long id, DataO
         final String url = jsonObject.getString("url");
         final String proxyUrl = jsonObject.getString("proxy_url");
         final String filename = jsonObject.getString("filename");
+        final String contentType = jsonObject.getString("content-type", null);
```
MinnDevelopment, line 1286:
```suggestion
        final String contentType = jsonObject.getString("content_type", null);
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2282,6 +2284,18 @@ public String getFileExtension()
             return index == 0 || index == fileName.length() ? null : fileName.substring(index);
         }
 
+        /**
+         * The Content-Type of this file.
+         * <br>This is the  <a href="https://en.wikipedia.org/wiki/Media_type" target="_blank">Media type</a> of the file that would be used in an HTTP request or similar.
```
caneleex, line 2289:
```suggestion
         * <br>This is the <a href="https://en.wikipedia.org/wiki/Media_type" target="_blank">Media type</a> of the file that would be used in an HTTP request or similar.
```

Andre601:
I assume this could simplify the isImage check (or whatever it wss called).
At least for when the content type is there.

src/main/java/net/dv8tion/jda/api/entities/Message.java
DV8FromTheWorld: @MinnDevelopment This doesn't appear like it was actually incorporated. Why was it marked `resolved`?

MinnDevelopment: Because it changes nothing of value.

Issue #1556, opened 2021-03-23
32 character limit for names should probably be 32 codepoints


ToxicMushroom:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

Im able to configure my name on dicord like this: `𝐸𝐿𝐼𝒮𝐸𝐸𝐿𝐸𝐿𝐼𝒮𝐸𝐸𝐿𝐼𝒮𝐸𝐸𝐿𝐼𝒮𝐸𝐸𝐿𝐼𝒮𝐸𝐸𝐿𝐼𝒮𝐸`
Which looks like 32 characters but with utf-8 this is actually 64 chars in length.
It still is 32 codepoints but jda checks the character count in some functions instead.
`JDA#getUserByTag(String username, String discriminator)` for example

### Expected Behavior
JDA should not throw an error when trying to supply usernames that fit into discord.
I doubt discord will change this behaviour so we should support it perhaps.

You could count codepoints on the string with `String#codePoints().count()`
But I'm unsure if this would allow inputs that would be too big for discord.

### Code Example or Reproduction Steps
```kotlin
shardManager.getUserByTag("𝐸𝐿𝐼𝒮𝐸𝐸𝐿𝐸𝐿𝐼𝒮𝐸𝐸𝐿𝐼𝒮𝐸𝐸𝐿𝐼𝒮𝐸𝐸𝐿𝐼𝒮𝐸𝐸𝐿𝐼𝒮𝐸#5602")
```

### Code for JDABuilder or DefaultShardManagerBuilder Used
N/A

### Exception or Error
```
java.lang.IllegalArgumentException: Username must be between 2 and 32 characters in length!
	at net.dv8tion.jda.internal.utils.Checks.check(Checks.java:43)
	at net.dv8tion.jda.api.JDA.getUserByTag(JDA.java:666)
	at net.dv8tion.jda.api.JDA.getUserByTag(JDA.java:635)
	at net.dv8tion.jda.api.sharding.ShardManager.lambda$getUserByTag$12(ShardManager.java:612)
[...]

PR #1579, opened 2021-04-07
Changed getUserByTag(String username, String discriminator) to use codepoints when checking username length

miiyu:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1556

## Description

Changes the length check to use `String#codePointCount(int beginIndex, int endIndex)` to give compatibility for usernames to include surrogate pairs.


duncte123 (commented):
Since you're calling the code point count twice I personally would put it in a variable

miiyu:
I was considering doing that but it didn't seem consistent with how it's done in other places to me so I stuck with simply replacing the `username.length()` calls.

DV8FromTheWorld (approved):

PR #1585, opened 2021-04-09
Deprecate UserFlag#SYSTEM and add User#isSystem

RedDaedalus:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Deprecates `UserFlag.SYSTEM` as it is an internal flag now and won't show up on any user (see [discord-api-docs issue 2494](https://github.com/discord/discord-api-docs/issues/2494#issuecomment-816970401)), and adds `User#isSystem` as a replacement.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -16,6 +16,7 @@
 package net.dv8tion.jda.api.entities;
 
 
+import net.dv8tion.jda.annotations.DeprecatedSince;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.requests.RestAction;
 import net.dv8tion.jda.api.utils.MiscUtil;
@@ -347,6 +348,8 @@ public interface User extends IMentionable, IFakeable
 
         EARLY_SUPPORTER(    9, "Early Supporter"),
         TEAM_USER(         10, "Team User"),
+        @Deprecated
+        @DeprecatedSince("4.2.0")
         SYSTEM(            12, "System User"),
         BUG_HUNTER_LEVEL_2(14, "Bug Hunter Level 2"),
         VERIFIED_BOT(      16, "Verified Bot"),
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -17,6 +17,7 @@ package net.dv8tion.jda.api.entities;
 
 
 import net.dv8tion.jda.annotations.DeprecatedSince;
+import net.dv8tion.jda.annotations.ForRemoval;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.requests.RestAction;
 import net.dv8tion.jda.api.utils.MiscUtil;
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -360,7 +360,8 @@ public interface User extends IMentionable, IFakeable
         EARLY_SUPPORTER(    9, "Early Supporter"),
         TEAM_USER(         10, "Team User"),
         @Deprecated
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -118,7 +118,8 @@ public SelfUser createSelfUser(DataObject self)
                 .setName(self.getString("username"))
                 .setDiscriminator(self.getString("discriminator"))
                 .setAvatarId(self.getString("avatar", null))
-                .setBot(self.getBoolean("bot"));
+                .setBot(self.getBoolean("bot"))
+                .setSystem(self.getBoolean("system", false));
```
MinnDevelopment, line 122: Isn't this always false?

RedDaedalus: It's marked as optional on the docs so I figured it was best to add a fallback.

RedDaedalus: Just checked, it's undefined for normal users: `{"id": "86699011792191488", "username": "Minn", "avatar": "f3bc77b1e276fba2e1485a915141564b", "discriminator": "6688", "public_flags": 512}`, although so is bot, so I guess `getBoolean` just decodes undefined as false.

MinnDevelopment: I was more referring to self user never being system=true

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -347,6 +359,8 @@ default String getEffectiveAvatarUrl()
 
         EARLY_SUPPORTER(    9, "Early Supporter"),
         TEAM_USER(         10, "Team User"),
+        @Deprecated
+        @DeprecatedSince("4.2.0")
```
MinnDevelopment, line 363:
```suggestion
        @ForRemoval
        @DeprecatedSince("4.3.0")
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -298,6 +299,17 @@ default String getEffectiveAvatarUrl()
...
     boolean isBot();
 
+    /**
+     * Returns whether or not the given user is a System account, which includes the urgent message account
+     * and the community updates bot.
+     *
+     * @throws UnsupportedOperationException
+     *         If this User was created with {@link #fromId(long)}
+     *
+     * @return If the User's account is marked as System
```
MinnDevelopment, line 309:
```suggestion
     * @return Whether the User's account is marked as System
```

MinnDevelopment (changes requested):

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -16,6 +16,7 @@
 package net.dv8tion.jda.api.entities;
 
 
+import net.dv8tion.jda.annotations.DeprecatedSince;
```
MinnDevelopment, line 19:
```suggestion
import net.dv8tion.jda.annotations.DeprecatedSince;
import net.dv8tion.jda.annotations.ForRemoval;
import net.dv8tion.jda.annotations.ReplaceWith;
```

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -347,6 +359,9 @@ default String getEffectiveAvatarUrl()
 
         EARLY_SUPPORTER(    9, "Early Supporter"),
         TEAM_USER(         10, "Team User"),
+        @Deprecated
+        @ForRemoval
+        @DeprecatedSince("4.3.0")
```
MinnDevelopment, line 364:
```suggestion
        @Deprecated
        @ForRemoval
        @ReplaceWith("User.isSystem()")
        @DeprecatedSince("4.3.0")
```

PR #1594, opened 2021-04-14
Update Invite to new vanity one

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: README.md change 

Closes Issue: NaN

## Description

Since the JDA discord is now verified and has a vanity URL would it be the best to update the one on the readme, no?


MinnDevelopment:
We want to keep the current invite.

DV8FromTheWorld:
We don't want to upgrade the invite link because it is unclear if it will break at some point in the future.

PR #1604, opened 2021-04-20
Fix typo in ShardManager

Nhalrath:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR rewrites some parts of the docs which will make it a bit easier for the readers to read.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
+++ b/src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
@@ -209,7 +209,7 @@ public interface ShardManager
 
     /**
      * The average time in milliseconds between all shards that discord took to respond to our last heartbeat.
-     * This roughly represents the WebSocket ping of this session. If there is no shard running, this will return {@code -1}.
+     * This roughly represents the WebSocket ping of this session. If there are no shards running, this will return {@code -1}.
      *
      * <p><b>{@link net.dv8tion.jda.api.requests.RestAction RestAction} request times do not
      * correlate to this value!</b>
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
```diff
@@ -209,7 +209,7 @@ default int getShardsTotal()
 
     /**
      * The average time in milliseconds between all shards that discord took to respond to our last heartbeat.
-     * This roughly represents the WebSocket ping of this session. If there is no shard running this wil return {@code -1}.
+     * This roughly represents the WebSocket ping of this session. If there is no shard running this will return {@code -1}.
```
Andre601, line 212: Something I now wonder when reading this is, what counts as a "running shard"? Like does it have to be connected, or will a disconnected shard also still return the last fetched ping? And if so, does this not make this a condition where -1 is barely returned (At most on startup I suppose?)
Would this perhaps even be a relict from before a JDA instance always having a shard active (Just a single shard with ID 0 if JDABuilder is used)?

Nhalrath: I never encountered having a -1 ping before, so I'm not sure. This might never return -1 at all as there's no shard running to take care of the request thingie(?)

src/main/java/net/dv8tion/jda/api/sharding/ShardManager.java
```diff
@@ -209,7 +209,7 @@ default int getShardsTotal()
 
     /**
      * The average time in milliseconds between all shards that discord took to respond to our last heartbeat.
-     * This roughly represents the WebSocket ping of this session. If there is no shard running this wil return {@code -1}.
+     * This roughly represents the WebSocket ping of this session. If there is no shard running, this will return {@code -1}.
```
DV8FromTheWorld, line 212:
```suggestion
     * This roughly represents the WebSocket ping of this session. If there are no shards running, this will return {@code -1}.
```

DV8FromTheWorld (approved):

PR #1605, opened 2021-04-20
Add getMentionedX for all guild channels

Simon-official:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds `getMentionedX()` and `getMentionedXBag()` for all GuildChannels, which allows to retrieve mentions for a specific Channel type in a message. I'm not sure about the usefulness of the `getMentionedXBag()` methods as these could be easily replicated without the need of an extra method. Also, I'm not sure which characters to use [here](https://github.com/Simon-official/JDA/blob/feature/mentioned-channels/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java#L723).

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
@@ -78,7 +78,10 @@ public class ReceivedMessage extends AbstractMessage
     protected List<Member> memberMentions = null;
     protected List<Emote> emoteMentions = null;
     protected List<Role> roleMentions = null;
-    protected List<TextChannel> channelMentions = null;
+    protected List<TextChannel> textChannelMentions = null;
+    protected List<VoiceChannel> voiceChannelMentions = null;
+    protected List<StoreChannel> storeChannelMentions = null;
+    protected List<Category> categoryMentions = null;
     protected List<String> invites = null;
 
     public ReceivedMessage(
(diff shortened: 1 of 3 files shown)


RedDaedalus:
You can't mention categories, so the `getMentionedCategories` method could probably be removed.

Andre601:
> You can't mention categories, so the `getMentionedCategories` method could probably be removed.

You kinda can but it's not directly supported by the Discord client.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -612,10 +716,22 @@ public String getContentDisplay()
...
                 tmp = tmp.replace(emote.getAsMention(), ":" + emote.getName() + ":");
             }
-            for (TextChannel mentionedChannel : getMentionedChannels())
+            for (TextChannel mentionedChannel : getMentionedTextChannels())
+            {
+                tmp = tmp.replace(mentionedChannel.getAsMention(), '#' + mentionedChannel.getName());
+            }
+            for (VoiceChannel mentionedChannel : getMentionedVoiceChannels())
+            {
+                tmp = tmp.replace(mentionedChannel.getAsMention(), mentionedChannel.getName());
```
granny, line 725: > Also, I'm not sure which characters to use [here](https://github.com/Simon-official/JDA/blob/feature/mentioned-channels/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java#L723).

```suggestion
                tmp = tmp.replace(mentionedChannel.getAsMention(), '#!' + mentionedChannel.getName());
```
Voice channels can be mentioned with the prefix `#!`

Simon-official: This method should return how the client displays something, not how to mention it. See the [docs](https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/api/entities/Message.html#getContentDisplay())
The client doesn't use ``#!`` but instead the following: 
Screenshot 2021-07-01 113735

DV8FromTheWorld:
I'm placing this on hold. One of the things we're discussing in the JDA 5 rework is something like `Message#getMentions() : MessageMentions` which would move all mention methods into a separate object.

MinnDevelopment:
Superseded by #1835

PR #1616, opened 2021-05-04
Make String.format usage locale safe

Scarsz:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Replaces internal String#format calls in JDA to use a helper method that has the locale set to Locale.ROOT. This fixes issues with JDA not being locale-safe, such as replacing `8` in the API version to the Arabic numeral 8, when the default locale is Arabic.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/Requester.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/Requester.java
@@ -22,6 +22,7 @@ import net.dv8tion.jda.api.requests.Request;
 import net.dv8tion.jda.api.requests.Response;
 import net.dv8tion.jda.internal.JDAImpl;
 import net.dv8tion.jda.internal.requests.ratelimit.BotRateLimiter;
+import net.dv8tion.jda.internal.utils.Helpers;
 import net.dv8tion.jda.internal.utils.JDALogger;
 import net.dv8tion.jda.internal.utils.config.AuthorizationConfig;
 import okhttp3.Call;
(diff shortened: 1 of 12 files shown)


src/main/java/net/dv8tion/jda/internal/requests/Requester.java
```diff
@@ -47,7 +49,7 @@
 public class Requester
 {
     public static final Logger LOG = JDALogger.getLog(Requester.class);
-    public static final String DISCORD_API_PREFIX = String.format("https://discord.com/api/v%d/", JDAInfo.DISCORD_REST_VERSION);
+    public static final String DISCORD_API_PREFIX = Helpers.format("https://discord.com/api/v%d/", JDAInfo.DISCORD_REST_VERSION);
```
Andre601, line 52: Since this seems to be the only place where this helper is used, why have this method in the first place and not apply the Locale.ROOT directly here?

Scarsz:
Because this PR is a draft. It’s not finished yet. The method will be used
in more places where the locale might be problematic.

src/main/java/net/dv8tion/jda/internal/requests/Requester.java
Scarsz: Because this PR is a draft. It’s not finished yet. The method will be used in more places where the locale might be problematic.

MinnDevelopment:
Any update on this?

Scarsz:
Was waiting to possibly hear back from the original user who was having this issue. They haven't reported any other problems so far but I'll go ahead and make the rest of the replacements today or tomorrow

Scarsz:
I changed all of the String#format calls that dealt with numbers over to the helper method. I'm not sure what the best solution to the whole locale problem as a whole would be though.

- There could be a test to check for String#format calls using numbers
- Ideally, there would also be a helper method + test for other string methods such as #toUpperCase/toLowerCase (capital `i` in Turkish is not `I`)

Although, truly, I'm not sure the rest of these potential problems matter much. Nobody has reported this in JDA since its inception until now, so it's clearly very low priority. I feel like having more helper methods as mentioned above wouldn't be worth the impact to code readability. Open to thoughts.

MinnDevelopment (approved):

Issue #1625, opened 2021-05-22
Add Guild#getNSFWLevel


MinnDevelopment:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request

See: https://github.com/discord/discord-api-docs/pull/2976

DV8FromTheWorld:
Completed via https://github.com/DV8FromTheWorld/JDA/pull/1793

PR #1626, opened 2021-05-22
Adds nsfw_level

ishwi:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: #1625 

## Description

Adds the enum NSFWLevel to the Guild object and an event for when the level changes. 
Initially added the level also to the GuildManager and to AuditLogKey but upon further clarification the level is set automatically by discord.

Pending to document what each level means.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2357,12 +2357,12 @@ public interface Guild extends ISnowflake
     ExplicitContentLevel getExplicitContentLevel();
 
     /**
-     * Returns the level of multifactor authentication required to execute administrator restricted functions in this guild.
+     * Returns the NSFW Level that this guild is classified with.
      * <br>For a short description of the different values, see {@link net.dv8tion.jda.api.entities.Guild.NSFWLevel NSFW Level}.
      * <p>
      * This value cannot be modified directly.
      *
-     * @return The NSFWLevel required by this Guild.
+     * @return The NSFWLevel of this guild.
      */
     @Nonnull
     NSFWLevel getNSFWLevel();
@@ -5422,8 +5422,13 @@ public interface Guild extends ISnowflake
         }
     }
     /**
-     * Represents the Guild NSFW level.
-      */
+     * Represents the NSFW level for this guild.
+     * <p>
+     * <br><b>DEFAULT</b>   {@literal ->}  Is yet to be classified by discord.
+     * <br><b>EXPLICIT</b>  {@literal ->} Is classified as a NSFW server
+     * <br><b>SAFE</b>  {@literal ->}  Doesn't classify as a NSFW server
+     * <br><b>AGE_RESTRICTED</b>  {@literal ->} Is classified as NSFW and has an age restriction in place
+     */
     enum NSFWLevel
     {
         DEFAULT(0),
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2357,6 +2357,17 @@ public interface Guild extends ISnowflake
     ExplicitContentLevel getExplicitContentLevel();
 
     /**
+     * Returns the level of multifactor authentication required to execute administrator restricted functions in this guild.
+     * <br>For a short description of the different values, see {@link net.dv8tion.jda.api.entities.Guild.NSFWLevel NSFW Level}.
+     * <p>
+     * This value cannot be modified directly.
+     *
+     * @return The NSFWLevel required by this Guild.
+     */
+    @Nonnull
+    NSFWLevel getNSFWLevel();
+
+    /**
      * Checks if the current Verification-level of this guild allows JDA to send messages to it.
      *
      * @return True if Verification-level allows sending of messages, false if not.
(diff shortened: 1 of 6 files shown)


src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -2356,6 +2356,17 @@ default Emote getEmoteById(long id)
     @Nonnull
     ExplicitContentLevel getExplicitContentLevel();
 
+    /**
+     * Returns the level of multifactor authentication required to execute administrator restricted functions in this guild.
```
Andre601, line 2360: How is the NSFW level connected to administrator-specific actions?

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -105,6 +105,7 @@
     private NotificationLevel defaultNotificationLevel = NotificationLevel.UNKNOWN;
     private MFALevel mfaLevel = MFALevel.UNKNOWN;
     private ExplicitContentLevel explicitContentLevel = ExplicitContentLevel.UNKNOWN;
+    private NSFWLevel nsfwLevel = NSFWLevel.UNKNOWN;
```
Andre601, line 108: I would assume `NSFWLevel.DEFAULT` makes more sense here since... well... it's the default a Server has...

But this is something Minn should give his input for I would say...

Xirado: Since the `MFALevel` here is also UNKNOWN, i think this is pretty much expected.

src/main/java/net/dv8tion/jda/api/entities/Guild.java
ishwi: It was copy pasted from the MFALevel. Changed it on last commit.

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -5410,6 +5421,59 @@ public String toString()
             return "GuildBan:" + user + (reason == null ? "" : '(' + reason + ')');
         }
     }
+    /**
+     * Represents the NSFW level for this guild.
+     * <p>
+     * <br><b>DEFAULT</b>   {@literal ->}  Is yet to be classified by discord.
+     * <br><b>EXPLICIT</b>  {@literal ->} Is classified as a NSFW server
+     * <br><b>SAFE</b>  {@literal ->}  Doesn't classify as a NSFW server
+     * <br><b>AGE_RESTRICTED</b>  {@literal ->} Is classified as NSFW and has an age restriction in place
```
MinnDevelopment, line 5430: Move these descriptions to the enum constants instead

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1643,6 +1652,11 @@ public GuildImpl setExplicitContentLevel(ExplicitContentLevel level)
         this.explicitContentLevel = level;
         return this;
     }
+    public GuildImpl setNSFWLevel(NSFWLevel nsfwLevel)
```
MinnDevelopment, line 1655:
```suggestion

    public GuildImpl setNSFWLevel(NSFWLevel nsfwLevel)
```

src/main/java/net/dv8tion/jda/internal/handle/GuildUpdateHandler.java
```diff
@@ -79,6 +79,7 @@ protected Long handleInternally(DataObject content)
         Guild.VerificationLevel verificationLevel = Guild.VerificationLevel.fromKey(content.getInt("verification_level"));
         Guild.NotificationLevel notificationLevel = Guild.NotificationLevel.fromKey(content.getInt("default_message_notifications"));
         Guild.MFALevel mfaLevel = Guild.MFALevel.fromKey(content.getInt("mfa_level"));
+        Guild.NSFWLevel nsfwLevel = Guild.NSFWLevel.fromKey(content.getInt("nsfw_level"));
```
MinnDevelopment, line 82:
```suggestion
        Guild.NSFWLevel nsfwLevel = Guild.NSFWLevel.fromKey(content.getInt("nsfw_level", -1));
```

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
MinnDevelopment: UNKNOWN is better. We don't want to incorrectly classify a guild as safe or otherwise.

src/main/java/net/dv8tion/jda/api/entities/Guild.java
ishwi: MFALevel has the descripctions like that. Should I also move them to the constants?

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -2356,6 +2356,17 @@ default Emote getEmoteById(long id)
     @Nonnull
     ExplicitContentLevel getExplicitContentLevel();
 
+    /**
+     * Returns the NSFW Level that this guild is classified with.
+     * <br>For a short description of the different values, see {@link net.dv8tion.jda.api.entities.Guild.NSFWLevel NSFW Level}.
+     * <p>
+     * This value cannot be modified directly.
```
DV8FromTheWorld, line 2363: When we have more information we should indicate _how_ this value gets changed if developers can't change it themselves given the fact that channel specific `nsfw` is modifyable.

MinnDevelopment (changes requested):
This seems to be missing handling inside `EntityBuilder#createGuild`

ishwi:
> This seems to be missing handling inside `EntityBuilder#createGuild`

Added the handling and decided to use 0 as the value in case the field is not found just like with the rest of the enums there.

DV8FromTheWorld:
Bump. Please comment on or resolve existing comments.
Afterwards this can probably be merged.

DV8FromTheWorld:
Closing this for lack of activity. Feel free to open another PR if still interested.

PR #1628, opened 2021-05-22
Initial pass on discord together

anweisen:
## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: _____

Closes Issue: NaN

## Description

This pr provides us with the functionality to create invites for applications and streams.
This allows us to use the discord together features.

More information and documentation will follow soon.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Invite.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Invite.java
@@ -143,6 +143,14 @@ public interface Invite
     Invite.InviteType getType();
 
     /**
+     * The target type of this invite.
+     *
+     * @return The invite's type
+     */
+    @Nonnull
+    Invite.TargetType getTargetType();
+
+    /**
      * An {@link net.dv8tion.jda.api.entities.Invite.Channel Invite.Channel} object
      * containing information about this invite's origin channel.
      *
@@ -502,5 +510,40 @@ public interface Invite
         GUILD,
         GROUP,
         UNKNOWN
+    }
+
+    enum TargetType
+    {
+        STREAM(1),
+
+        EMBEDDED_APPLICATION(2),
+
+        /**
+         * Unknown invite target type.
+         * Either this invite does not have a target type or Discord implemented a new invite target type and JDA has not yet implemented support for it.
+         */
+        UNKNOWN(-1);
+
+        private final int id;
+
+        TargetType(int id)
+        {
+            this.id = id;
+        }
+
+        public int getId()
+        {
+            return id;
+        }
+
+        public static TargetType fromId(int id) {
+            for (TargetType type : values())
+            {
+                if (type.id == id)
+                    return type;
+            }
+            return UNKNOWN;
+        }
+
     }
 }
--- a/src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
+++ b/src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
@@ -18,6 +18,8 @@ package net.dv8tion.jda.api.requests.restaction;
 
 import net.dv8tion.jda.api.entities.GuildChannel;
 import net.dv8tion.jda.api.entities.Invite;
+import net.dv8tion.jda.api.entities.User;
+import net.dv8tion.jda.api.utils.MiscUtil;
 
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
@@ -118,4 +120,99 @@ public interface InviteAction extends AuditableRestAction<Invite>
     @Nonnull
     @CheckReturnValue
     InviteAction setUnique(@Nullable final Boolean unique);
+
+    /**
+     * Sets the type of target for this voice channel invite.
+     *
+     * @param type
+     *        The target type for this voice channel invite.
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If this channel is not a voice channel.
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    InviteAction setTargetType(@Nullable final Invite.TargetType type);
+
+    /**
+     * Sets the id of the targeted application.
+     * This will automatically {@link #setTargetType(Invite.TargetType) set the target type} to {@link Invite.TargetType#EMBEDDED_APPLICATION}.
+     *
+     * @param applicationId
+     *        The id of the embedded application or {@code null} to use none, requires type {@link Invite.TargetType#EMBEDDED_APPLICATION}
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    InviteAction setTargetApplication(@Nullable final Long applicationId);
+
+    /**
+     * Sets the id of the targeted application.
+     * This will automatically {@link #setTargetType(Invite.TargetType) set the target type} to {@link Invite.TargetType#EMBEDDED_APPLICATION}.
+     *
+     * @param applicationId
+     *        The id of the embedded application or {@code null} to use none, requires type {@link Invite.TargetType#EMBEDDED_APPLICATION}
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    default InviteAction setTargetApplication(@Nullable final String applicationId) {
+        return setTargetApplication(MiscUtil.parseSnowflake(applicationId));
+    }
+
+    /**
+     * Sets the user whose stream to display for this invite.
+     * This will automatically {@link #setTargetType(Invite.TargetType) set the target type} to {@link Invite.TargetType#STREAM}.
+     * The user must be streaming in the same channel.
+     *
+     * @param userId
+     *        The id of the user whose stream to display.
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    InviteAction setTargetUser(@Nullable final Long userId);
+
+    /**
+     * Sets the user whose stream to display for this invite.
+     * This will automatically {@link #setTargetType(Invite.TargetType) set the target type} to {@link Invite.TargetType#STREAM}.
+     * The user must be streaming in the same channel.
+     *
+     * @param userId
+     *        The id of the user whose stream to display.
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided ID is null
+     * @throws java.lang.NumberFormatException
+     *         If the provided ID is not a snowflake
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    default InviteAction setTargetUser(@Nonnull String userId) {
+        return setTargetUser(MiscUtil.parseSnowflake(userId));
+    }
+
+    /**
+     * Sets the user whose stream to display for this invite.
+     * This will automatically {@link #setTargetType(Invite.TargetType) set the target type} to {@link Invite.TargetType#STREAM}.
+     * The user must be streaming in the same channel.
+     *
+     * @param user
+     *        The user whose stream to display.
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    default InviteAction setTargetUser(@Nonnull User user) {
+        return setTargetUser(user.getIdLong());
+    }
+
 }
--- a/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
@@ -1556,6 +1556,7 @@ public class EntityBuilder
 
         final DataObject channelObject = object.getObject("channel");
         final ChannelType channelType = ChannelType.fromId(channelObject.getInt("type"));
+        final Invite.TargetType targetType = Invite.TargetType.fromId(object.getInt("target_type", -1));
 
         final Invite.InviteType type;
         final Invite.Guild guild;
@@ -1646,7 +1647,7 @@ public class EntityBuilder
 
         return new InviteImpl(getJDA(), code, expanded, inviter,
                               maxAge, maxUses, temporary,
-                              timeCreated, uses, channel, guild, group, type);
+                              timeCreated, uses, channel, guild, group, type, targetType);
     }
 
     public ApplicationInfo createApplicationInfo(DataObject object)
--- a/src/main/java/net/dv8tion/jda/internal/handle/InviteCreateHandler.java
+++ b/src/main/java/net/dv8tion/jda/internal/handle/InviteCreateHandler.java
@@ -74,7 +74,9 @@ public class InviteCreateHandler extends SocketHandler
         InviteImpl.ChannelImpl channel = new InviteImpl.ChannelImpl(realChannel);
         InviteImpl.GuildImpl guild = new InviteImpl.GuildImpl(realGuild);
 
-        Invite invite = new InviteImpl(getJDA(), code, expanded, inviter, maxAge, maxUses, temporary, creationTime, 0, channel, guild, null, Invite.InviteType.GUILD);
+        Invite.TargetType targetType = Invite.TargetType.fromId(content.getInt("target_type", -1));
+
+        Invite invite = new InviteImpl(getJDA(), code, expanded, inviter, maxAge, maxUses, temporary, creationTime, 0, channel, guild, null, Invite.InviteType.GUILD, targetType);
         getJDA().handleEvent(
             new GuildInviteCreateEvent(
                 getJDA(), responseNumber,
--- a/src/main/java/net/dv8tion/jda/internal/entities/InviteImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/InviteImpl.java
@@ -54,10 +54,11 @@ public class InviteImpl implements Invite
     private final OffsetDateTime timeCreated;
     private final int uses;
     private final Invite.InviteType type;
+    private final Invite.TargetType targetType;
 
     public InviteImpl(final JDAImpl api, final String code, final boolean expanded, final User inviter,
             final int maxAge, final int maxUses, final boolean temporary, final OffsetDateTime timeCreated,
-            final int uses, final Channel channel, final Guild guild, final Group group, final Invite.InviteType type)
+            final int uses, final Channel channel, final Guild guild, final Group group, final Invite.InviteType type, Invite.TargetType targetType)
     {
         this.api = api;
         this.code = code;
@@ -72,6 +73,7 @@ public class InviteImpl implements Invite
         this.guild = guild;
         this.group = group;
         this.type = type;
+        this.targetType = targetType;
     }
 
     public static RestAction<Invite> resolve(final JDA api, final String code, final boolean withCounts)
@@ -155,6 +157,13 @@ public class InviteImpl implements Invite
     public Invite.InviteType getType()
     {
         return this.type;
+    }
+
+    @Nonnull
+    @Override
+    public TargetType getTargetType()
+    {
+        return this.targetType;
     }
 
     @Override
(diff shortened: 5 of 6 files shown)


Andre601:
Is this an actual planned feature of Discord?
The docs don't show any PRs for it...

KxmischesDomi:
> Is this an actual planned feature of Discord?
> The docs don't show any PRs for it...

It is already implemented in the desktop app.
Look at this: https://www.reddit.com/r/discordapp/comments/jtnso2/social_party_games_on_discord/
There is still no way of creating these invites with the discord app.

anweisen:
> > Is this an actual planned feature of Discord?
> > The docs don't show any PRs for it...

> It is already implemented in the desktop app.
> Look at this: https://www.reddit.com/r/discordapp/comments/jtnso2/social_party_games_on_discord/
> There is still no way of creating these invites with the discord app.

The corresponding properties of [the invite object](https://discord.com/developers/docs/resources/invite) and in the [creation request](https://discord.com/developers/docs/resources/channel#create-channel-invite) are already in the docs

Sanduhr32:
Question: Why do you used boxed longs instead of native once?

anweisen:
> Question: Why do you used boxed longs instead of native once?

I thought it would be better to use them as they can be null and this property is not required
They were also used for other properties which are not required like maxAge and maxUsed

RedDaedalus:
> Is this an actual planned feature of Discord?
> The docs don't show any PRs for it...

This is also enabled on the [official Poker Night server](https://discord.gg/pokernight).

DV8FromTheWorld:
JDA only implements features exposed in the discord public documentation. If this feature gets documented there then we can revisit this.

Setting this PR on freezer until then.

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -492,6 +519,73 @@ default String getUrl()
...
+        @Nullable
+        String getIconUrl();
+
+        /**
+         * If this application has a max participant count.
+         * If this application does not have max participant count {@link #getMaxParticipants() getMaxParticipants()} will return {@code -1}.
+         *
+         * @return If this application has a max participant count set
+         */
+        boolean hasMaxParticipants();
```
MinnDevelopment, line 579: This method is redundant if the getter encodes this with -1

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -503,4 +597,44 @@ default String getUrl()
...
+        {
+            this.id = id;
+        }
+
+        public int getId()
+        {
+            return id;
+        }
+
+        public static TargetType fromId(int id) {
```
MinnDevelopment, line 630: Missing docs and annotation

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -503,4 +597,44 @@ default String getUrl()
...
+        UNKNOWN(-1);
+
+        private final int id;
+
+        TargetType(int id)
+        {
+            this.id = id;
+        }
+
+        public int getId()
```
MinnDevelopment, line 625: Missing docs

src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
```diff
@@ -118,4 +120,100 @@
...
+     *        The target type for this voice channel invite.
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If this channel is not a voice channel.
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    InviteAction setTargetType(@Nullable final Invite.TargetType type);
```
MinnDevelopment, line 137: Does it make sense to just set the type without user/application?

src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
```diff
@@ -118,4 +120,100 @@
...
+     *
+     * @param applicationId
+     *        The id of the embedded application or {@code null} or {@link Invite.TargetType#UNKNOWN} to use none.
+     *        Requires type {@link Invite.TargetType#EMBEDDED_APPLICATION}
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    InviteAction setTargetApplication(@Nullable final Long applicationId);
```
MinnDevelopment, line 151: Making two methods with the same name accept nullable parameters is bad since you run into ambiguity. I'd just use `0` instead of null for this.

src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
```diff
@@ -118,4 +120,100 @@
...
+     *
+     * @param applicationId
+     *        The id of the embedded application or {@code null} to use none, requires type {@link Invite.TargetType#EMBEDDED_APPLICATION}
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    default InviteAction setTargetApplication(@Nullable final String applicationId) {
+        return setTargetApplication(MiscUtil.parseSnowflake(applicationId));
```
MinnDevelopment, line 165: This throws when applicationId is null

src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
```diff
@@ -118,4 +120,100 @@
...
+    InviteAction setTargetApplication(@Nullable final Long applicationId);
+
+    /**
+     * Sets the id of the targeted application.
+     * This will automatically {@link #setTargetType(Invite.TargetType) set the target type} to {@link Invite.TargetType#EMBEDDED_APPLICATION}.
+     *
+     * @param applicationId
+     *        The id of the embedded application or {@code null} to use none, requires type {@link Invite.TargetType#EMBEDDED_APPLICATION}
+     *
+     * @return The current InviteAction for chaining.
```
MinnDevelopment, line 160: Missing throws docs for invalid ids

src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
```diff
@@ -118,4 +120,100 @@
...
+     * The user must be streaming in the same channel.
+     *
+     * @param userId
+     *        The id of the user whose stream to display.
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    InviteAction setTargetUser(@Nullable final Long userId);
```
MinnDevelopment, line 180: Same as above

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1618,6 +1621,31 @@ else if (channelType.isGuild())
             group = null;
         }
 
+        if (targetType == Invite.TargetType.STREAM) {
```
MinnDevelopment, line 1624: Use a switch

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -137,11 +137,19 @@
...
     @Nonnull
     Invite.InviteType getType();
 
+    /**
+     * The target type of this invite.
+     *
+     * @return The invite's target type
+     */
+    @Nonnull
+    Invite.TargetType getTargetType();
```
MinnDevelopment, line 151: What if there is no target?

src/main/java/net/dv8tion/jda/internal/handle/InviteCreateHandler.java
```diff
@@ -74,7 +74,32 @@ protected Long handleInternally(DataObject content)
         InviteImpl.ChannelImpl channel = new InviteImpl.ChannelImpl(realChannel);
         InviteImpl.GuildImpl guild = new InviteImpl.GuildImpl(realGuild);
 
-        Invite invite = new InviteImpl(getJDA(), code, expanded, inviter, maxAge, maxUses, temporary, creationTime, 0, channel, guild, null, Invite.InviteType.GUILD);
+        final Invite.TargetType targetType = Invite.TargetType.fromId(content.getInt("target_type", -1));
+        final User targetUser;
+        final InviteImpl.EmbeddedApplicationImpl application;
+
+        if (targetType == Invite.TargetType.STREAM)
```
MinnDevelopment, line 81: Use a switch

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -492,6 +519,73 @@ default String getUrl()
...
+         *
+         * @return The application's icon id
+         *
+         * @see    #getIconUrl()
+         */
+        @Nullable
+        String getIconId();
+
+        /**
+         * The icon url of this application or {@code null} if thr application has no icon.
```
jan-tennert, line 564:
```suggestion
         * The icon url of this application or {@code null} if the application has no icon.
```

src/main/java/net/dv8tion/jda/api/entities/Invite.java
anweisen: > What if there is no target?

If there is not target it will be TargetType.UNKNOWN

MinnDevelopment: That should be documented.

anweisen: Yeah added that in the last commit

src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
```diff
@@ -121,71 +121,62 @@
...
-     */
-    @Nonnull
-    @CheckReturnValue
-    InviteAction setTargetType(@Nullable final Invite.TargetType type);
-
     /**
      * Sets the id of the targeted application.
-     * This will automatically {@link #setTargetType(Invite.TargetType) set the target type} to {@link Invite.TargetType#EMBEDDED_APPLICATION}.
+     * The invite has to point to a voice channel.
+     * The invite will have the {@link Invite.TargetType#STREAM} target.
```
MinnDevelopment, line 127: Is it correct that this was changed from EMBEDDED_APPLICATION to STREAM?

anweisen: Uhm no, done this by accident, copy pasting is bad

src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
```diff
@@ -118,4 +120,91 @@
...
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided ID is null
+     * @throws java.lang.NumberFormatException
+     *         If the provided ID is not a snowflake
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    default InviteAction setTargetApplication(@Nonnull final String applicationId) {
```
MinnDevelopment, line 155:
```suggestion
    default InviteAction setTargetApplication(@Nonnull final String applicationId)
    {
```

src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
```diff
@@ -118,4 +120,91 @@
...
+     * @throws java.lang.IllegalArgumentException
+     *         If the provided ID is null
+     * @throws java.lang.NumberFormatException
+     *         If the provided ID is not a snowflake
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    default InviteAction setTargetUser(@Nonnull final String userId) {
```
MinnDevelopment, line 190:
```suggestion
    default InviteAction setTargetUser(@Nonnull final String userId)
    {
```

src/main/java/net/dv8tion/jda/api/requests/restaction/InviteAction.java
```diff
@@ -118,4 +120,91 @@
...
+     * The invite will have the {@link Invite.TargetType#STREAM} target.
+     *
+     * @param user
+     *        The user whose stream to target.
+     *
+     * @return The current InviteAction for chaining.
+     */
+    @Nonnull
+    @CheckReturnValue
+    default InviteAction setTargetUser(@Nonnull final User user) {
```
MinnDevelopment, line 206:
```suggestion
    default InviteAction setTargetUser(@Nonnull final User user)
    {
```

MinnDevelopment (changes requested):

MinnDevelopment (approved):

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -503,4 +589,70 @@ default String getUrl()
...
+
+    /**
+     * Enum representing the target type of an invite.
+     *
+     * @see #getTargetType()
+     */
+    enum TargetType
+    {
+        /**
+         * The invite points to an users stream in a voice channel.
```
DV8FromTheWorld, line 601:
```suggestion
         * The invite points to a user's stream in a voice channel.
```

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -503,4 +589,70 @@ default String getUrl()
         GROUP,
         UNKNOWN
     }
+
+    /**
+     * Enum representing the target type of an invite.
```
DV8FromTheWorld, line 594: Either this or `Invite#getTargetType` need to have some elaboration on what a "TargetType" _is_.
The docs for them are essentially "The thing is the thing". Not particularly helpful.

Something like "A TargetType indicates additional action to be taken by the client on accepting the invite, typically connecting external services or launching external applications depending on the specific TargetType." would be more descriptive.

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -503,4 +589,70 @@ default String getUrl()
...
+         *
+         * @see Invite#getTargetApplication()
+         */
+        EMBEDDED_APPLICATION(2),
+
+        /**
+         * Unknown invite target type.
+         * Either this invite does not have a target type or Discord implemented a new invite target type and JDA has not yet implemented support for it.
+         */
+        UNKNOWN(-1);
```
DV8FromTheWorld, line 620: Is there no type `0` to represent "No target type"?
It may be useful to represent the the invite has no target at all compared to "we don't know what the target is due to not implementing it"

src/main/java/net/dv8tion/jda/internal/handle/InviteCreateHandler.java
```diff
@@ -74,7 +74,31 @@ protected Long handleInternally(DataObject content)
...
+            DataObject applicationObject = content.getObject("target_application");
+            application = new InviteImpl.EmbeddedApplicationImpl(
+                    applicationObject.getString("icon", null), applicationObject.getString("name"), applicationObject.getString("description"),
+                    applicationObject.getString("summary"), applicationObject.getLong("id"), applicationObject.getInt("max_participants", -1)
+            );
+            targetUser = null;
+            break;
+        default:
+            application = null;
+            targetUser = null;
```
DV8FromTheWorld, line 98:
```suggestion
            case STREAM:
                DataObject targetUserObject = content.getObject("target_user");
                targetUser = getJDA().getEntityBuilder().createUser(targetUserObject);
                application = null;
                break;
            case EMBEDDED_APPLICATION:
                DataObject applicationObject = content.getObject("target_application");
                application = new InviteImpl.EmbeddedApplicationImpl(
                        applicationObject.getString("icon", null), applicationObject.getString("name"), applicationObject.getString("description"),
                        applicationObject.getString("summary"), applicationObject.getLong("id"), applicationObject.getInt("max_participants", -1)
                );
                targetUser = null;
                break;
            default:
                application = null;
                targetUser = null;
```

DV8FromTheWorld (changes requested):
I'm definitely nervous about this API.
All of these setters _implicitly_ set the TargetType based on tangentially related data passed in.

What happens when discord adds a TargetType `3` that _also_ only needs a user?

Feels like something like `setTarget(Target.STREAM.of(User))` and `setTarget(Target.EMBEDDED_APPLICATION.of(long))` would be more likely to work further into the future.

Note: I don't like the above proposed builders, they were purely for example purposes. The point was the explicit targetting instead of the implicit targeting via tangential information

anweisen:
> I'm definitely nervous about this API.
> All of these setters _implicitly_ set the TargetType based on tangentially related data passed in.
> 
> What happens when discord adds a TargetType `3` that _also_ only needs a user?
> 
> Feels like something like `setTarget(Target.STREAM.of(User))` and `setTarget(Target.EMBEDDED_APPLICATION.of(long))` would be more likely to work further into the future.
> 
> Note: I don't like the above proposed builders, they were purely for example purposes. The point was the explicit targetting instead of the implicit targeting via tangential information

Maybe it would be better to use it like this:
  `setTargetType(TargetType.STREAM)`
  `setTargetUser(user)`

`setTargetUser` will then not set the target type to `STREAM`, so it has to be done by hand.
This would allows us to integrate a new type which uses the same setters.
But this approach is not ideal because you don't really know what you should/can set and what is for other target types.

src/main/java/net/dv8tion/jda/internal/handle/InviteCreateHandler.java
anweisen: I didn't indent these cases as they weren't indented in other switches I saw (eg in the EntityBuilder class) and the indention for them are set to false in the editorconfig. Does it make sense to indent them?

src/main/java/net/dv8tion/jda/api/entities/Invite.java
anweisen: When no target type is set for the invite discord just leaves this property out.
We could use `0` as default value which represents `TargetType.NONE` or something like that.

MinnDevelopment:
I think it's ok the way it is now, if in the future new target types get introduced we could add an overload that accepts a target type parameter. Currently, this is not needed since you can only have one target type per user/application anyway.

src/main/java/net/dv8tion/jda/api/entities/Invite.java
anweisen: > We could use `0` as default value which represents `TargetType.NONE` or something like that.

Added that in the latest commit, would it better to use `null` in this place?

carstendachsbacher:
Any updates on this? I would love to use this feature

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -503,4 +594,75 @@ default String getUrl()
...
+        /**
+         * Static accessor for retrieving a target type based on its Discord id key.
+         *
+         * @param  id
+         *         The id key of the requested target type.
+         *
+         * @return The TargetType that is referred to by the provided key. If the id key is unknown, {@link #UNKNOWN} is returned.
+         */
+        @Nonnull
+        public static TargetType fromId(int id) {
```
MinnDevelopment, line 658:
```suggestion
        public static TargetType fromId(int id)
        {
```

MinnDevelopment (approved):
I think we can merge this. @DV8FromTheWorld any reason to still hold this up?

DV8FromTheWorld (commented):
Some questions

DV8FromTheWorld:
After all this time I _still_ don't think that having `setTargetApplication` and `setTargetUser` and their respective getters are the correct move.

`setTarget(InviteTarget.APPLICATION.fromId(...))` and `getTarget().getType/getUser/getApplication` seems better.

Theres also nothing that indicates to the users that `.setTargetUser(...).setTargetApplication(...)` is wrong.

Xirado:
@MinnDevelopment Very weird. Maybe it's just my client? Me and my friends can't click the invite link even though it says Poker below it

PR #1631, opened 2021-05-25
Move reason setter to RestAction interface

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This moves the reason setter from `AuditableRestAction` into the `RestAction` interface. This does not mean all `RestActions` support the reason header, but they will just not do anything when a reason is supplied.


MinnDevelopment:
We have decided against this, for now we will stick to `AuditableRestAction`.

PR #1638, opened 2021-05-30
Remove nonsense runtime dependencies

iRebbok:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: Runtime dependencies & code annotations

Closes Issue: NaN

## Description

First of all, this PR removes the `org.jetbrains:annotations` & `com.google.code.findbugs:jsr` dependencies from the runtime classpath. That being said, they are never used at runtime, and it's absolutely nonsense to have them in the ABI. These are just a few annotations used by various IDEs.

Second of all, `org.slf4j:slf4j-api` is now an implementation dependency, not an API one. That dependency isn't a part of API (not method return type, not public field type, etc) and is used by the JDA's internal parts, which is the implementation dependency configuration for.

Third of all, all the `@Nonnull` & `@Nullable` annotations were replaced with the JetBrains' `@NotNull` & `@Nullable` ones. Or isn't it why the `org.jetbrains:annotations` dependency was added for?

-----

A brief summary of the size changes with these changes. Based on the [development](https://github.com/DV8FromTheWorld/JDA/tree/1dad81d1e5c8bc5f1524ba12efbc555d2e684fba) branch.

File | Size Before | Size After | Decrease |
-- | :---: | :---: | :---: |
JDA-4.2.1_DEV.jar | 1235Kb | 1256Kb | +21Kb
JDA-4.2.1_DEV-withDependencies.jar | 10,164Kb | 10,148Kb | -16Kb
JDA-4.2.1_DEV-withDependencies-min.jar | 4,419Kb | 4,395Kb | -24Kb
JDA-4.2.1_DEV-withDependencies-no-opus.jar | 6,861Kb | 6,845Kb | -16Kb

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -61,11 +61,11 @@ dependencies {
     /* ABI dependencies */
 
     //Code safety
-    api("com.google.code.findbugs:jsr305:3.0.2")
-    api("org.jetbrains:annotations:16.0.1")
+    compileOnly("com.google.code.findbugs:jsr305:3.0.2")
+    compileOnly("org.jetbrains:annotations:21.0.1")
 
     //Logger
-    api("org.slf4j:slf4j-api:1.7.25")
+    implementation("org.slf4j:slf4j-api:1.7.25")
 
     //Web Connection Support
     api("com.neovisionaries:nv-websocket-client:2.14")
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -23,9 +23,9 @@ import net.dv8tion.jda.api.entities.Role;
 import net.dv8tion.jda.internal.entities.EntityBuilder;
 import net.dv8tion.jda.internal.utils.Checks;
 import net.dv8tion.jda.internal.utils.Helpers;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import java.awt.*;
 import java.time.*;
 import java.time.temporal.TemporalAccessor;
@@ -114,7 +114,7 @@ public class EmbedBuilder
      *
      * @return the built, sendable {@link net.dv8tion.jda.api.entities.MessageEmbed}
      */
-    @Nonnull
+    @NotNull
     public MessageEmbed build()
     {
         if (isEmpty())
@@ -135,7 +135,7 @@ public class EmbedBuilder
      *
      * @return The current EmbedBuilder with default values
      */
-    @Nonnull
+    @NotNull
     public EmbedBuilder clear()
     {
         description.setLength(0);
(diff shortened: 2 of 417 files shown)


MinnDevelopment:
This change is breaking anything that depends on these transitive dependencies of JDA. Also changing all the annotations is useless and I don't want to even read the changelog here, that is true nonsense.

iRebbok:
Applications that depend on JDA mustn't use these transitive dependencies, firstly, at runtime, secondly, without explicitly specifying. That makes no sense to provide them as a part of the API, e.g. if JDA was a SDK you develop on, it'd be approvable.

Anyway, leaving here a trick to remove these from the runtime classpath:
```kts
    implementation("net.dv8tion:JDA:$jdaVersion") {
        // Exclude non-runtime dependencies
        exclude(module = "jsr305")
        exclude(module = "annotations")
    }
``

Andre601:
I personally believe (with my limited knowledge) that removing those transitive dependencies from the final JDA available would cause issues where the IDE complains about f.e. an Event being annotated as NonNull and you would need to find a replacement annotation to fix this.

I seriously see no benefit here either in removing those dependencies from the final jars other than final jar size which shouldn't matter that much in most cases.

iRebbok:
> I personally believe (with my limited knowledge) that removing those transitive dependencies from the final JDA available would cause issues where the IDE complains about f.e. an Event being annotated as NonNull and you would need to find a replacement annotation to fix this.
> 
> I seriously see no benefit here either in removing those dependencies from the final jars other than final jar size which shouldn't matter that much in most cases.

If you're talking about when you annotate `@NotNull` somewhere in your code, the IDE will complain because it can't resolve that module, you'll need to add that compile-time dependency in your gradle build script.

But if you're meaning that the IDE would complain because it can't correctly treat such annotations, then it's wrong. It's better for you to think about how the IDE handles these annotations. Reading a metadata from a `.class` isn't an issue for IDE's.

It's anyway a compile-time dependency that for no reason was included in the ABI. These annotations not even have a runtime retention policy and don't affect runtime in any way.

axieum:
I'm not sure why this was shut down so quickly. The PR does indeed clean up the dependencies, removing those unnecessary from the final distribution. There's even some that I didn't even notice myself - so kudos to you @iRebbok!

After checking JetBrain's docs, they say that the `jetbrains-annotations` **should** be `compileOnly`. This PR resolves that.

I found this, as I'm shadowing JDA in my project and relocating a bunch of the transitive dependencies from the classpath into my own as to not conflict, and there are a bunch of seemingly unnecessary dependencies (i.e. JetBrains annotations and javax annotations). Might be worth revisiting this PR, choosing one of `javax` or `jetbrains` annotations for consistency and making them `compileOnly` as their docs say so.

Horstexplorer:
While including those dependencies might have been a wrong decision in the first place removing them from the lib at this point in time will break it for anyone who -for whatever reason- depend on them to be there. 
As we have literally no gain of value by removing them there also is no real reason to force it to break for others. We might aswell just wait for the next major release to clear those out.

PR #1641, opened 2021-05-31
Upgrading gradle wrapper version to 7.0.2

aferg:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: \ build

Closes Issue: NaN

## Description

This pull request updates the Gradle wrapper to version 7.02 from 6.0.1. In order to do so, I needed to replace a few deprecated functionalities with their new ones, but it wasn't as hairy as I initially thought it would be :)  As a result of the upgrade to Gradle version being greater than 6.5, developers should now benefit from faster builds due to the Gradle virtual file system watching feature, which is enabled by default.


MinnDevelopment:
What is with all these `Encountered duplicate path` warnings that are spammed during the build?

aferg:
> What is with all these `Encountered duplicate path` warnings that are spammed during the build?

Yeah so my understanding is that Gradle 6 allowed duplicate paths by default, but Gradle 7 doesn't anymore and by default fails outright if there are any. There are a few options that you can set as the duplicates strategy and I wasn't sure if I should just outright ignore or not, so that's why I set it to warning. Meant to mention that, but it slipped my mind. Sorry about that. I think it can probably just be DuplicatesStrategy.EXCLUDE but wanted to discuss that first before excluding anything.

MinnDevelopment:
Ideally, the output jar should be identical. Whichever strategy achieves that without spamming the logs with warnings is fine by me.

aferg:
Ok thanks for the feedback Florian :) I will make the change and submit the update a bit later in the week when I have some more time

DV8FromTheWorld:
Bump

MinnDevelopment:
Closing due to inactivity.

PR #1648, opened 2021-06-05
Unify reply methods

Nick-kel:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation

## Description

Currently, there are several methods to reply to interactions. For TextChannels, there are only two methods (TextChannel#sendMessage and TextChannel#sendMessageFormat) and not an extra one for embeds. For interactons, there is an extra one. It would be better and more consistent to other methods (like the TextChannel methods) to unify them so that there isn't more an extra method for embeds.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/Interaction.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/Interaction.java
@@ -272,7 +272,7 @@ public interface Interaction extends ISnowflake
      */
     @Nonnull
     @CheckReturnValue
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
These names were chosen by intention. Java has restrictions to overloading due to type erasure. The reason we use `replyEmbeds` is that future overloads will use `Collection<? extends ComponentLayout>` and similar which would otherwise conflict with `Collection<MessageEmbed>` due to the type erasure.

This would also be a massive breaking change which is not acceptable.

PR #1649, opened 2021-06-05
Implement per-guild avatars

RedDaedalus:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This pull request implements per-guild avatars ([#3081](https://github.com/discord/discord-api-docs/pull/3081)). These additions have been tested and do work. It is still unclear whether bots will be able to access these.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdateAvatarEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdateAvatarEvent.java
@@ -1,0 +1,54 @@
+package net.dv8tion.jda.api.events.guild.member.update;
+
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.entities.Member;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Indicates that a {@link net.dv8tion.jda.api.entities.Member Member} updated their {@link net.dv8tion.jda.api.entities.Guild Guild} avatar.
+ *
+ * <p>Can be used to retrieve members who change their per guild avatar, triggering guild, the old avatar id and the new avatar id.
+ *
+ * <p>Identifier: {@code avatar}
+ *
+ * <h2>Requirements</h2>
+ *
+ * <p>This event requires the {@link net.dv8tion.jda.api.requests.GatewayIntent#GUILD_MEMBERS GUILD_MEMBERS} intent to be enabled.
+ * <br>{@link net.dv8tion.jda.api.JDABuilder#createDefault(String) createDefault(String)} and
+ * {@link net.dv8tion.jda.api.JDABuilder#createLight(String) createLight(String)} disable this by default!
+ *
+ * <p>Additionally, this event requires the {@link net.dv8tion.jda.api.utils.MemberCachePolicy MemberCachePolicy}
+ * to cache the updated members. Discord does not specifically tell us about the updates, but merely tells us the
+ * member was updated and gives us the updated member object. In order to fire a specific event like this we
+ * need to have the old member cached to compare against.
+ */
+public class GuildMemberUpdateAvatarEvent extends GenericGuildMemberUpdateEvent<String>
+{
+    public static final String IDENTIFIER = "avatar";
+
+    public GuildMemberUpdateAvatarEvent(@NotNull JDA api, long responseNumber, @NotNull Member member, @Nullable String oldAvatarId)
+    {
+        super(api, responseNumber, member, oldAvatarId, member.getAvatarId(), IDENTIFIER);
+    }
+
+    /**
+     * The old avatar id
+     *
+     * @return The old avatar id
+     */
+    public String getOldAvatarId()
+    {
+        return getOldValue();
+    }
+
+    /**
+     * The new avatar id
+     *
+     * @return The new avatar id
+     */
+    public String getNewAvatarId()
+    {
+        return getNewValue();
+    }
+}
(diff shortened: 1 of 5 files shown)


src/main/java/net/dv8tion/jda/internal/requests/Route.java
```diff
@@ -136,7 +137,7 @@
         public static final Route MODIFY_MEMBER =      new Route(PATCH,  "guilds/{guild_id}/members/{user_id}");
         public static final Route ADD_MEMBER =         new Route(PUT,    "guilds/{guild_id}/members/{user_id}");
         public static final Route GET_MEMBER =         new Route(GET,    "guilds/{guild_id}/members/{user_id}");
-        public static final Route MODIFY_SELF_NICK =   new Route(PATCH,  "guilds/{guild_id}/members/@me/nick");
+        public static final Route MODIFY_SELF =        new Route(PATCH, "guilds/{guild_id}/members/@me");
```
averen, line 140: A space was deleted before the route path
```suggestion
        public static final Route MODIFY_SELF =        new Route(PATCH,  "guilds/{guild_id}/members/@me");
```

Andre601:
The docs PR says it's not available for bots yet.
And when it isn't available for bots is there imo no need for this PR.

Sanduhr32:
> The docs PR says it's not available for bots yet.
> And when it isn't available for bots is there imo no need for this PR.

Where does it introduce a way for a bot to set a guild avatar Andre? Correct non yet. This is basically read-only support for now. Why would it matter if bots could set per-guild avatars?
Afaict this is basic API functionality for clients and bots.

Andre601:
> Where does it introduce a way for a bot to set a guild avatar Andre? Correct non yet. This is basically read-only support for now. Why would it matter if bots could set per-guild avatars?
> Afaict this is basic API functionality for clients and bots.

The PR comment says `Per guild avatars are rolled out to a small percentage of users and do not seem to be currently accessible to bots.` which can mean both that bots can't set a per-guild avatar but also that they can't access per-guild avatars of users yet.

RedDaedalus:
> > Where does it introduce a way for a bot to set a guild avatar Andre? Correct non yet. This is basically read-only support for now. Why would it matter if bots could set per-guild avatars?
> > Afaict this is basic API functionality for clients and bots.
> 
> The PR comment says `Per guild avatars are rolled out to a small percentage of users and do not seem to be currently accessible to bots.` which can mean both that bots can't set a per-guild avatar but also that they can't access per-guild avatars of users yet.

You are misinterpreting this, that is specifically referring to setting them, anyone can read them:
Screen Shot 2021-06-06 at 11 33 13 AM

src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdateAvatarEvent.java
```diff
@@ -0,0 +1,54 @@
...
+ * {@link net.dv8tion.jda.api.JDABuilder#createLight(String) createLight(String)} disable this by default!
+ *
+ * <p>Additionally, this event requires the {@link net.dv8tion.jda.api.utils.MemberCachePolicy MemberCachePolicy}
+ * to cache the updated members. Discord does not specifically tell us about the updates, but merely tells us the
+ * member was updated and gives us the updated member object. In order to fire a specific event like this we
+ * need to have the old member cached to compare against.
+ */
+public class GuildMemberUpdateAvatarEvent extends GenericGuildMemberUpdateEvent<String>
+{
+    public static final String IDENTIFIER = "avatar";
```
Andre601, line 28: Since the `UserUpdateAvatarEvent` also uses "avatar" as its identifier, wouldn't it be better to use something like `guild_avatar` to prevent possible confusion?

RedDaedalus: I was under the impression that the identifier is supposed to match the API field.

Andre601: Not sure what or *if* there are specific requirements for the identifier String.
The only usage I saw so far was in the Generic Event class for the `toString()` method

RedDaedalus: I've made this change, if it's wrong I'm sure Minn will leave a review.

src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdateAvatarEvent.java
```diff
@@ -0,0 +1,54 @@
+package net.dv8tion.jda.api.events.guild.member.update;
```
MinnDevelopment, line 1: Missing copyright

src/main/java/net/dv8tion/jda/api/events/guild/member/update/GuildMemberUpdateAvatarEvent.java
```diff
@@ -0,0 +1,54 @@
...
+    {
+        return getOldValue();
+    }
+
+    /**
+     * The new avatar id
+     *
+     * @return The new avatar id
+     */
+    public String getNewAvatarId()
```
MinnDevelopment, line 50: Missing nullability annotations

MinnDevelopment: Would be nice to also have getters for the URL

RedDaedalus:
The [docs pr](https://github.com/discord/discord-api-docs/pull/3081) has been merged and reviews have been addressed.

PR #1653, opened 2021-06-09
Implement Button.of(ButtonStyle, String, String, Emoji) to quickly get a button with a label and emoji.

Arraying:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds a handy little method to quickly construct a button with both an emoji and label (which I found to be a common use case) without having to chain multiple methods. The signature of this method is `static Button of(@Nonnull ButtonStyle style, @Nonnull String idOrUrl, @Nonnull String label, @Nonnull Emoji emoji)`. I have good-weather tested it against links as well as normal buttons; the preconditions have been taken from `Button.of(ButtonStyle, String, String)` and `Button.of(ButtonStyle, String, Emoji)`.


src/main/java/net/dv8tion/jda/api/interactions/components/Button.java
```diff
@@ -568,4 +568,40 @@ static Button of(@Nonnull ButtonStyle style, @Nonnull String idOrUrl, @Nonnull E
         Checks.notLonger(idOrUrl, 100, "Id");
         return new ButtonImpl(idOrUrl, "", style, false, emoji);
     }
+
+    /**
+     * Create a button with the provided {@link ButtonStyle style}, URL or ID, label and emoji.
+     * <br>The button is enabled.
```
arynxd, line 574:
```suggest
     * Create an enabled button with the provided {@link ButtonStyle style}, URL or ID, label and emoji.

```
clarification

src/main/java/net/dv8tion/jda/api/interactions/components/Button.java
```diff
@@ -568,4 +568,40 @@ static Button of(@Nonnull ButtonStyle style, @Nonnull String idOrUrl, @Nonnull E
...
+        Checks.notNull(style, "Style");
+        Checks.notNull(label, "Label");
+        Checks.notLonger(label, 80, "Label");
+        Checks.notNull(emoji, "Emoji");
+        if (style == ButtonStyle.LINK)
+            return link(idOrUrl, emoji).withLabel(label);
+        Checks.notEmpty(idOrUrl, "Id");
+        Checks.notLonger(idOrUrl, 100, "Id");
+        return new ButtonImpl(idOrUrl, label, style, false, emoji);
+    }
```
MinnDevelopment, line 606:
```suggestion
    static Button of(@Nonnull ButtonStyle style, @Nonnull String idOrUrl, @Nullable String label, @Nullable Emoji emoji)
    {
        if (label != null)
            return of(style, idOrUrl, label).withEmoji(emoji);
        if (emoji != null)
            return of(style, idOrUrl, emoji);
        throw new IllegalArgumentException("Cannot build a button without a label and without an emoji. It must have at least one!");
    }
```

src/main/java/net/dv8tion/jda/api/interactions/components/Button.java
```diff
@@ -568,4 +568,40 @@ static Button of(@Nonnull ButtonStyle style, @Nonnull String idOrUrl, @Nonnull E
...
+     *         The button style
+     * @param  idOrUrl
+     *         Either the ID or URL for this button
+     * @param  label
+     *         The text to display on the button
+     * @param  emoji
+     *         The emoji to use as the button label
+     *
+     * @throws IllegalArgumentException
+     *         If any argument is empty or null, the label is longer than 80 characters, the id is longer than 100 characters, or the url is longer than 512 characters
```
MinnDevelopment, line 589: This should be a list and updated to allow label or emoji to be null, but not both.

src/main/java/net/dv8tion/jda/api/interactions/components/Button.java
```diff
@@ -568,4 +568,40 @@ static Button of(@Nonnull ButtonStyle style, @Nonnull String idOrUrl, @Nonnull E
         Checks.notLonger(idOrUrl, 100, "Id");
         return new ButtonImpl(idOrUrl, "", style, false, emoji);
     }
+
+    /**
+     * Create an enabled button with the provided {@link ButtonStyle style}, URL or ID, label and emoji.
+     
+     * You can use {@link #asDisabled()} to disable it.
```
MinnDevelopment, line 575:
```suggestion
     * <p>You can use {@link #asDisabled()} to disable it.
```

MinnDevelopment (changes requested):
I think emoji and label should both be nullable to allow for better dynamics. Also, you shouldn't copy code from other methods, just call the methods instead. With these changes, we at least introduce a new way to build buttons rather than just a "quicker" way.

src/main/java/net/dv8tion/jda/api/interactions/components/Button.java
Arraying: What was the reasoning behind not making label nor emoji nullable in the other two `of` methods? I think, if I were to make it nullable, it should at least be consistent with the other methods that I will be calling internally (to fix the code repetition). 

Also, in the JavaDoc for the `of` method with just the Emoji, in the description, only `ButtonStyle` is linked and `Emoji` is not. Shall I also link it to the correct class in this method and my method?

MinnDevelopment: The reason is simply that you should not be able to make a button without emoji and without label. This is done by enforcing each method to accept one non-null parameter.

MinnDevelopment (approved):

PR #1660, opened 2021-06-11
add missing checks to Message#suppressEmbeds

caneleex:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this PR adds missing checks to Message#suppressEmbeds. I'm not sure about throwing `PermissionException` in the private channel case but it's in the docs


src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -862,6 +863,14 @@ else if (!getGuild().getSelfMember()
     @Override
     public AuditableRestAction<Void> suppressEmbeds(boolean suppressed)
     {
+        if (!getJDA().getSelfUser().equals(getAuthor()))
+        {
+            TextChannel textChannel = getTextChannel();
+            if (isFromType(ChannelType.PRIVATE))
```
MinnDevelopment, line 869: It looks like you call `getTextChannel()` before checking the type, this means it will throw before this check ever happens.

PR #1672, opened 2021-06-16
Add check and recovery for missing self members

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Very rarely, discord sends us guilds with missing self members. This pull request adds a check for this case and a recovery attempt. We make a gateway request to load the self member after creating the guild object, which should in theory be able to recover the self member.


DV8FromTheWorld:
I wonder if we should use the GuildLock system (or another form of waiting) before releasing the Guild for use if we need to re-request the self-member 🤔

MinnDevelopment:
This is a pretty rare API error which might not even recover successfully. I don't think we should prevent ready event because of a buggy guild initialization.

DV8FromTheWorld:
Where are we on this? Are we confident enough to go ahead and merge this in for inclusion in the v4 LTS?

MinnDevelopment:
Should be good to go.

PR #1680, opened 2021-06-20
Fixed the issue with last character handling for cases when the length of the message is 4001 characters

maxtrap:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This fixes the issue that happens when `MessageBuilder#buildAll` is called on a message with 2001+ characters. Once the message is split, the last character of the last message ends up being cut off. The reason this happens is because the endIndex of the substring is set as builder.length()-1, even though a substring's end index is exclusive. As a result, the last char in the StringBuilder of the message ends up being cut off. 

## Tested

With the following sample code
```java
        StringBuilder db = new StringBuilder();
        for (int i = 0; i < 4000; i++) {
            if (i != 1999 && i != 3999)
                db.append('a');
            else db.append('b');
        }
        db.append('c');

        new MessageBuilder(db.toString())
                .buildAll(MessageBuilder.SplitPolicy.ANYWHERE)
                .forEach(replyMessage -> getServer().getTextChannelById(CHANNEL_ID).sendMessage(replyMessage.getContentRaw()).queue());
```

Existing implementation loses last 'c' letter, by generating 2 messages 
image

And proposed implementation correctly generates 3 messages
image

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/MessageBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/MessageBuilder.java
@@ -1274,7 +1274,7 @@ public class MessageBuilder implements Appendable
 
         LinkedList<Message> messages = new LinkedList<>();
 
-        if (builder.length() <= 2000) {
+        if (builder.length() <= Message.MAX_CONTENT_LENGTH) {
             messages.add(this.build());
             return messages;
         }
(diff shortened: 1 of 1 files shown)


DManstrator:
Good catch!

Small suggestion: Use the constant from the Message interface ([`Message.MAX_CONTENT_LENGTH`](https://github.com/DV8FromTheWorld/JDA/blob/8bfe22fd24f012c47bf9e1e8111ee878319b435d/src/main/java/net/dv8tion/jda/api/entities/Message.java#L141-L147)) instead of the `2000`.

Almighty-Alpaca (approved):
I agree with the comment from @DManstrator but otherwise lgtm

maxtrap:
Thanks for review.
Addressed comment by replacing hardcoded number with constant.
Please take a look again.

DManstrator:
> otherwise lgtm

`lgtm` means "looks good to me". So it is fine now. :+1:

PR #1682, opened 2021-06-20
Remove last CLIENT AccountType occurrences

DManstrator:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [x] Other: Examples

## Description

Removes the last occurrences of the `CLIENT` `AccountType` by removing a constant and its mention from the [`MessageListenerExample`](https://github.com/DV8FromTheWorld/JDA/blob/development/src/examples/java/MessageListenerExample.java).

Also removed the `Group` mention from the Example and added the `StoreChannel` instead.

### DIFF
--- a/src/examples/java/MessageListenerExample.java
+++ b/src/examples/java/MessageListenerExample.java
@@ -86,7 +86,7 @@ public class MessageListenerExample extends ListenerAdapter
         User author = event.getAuthor();                //The user that sent the message
         Message message = event.getMessage();           //The message that was received.
         MessageChannel channel = event.getChannel();    //This is the MessageChannel that the message was sent to.
-                                                        //  This could be a TextChannel, PrivateChannel, or Group!
+                                                        //  This could be a channel from a guild (TextChannel or a StoreChannel) or a PrivateChannel!
 
         String msg = message.getContentDisplay();       //This returns a human readable version of the Message. Similar to
                                                         //  what you would see in the client.
--- a/src/examples/java/MessageListenerExample.java
+++ b/src/examples/java/MessageListenerExample.java
@@ -86,7 +86,7 @@ public class MessageListenerExample extends ListenerAdapter
         User author = event.getAuthor();                //The user that sent the message
         Message message = event.getMessage();           //The message that was received.
         MessageChannel channel = event.getChannel();    //This is the MessageChannel that the message was sent to.
-                                                        //  This could be a channel from a guild (TextChannel or a StoreChannel) or a PrivateChannel!
+                                                        //  This could be a TextChannel or a PrivateChannel!
 
         String msg = message.getContentDisplay();       //This returns a human readable version of the Message. Similar to
                                                         //  what you would see in the client.
--- a/src/examples/java/MessageListenerExample.java
+++ b/src/examples/java/MessageListenerExample.java
@@ -35,8 +35,7 @@ public class MessageListenerExample extends ListenerAdapter
      */
     public static void main(String[] args)
     {
-        //We construct a builder for a BOT account. If we wanted to use a CLIENT account
-        // we would use AccountType.CLIENT
+        //We construct a builder for the bot.
         try
         {
             JDA jda = JDABuilder.createDefault("Your-Token-Goes-Here") // The token of the account that is logging in.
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/entities/MessageEmbed.java
```diff
@@ -93,17 +93,6 @@
...
-    /**
-     * The maximum amount of total visible characters an embed can have
-     * <br>This limit depends on the current {@link net.dv8tion.jda.api.AccountType AccountType} and applies to CLIENT
-     *
-     * @see net.dv8tion.jda.api.EmbedBuilder#setDescription(CharSequence)
-     * @see net.dv8tion.jda.api.EmbedBuilder#setTitle(String)
-     * @see net.dv8tion.jda.api.EmbedBuilder#setFooter(String, String)
-     * @see net.dv8tion.jda.api.EmbedBuilder#addField(String, String, boolean)
-     */
-    public static final int EMBED_MAX_LENGTH_CLIENT = 2000;
```
MinnDevelopment, line 105: I don't see why we would remove this? It can still be used to know the limits for a client account.

src/examples/java/MessageListenerExample.java
```diff
@@ -87,13 +86,13 @@ public void onMessageReceived(MessageReceivedEvent event)
         User author = event.getAuthor();                //The user that sent the message
         Message message = event.getMessage();           //The message that was received.
         MessageChannel channel = event.getChannel();    //This is the MessageChannel that the message was sent to.
-                                                        //  This could be a TextChannel, PrivateChannel, or Group!
+                                                        //  This could be a channel from a guild (TextChannel or a StoreChannel) or a PrivateChannel!
```
MinnDevelopment, line 89: Last I checked store channels are not message channels.

src/main/java/net/dv8tion/jda/api/entities/MessageEmbed.java
DManstrator: But isn't JDA the wrong place for that since it completely dropped Client support? Not that it would be in interest of any JDA developer.

I looked at the Discord Documentation and I can't find any mention about that, perhaps the limit was dropped too?

src/examples/java/MessageListenerExample.java
DManstrator: Yeah, my bad, I missread the docs. I saw both `StoreChannel` and `TextChannel` extend `GuildChannel` and confused that with `MessageChannel`. Guess I also confused the `StoreChannel` with the News / Announcements Channel. Sorry for that, I'll revert that.

src/main/java/net/dv8tion/jda/api/entities/MessageEmbed.java
MinnDevelopment: Doesn't matter anyway since this would be a breaking change.

DManstrator: Yeah, true. Then I would add a deprecation notice that it will be removed in 4.4.0 or so. Okay?

Imo it should still be completely removed at some point.

RedDaedalus: This field could still be useful for configuring max string lengths in a database or something, I don't really think it needs to be removed or deprecated.

DManstrator: The value of `MessageEmbed.EMBED_MAX_LENGTH_CLIENT` is identical to ([`Message.MAX_CONTENT_LENGTH`](https://github.com/DV8FromTheWorld/JDA/blob/8bfe22fd24f012c47bf9e1e8111ee878319b435d/src/main/java/net/dv8tion/jda/api/entities/Message.java#L141-L147)). So this constant can be used instead if it is really needed. Users can also only use Embeds if they use a selfbot or something similar so it makes even less sense for this constant to exist imo.

Andre601: I myself use this in different places where I make pagination embeds to check when I should split and end in a new embed, so it can have its uses.

Issue #1679, opened 2021-06-19
Split canReceiveCombined() in two functions


BaybaraEvgeny:
- [ ✅] I have checked for similar issues.
- [ ✅] I have updated to the [latest JDA version][download].
- [✅ ] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request

Function AudioReceiveHandler::canReceiveCombined() is called every 20 ms, but if the first call returns false, then CombinedExecutor never will be set up. 
I want to have receive combined voice feature but I don't want to receive every single packet. I want to decide if I will receive next packet or not, just like AudioSendHandler::canProvide() do every 20 ms. Also I want to maximize performance.
This function has two meanings(setting up CombinedExecutor and receiving next packet), which is wrong.
### Example Use-Case
I suggest splitting this function in two: SetCombinedExecutor() and canReceiveCombined().

DV8FromTheWorld:
PR was rejected. See PR for details.

PR #1685, opened 2021-06-20
setCombinedExecutor()

BaybaraEvgeny:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation

Closes Issue: https://github.com/DV8FromTheWorld/JDA/issues/1679

## Description

Splits canReceiveCombined() funtion in two

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
+++ b/src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
@@ -43,10 +43,13 @@ public interface AudioReceiveHandler
     }
 
     /**
-     * If this method returns true, then JDA will generate combined audio data and provide it to the handler.
+     * If this method returns true JDA will attempt to provide audio data to this handler by calling
+     * {@link #handleCombinedAudio(CombinedAudio)}. The return value is checked each time JDA attempts receive audio, so if
+     * the developer wanted to start and stop receiving audio it could be done by changing the value returned
+     * by this method at runtime.
      * <br><b>Only enable if you specifically want combined audio because combining audio is costly if unused.</b>
      *
-     * @return If true, JDA will receive next combined packet
+     * @return If true, JDA will attempt to receive next combined packet
      */
     default boolean canReceiveCombined()
     {
--- a/src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
+++ b/src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
@@ -39,7 +39,7 @@ public interface AudioReceiveHandler
      */
     default boolean canSupportCombinedAudio()
     {
-        return true;
+        return canReceiveCombined();
     }
 
     /**
--- a/src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
+++ b/src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
@@ -35,7 +35,7 @@ public interface AudioReceiveHandler
      * If this method returns true, then JDA will generate combined audio data and provide it to the handler.
      * <br><b>Only enable if you specifically want combined audio because combining audio is costly if unused.</b>
      *
-     * @return If true, JDA will combine the next 20ms of user audio
+     * @return If true, JDA enables subsystems to combine all user audio into a single provided data packet.
      */
     default boolean canSupportCombinedAudio()
     {
@@ -49,7 +49,7 @@ public interface AudioReceiveHandler
      * by this method at runtime.
      * <br><b>Only enable if you specifically want combined audio because combining audio is costly if unused.</b>
      *
-     * @return If true, JDA will attempt to receive next combined packet
+     * @return If true, JDA will combine the next 20ms of user audio
      */
     default boolean canReceiveCombined()
     {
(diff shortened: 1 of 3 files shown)


arynxd:
PRs should target the development branch. see [here](https://github.com/DV8FromTheWorld/JDA/wiki/5%29-Contributing#making-changes)

src/main/java/net/dv8tion/jda/internal/audio/AudioConnection.java
```diff
@@ -468,7 +468,7 @@ else if (couldReceive)
             receiveThread.start();
         }
 
-        if (receiveHandler.canReceiveCombined())
+        if (receiveHandler.setCombinedExecutor())
```
Horstexplorer, line 471: Is there a reason for why this method follows a setter like naming scheme while not actually being a setter?

BaybaraEvgeny: fixed

src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
```diff
@@ -37,6 +37,20 @@
      *
      * @return If true, JDA enables subsystems to combine all user audio into a single provided data packet.
      */
+    default boolean shouldCombineAudio()
+    {
+        return false;
```
MinnDevelopment:
```suggestion
        return canReceiveCombined();
```

src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
```diff
@@ -37,6 +37,20 @@
      *
      * @return If true, JDA enables subsystems to combine all user audio into a single provided data packet.
      */
+    default boolean shouldCombineAudio()
```
MinnDevelopment, line 40: I think a better name would be `canSupportCombinedAudio()`

src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
```diff
@@ -37,6 +37,20 @@
...
+    }
+
+    /**
+     * If this method returns true JDA will attempt to provide audio data to this handler by calling
+     * {@link #handleCombinedAudio(CombinedAudio)}. The return value is checked each time JDA attempts receive audio, so if
+     * the developer wanted to start and stop receiving audio it could be done by changing the value returned
+     * by this method at runtime.
+     * <br><b>Only enable if you specifically want combined audio because combining audio is costly if unused.</b>
+     *
+     * @return If true, JDA will attempt to receive next combined packet
```
MinnDevelopment, line 52:
```suggestion
     * @return If true, JDA will combine the next 20ms of user audio
```

src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
```diff
@@ -37,6 +37,20 @@
...
+    default boolean shouldCombineAudio()
+    {
+        return false;
+    }
+
+    /**
+     * If this method returns true JDA will attempt to provide audio data to this handler by calling
+     * {@link #handleCombinedAudio(CombinedAudio)}. The return value is checked each time JDA attempts receive audio, so if
+     * the developer wanted to start and stop receiving audio it could be done by changing the value returned
+     * by this method at runtime.
```
MinnDevelopment, line 49:
```suggestion
     * If this method returns true JDA will attempt to provide audio data to this handler by calling
     * {@link #handleCombinedAudio(CombinedAudio)}. The return value is checked each time JDA receives audio, so if
     * the developer wanted to start and stop receiving audio it could be done by changing the value returned
     * by this method at runtime.
```

src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
```diff
@@ -37,6 +37,20 @@
      *
      * @return If true, JDA enables subsystems to combine all user audio into a single provided data packet.
      */
+    default boolean shouldCombineAudio()
+    {
+        return true;
```
MinnDevelopment, line 42:
```suggestion
        return canReceiveCombined();
```

MinnDevelopment (changes requested):

src/main/java/net/dv8tion/jda/api/audio/AudioReceiveHandler.java
```diff
@@ -39,7 +39,7 @@
      */
     default boolean shouldCombineAudio()
     {
-        return false;
+        return true;
```
BaybaraEvgeny, line 42: If default is false then everyone's apps will break and won't receive combined audio

DV8FromTheWorld:
This can be replicated by calling `AudioManager#setReceivingHandler` each time you modify `canReceiveCombined`.

Part of my problem with this change is that add `canSupportCombinedAudio` doesn't fix the problem, it just kicks the can. If `canSupportCombinedAudio` starts as `false` and then changed to `true` you're stuck with the same problem.

As such, I don't know that I see a reason to add it in, at least not in this form.

DV8FromTheWorld:
Going to close this given the emoji responses seem to be okay with this not being included.

PR #1686, opened 2021-06-21
Implement access and retrieval of message_reference

arynxd:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds support for retrieving referenced message in cases where discord does not provide one. It also adds a MessageReference class which wraps the message_reference value seen in a message's json. It also adds some getters for referenced message id


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -226,6 +226,44 @@
     @Nullable
     Message getReferencedMessage();
 
+    /**
+     * This returns the id of the referenced message, never null
```
DV8FromTheWorld, line 230:
```suggestion
     * This returns the id of the referenced message.
```

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1232,15 +1232,29 @@ else if (MISSING_CHANNEL.equals(ex.getMessage()))
...
+        if (!jsonObject.isNull("message_reference"))
+        {
+            DataObject messageReferenceJson = jsonObject.getObject("message_reference");
+
+            messageReference = new MessageReference(
+                    messageReferenceJson.getLong("message_id"),
+                    messageReferenceJson.getLong("channel_id"),
+                    referencedMessage
+            );
+        }
```
DV8FromTheWorld, line 1247: Lets move this up closer to the `referenced_message` code
Lets also add some inline documentation explaining (briefly) the difference between `referenced_message` and `message_reference`

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -88,15 +90,17 @@
...
         boolean fromWebhook, boolean mentionsEveryone, TLongSet mentionedUsers, TLongSet mentionedRoles, boolean tts, boolean pinned,
         String content, String nonce, User author, Member member, MessageActivity activity, OffsetDateTime editTime,
         List<MessageReaction> reactions, List<Attachment> attachments, List<MessageEmbed> embeds, List<MessageSticker> stickers, List<ActionRow> components, int flags)
     {
         super(content, nonce, tts);
         this.id = id;
         this.channel = channel;
-        this.referencedMessage = referencedMessage;
+        this.messageReference = messageReference;
+        this.referencedMessage = (messageReference != null) ? messageReference.getReferencedMessage() : null;
```
DV8FromTheWorld, line 102: Why not just do `messageReference.getReferenedMessage()` where needed in the code instead of storing it here?

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -135,6 +139,52 @@ public Message getReferencedMessage()
         return referencedMessage;
     }
 
+    @Override
+    public long getReferencedMessageIdLong()
+    {
+        if (messageReference == null)
+        {
+            throw new IllegalStateException("Cannot get an id for a non-existent reference.");
```
DV8FromTheWorld, line 147: Thoughts?
```suggestion
            throw new IllegalStateException("This message does not reference another message!");
```

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -135,6 +139,52 @@ public Message getReferencedMessage()
...
+    @Override
+    public RestAction<Message> retrieveReferencedMessage()
+    {
+        if (messageReference == null) {
+            throw new IllegalStateException("Cannot get a retrieve a message for a non-existent reference.");
+        }
+
+        if (referencedMessage != null) {
+            return new CompletedRestAction<>(getJDA(), referencedMessage);
+        }
```
DV8FromTheWorld, line 180: I'm not sure about the implications of this.
Retrieve typically means "go get it from the API", thus it is always up-to-date.

The functionality of this method indicates that if we initially constructed the `Message` with a full referenced message then this method cannot be used to get an updated version of that message. The developer would have to know that in some cases this method can be used to get updated versions of the message and other times it cannot be.

DV8FromTheWorld (changes requested):

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -226,6 +226,44 @@
...
+     */
+    @Nonnull
+    String getReferencedMessageId();
+
+    /**
+     * This determines whether a message has a reference, for use with {@link Message#retrieveReferencedMessage()}
+     *
+     * @return Whether this message has a reference attached to it
+     */
+    boolean hasReferencedMessage();
```
MinnDevelopment, line 265: Wouldn't it be better to simply have a single getter for all of this instead of 5?

```java
@Nullable
MessageReference getMessageReference();
```

MinnDevelopment:
Bump

arynxd:
Sorry about the late commit, i never received the notifications for this PR.

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
+package net.dv8tion.jda.internal.entities;
```
MinnDevelopment, line 1: Copyright

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1219,6 +1219,20 @@ else if (MISSING_CHANNEL.equals(ex.getMessage()))
...
+
+        if (!jsonObject.isNull("message_reference")) // always contains the channel + message id for a referenced message
+        {                                                // used for when referenced_message is not provided
+            DataObject messageReferenceJson = jsonObject.getObject("message_reference");
+
+            messageReference = new MessageReference(
+                    messageReferenceJson.getLong("message_id"),
+                    messageReferenceJson.getLong("channel_id"),
+                    referencedMessage,
+                    api
```
MinnDevelopment, line 1232: What about guild_id?

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -214,17 +215,29 @@
...
-     * This can be null even if the type is {@link MessageType#INLINE_REPLY INLINE_REPLY}, when the message it references doesn't exist or discord wasn't able to resolve it in time.
+     * @throws java.lang.IllegalStateException
+     *         If this message does not have a reference
      *
-     * @return The referenced message, or null
+     * @return The referenced message id
      */
-    @Nullable
-    Message getReferencedMessage();
+    long getReferencedMessageIdLong();
```
MinnDevelopment, line 229: Why would I use this over `getMessageReference().getMessageIdLong()`?

src/main/java/net/dv8tion/jda/internal/entities/AbstractMessage.java
```diff
@@ -104,9 +107,27 @@ protected void appendFormat(Formatter formatter, int width, int precision, boole
...
+    {
+        unsupported();
+        return null;
+    }
+
     @Override
-    public Message getReferencedMessage()
+    public long getReferencedMessageIdLong()
     {
+        unsupported();
```
MinnDevelopment, line 122: This isn't implemented anywhere, is it?

src/main/java/net/dv8tion/jda/internal/entities/AbstractMessage.java
```diff
@@ -104,9 +107,27 @@ protected void appendFormat(Formatter formatter, int width, int precision, boole
         }
     }
 
+
+    @NotNull
```
MinnDevelopment, line 111: We use a different annotation from `javax.annotations` called `Nonnull`

src/main/java/net/dv8tion/jda/internal/entities/AbstractMessage.java
```diff
@@ -32,6 +32,9 @@
 import java.time.OffsetDateTime;
 import java.util.*;
 
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
```
MinnDevelopment, line 36: We use javax.annotations

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Retrieves the referenced message for this message. If one is already present, it will be returned.
+     *
+     * @throws java.lang.IllegalStateException
+     *         If this message does not have a reference
+     *
+     * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.Message}
+     */
+    @Nonnull
+    public RestAction<Message> retrieveReferencedMessage() {
```
MinnDevelopment, line 39:
```suggestion
    public RestAction<Message> retrieveReferencedMessage()
    {
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+    public MessageReference(long messageId, long channelId, @Nullable Message referencedMessage, JDA api)
+    {
+        this.messageId = messageId;
+        this.channelId = channelId;
+        this.referencedMessage = referencedMessage;
+        this.api = api;
+    }
+
+    /**
+     * Retrieves the referenced message for this message. If one is already present, it will be returned.
```
MinnDevelopment, line 31: Missing docs for error responses

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
MinnDevelopment: How about just `resolve()` instead of this long name? `message.getMessageReference().retrieveReferencedMessage()` seems way too long and unnecessarily redundant.

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * This will return the channel from which this message originates.
+     * <br>
+     * Messages from other servers can be referenced, in which case JDA may not have the channel cached.
+     *
+     * @return The channel for this message reference, or null
+     */
+    @Nullable
+    public TextChannel getChannel() {
+        return api.getTextChannelById(getChannelIdLong());
```
MinnDevelopment, line 71: This is not good. What if the message came from a private channel? Why is this not stored in the object?

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * This will return the channel from which this message originates.
+     * <br>
+     * Messages from other servers can be referenced, in which case JDA may not have the channel cached.
+     *
+     * @return The channel for this message reference, or null
+     */
+    @Nullable
+    public TextChannel getChannel() {
```
MinnDevelopment, line 70:
```suggestion
    public TextChannel getChannel()
    {
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+        return api.getTextChannelById(getChannelIdLong());
+    }
+
+    /**
+     * Returns the message id for this reference.
+     * @return The message id
```
MinnDevelopment, line 76:
```suggestion
     *
     * @return The message id
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public long getMessageIdLong()
+    {
+        return messageId;
+    }
+
+    /**
+     * Returns the channel id for this reference.
+     * @return The channel id
```
MinnDevelopment, line 85:
```suggestion
     *
     * @return The channel id
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public long getChannelIdLong()
+    {
+        return channelId;
+    }
+
+    /**
+     * Returns the message id for this reference.
+     * @return The message id
```
MinnDevelopment, line 94:
```suggestion
     *
     * @return The message id
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public String getMessageId()
+    {
+        return Long.toUnsignedString(getMessageIdLong());
+    }
+
+    /**
+     * Returns the channel id for this reference.
+     * @return The channel id
```
MinnDevelopment, line 103:
```suggestion
     *
     * @return The channel id
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public String getChannelId()
+    {
+        return Long.toUnsignedString(getChannelIdLong());
+    }
+
+    /**
+     * Returns the JDA instance related to this message reference.
+     * @return The corresponding JDA instance
```
MinnDevelopment, line 112:
```suggestion
     *
     * @return The corresponding JDA instance
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public JDA getJDA() {
```
MinnDevelopment, line 114:
```suggestion
    public JDA getJDA()
    {
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+import net.dv8tion.jda.api.entities.TextChannel;
+import net.dv8tion.jda.api.requests.RestAction;
+import net.dv8tion.jda.internal.JDAImpl;
+import net.dv8tion.jda.internal.requests.RestActionImpl;
+import net.dv8tion.jda.internal.requests.Route;
+import org.jetbrains.annotations.Nullable;
+
+import javax.annotation.Nonnull;
+
+public class MessageReference
```
MinnDevelopment, line 15: Class level docs, and most of the methods are missing annotations

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+    @Nullable
+    public Message getReferencedMessage()
+    {
+        return referencedMessage;
+    }
+
+    /**
+     * This will return the channel from which this message originates.
+     * <br>
+     * Messages from other servers can be referenced, in which case JDA may not have the channel cached.
```
MinnDevelopment, line 65:
```suggestion
     * <br>Messages from other guilds can be referenced, in which case JDA may not have the channel cached.
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+     * @return The referenced message, or null
+     */
+    @Nullable
+    public Message getReferencedMessage()
+    {
+        return referencedMessage;
+    }
+
+    /**
+     * This will return the channel from which this message originates.
```
MinnDevelopment, line 63:
```suggestion
     * The channel from which this message originates.
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
```diff
@@ -0,0 +1,117 @@
...
+     * <br>
+     * Messages from other servers can be referenced, in which case JDA may not have the channel cached.
+     *
+     * @return The channel for this message reference, or null
```
MinnDevelopment, line 67:
```suggestion
     * @return The origin channel for this message reference, or null
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -214,17 +215,29 @@
...
-     * Usually, this is a {@link MessageType#INLINE_REPLY INLINE_REPLY} reference.
-     * This can be null even if the type is {@link MessageType#INLINE_REPLY INLINE_REPLY}, when the message it references doesn't exist or discord wasn't able to resolve it in time.
+     * @throws java.lang.IllegalStateException
+     *         If this message does not have a reference
      *
-     * @return The referenced message, or null
+     * @return The referenced message id
      */
-    @Nullable
-    Message getReferencedMessage();
```
MinnDevelopment, line 227: Removing this is a breaking change, please add it back.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -88,15 +89,16 @@
...
         boolean fromWebhook, boolean mentionsEveryone, TLongSet mentionedUsers, TLongSet mentionedRoles, boolean tts, boolean pinned,
         String content, String nonce, User author, Member member, MessageActivity activity, OffsetDateTime editTime,
         List<MessageReaction> reactions, List<Attachment> attachments, List<MessageEmbed> embeds, List<MessageSticker> stickers, List<ActionRow> components, int flags)
     {
         super(content, nonce, tts);
         this.id = id;
         this.channel = channel;
-        this.referencedMessage = referencedMessage;
+        this.messageReference = messageReference;
+
```
MinnDevelopment, line 101:
```suggestion
```

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -129,10 +131,15 @@ public JDA getJDA()
         return api;
     }
 
+    /**
+     * Returns the {@link MessageReference} for this Message. This will be null if this Message has no reference.
+     * @return The message reference.
```
MinnDevelopment, line 136:
```suggestion
     *
     * @return The message reference.
```

MinnDevelopment: You don't need to add documentation to implementations of an interface that already has docs

src/main/java/net/dv8tion/jda/internal/entities/SystemMessage.java
```diff
@@ -29,16 +29,17 @@
...
         boolean tts, boolean pinned,
         String content, String nonce, User author, Member member, MessageActivity activity, OffsetDateTime editTime,
         List<MessageReaction> reactions, List<Attachment> attachments, List<MessageEmbed> embeds, List<MessageSticker> stickers, int flags)
     {
-        super(id, channel, type, null, fromWebhook, mentionsEveryone, mentionedUsers, mentionedRoles,
+        super(id, channel, type, messageReference, fromWebhook, mentionsEveryone, mentionedUsers, mentionedRoles,
             tts, pinned, content, nonce, author, member, activity, editTime, reactions, attachments, embeds, stickers, Collections.emptyList(), flags);
     }
 
+
```
MinnDevelopment, line 42:
```suggestion
```

src/main/java/net/dv8tion/jda/internal/entities/MessageReference.java
MinnDevelopment: Why is this internal? You expose it in the interface so it has to be in api.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1232,15 +1246,16 @@ else if (MISSING_CHANNEL.equals(ex.getMessage()))
 
         if (type == MessageType.UNKNOWN)
             throw new IllegalArgumentException(UNKNOWN_MESSAGE_TYPE);
+
```
MinnDevelopment, line 1249:
```suggestion
```

MinnDevelopment (changes requested):
Please add back the things for referenced message that you removed.

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,234 @@
...
+    public long getChannelIdLong()
+    {
+        return channelId;
+    }
+
+    /**
+     * Returns the guild id for this reference.
+     * @return The guild id
+     */
+    public long getGuildIdLong()
```
MinnDevelopment, line 175: Is this always present?

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,234 @@
...
+    {
+        return Long.toUnsignedString(getChannelIdLong());
+    }
+
+    /**
+     * Returns the guild id for this reference.
+     * @return The guild id
+     */
+    @Nonnull
+    public String getGuildId() {
```
MinnDevelopment, line 207:
```suggestion
    public String getGuildId()
    {
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,234 @@
...
+     *
+     * @return The corresponding JDA instance
+     */
+    @Nonnull
+    public JDA getJDA()
+    {
+        return api;
+    }
+
+    private void checkPermission(Permission permission) {
```
MinnDevelopment, line 222:
```suggestion
    private void checkPermission(Permission permission)
    {
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,234 @@
...
+        if (guild == null)
+        {
+            return;
+        }
```
MinnDevelopment, line 226:
```suggestion
            return;
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,234 @@
...
+
+        Member selfMember = guild.getSelfMember();
+        if (!selfMember.hasPermission(permission))
+        {
+            throw new InsufficientPermissionException(guild, permission);
+        }
```
MinnDevelopment, line 232:
```suggestion
            throw new InsufficientPermissionException((GuildChannel) channel, permission);
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    private void checkPermission(Permission permission) {
+        if (guild == null)
+        {
+            return;
+        }
+
+        Member selfMember = guild.getSelfMember();
+        if (!selfMember.hasPermission(permission))
```
MinnDevelopment, line 229: Check for `hasAccess` explicitly and throw a `MissingAccessException`

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,234 @@
...
+    public RestAction<Message> resolve()
+    {
+        checkPermission(Permission.MESSAGE_READ);
+        checkPermission(Permission.MESSAGE_HISTORY);
+
+        Route.CompiledRoute route = Route.Messages.GET_MESSAGE.compile(getChannelId(), getMessageId());
+        JDAImpl jda = (JDAImpl) getJDA();
+
+        return new RestActionImpl<>(jda, route,
+                (response, request) -> jda.getEntityBuilder().createMessage(response.getObject(), getChannel(), false));
```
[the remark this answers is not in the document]
arynxd: should this have a boolean overload for updating the referenced message?

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
arynxd: Not when sent in DMs. The docs also state that message_id is not present on [Channel Follow Add](https://discord.com/developers/docs/resources/channel#message-types-channel-follow-add-messages).

 How should JDA handle the getters in this case? Would nullable Long be an acceptable solution?

MinnDevelopment: I would return 0 and explicitly document this behavior

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+        checkPermission(Permission.MESSAGE_READ);
+        checkPermission(Permission.MESSAGE_HISTORY);
+
+        JDAImpl jda = (JDAImpl) getJDA();
+        Message referenced = getMessage();
+
+        if (referenced != null)
+        {
+            return new CompletedRestAction<>(jda, referenced);
+        }
```
MinnDevelopment, line 112:
```suggestion
            return new CompletedRestAction<>(jda, referenced);
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+        Route.CompiledRoute route = Route.Messages.GET_MESSAGE.compile(getChannelId(), getMessageId());
+
+
+        return new RestActionImpl<>(jda, route,
+                (response, request) ->
+                {
+                    Message created = jda.getEntityBuilder().createMessage(response.getObject(), getChannel(), false);
+                    this.referencedMessage = created;
+                    return created;
+                });
```
MinnDevelopment, line 123:
```suggestion
        return new RestActionImpl<>(jda, route, (response, request) ->{
            Message created = jda.getEntityBuilder().createMessage(response.getObject(), getChannel(), false);
            this.referencedMessage = created;
            return created;
        });
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    /**
+     * Referenced message.
```
MinnDevelopment, line 127:
```suggestion
     * The resolved message, if available.
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+     *
+     * <p>This will have different meaning depending on the {@link Message#getType() type} of message.
+     * Usually, this is a {@link MessageType#INLINE_REPLY INLINE_REPLY} reference.
+     * This can be null even if the type is {@link MessageType#INLINE_REPLY INLINE_REPLY}, when the message it references doesn't exist or discord wasn't able to resolve it in time.
+     * @see #resolve()
+     *
+     * @return The referenced message, or null
```
MinnDevelopment, line 134:
```suggestion
     * @return The referenced message, or null if this is not available
     *
     * @see    #resolve()
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+    public Message getMessage()
+    {
+        return referencedMessage;
+    }
+
+    /**
+     * The channel from which this message originates.
+     * <br>Messages from other guilds can be referenced, in which case JDA may not have the channel cached.
+     *
+     * @return The origin channel for this message reference.
```
MinnDevelopment, line 146:
```suggestion
     * @return The origin channel for this message reference, or null if this is not available
     *
     * @see    #getChannelId()
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+        return channel;
+    }
+
+
+    /**
+     * The guild for this reference.
+     * <br>
+     * This will be null if the message did not come from a guild, or JDA did not have the guild cached
+     *
+     * @return The guild, or null.
```
MinnDevelopment, line 160:
```suggestion
     * @return The guild, or null if this is not available
     *
     * @see    #getGuildId()
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+    public MessageChannel getChannel()
+    {
+        return channel;
+    }
+
+
+    /**
+     * The guild for this reference.
+     * <br>
+     * This will be null if the message did not come from a guild, or JDA did not have the guild cached
```
MinnDevelopment, line 158:
```suggestion
     * <br>This will be null if the message did not come from a guild, or JDA did not have the guild cached
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+    private void checkPermission(Permission permission)
+    {
+        if (guild == null) return;
+
+        Member selfMember = guild.getSelfMember();
+        GuildChannel guildChannel = (GuildChannel) channel;
+
+        if (!selfMember.hasPermission(guildChannel, Permission.VIEW_CHANNEL))
+            throw new MissingAccessException(guildChannel, Permission.VIEW_CHANNEL);
+
```
MinnDevelopment, line 251:
```suggestion
```

src/main/java/net/dv8tion/jda/internal/entities/AbstractMessage.java
```diff
@@ -16,6 +16,8 @@
 
 package net.dv8tion.jda.internal.entities;
 
+import javax.annotation.Nullable;
```
MinnDevelopment, line 19: Move this down to the other imports for javax.annotation

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -129,10 +129,19 @@ public JDA getJDA()
...
-    public Message getReferencedMessage()
+    public MessageReference getMessageReference()
     {
-        return referencedMessage;
+        return messageReference;
+    }
+
+    @Nullable
+    @Override
+    public Message getReferencedMessage() {
```
MinnDevelopment, line 141: Your brace is on the wrong line, also make this a default implementation in the `Message` interface

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+public class MessageReference
+{
+    private final long messageId;
+    private final long channelId;
+    private final long guildId;
+
+    private final MessageChannel channel;
+    private final Guild guild;
+    private Message referencedMessage;
+    private final JDA api;
```
MinnDevelopment, line 44:
```suggestion
    private final JDA api;
    private final MessageChannel channel;
    private final Guild guild;
    private Message referencedMessage;
```

MinnDevelopment (changes requested):

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
arynxd: is line 117 not supposed to have a space after the ->?

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
arynxd: Is this not a breaking change?

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
MinnDevelopment: Yeah, fixed.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
MinnDevelopment: Not really? Unless you use internals, then it could potentially break invokevirtual but usually not.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -214,6 +214,17 @@
             "(?:\\?\\S*)?(?:#\\S*)?",                                      // Useless query or URN appendix
             Pattern.CASE_INSENSITIVE);
 
+    /**
+     * Returns the {@link MessageReference} for this Message. This will be null if this Message has no reference.
+     *
+     * You can access all the information about a reference through this object.
```
MinnDevelopment, line 220:
```suggestion
     * <p>You can access all the information about a reference through this object.
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -214,6 +214,17 @@
...
+     * Additionally you can retrieve the referenced Message if discord did not load it in time.
```
MinnDevelopment, line 221:
```suggestion
     * Additionally, you can retrieve the referenced Message if discord did not load it in time. This can be done with {@link MessageReference#resolve()}.
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -224,7 +235,12 @@
      * @return The referenced message, or null
      */
     @Nullable
-    Message getReferencedMessage();
+    default Message getReferencedMessage()
```
MinnDevelopment, line 238: Add a reference to getMessageReference in these docs to make it clear what the difference is and why it exists

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+
+        if (tc == null)
+        {
+            this.channel = api.getPrivateChannelById(channelId);
+            this.guild = null;
+        }
+        else
+        {
+            this.channel = tc;
+            this.guild = api.getGuildById(guildId);
```
MinnDevelopment, line 63:
```suggestion
            this.guild = tc.getGuild();
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+        this.messageId = messageId;
+        this.channelId = channelId;
+        this.guildId = guildId;
+        this.referencedMessage = referencedMessage;
+
+        TextChannel tc = api.getTextChannelById(channelId);
+
+        if (tc == null)
+        {
+            this.channel = api.getPrivateChannelById(channelId);
```
MinnDevelopment, line 57: It might be worth checking if the `referencedMessage` has a channel attached and using that directly if available

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+
+        JDAImpl jda = (JDAImpl) getJDA();
+        Message referenced = getMessage();
+
+        if (referenced != null)
+            return new CompletedRestAction<>(jda, referenced);
+
+        Route.CompiledRoute route = Route.Messages.GET_MESSAGE.compile(getChannelId(), getMessageId());
+
+
```
MinnDevelopment, line 114:
```suggestion
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,255 @@
...
+        return new RestActionImpl<>(jda, route, (response, request) -> {
+            Message created = jda.getEntityBuilder().createMessage(response.getObject(), getChannel(), false);
```
MinnDevelopment, line 116: This will most likely fail if the channel is not cached btw

src/main/java/net/dv8tion/jda/internal/entities/AbstractMessage.java
```diff
@@ -27,6 +27,8 @@
 import org.apache.commons.collections4.Bag;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
```
MinnDevelopment, line 31:
```suggestion
```

MinnDevelopment (changes requested):

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
arynxd: for what reason? channel is marked as nullable for createMessage.

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,234 @@
...
+     *          If this reference refers to a {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} and the logged in account does not have
+     *         <ul>
+     *             <li>{@link net.dv8tion.jda.api.Permission#MESSAGE_READ Permission.MESSAGE_READ}</li>
+     *             <li>{@link net.dv8tion.jda.api.Permission#MESSAGE_HISTORY Permission.MESSAGE_HISTORY}</li>
+     *         </ul>
+     *
+     * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.Message}
+     */
+    @Nonnull
+    public RestAction<Message> resolve()
```
arynxd, line 101: Should this have a boolean overload to force an update?

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
MinnDevelopment: Because `Message#getChannel` cannot be null.

https://github.com/DV8FromTheWorld/JDA/blob/b9d1a0e35180bce2516d492b748d1656648d0434/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java#L1156-L1164

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
MinnDevelopment: Could be added, sure.

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
arynxd: Ah I see, so what is the proposed solution? Since then we wouldn't be able to resolve messages from other guilds, which is a pretty big deal.

MinnDevelopment: For now, we will probably just have to throw if `getChannel()` is null

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -214,17 +214,36 @@
...
     /**
      * Referenced message.
      *
      * <p>This will have different meaning depending on the {@link #getType() type} of message.
      * Usually, this is a {@link MessageType#INLINE_REPLY INLINE_REPLY} reference.
      * This can be null even if the type is {@link MessageType#INLINE_REPLY INLINE_REPLY}, when the message it references doesn't exist or discord wasn't able to resolve it in time.
      *
+     * <p>This differs from a {@link MessageReference}, which contains the raw IDs attached to the reference, and allows you to retrieve the referenced message
+     *
+     * @see #getMessageReference()
```
MinnDevelopment, line 237: This should be below return separated by an empty line

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,303 @@
...
+        this.messageId = messageId;
+        this.channelId = channelId;
+        this.guildId = guildId;
+        this.referencedMessage = referencedMessage;
+
+        TextChannel tc = referencedMessage != null
+                ? referencedMessage.getTextChannel()
+                : api.getTextChannelById(channelId);
+
+        if (tc == null)
```
MinnDevelopment, line 57: To prepare for threads we should do:

```java
if (guildId == 0L)
    this.channel = api.getPrivateChannelById(channelId);
else
    this.channel = (MessageChannel) api.getGuildChannelById(channelId);
this.guild = api.getGuildById(guild); // is null if guildId = 0 anyway
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,303 @@
...
+     */
+    @Nonnull
+    public JDA getJDA()
+    {
+        return api;
+    }
+
+    private void checkPermission(Permission permission)
+    {
+        if (guild == null) return;
```
MinnDevelopment, line 293:
```suggestion
        if (guild == null || !(channel instanceof GuildChannel)) return;
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,303 @@
...
+
+        Member selfMember = guild.getSelfMember();
+        GuildChannel guildChannel = (GuildChannel) channel;
+
+        if (!selfMember.hasPermission(guildChannel, Permission.VIEW_CHANNEL))
```
MinnDevelopment, line 298:
```suggestion
        if (!selfMember.hasAccess(guildChannel))
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,303 @@
...
+        });
+    }
+
+    /**
+     * The resolved message, if available.
+     *
+     * <p>This will have different meaning depending on the {@link Message#getType() type} of message.
+     * Usually, this is a {@link MessageType#INLINE_REPLY INLINE_REPLY} reference.
+     * This can be null even if the type is {@link MessageType#INLINE_REPLY INLINE_REPLY}, when the message it references doesn't exist or discord wasn't able to resolve it in time.
+     * @see #resolve()
```
MinnDevelopment, line 176:
```suggestion
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,303 @@
...
+     *         in the {@link net.dv8tion.jda.api.entities.TextChannel TextChannel}.</li>
+     *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
+     *     <br>The message has already been deleted.</li>
+     *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_CHANNEL UNKNOWN_CHANNEL}
+     *     <br>The request was attempted after the channel was deleted.</li>
+     * </ul>
+     *
+     * @param update Whether to update the already stored message
```
MinnDevelopment, line 133:
```suggestion
     * @param  update
     *         Whether to update the already stored message
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,303 @@
...
+     *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
+     *     <br>The message has already been deleted.</li>
+     *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_CHANNEL UNKNOWN_CHANNEL}
+     *     <br>The request was attempted after the channel was deleted.</li>
+     * </ul>
+     *
+     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
+     *          If this reference refers to a {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} and the logged in account does not have
```
MinnDevelopment, line 94:
```suggestion
     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
     *         If this reference refers to a {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} and the logged in account does not have
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,303 @@
...
+            this.channel = tc;
+            this.guild = tc.getGuild();
+        }
+
+        this.api = api;
+    }
+
+    /**
+     * Retrieves the referenced message for this message.
+     * If the message already exists, it will be returned immediately.
```
MinnDevelopment, line 73:
```suggestion
     * <br>If the message already exists, it will be returned immediately.
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,303 @@
...
+     */
+    @Nonnull
+    public RestAction<Message> resolve()
+    {
+        return resolve(true);
+    }
+
+    /**
+     * Retrieves the referenced message for this message.
+     * If the message already exists, it will be returned immediately.
```
MinnDevelopment, line 113:
```suggestion
     * <br>If the message already exists, it will be returned immediately.
```

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
```diff
@@ -0,0 +1,303 @@
...
+     *     <br>The message has already been deleted.</li>
+     *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_CHANNEL UNKNOWN_CHANNEL}
+     *     <br>The request was attempted after the channel was deleted.</li>
+     * </ul>
+     *
+     * @param update Whether to update the already stored message
+     *
+     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
+     *          If this reference refers to a {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} and the logged in account does not have
```
MinnDevelopment, line 136:
```suggestion
     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
     *         If this reference refers to a {@link net.dv8tion.jda.api.entities.TextChannel TextChannel} and the logged in account does not have
```

MinnDevelopment (changes requested):
Almost there

src/main/java/net/dv8tion/jda/api/entities/MessageReference.java
arynxd: I assume we just trust that the API gave us the ID for a message channel? Is it worth instaneof checking it

MinnDevelopment: Well messages should only ever be inside message channels, if discord changes this that would break lots of things anyway.

MinnDevelopment (approved):

DV8FromTheWorld (approved):

Issue #1688, opened 2021-06-22
Double checking my results on a possible issue concerning #updateButton.


puttydotexe:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[wiki]: https://github.com/DV8FromTheWorld/JDA/wiki
[FAQ]: https://github.com/DV8FromTheWorld/JDA/wiki/10%29-FAQ

## General Troubleshooting

- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download]
- [x] I have checked the [wiki] and especially the [FAQ] section for similar questions.

## Question

Just trying to make sure that my finds are correct.

Whilst trying to edit a `ReplyAction` using `ButtonClickEvent#getMessage#editButton`, if called twice it fails to function correctly.

It can be shown [here ](https://i.gyazo.com/2b9e5480961501b9e32778ebbee3ba89.gif) in a graphical demonstration.

## Example Code

```java
event.editButton(Button.primary("BUG_1", "Bug").asDisabled()).complete();
event.editButton(Button.danger("BUG_2", "Bug").asDisabled()).complete();
```

I looked into the `#editButton` method being called and it apears it doesn't crossreference the existing `Button` and the new `Button` correctly.

The code modifications I used to `ButtonInteraction#editButton` along with results can be found below.

### Code Changes
```diff

  /**
     * Update the button with a new button instance.
     * <br>This only works for non-ephemeral messages where {@link #getMessage()} is available!
     *
     * <p>If this interaction is already acknowledged this will use {@link #getHook()}
     * and otherwise {@link #editComponents(Collection)} directly to acknowledge the interaction.
     *
     * @param  newButton
     *         The new button to use, or null to remove this button from the message entirely
     *
     * @throws IllegalStateException
     *         If this interaction was triggered by a button on an ephemeral message.
     *
     * @return {@link RestAction}
     */
    @Nonnull
    @CheckReturnValue
    default RestAction<Void> editButton(@Nullable Button newButton)
    {
        Message message = getMessage();
        if (message == null)
            throw new IllegalStateException("Cannot update button for ephemeral messages! Discord does not provide enough information to perform the update.");
        List<ActionRow> components = new ArrayList<>(message.getActionRows());
        String id = getComponentId();
        find: for (Iterator<ActionRow> rows = components.iterator(); rows.hasNext();)
        {
[...]

PR #1689, opened 2021-06-22
Fix `ButtonInteraction#editButton` not checking id's before edit.

puttydotexe:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1688

## Description

Changes the logic used to crossreference the old `Button` id before editing to new `Button`.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/components/ButtonInteraction.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/components/ButtonInteraction.java
@@ -79,7 +79,7 @@ public interface ButtonInteraction extends ComponentInteraction
             for (ListIterator<Component> it = row.listIterator(); it.hasNext();)
             {
                 Component component = it.next();
-                if (id.equals(component.getId()))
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/interactions/components/ButtonInteraction.java
```diff
@@ -79,7 +79,7 @@ default Button getComponent()
             for (ListIterator<Component> it = row.listIterator(); it.hasNext();)
             {
                 Component component = it.next();
-                if (id.equals(component.getId()))
+                if (component.getId().equals(newButton.getId()))
```
MinnDevelopment, line 82: This is actually wrong. This method intentionally does NOT check the id of newButton because it is valid to change the button id.

MinnDevelopment:
You misunderstood what `editButton` does. This method specifically updates the exact button that triggered the interaction, this is why it finds the button within the message which corresponds with the component id for the event. Your change breaks this logic.

PR #1691, opened 2021-06-26
Fix isEmpty checks for embeds

arynxd:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: Builder code

Closes Issue: NaN

## Description
This PR fixes some oversights for embed validation. Discord disallows embeds with no content (a space character or similar) and JDA didn't check this before sending the embed.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -175,7 +175,7 @@ public class EmbedBuilder
      */
     public int length()
     {
-        int length = description.length();
+        int length = description.toString().trim().length();
         synchronized (fields)
         {
             length = fields.stream().map(f -> f.getName().length() + f.getValue().length()).reduce(length, Integer::sum);
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -156,14 +156,14 @@ public class EmbedBuilder
      */
     public boolean isEmpty()
     {
-        return title == null
+        return (title == null || title.isEmpty())
             && timestamp == null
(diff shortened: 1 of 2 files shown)


Chew:
Shouldn't String#isBlank() be used instead which "Returns true if the string is empty or contains only white space codepoints, otherwise false."? Discord wouldn't accept white space characters

arynxd:
Just changed to the isBlank helper, thanks for the suggestion

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -156,14 +156,14 @@ public EmbedBuilder clear()
      */
     public boolean isEmpty()
     {
-        return title == null
+        return Helpers.isBlank(title)
```
MinnDevelopment, line 159:
```suggestion
        return (title == null || title.trim().isEmpty())
```

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -156,14 +156,14 @@ public EmbedBuilder clear()
...
             && timestamp == null
             && thumbnail == null
             && author == null
             && footer == null
             && image == null
             && color == Role.DEFAULT_COLOR_RAW
-            && description.length() == 0
+            && Helpers.isBlank(description)
```
MinnDevelopment, line 166:
```suggestion
            && (description == null || description.toString().trim().isEmpty())
```

MinnDevelopment (commented):
I think you also should update `EmbedBuilder#getLength` and `MessageEmbed#getLength` to properly check the trimmed length

arynxd (commented):
Apply changes

arynxd:
I broke my fork, so will re-open this PR

PR #1692, opened 2021-06-27
fix parsing stickers

caneleex:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this PR fixes parsing of stickers because discord likes to do breaking changes in prod. there are more changes but this PR is just to stop receiving errors

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
@@ -1463,7 +1463,7 @@ public class EntityBuilder
     {
         final long id = content.getLong("id");
         final String name = content.getString("name");
-        final String description = content.getString("description", null);
+        final String description = content.getString("description", "");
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1463,7 +1463,7 @@ public MessageSticker createSticker(DataObject content)
     {
         final long id = content.getLong("id");
         final String name = content.getString("name");
-        final String description = content.getString("description");
+        final String description = content.getString("description", null);
```
MinnDevelopment, line 1466: Could we make this backwards compatible for now by defaulting to an empty string?

caneleex: yes

PR #1698, opened 2021-06-29
Replace isEmpty check with isBlank in StageInstanceActionImpl#setTopic

LenderSchlender:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

When testing the latest JDA version (`4.3.0_286`), I noticed that when setting `EmbedBuilder#ZERO_WIDTH_SPACE` as the `StageInstanceAction` topic, I would only get an error when sending, not before.
This PR should fix this.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/restaction/StageInstanceActionImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/restaction/StageInstanceActionImpl.java
@@ -69,7 +69,7 @@ public class StageInstanceActionImpl extends RestActionImpl<StageInstance> imple
     @Override
     public StageInstanceAction setTopic(@Nonnull String topic)
     {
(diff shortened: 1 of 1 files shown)


arynxd:
> This PR should fix this.
have you tested it

LenderSchlender:
> > This PR should fix this.
> 
> have you tested it

I didn't test it at the time of creating this PR, but I tested it just now.
Apparently, Zero width space, and a few other invisible with no width characters don't count as whitespace (6 in total), so those still aren't checked, but it's still better than the old solution, which only checked if the String was empty.
There is probably still a better solution than this, but I don't know which one.
Maybe this should also be changed for `ChannelActionImpl#setName()`?

PR #1699, opened 2021-06-30
Braces at end of line instead of new line

DeveloperTK:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: Code-Style

Closes Issue: NaN

## Description

This is just something I came across when working with JDA. I changed the `.editorconfig` to put braces at the end of the line instead of a new line. I don't know if you would want to even review a huge change like that but It's something to think about since almost all open source java projects that I came across use the end of line brace style.


MinnDevelopment:
No.

DManstrator:
Code Style changes are very unlikely to get accepted, especially if they are changing that many files. I also prefer the style you chose but in the end it's the decision of the project maintainers which code style should be used (and enforced).

PR #1701, opened 2021-07-01
Change checks for emote name length to check minimum length

LenderSchlender:
Add Checks#notShorter(), and check for the name length when creating/updating emotes.

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds a `Checks#notShorter()` method, and adds checks for the name length when creating/updating emotes.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
@@ -1654,6 +1654,8 @@ public class GuildImpl implements Guild
     {
         checkPermission(Permission.MANAGE_EMOTES);
         Checks.notBlank(name, "Emote name");
+        Checks.notShorter(name, 2, "Emote name");
+        Checks.notLonger(name, 32, "Emote name");
         Checks.notNull(icon, "Emote icon");
         Checks.notNull(roles, "Roles");
 
--- a/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
@@ -1653,9 +1653,7 @@ public class GuildImpl implements Guild
     public AuditableRestAction<Emote> createEmote(@Nonnull String name, @Nonnull Icon icon, @Nonnull Role... roles)
     {
         checkPermission(Permission.MANAGE_EMOTES);
-        Checks.notBlank(name, "Emote name");
-        Checks.notShorter(name, 2, "Emote name");
-        Checks.notLonger(name, 32, "Emote name");
+        Checks.inRange(name, 2, 32, "Emote name");
         Checks.notNull(icon, "Emote icon");
         Checks.notNull(roles, "Roles");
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/internal/managers/EmoteManagerImpl.java
```diff
@@ -112,6 +112,7 @@ public EmoteManagerImpl setName(@Nonnull String name)
         Checks.notBlank(name, "Name");
         name = name.trim();
         Checks.notEmpty(name, "Name");
+        Checks.notShorter(name, 2, "Name");
```
MinnDevelopment, line 115: Doesn't make much sense to check empty and min length

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1654,6 +1654,8 @@ public RoleAction createRole()
     {
         checkPermission(Permission.MANAGE_EMOTES);
         Checks.notBlank(name, "Emote name");
+        Checks.notShorter(name, 2, "Emote name");
```
Andre601, line 1657: Wouldn't it be more reasonable to use `Checks.check` here instead of inventing a new tool for two cases?

Or alternatively, have a `Checks.length` method that checks a min and max length (Maybe with -1 making them optional?)?

MinnDevelopment: Better would be `Checks.inRange(string, min, max, ...)` maybe but using just the raw check is less efficient and would require writing the same error string on each check, which is undesirable.

src/main/java/net/dv8tion/jda/internal/managers/EmoteManagerImpl.java
LenderSchlender: Doesn't make sense for me either, I didn't really look at the existing checks, which was probably a bad idea.

src/main/java/net/dv8tion/jda/internal/utils/Checks.java
```diff
@@ -146,6 +146,15 @@ public static void noneNull(final Object[] argument, final String name)
         argument.forEach(it -> noWhitespace(it, name));
     }
 
+    public static void inRange(final String input, final int min, final int max, final String name)
+    {
+        notNull(input, name);
+        int length = Helpers.codePointLength(input);
+        check(length >= min && (max < 1 || length <= max),
+                "%s is not in bounds! A minimum length of %d characters, and a maximum length of %d characters is required! Provided: \"%s\"",
```
MinnDevelopment, line 154:
```suggestion
                "%s must be between %d and %d characters long! Provided: \"%s\"",
```

src/main/java/net/dv8tion/jda/internal/utils/Checks.java
```diff
@@ -146,6 +146,15 @@ public static void noneNull(final Object[] argument, final String name)
         argument.forEach(it -> noWhitespace(it, name));
     }
 
+    public static void inRange(final String input, final int min, final int max, final String name)
+    {
+        notNull(input, name);
+        int length = Helpers.codePointLength(input);
+        check(length >= min && (max < 1 || length <= max),
```
MinnDevelopment, line 153:
```suggestion
        check(min <= length && length <= max,
```

Issue #1702, opened 2021-07-02
Discord bumped the embed description char count to 4096


rakosi2:
## General Troubleshooting

- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request

Discord just bumped the embed description char count to 4096 as can be seen here: https://discord.com/developers/docs/resources/channel#embed-limits

rakosi2:
Closing because it was added in 4.3.0_294

PR #1703, opened 2021-07-02
Bump embed max description length

Andre601:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1702

## Description

Increases the `MessageEmbed.TEXT_MAX_LENGTH` to `4096` and also adds new value called `FOOTER_MAX_LENGTH` since the old Javadoc mentions the text value being for description AND footer.

Not sure if a new value shouldn't be added for the description while deprecating the text one for those who are using this int value for both footer and description...

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -690,7 +690,7 @@ public class EmbedBuilder
         }
         else
         {
-            Checks.check(text.length() <= MessageEmbed.TEXT_MAX_LENGTH, "Text cannot be longer than %d characters.", MessageEmbed.TEXT_MAX_LENGTH);
+            Checks.check(text.length() <= MessageEmbed.FOOTER_MAX_LENGTH, "Text cannot be longer than %d characters.", MessageEmbed.FOOTER_MAX_LENGTH);
             urlCheck(iconUrl);
             this.footer = new MessageEmbed.Footer(text, iconUrl, null);
         }
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -117,8 +117,8 @@ public class EmbedBuilder
     {
         if (isEmpty())
             throw new IllegalStateException("Cannot build an empty embed!");
-        if (description.length() > MessageEmbed.TEXT_MAX_LENGTH)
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -690,7 +690,7 @@ public EmbedBuilder setFooter(@Nullable String text, @Nullable String iconUrl)
         }
         else
         {
-            Checks.check(text.length() <= MessageEmbed.TEXT_MAX_LENGTH, "Text cannot be longer than %d characters.", MessageEmbed.TEXT_MAX_LENGTH);
+            Checks.check(text.length() <= MessageEmbed.FOOTER_MAX_LENGTH, "Text cannot be longer than %d characters.", MessageEmbed.FOOTER_MAX_LENGTH);
```
Chew, line 693: Should this also be "Footer cannot be longer than %d characters"?

Andre601: I wondered about this too... I assume it's honestly a copy-paste, but I won't just change this text unless told so.

MinnDevelopment: The parameter is called text

src/main/java/net/dv8tion/jda/api/entities/MessageEmbed.java
```diff
@@ -66,12 +66,18 @@
...
 
     /**
-     * The maximum length the description and footer of an embed can have
+     * The maximum length the description of an embed can have
      *
-     * @see net.dv8tion.jda.api.EmbedBuilder#setFooter(String, String) EmbedBuilder.setFooter(text, iconUrl)
      * @see net.dv8tion.jda.api.EmbedBuilder#setDescription(CharSequence) EmbedBuilder.setDescription(text)
      */
-    public static final int TEXT_MAX_LENGTH = 2048;
+    public static final int TEXT_MAX_LENGTH = 4096;
```
MinnDevelopment, line 73: I feel like it would be better to make a DESCRIPTION_MAX_LENGTH and keep text as the footer maximum. That would also be less breaking

Andre601: Done

MinnDevelopment (approved):

PR #1730, opened 2021-07-15
Fixed a parse error on empty selection menu events

fratik:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Discord doesn't send a `values` field in the `data` object when a user unchecks all options in the Selection Menu, causing an "unexpected Json-parse error"
```
[WARN] [JDA [0 / 1] MainWS-ReadThread] Got an unexpected Json-parse error. Please redirect following message to the devs:
        Unable to resolve value with key values to type DataArray: null
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/interactions/SelectionMenuInteractionImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/interactions/SelectionMenuInteractionImpl.java
@@ -37,16 +37,9 @@ public class SelectionMenuInteractionImpl extends ComponentInteractionImpl imple
     public SelectionMenuInteractionImpl(JDAImpl jda, DataObject data)
     {
         super(jda, data);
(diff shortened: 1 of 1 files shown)


Andre601:
Wouldn't it be a better solution to utilize `optArray` and then check for if it is empty or not?
https://github.com/DV8FromTheWorld/JDA/blob/c9f76f359db25c50b137865a72de7cc809bb3f99/src/main/java/net/dv8tion/jda/api/utils/data/DataObject.java#L335-L349

fratik:
Like this?

MinnDevelopment:
This is actually a discord bug, I reported it, and it should be fixed soon.

fratik:
It works now, so I guess they fixed it

Issue #1731, opened 2021-07-16
Message#getStickers always returns empty list


duncte123:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

With the introduction of the new `sticker_items` field on messages the old `stickers` field was removed, causing stickers to not get parsed anymore

### Expected Behavior

Message#getStickers to not be empty when attaching a sticker from a guild

### Code Example or Reproduction Steps

N/A

### Code for JDABuilder or DefaultShardManagerBuilder Used

N/A

### Exception or Error

N/A

MinnDevelopment:
This is fixed in 4.3.0_297

PR #1732, opened 2021-07-16
Fix outdated sticker field name

Andre601:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1731 

## Description

Updated `stickers` to `sticker_items` in the EntityBuilder

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -914,6 +914,7 @@ public interface Message extends ISnowflake, Formattable
 
     /**
      * All {@link net.dv8tion.jda.api.entities.MessageSticker MessageStickers} that are in this Message.
+     * <br>The MessageStickers returns may only contain necessary information such as id, format type and name.
      *
      * @throws java.lang.UnsupportedOperationException
      *         If this is not a Received Message from {@link net.dv8tion.jda.api.entities.MessageType#DEFAULT MessageType.DEFAULT}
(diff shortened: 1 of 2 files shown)


MinnDevelopment:
This needs to adjust the documentation on `Message#getStickers` to state that these stickers are partial, and limited to just the necessary information to render it. This means you only get the id, format type, and name.

----

Unrelated to this pull request:

I noticed that, when sending a message, you still receive the `stickers` array in addition to `sticker_items`. Although, this is deprecated so we can probably just ignore it.

Another thing, the guild object might also contain rich sticker objects, which we could potentially add support for if discord would document it. There also is a `GET/stickers/:sticker_id/guild` endpoint that seems to be undocumented as well.

Andre601:
> This needs to adjust the documentation on `Message#getStickers` to state that these stickers are partial, and limited to just the necessary information to render it. This means you only get the id, format type, and name.

Adressed this.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -914,6 +914,7 @@ default Button getButtonById(@Nonnull String id)
 
     /**
      * All {@link net.dv8tion.jda.api.entities.MessageSticker MessageStickers} that are in this Message.
+     * <br>The MessageStickers returns may only contain necessary information such as id, format type and name.
```
MinnDevelopment, line 917:
```suggestion
     * <br>The returned MessageStickers may only contain necessary information such as the sticker id, format type, name, and icon url.
```

MinnDevelopment (changes requested):

PR #1738, opened 2021-07-20
Add copyFrom to EmbedBuilder

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This enables users to replace an EmbedBuilder's content with another embed or another builder. 
This may solve situations such as when you need to fill a given EmbedBuilder but already have an embed ready, and replacing the old EmbedBuilder's variable by the new one is not possible.

Also adds a missing constructor

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -137,7 +137,7 @@ public class EmbedBuilder
      * @param  builder
      *         the existing builder
      */
-    public void reset(EmbedBuilder builder)
+    public void reset(@Nullable EmbedBuilder builder)
     {
         if (builder != null)
         {
@@ -161,7 +161,7 @@ public class EmbedBuilder
      * @param  embed
      *         the existing embed
      */
-    public void reset(MessageEmbed embed)
+    public void reset(@Nullable MessageEmbed embed)
     {
         if(embed != null)
         {
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -62,6 +62,12 @@ public class EmbedBuilder
      */
     public EmbedBuilder() { }
 
+    /**
+     * Creates an EmbedBuilder using fields from an existing builder
+     *
+     * @param  builder
+     *         the existing builder
+     */
     public EmbedBuilder(@Nullable EmbedBuilder builder)
     {
         reset(builder);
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -149,6 +130,55 @@ public EmbedBuilder clear()
...
     }
 
+    /**
+     * Resets this builder to the state of the given builder.
+     * <br>All the parts of the given builder will be applied to this one.
+     *
+     * @param  builder
+     *         the existing builder
+     */
+    public void reset(EmbedBuilder builder)
```
anweisen, line 140: Maybe add Nullable annotation for consistency

I also think that reset is not the right method name, as it gives the impression that it resets all the builders data back to nothing, but this is not case because it applies the data of the given builder to this one

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -149,6 +130,55 @@ public EmbedBuilder clear()
...
+    }
+
+    /**
+     * Resets this builder to the state of the given embed.
+     * <br>All the parts of the given embed will be applied to this builder.
+     *
+     * @param  embed
+     *         the existing embed
+     */
+    public void reset(MessageEmbed embed)
```
anweisen, line 164: Same as above

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -149,6 +130,56 @@ public EmbedBuilder clear()
...
+            this.color = builder.color;
+            this.thumbnail = builder.thumbnail;
+            this.author = builder.author;
+            this.footer = builder.footer;
+            this.image = builder.image;
+        }
+    }
+
+    /**
+     * Resets this builder to the state of the given embed.
```
DV8FromTheWorld, line 159: Please update this documentation as it isn't _resetting_ anything.

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -149,6 +130,56 @@ public EmbedBuilder clear()
         return this;
     }
 
+    /**
+     * Resets this builder to the state of the given builder.
```
DV8FromTheWorld, line 134: Please update this documentation as it isn't _resetting_ anything.

DV8FromTheWorld (changes requested):

DV8FromTheWorld (approved):

Issue #1744, opened 2021-07-27
Add support for new NUMBER slash command option type


nikammerlaan:
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request

Add support for the new NUMBER option type for slash commands.

### Example Use-Case

`/example input:1.5`

MinnDevelopment:
This is supported as of [4.3.0_310](https://ci.dv8tion.net/job/JDA/310/)

PR #1745, opened 2021-07-28
Add NUMBER support

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1744 

## Description

Adds support for the `NUMBER` (type 10) option for Slashcommands.
This also adds a `DataObject#getAsDouble(String)` method. Didn't add one for default values yet.

If I [read the docs correctly](https://discord.com/developers/docs/interactions/slash-commands#application-command-object-application-command-option-choice-structure) is number one option that allows choices so I set the boolean for it.

I am not sure if I got all the places necessary updated, so please inform me if there are other ones missing.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
@@ -611,7 +611,7 @@ public class Command implements ISnowflake
         }
 
         /**
-         * The options for this subcommand, or the subcommands whtin this group.
+         * The options for this subcommand, or the subcommands within this group.
          *
          * @return Immutable list of Options
          */
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
@@ -437,7 +437,7 @@ public class Command implements ISnowflake
         /**
          * The value of this choice.
          * 
-         * @return The double value
+         * @return The double value, or NaN if this is not a numeric choice value
          */
         public double getAsDouble()
         {
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
@@ -347,9 +347,9 @@ public class Command implements ISnowflake
     public static class Choice
     {
         private final String name;
-        private final long intValue;
-        private final double doubleValue;
-        private final String stringValue;
+        private long intValue = 0;
+        private double doubleValue = 0.0;
+        private String stringValue = null;
 
         /**
          * Create a Choice tuple
@@ -362,9 +362,7 @@ public class Command implements ISnowflake
         public Choice(@Nonnull String name, long value)
         {
             this.name = name;
-            this.intValue = value;
-            this.stringValue = Long.toString(value);
-            this.doubleValue = 0.0d;
+            setIntValue(value);
         }
 
         /**
@@ -378,9 +376,7 @@ public class Command implements ISnowflake
         public Choice(@Nonnull String name, double value)
         {
             this.name = name;
-            this.intValue = 0;
-            this.stringValue = Double.toString(value);
-            this.doubleValue = value;
+            setDoubleValue(value);
         }
 
         /**
@@ -394,9 +390,7 @@ public class Command implements ISnowflake
         public Choice(@Nonnull String name, @Nonnull String value)
         {
             this.name = name;
-            this.intValue = 0;
-            this.stringValue = value;
-            this.doubleValue = 0.0d;
+            setStringValue(value);
         }
 
         /**
@@ -416,21 +410,15 @@ public class Command implements ISnowflake
             this.name = json.getString("name");
             if (json.isType("value", DataType.INT))
             {
-                this.doubleValue = 0.0d;
-                this.intValue = json.getLong("value");
-                this.stringValue = Long.toString(intValue); // does this make sense?
+                setIntValue(json.getLong("value"));
             }
             else if (json.isType("value", DataType.FLOAT))
             {
-                this.doubleValue = json.getDouble("value");
-                this.intValue = 0;
-                this.stringValue = Double.toString(doubleValue);
+                setDoubleValue(json.getDouble("value"));
             }
             else
             {
-                this.doubleValue = 0.0d;
-                this.intValue = 0;
-                this.stringValue = json.getString("value");
+                setStringValue(json.getString("value"));
             }
         }
 
@@ -496,6 +484,27 @@ public class Command implements ISnowflake
         public String toString()
         {
             return "Choice(" + name + "," + stringValue + ")";
+        }
+        
+        private void setIntValue(long value)
+        {
+            this.doubleValue = value;
+            this.intValue = value;
+            this.stringValue = Long.toString(value);
+        }
+        
+        private void setDoubleValue(double value)
+        {
+            this.doubleValue = value;
+            this.intValue = (int) value;
+            this.stringValue = Double.toString(value);
+        }
+        
+        private void setStringValue(@Nonnull String value)
+        {
+            this.doubleValue = Double.NaN;
+            this.intValue = 0;
+            this.stringValue = value;
         }
     }
 
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
@@ -348,6 +348,7 @@ public class Command implements ISnowflake
     {
         private final String name;
         private final long intValue;
+        private final double doubleValue;
         private final String stringValue;
 
         /**
@@ -363,6 +364,23 @@ public class Command implements ISnowflake
             this.name = name;
             this.intValue = value;
             this.stringValue = Long.toString(value);
+            this.doubleValue = 0.0d;
+        }
+
+        /**
+         * Create a Choice tuple
+         * 
+         * @param name
+         *        The display name of this choice
+         * @param value
+         *        The double value you receive in a command option
+         */
+        public Choice(@Nonnull String name, double value)
+        {
+            this.name = name;
+            this.intValue = 0;
+            this.stringValue = Double.toString(value);
+            this.doubleValue = value;
         }
 
         /**
@@ -378,6 +396,7 @@ public class Command implements ISnowflake
             this.name = name;
             this.intValue = 0;
             this.stringValue = value;
+            this.doubleValue = 0.0d;
         }
 
         /**
@@ -397,11 +416,19 @@ public class Command implements ISnowflake
             this.name = json.getString("name");
             if (json.isType("value", DataType.INT))
             {
+                this.doubleValue = 0.0d;
                 this.intValue = json.getLong("value");
                 this.stringValue = Long.toString(intValue); // does this make sense?
             }
+            else if (json.isType("value", DataType.FLOAT))
+            {
+                this.doubleValue = json.getDouble("value");
+                this.intValue = 0;
+                this.stringValue = Double.toString(doubleValue);
+            }
             else
             {
+                this.doubleValue = 0.0d;
                 this.intValue = 0;
                 this.stringValue = json.getString("value");
             }
@@ -419,6 +446,16 @@ public class Command implements ISnowflake
             return name;
         }
 
+        /**
+         * The value of this choice.
+         * 
+         * @return The double value
+         */
+        public double getAsDouble()
+        {
+            return doubleValue;
+        }
+        
         /**
          * The value of this choice.
          *
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
@@ -496,7 +496,7 @@ public class Command implements ISnowflake
         private void setDoubleValue(double value)
         {
             this.doubleValue = value;
-            this.intValue = (int) value;
+            this.intValue = (long) value;
             this.stringValue = Double.toString(value);
         }
         
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
@@ -348,7 +348,7 @@ public class Command implements ISnowflake
     {
         private final String name;
         private long intValue = 0;
-        private double doubleValue = 0.0;
+        private double doubleValue = Double.NaN;
         private String stringValue = null;
 
         /**
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
@@ -124,6 +124,18 @@ public class OptionMapping
                 return data.getLong("value");
         }
     }
+    
+    public double getAsDouble()
+    {
+        switch (type)
+        {
+            default:
+                throw new IllegalStateException("Cannot convert option of type " + type + " to double");
+            case STRING:
+            case NUMBER:
+                return data.getDouble("value");
+        }
+    }
 
     /**
      * The resolved {@link IMentionable} instance for this option value.
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
@@ -132,6 +132,7 @@ public class OptionMapping
             default:
                 throw new IllegalStateException("Cannot convert option of type " + type + " to double");
             case STRING:
+            case INTEGER:
             case NUMBER:
                 return data.getDouble("value");
         }
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
@@ -124,7 +124,17 @@ public class OptionMapping
                 return data.getLong("value");
         }
     }
-    
+
+    /**
+     * The double value for this option.
+     * 
+     * @throws IllegalStateException
+     *         If this option {@link #getType() type} cannot be converted to a double
+     * @throws NumberFormatException
+     *         If this option is of type {@link OptionType#STRING STRING} and could not be parsed to a valid double value
+     * 
+     * @return The double value
+     */
     public double getAsDouble()
     {
         switch (type)
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
@@ -36,12 +36,27 @@ public class OptionData implements SerializableData
     /**
      * The highest positive amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
      */
-    public static final double MAX_POSITIVE_NUMBER = 0x1.0p53;
+    public static final double MAX_POSITIVE_NUMBER = (1L << 55);
     
     /**
      * The smallest negative amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
      */
-    public static final double MIN_NEGATIVE_NUMBER = -0x1.0p53;
+    public static final double MIN_NEGATIVE_NUMBER = -(1L << 55);
+
+    /**
+     * The highest length the name of an option can be.
+     */
+    public static final int MAX_NAME_LENGTH = 32;
+
+    /**
+     * The highest length the description of an option can be.
+     */
+    public static final int MAX_DESCRIPTION_LENGTH = 100;
+
+    /**
+     * The highest length the {@link #addChoice(String, String) String value} of a Choice can be.
+     */
+    public static final int MAX_VALUE_LENGTH = 100;
     
     private final OptionType type;
     private String name, description;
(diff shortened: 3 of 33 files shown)


src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
```diff
@@ -124,6 +124,18 @@ public long getAsLong()
...
     }
+    
+    public double getAsDouble()
+    {
+        switch (type)
+        {
+            default:
+                throw new IllegalStateException("Cannot convert option of type " + type + " to double");
+            case STRING:
+            case NUMBER:
```
Chew, line 135: Can an integer not be converted to a number?

MinnDevelopment (changes requested):
This also needs to address choices which can support double too

Andre601:
I applied some new changes, but I'm not sure about a few things.

For once am I not sure how we should address `fromData`'s handling of the value.
The if seems to check if it is a Number instance (Which double counts towards), but then only gets the value as int.
So I'm not sure if we could just change this to `if (value instanceof Integer/Long)` and add an else-if for the Double case...

src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
Andre601: Added this.

DV8FromTheWorld:
I'm wondering if we should be using `getNumber` instead of `getDouble` to better match the Discord docs and also the `OptionType`.

It is definitely confusing that there is an `OptionType.NUMBER` but not `getNumber`

Andre601:
> I'm wondering if we should be using `getNumber` instead of `getDouble` to better match the Discord docs and also the `OptionType`.
> 
> It is definitely confusing that there is an `OptionType.NUMBER` but not `getNumber`

I think a change to getNumber would be good. Will commit this real quick.
On another note, wouldn't it be an idea to make all the `getX(String)` methods call `getX(String, X)` to reduce duplicate code?
The second one could have a check for if both the value and defaultValue are null and throw in such a case, because then it could be a simple `getX(String, null)` call for the main method.

Downside I see here is, that we would need to use the non-primitive options for the defaultValue to have the nullability, which would probably be breaking...

But I feel like the duplicated code stuff should be reduced a bit...

Andre601:
After actually looking do I wonder where you want `getDouble` changed to `getNumber`

The Choice option has `getAsDouble` and that one matches the naming of the other options which are `getAsString` for String and `getAsLong` for long
By the logic you provided would we be required to also change `getAsLong` to `getAsInt` because Discord treats the info as Integer and the DataType is also declared as `INT`

The `getDouble` in the DataObject is also fine since there are other methods to get as string, (unsigned) integer and (unsigned) long...

I'm just a bit confused where stuff should change...

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -226,6 +228,35 @@ public OptionData setRequired(boolean required)
...
+     * 
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData addChoice(@Nonnull String name, double value)
+    {
+        Checks.notEmpty(name, "Name");
+        Checks.notLonger(name, 100, "Name");
+        Checks.check(value >= -0x1.0p53, "Double value may not be lower than -2^53");
+        Checks.check(value <= 0x1.0p53, "Double value may not be larger than 2^53");
```
MinnDevelopment, line 252: These should be public constants with docs references

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -226,6 +228,35 @@ public OptionData setRequired(boolean required)
...
+     * <br>The user can only provide one of the choices and cannot specify any other value.
+     * 
+     * @param  name
+     *         The name used in the client
+     * @param  value
+     *         The value received in {@link net.dv8tion.jda.api.interactions.commands.OptionMapping OptionMapping}
+     * 
+     * @throws IllegalArgumentException
+     *         If the name is null, empty or longer than 100 Characters.
+     *         Also thrown if this is not an option of type {@link OptionType#NUMBER} or more than 25 choices are provided.
```
MinnDevelopment, line 242: This says nothing about the valid range

src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
```diff
@@ -363,6 +364,23 @@ public Choice(@Nonnull String name, long value)
             this.name = name;
             this.intValue = value;
             this.stringValue = Long.toString(value);
+            this.doubleValue = 0.0d;
```
MinnDevelopment, line 367: Can't this also use the long value? Any integer is also a valid float

src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
```diff
@@ -378,6 +396,7 @@ public Choice(@Nonnull String name, @Nonnull String value)
             this.name = name;
             this.intValue = 0;
             this.stringValue = value;
+            this.doubleValue = 0.0d;
```
MinnDevelopment, line 399: Should be use NaN here?

src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
```diff
@@ -124,6 +124,19 @@ public long getAsLong()
                 return data.getLong("value");
         }
     }
+    
+    public double getAsDouble()
```
MinnDevelopment, line 128: This is missing docs

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -380,8 +413,10 @@ public static OptionData fromData(@Nonnull DataObject json)
                 choices1.stream(DataArray::getObject).forEach(o ->
                 {
                     Object value = o.get("value");
-                    if (value instanceof Number)
+                    if (value instanceof Long)
                         option.addChoice(o.getString("name"), ((Number) value).intValue());
+                    else if (value instanceof Double)
+                        option.addChoice(o.getString("name"), ((Number) value).doubleValue());
```
MinnDevelopment, line 419:
```suggestion
                        option.addChoice(o.getString("name"), (Double) value);
```

src/main/java/net/dv8tion/jda/api/utils/data/DataObject.java
```diff
@@ -608,6 +608,43 @@ public int getUnsignedInt(@Nonnull String key, int defaultValue)
...
+    public double getDouble(@Nonnull String key)
+    {
+        Double value = get(Double.class, key, Double::parseDouble, Number::doubleValue);
+        if(value == null)
+            throw valueError(key, "double");
+        return value;
+    }
+
+    /**
+     * Resolves a double to a key.
```
MinnDevelopment, line 631:
```suggestion
     * Resolves a double to a key.
     *
```

MinnDevelopment (changes requested):

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
Andre601: Would it in that case not also be a good idea to use a constant for the name?

Andre601:
I created some private setters as suggested by @DV8FromTheWorld on Discord.
I'm not too sure about this because to achieve this did I had to remove the final from the values and apply default values instead.

src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
```diff
@@ -419,6 +434,16 @@ public String getName()
             return name;
         }
 
+        /**
+         * The value of this choice.
+         * 
+         * @return The double value
```
MinnDevelopment, line 440:
```suggestion
         * @return The double value, or NaN if this is not a numeric choice value
```

src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
```diff
@@ -460,6 +485,27 @@ public String toString()
...
+        {
+            this.doubleValue = value;
+            this.intValue = value;
+            this.stringValue = Long.toString(value);
+        }
+        
+        private void setDoubleValue(double value)
+        {
+            this.doubleValue = value;
+            this.intValue = (int) value;
```
MinnDevelopment, line 499: Why not long?

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -32,6 +32,16 @@
  */
 public class OptionData implements SerializableData
 {
+    /**
+     * The maximum positive amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
+     */
+    public static final double MAX_POSITIVE_NUMBER = 0x1.0p53;
+    
+    /**
+     * The maximum negative amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
```
MinnDevelopment, line 41:
```suggestion
     * The smallest negative amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -380,8 +424,10 @@ public static OptionData fromData(@Nonnull DataObject json)
                 choices1.stream(DataArray::getObject).forEach(o ->
                 {
                     Object value = o.get("value");
-                    if (value instanceof Number)
+                    if (value instanceof Long)
                         option.addChoice(o.getString("name"), ((Number) value).intValue());
+                    else if (value instanceof Double)
```
MinnDevelopment, line 429: Can this also be instanceof Float? (see etf, jackson inconsistency, etc.)

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -380,8 +424,10 @@ public static OptionData fromData(@Nonnull DataObject json)
                 choices1.stream(DataArray::getObject).forEach(o ->
                 {
                     Object value = o.get("value");
-                    if (value instanceof Number)
+                    if (value instanceof Long)
```
MinnDevelopment, line 427: There is no way jackons will parse all integers into longs

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
MinnDevelopment: Maybe we should do:

```java
if (value instanceof Float || value instanceof Double)
  ...
else if (value instanceof Number) // can only be int type
  ...
```

or

```java
if (o.isType("value", FLOAT))
  ...
else if (o.isType("value", INTEGER))
  ...
```

Make sure to actually test that these checks work correctly

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
Andre601: So return back to Number?

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
Andre601: Not sure how I would go about testing those...
Btw, the tests will probably also require some tinkering as they only test for strings and longs right now but not floats/doubles.

src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
Andre601: Oversight from me.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
Andre601: I've gone ahead and used the `if (o.isType(...))` setup you suggested, because it feels the cleanest to me compared to an instanceof check.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -32,6 +33,16 @@
...
 {
+    /**
+     * The highest positive amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
+     */
+    public static final double MAX_POSITIVE_NUMBER = 0x1.0p53;
+    
+    /**
+     * The smallest negative amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
+     */
+    public static final double MAX_NEGATIVE_NUMBER = -0x1.0p53;
```
MinnDevelopment, line 44:
```suggestion
    public static final double MIN_NEGATIVE_NUMBER = -0x1.0p53;
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -226,6 +239,36 @@ public OptionData setRequired(boolean required)
...
+     *         {@link #MAX_NEGATIVE_NUMBER} or higher than {@link #MAX_POSITIVE_NUMBER}, or more than 25 choices are provided.
+     * 
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData addChoice(@Nonnull String name, double value)
+    {
+        Checks.notEmpty(name, "Name");
+        Checks.notLonger(name, 100, "Name");
+        Checks.check(value >= MAX_NEGATIVE_NUMBER, "Double value may not be lower than %f", MAX_NEGATIVE_NUMBER);
```
MinnDevelopment, line 263:
```suggestion
        Checks.check(value >= MIN_NEGATIVE_NUMBER, "Double value may not be lower than %f", MIN_NEGATIVE_NUMBER);
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -226,6 +239,36 @@ public OptionData setRequired(boolean required)
...
+     * 
+     * @param  name
+     *         The name used in the client
+     * @param  value
+     *         The value received in {@link net.dv8tion.jda.api.interactions.commands.OptionMapping OptionMapping}
+     * 
+     * @throws IllegalArgumentException
+     *         If the name is null, empty or longer than 100 Characters.
+     *         Also thrown if this is not an option of type {@link OptionType#NUMBER}, the double is smaller than
+     *         {@link #MAX_NEGATIVE_NUMBER} or higher than {@link #MAX_POSITIVE_NUMBER}, or more than 25 choices are provided.
```
MinnDevelopment, line 254:
```suggestion
     *         {@link #MIN_NEGATIVE_NUMBER} or greater than {@link #MAX_POSITIVE_NUMBER}, or more than 25 choices are provided.
```

MinnDevelopment (changes requested):
There seems to be a class cast error on this test https://github.com/DV8FromTheWorld/JDA/blob/development/src/test/java/CommandDataTest.java#L186

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -176,7 +176,7 @@ public boolean isRequired()
                         return new Command.Choice(entry.getKey(), entry.getValue().toString());
                     else if (entry.getValue() instanceof Double)
                         return new Command.Choice(entry.getKey(), (double) entry.getValue());
-                    return new Command.Choice(entry.getKey(), (long) entry.getValue());
+                    return new Command.Choice(entry.getKey(), ((Number) entry.getValue()).longValue());
```
MinnDevelopment, line 179: Did you test that the cast for double above this works?

Andre601: Can't even test because apparently "Release version 8 not supported" lol

Andre601: I changed it to `((Number) entry.getValue()).doubleValue()` just in case.

MinnDevelopment (approved):

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -32,6 +33,36 @@
...
+     */
+    public static final int MAX_NAME_LENGTH = 32;
+
+    /**
+     * The highest length the description of an option can be.
+     */
+    public static final int MAX_DESCRIPTION_LENGTH = 100;
+
+    /**
+     * The highest length the {@link #addChoice(String, String) String value} of a Choice can be.
```
DV8FromTheWorld, line 57: I question this description and variable name.
Isn't this actually `MAX_CHOICE_VALUE_LENGTH` ?

I'd update the variable name and docs accordingly

DV8FromTheWorld: I realize now that it is named this way because the others also don't prefix the fact that they are for choices 🤔

Andre601: I made corrections to the docs and renamed the var.

If you have further suggestions, maybe provide an example for a good description, because I'm not quite sure here...

Also, if we wanted to be really accurate would this probs need to be named `MAX_CHOICE_STRING_VALUE_LENGTH` or something...

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -32,6 +33,36 @@
  */
 public class OptionData implements SerializableData
 {
+    /**
+     * The highest positive amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
+     */
+    public static final double MAX_POSITIVE_NUMBER = (1L << 53);
```
oliver276, line 39: This number is an exclusive upper bound.  The client will fail when using an option that is this value or higher.
image
Numbers below this will succeed.
image

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -226,6 +262,41 @@ public OptionData setRequired(boolean required)
...
+     * 
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData addChoice(@Nonnull String name, double value)
+    {
+        Checks.notEmpty(name, "Name");
+        Checks.notLonger(name, MAX_NAME_LENGTH, "Name");
+        Checks.check(value >= MIN_NEGATIVE_NUMBER, "Double value may not be lower than %f", MIN_NEGATIVE_NUMBER);
+        Checks.check(value <= MAX_POSITIVE_NUMBER, "Double value may not be larger than %f", MAX_POSITIVE_NUMBER);
```
oliver276, line 292: Refer to earlier comments; just putting this here, as it may be easier to change these checks than change the definitions.  I'm not sure which is better for readability, too.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -32,6 +33,36 @@
...
 {
+    /**
+     * The highest positive amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
+     */
+    public static final double MAX_POSITIVE_NUMBER = (1L << 53);
+    
+    /**
+     * The smallest negative amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
+     */
+    public static final double MIN_NEGATIVE_NUMBER = -(1L << 53);
```
oliver276, line 44: Same as above... or below... this is confusing.

Numbers this size or more negative than this will fail:
image

Numbers less negative than this will succeed:
image

oliver276: Also can I suggest changing the comment to read "most negative amount", or "largest negative value" or similar?

Chew (commented):
My suggestions, great work!

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
Andre601: So what now?

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -55,16 +70,17 @@
      * @param  type
      *         The {@link OptionType}
      * @param  name
-     *         The option name, 1-32 lowercase alphanumeric characters
+     *         The option name, up to {@value #MAX_NAME_LENGTH} alphanumeric (with dash) lowercase characters, as
+     *         defined by {@link #MAX_NAME_LENGTH}
      * @param  description
-     *         The option description, 1-100 characters
+     *         The option description, up to {@value #MAX_DESCRIPTION_LENGTH} characters, as defined by {@link #MAX_DESCRIPTION_LENGTH}
```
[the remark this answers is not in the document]
Andre601: Refer to the thread in the server about docs for the context

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -36,12 +36,27 @@
...
     /**
      * The smallest negative amount Discord allows the {@link OptionType#NUMBER NUMBER} type to be.
      */
-    public static final double MIN_NEGATIVE_NUMBER = -0x1.0p53;
+    public static final double MIN_NEGATIVE_NUMBER = -(1L << 55);
+
+    /**
+     * The highest length the name of an option can be.
+     */
+    public static final int MAX_NAME_LENGTH = 32;
```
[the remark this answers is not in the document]
Andre601: It's consistent with the other namings used throughout the class and with other classes.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -244,22 +262,27 @@ public OptionData setRequired(boolean required)
...
+     *         </ul>
      * 
      * @return The OptionData instance, for chaining
      */
     @Nonnull
     public OptionData addChoice(@Nonnull String name, double value)
     {
         Checks.notEmpty(name, "Name");
-        Checks.notLonger(name, 100, "Name");
+        Checks.notLonger(name, MAX_NAME_LENGTH, "Name");
```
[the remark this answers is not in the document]
Andre601: What do you mean exactly?

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
DV8FromTheWorld: The constant should define what is the largest _valid_ value.
So ensure that the value itself is valid if it were used

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
DV8FromTheWorld: Copying previous statement: The constant should represent the largest _valid_ value.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
DV8FromTheWorld: > Also can I suggest changing the comment to read "most negative amount", or "largest negative value" or similar?
Good point. It isn't a "small" negative number.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
Andre601: Repeating myself:
> So what now? Decrease 53 by one?

To be more precise: Use `(1L << 53) - 1` for positive and `(1L << 53) + 1` for negative value?

DV8FromTheWorld: Yup, which it looks like you've already done 👍

Issue #1746, opened 2021-07-28
Add support for Interaction field


TopiSenpai:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request

Add support for the `interaction` field in the message object.
<https://discord.com/developers/docs/interactions/slash-commands#message-interaction-object>

### Example Use-Case

could be used to see which interaction/slash command a button originates from

for button with custom id `accept` from slash command `ban` run this code etc

PR #1747, opened 2021-07-28
add Interaction Object for Messages

TopiSenpai:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1746

## Description

This pr implements the message interaction object <https://discord.com/developers/docs/interactions/slash-commands#message-interaction-object>

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -2495,6 +2495,8 @@ public interface Message extends ISnowflake, Formattable
     /**
      * This is sent on the message object when the message is a response to an {@link net.dv8tion.jda.api.interactions.Interaction Interaction} without an existing message.
      *
+     * <p>This means responses to Message Components do not include this property, instead including a message reference object as components always exist on preexisting messages.
+     *
      * @throws java.lang.UnsupportedOperationException
      *         If this is a system message
      *
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -22,6 +22,7 @@ import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.Permission;
 import net.dv8tion.jda.api.exceptions.HttpException;
 import net.dv8tion.jda.api.exceptions.InsufficientPermissionException;
+import net.dv8tion.jda.api.interactions.InteractionType;
 import net.dv8tion.jda.api.interactions.components.ActionRow;
 import net.dv8tion.jda.api.interactions.components.Button;
 import net.dv8tion.jda.api.interactions.components.ComponentLayout;
@@ -2471,6 +2472,9 @@ public interface Message extends ISnowflake, Formattable
     @Nonnull
     MessageType getType();
 
+    @Nullable
+    Interaction getInteraction();
+
     /**
      * Mention constants, useful for use with {@link java.util.regex.Pattern Patterns}
      */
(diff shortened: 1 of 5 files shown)


TopiSenpai:
this should be all changes needed and I'll do docs once the implementation is fine

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -3035,4 +3039,58 @@ public boolean isSpoiler()
...
+    class Interaction implements ISnowflake
+    {
+
+        private final long id;
+        private final int type;
+        private final String name;
+        private final User user;
+        private final Member member;
+
+        public Interaction(long id, int type, String name, User user, Member member)
```
Sanduhr32, line 3052: I personally think, it's not optimal to have both arguments for user and member.

A user will basically always exist, so it's redundant to pass it when you also pass a non-null member, since you get that said user from the member on creation.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1521,6 +1532,29 @@ public MessageSticker createSticker(DataObject content)
...
+        final String name = content.getString("name");
+        DataObject userJson = content.getObject("user");
+        User user = null;
+        MemberImpl member = null;
+        if (!content.isNull("member") && guildImpl != null)
+        {
+            DataObject memberJson = content.getObject("member");
+            memberJson.put("user", userJson);
+            member = createMember(guildImpl, memberJson);
+            user = member.getUser();
```
Sanduhr32, line 1548: said user from member

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1521,6 +1532,29 @@ public MessageSticker createSticker(DataObject content)
...
+        }
+        else
+        {
+            user = createUser(userJson);
+        }
+
+        return new Message.Interaction(id, type, name, user, member);
```
Sanduhr32, line 1555: said redundant arguments, I think

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1521,6 +1532,29 @@ public MessageSticker createSticker(DataObject content)
...
+        if (!content.isNull("member") && guildImpl != null)
+        {
+            DataObject memberJson = content.getObject("member");
+            memberJson.put("user", userJson);
+            member = createMember(guildImpl, memberJson);
+            user = member.getUser();
+        }
+        else
+        {
+            user = createUser(userJson);
```
Sanduhr32, line 1552: user provided, if guild or member are null

Sanduhr32 (commented):
I'm unsure how important this is, so I'm just leaving this as a neutral comment

src/main/java/net/dv8tion/jda/api/entities/Message.java
DV8FromTheWorld: Are you suggesting that we have 2 constructors:
`Interaction(long, int, String, User)`
and
`Interaction(long, int, String, Member)` instead?

Sanduhr32: Possible yes, even tho the member / user argument is kind of ambiguous (because you easily could provide a less detailed user instead of a member), but since this is an internal constructor, I probably would go with this it.

arynxd: > per guild avatars and banners are possible or have been teased

They may be in some sort of beta, I've seen them in the wild

DV8FromTheWorld: The current implementation is probably fine.

DV8FromTheWorld:
This PR needs docs to be merged. Looks okay otherwise.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -3055,4 +3067,88 @@ public boolean isSpoiler()
         }
 
     }
+
+    /**
+     * Represents an {@link net.dv8tion.jda.api.interactions.Interaction Interaction} provided with a {@link net.dv8tion.jda.api.entities.Message Message}.
+     */
+    class Interaction implements ISnowflake
+    {
+
```
MinnDevelopment, line 3076:
```suggestion
```

MinnDevelopment (approved):

PR #1753, opened 2021-08-01
Bump selection option limits and extract max length values of SelectOption and Button

Alf-Melmac:
## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

Introduces public fields in SelectOption that contain the maximum length of Label, Value and Description.
Introduces public fields in Button that contain the maximum length of Label, ID and URL.
Identical to MessageEmbed, these fields can then be used by developers without having to update themselves when changes occur.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/components/Button.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/components/Button.java
@@ -188,7 +188,7 @@ public interface Button extends Component
      *         The label to use
      *
      * @throws IllegalArgumentException
-     *         If the label is not between 1-80 characters
+     *         If the label is null, empty, or longer than {@value #LABEL_MAX_LENGTH} characters
      *
      * @return New button with the changed label
      */
@@ -208,7 +208,7 @@ public interface Button extends Component
      *         The id to use
      *
      * @throws IllegalArgumentException
-     *         If the id is not between 1-100 characters
+     *         If the id is null, empty, or longer than {@value #ID_MAX_LENGTH} characters
      *
      * @return New button with the changed id
      */
@@ -228,7 +228,7 @@ public interface Button extends Component
      *         The url to use
      *
      * @throws IllegalArgumentException
-     *         If the url is null, empty, or longer than 512 characters
+     *         If the url is null, empty, or longer than {@value #URL_MAX_LENGTH} characters
      *
      * @return New button with the changed url
      */
@@ -278,7 +278,7 @@ public interface Button extends Component
      *         The text to display on the button
      *
      * @throws IllegalArgumentException
-     *         If any argument is empty or null, the label is longer than 80 characters, or the id is longer than 100 characters
+     *         If any argument is empty or null, the label is longer than {@value #LABEL_MAX_LENGTH} characters, or the id is longer than {@value #ID_MAX_LENGTH} characters
      *
      * @return The button instance
      */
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/api/interactions/components/Button.java
```diff
@@ -68,6 +68,21 @@
  */
 public interface Button extends Component
 {
+    /**
+     * The maximum length a button label can have
+     */
+    int LABEL_MAX_LENGTH = 80;
```
DV8FromTheWorld, line 74: Add `@see` tags to all constants similar to how `MessageEmbed` does

src/main/java/net/dv8tion/jda/api/interactions/components/selections/SelectOption.java
```diff
@@ -30,6 +30,21 @@
  */
 public class SelectOption implements SerializableData
 {
+    /**
+     * The maximum length a select option label can have
+     */
+    public static final int LABEL_MAX_LENGTH = 25;
```
DV8FromTheWorld, line 36: Add `@see` tag to all constants

Alf-Melmac: SelectOption has neither a builder nor a setter. Do you want me to reference the copy constructors withLabel, withValue and withDiscription?

src/main/java/net/dv8tion/jda/api/interactions/components/Button.java
Alf-Melmac: Unlike MessagEmbed, there is no builder or setter for buttons.
Imo linking to the copy constructors adds no value

MinnDevelopment: I think a more valuable link would be from `withLabel` to the constant, especially the throws descriptions currently still use the hardcoded 80 instead of the constant. You can use `{@value #LABEL_MAX_LENGTH}` to render the value of a constant in documentation.

DV8FromTheWorld: `@value` will render in the actual value, but should we instead stick with `@link` so that developers know that there is a variable that they can compare against? Seems like it might be more valuable

Alf-Melmac: Values are highlighted and I don't have to look in the variable definition to know which value is there.
Links, however, would be consistent with embeds.

DV8FromTheWorld: This exact wording isn't necessarily what we need, but I think this documentation is more explicit and thus better
https://github.com/DV8FromTheWorld/JDA/blob/d64de79db8c7a1581c4e9ab44472193d25327672/src/main/java/net/dv8tion/jda/api/entities/MessageEmbed.java#L381

It renders out as 
image

That feels more explicit as to the fact that there is a variable you can rely on than 
image

Alf-Melmac: I think I've caught all the spots now. While editing i unified the exception documentation. Ready to review again

Alf-Melmac:
Resolved merge conflict

MinnDevelopment:
It seems select option label and description limits are being raised to 100: https://github.com/discord/discord-api-docs/discussions/3409#discussioncomment-1128066
We should probably address that in this PR as well.

Issue #1298, opened 2020-05-14
Add option to escape all markdown characters to MarkdownSanitizer


Andre601:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request
When you use `MarkdownSanitized.escape(String)` on a text that only contains a single icon used in markdown does MarkdownSanitizer not escape it, causing weird Markdown formatting to happen.

For example:
Assume we have a user with the name `_ User`
When I would have the message `Andre_601 is giving _ User _1 Reputation Point_` and want the MarkdownSanitizer to escape the username, will it not do that as it doesn't find two _ to escape.

This causes the above message to be displayed like this:
![image](https://user-images.githubusercontent.com/11576465/81928638-e0d00500-95e5-11ea-9b23-0f45628801b4.png)

When in actuality it should be displayed like this:
![image](https://user-images.githubusercontent.com/11576465/81928685-f3e2d500-95e5-11ea-8f60-e697d3292b00.png)

Having an option to escape every character used for markdown, even when it's a single one, would be a useful thing as you may want to escape those characters in usernames or similar, while still having markdown formatting used.
My idea would be a boolean option that when set to true, would escape every _, *, ~, >, etc.

### Example Use-Case
Use-case has explained above.

DV8FromTheWorld:
```
String someString = "Andre_601"
String fullyEscaped = someString.replaceAll("([_`~*>])", "\\\\$1")
```

DV8FromTheWorld:
That being said, this issue seems like an easy issue for a simple PR.

RicardormDev:
I wrote this at a server and it seems that the problem is persisting, this might be an error from discord.

![image](https://user-images.githubusercontent.com/8129679/98714259-4b23f100-234e-11eb-8aa0-861e9670ede0.png)

PR #1754, opened 2021-08-02
Feature: Markdown escape all

SweetSofiMC:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1298

## Description

Added method to escape all markdown tokens in a given String, without checking for region ends, ignoring already escaped characters.

This pull request also contains tests for the described method.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
@@ -266,7 +266,14 @@ public class MarkdownSanitizer
                         builder.append("\\`");
                         break;
                     case '|':
-                        builder.append("\\|");
+                        if(i+1 < sequence.length()){
+                            if(sequence.charAt(i+1) == '|'){
+                                builder.append("\\|\\|");
+                                i++;
+                                continue;
+                            }
+                        }
+                        builder.append("|");
                         break;
                     case '~':
                         builder.append("\\~");
--- a/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
@@ -189,6 +189,76 @@ public class MarkdownSanitizer
     }
 
     /**
+     * Escapes every single markdown formatting token found in the provided string.
+     * <br>Example: {@code escape("**Hello _World_", true)}
+     *
+     * @param sequence
+     *        The string to sanitize
+     * @param single
+     *        Whether it should scape single tokens or not.
+     *
+     * @throws java.lang.IllegalArgumentException
+     *         If provided with null sequence
+     *
+     * @return The string with escaped markdown
+     */
+    @Nonnull
+    public static String escape(@Nonnull String sequence, boolean single)
+    {
+        Checks.notNull(sequence, "Input");
+        if(single)
+        {
+            StringBuilder builder = new StringBuilder();
+            boolean escaped = false;
+            for (int i = 0; i < sequence.length(); i++)
+            {
+                char current = sequence.charAt(i);
+                if(!escaped)
+                {
+                    switch (current)
+                    {
+                    case '*':
+                        builder.append("\\*");
+                        break;
+                    case '_':
+                        builder.append("\\_");
+                        break;
+                    case '`':
+                        builder.append("\\`");
+                        break;
+                    case '|':
+                        builder.append("\\|");
+                        break;
+                    case '~':
+                        builder.append("\\~");
+                        break;
+                    case '>':
+                        builder.append("\\>");
+                        break;
+                    case '\\':
+                        builder.append(current);
+                        escaped = true;
+                        break;
+                    default:
+                        builder.append(current);
+                    }
+                }
+                else
+                {
+                    builder.append(current);
+                    escaped = false;
+                }
+
+            }
+            return builder.toString();
+        }
+        else
+        {
+            return escape(sequence);
+        }
+    }
+
+    /**
      * Switches the used {@link net.dv8tion.jda.api.utils.MarkdownSanitizer.SanitizationStrategy}.
      *
      * @param  strategy
--- a/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
@@ -210,9 +210,48 @@ public class MarkdownSanitizer
         {
             StringBuilder builder = new StringBuilder();
             boolean escaped = false;
+            boolean newline = true;
             for (int i = 0; i < sequence.length(); i++)
             {
                 char current = sequence.charAt(i);
+                if(newline)
+                {
+                    if(current == '>')
+                    {
+                        if(i+1 < sequence.length())
+                        {
+                            if(sequence.charAt(i+1) == ' ')
+                            {
+                                builder.append("\\>");
+                            }
+                            else
+                            {
+                                if(i + 3 < sequence.length())
+                                {
+                                    if(sequence.startsWith(">>> ", i))
+                                    {
+                                        builder.append("\\>\\>\\> ");
+                                        i += 3;
+                                        newline = false;
+                                        continue;
+                                    }
+                                }
+                                builder.append(current);
+                            }
+                            newline = false;
+                            continue;
+                        }
+                        else
+                            escaped = true;
+                    }
+                    else if(current == ' ')
+                    {
+                        builder.append(current);
+                        continue;
+                    }
+                    newline=false;
+                }
+
                 if(!escaped)
                 {
                     switch (current)
@@ -232,12 +271,13 @@ public class MarkdownSanitizer
                     case '~':
                         builder.append("\\~");
                         break;
-                    case '>':
-                        builder.append("\\>");
-                        break;
                     case '\\':
                         builder.append(current);
                         escaped = true;
+                        break;
+                    case '\n':
+                        builder.append(current);
+                        newline = true;
                         break;
                     default:
                         builder.append(current);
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
```diff
@@ -188,6 +188,76 @@ public static String escape(@Nonnull String sequence, int ignored)
...
+                    case '`':
+                        builder.append("\\`");
+                        break;
+                    case '|':
+                        builder.append("\\|");
+                        break;
+                    case '~':
+                        builder.append("\\~");
+                        break;
+                    case '>':
```
MinnDevelopment, line 235: How does this prevent escaping a message like `hello 7 > 5` as `hello 7 \> 5`? What about `>hello` without a space, which would not be a quote?

SweetSofiMC: It doesn't, but as of now, the behavior of Discord allows to escape markdown characters even if they are not part of a valid Markdown token that modifies the text, it just displays it as a character without the `\`.

SweetSofiMC: i.e.
This: 
image
Generates this:
image

MinnDevelopment: This might be acceptable in some cases, but not if the message contains a link:

image

image

src/test/java/MarkdownTest.java
```diff
@@ -391,3 +391,74 @@ public void testQuote()
...
+    @Test
+    public void testTilde()
+    {
+        Assertions.assertEquals("Hello\\~\\~World",MarkdownSanitizer.escape("Hello~~World",true));
+
+        Assertions.assertEquals("Hello\\~\\~World",MarkdownSanitizer.escape("Hello\\~\\~World",true));
+    }
+
+    @Test
+    public void testAngle()
```
MinnDevelopment, line 453: Could you rename this to `testQuote` and add a test for `Hello > World` to make sure it doesn't escape it incorrectly?

src/test/java/MarkdownTest.java
```diff
@@ -391,3 +391,74 @@ public void testQuote()
...
+        Assertions.assertEquals("Hello\\`World",MarkdownSanitizer.escape("Hello\\`World",true));
+        Assertions.assertEquals("Hello\\`\\`World",MarkdownSanitizer.escape("Hello\\`\\`World",true));
+        Assertions.assertEquals("Hello\\`\\`\\`World",MarkdownSanitizer.escape("Hello\\`\\`\\`World",true));
+    }
+
+    @Test
+    public void testPipe()
+    {
+        Assertions.assertEquals("Hello\\|\\|World",MarkdownSanitizer.escape("Hello||World",true));
+        Assertions.assertEquals("Hello\\|World",MarkdownSanitizer.escape("Hello|World",true));
```
MinnDevelopment, line 438: Should this really escape?

src/test/java/MarkdownTest.java
```diff
@@ -391,3 +391,74 @@ public void testQuote()
...
+        Assertions.assertEquals("Hello\\`\\`World",MarkdownSanitizer.escape("Hello``World",true));
+        Assertions.assertEquals("Hello\\`\\`\\`World",MarkdownSanitizer.escape("Hello```World",true));
+
+        Assertions.assertEquals("Hello\\`World",MarkdownSanitizer.escape("Hello\\`World",true));
+        Assertions.assertEquals("Hello\\`\\`World",MarkdownSanitizer.escape("Hello\\`\\`World",true));
+        Assertions.assertEquals("Hello\\`\\`\\`World",MarkdownSanitizer.escape("Hello\\`\\`\\`World",true));
+    }
+
+    @Test
+    public void testPipe()
```
MinnDevelopment, line 435:
```suggestion
    public void testSpoiler()
```

src/test/java/MarkdownTest.java
```diff
@@ -391,3 +391,74 @@ public void testQuote()
...
+        Assertions.assertEquals("Hello\\_\\_World",MarkdownSanitizer.escape("Hello__World",true));
+        Assertions.assertEquals("Hello\\_\\_\\_World",MarkdownSanitizer.escape("Hello___World",true));
+
+        Assertions.assertEquals("Hello\\_World",MarkdownSanitizer.escape("Hello\\_World",true));
+        Assertions.assertEquals("Hello\\_\\_World",MarkdownSanitizer.escape("Hello\\_\\_World",true));
+        Assertions.assertEquals("Hello\\_\\_\\_World",MarkdownSanitizer.escape("Hello\\_\\_\\_World",true));
+    }
+
+    @Test
+    public void testBackQuote()
```
MinnDevelopment, line 423:
```suggestion
    public void testCodeblock()
```

src/test/java/MarkdownTest.java
```diff
@@ -391,3 +391,74 @@ public void testQuote()
         Assertions.assertEquals("\\> \\_Hello \n\\> World\\_", markdown.compute("> _Hello \n> World_"));
     }
 }
+
+class EscapeMarkdownAllTest
+{
+    @Test
+    public void testAsterisk()
+    {
+
```
MinnDevelopment, line 400:
```suggestion
```

src/test/java/MarkdownTest.java
```diff
@@ -391,3 +391,74 @@ public void testQuote()
...
+        Assertions.assertEquals("Hello\\|\\|World",MarkdownSanitizer.escape("Hello\\|\\|World",true));
+        Assertions.assertEquals("Hello\\|World",MarkdownSanitizer.escape("Hello\\|World",true));
+    }
+
+    @Test
+    public void testTilde()
+    {
+        Assertions.assertEquals("Hello\\~\\~World",MarkdownSanitizer.escape("Hello~~World",true));
+
```
MinnDevelopment, line 448:
```suggestion
```

src/test/java/MarkdownTest.java
```diff
@@ -391,3 +391,74 @@ public void testQuote()
...
+    {
+        Assertions.assertEquals("Hello\\|\\|World",MarkdownSanitizer.escape("Hello||World",true));
+        Assertions.assertEquals("Hello\\|World",MarkdownSanitizer.escape("Hello|World",true));
+
+        Assertions.assertEquals("Hello\\|\\|World",MarkdownSanitizer.escape("Hello\\|\\|World",true));
+        Assertions.assertEquals("Hello\\|World",MarkdownSanitizer.escape("Hello\\|World",true));
+    }
+
+    @Test
+    public void testTilde()
```
MinnDevelopment, line 445:
```suggestion
    public void testStrike()
```

src/test/java/MarkdownTest.java
```diff
@@ -391,3 +391,74 @@ public void testQuote()
...
+
+class EscapeMarkdownAllTest
+{
+    @Test
+    public void testAsterisk()
+    {
+
+        Assertions.assertEquals("Hello\\*World",MarkdownSanitizer.escape("Hello*World",true));
+        Assertions.assertEquals("Hello\\*\\*World",MarkdownSanitizer.escape("Hello**World",true));
+        Assertions.assertEquals("Hello\\*\\*\\*World",MarkdownSanitizer.escape("Hello***World",true));
```
MinnDevelopment, line 403: Usually we always add a space after a comma, since it looks less cluttered

MinnDevelopment (changes requested):
Sorry for the delay, here is my review

DV8FromTheWorld:
Bump

src/main/java/net/dv8tion/jda/api/utils/MarkdownSanitizer.java
```diff
@@ -188,6 +188,123 @@ public static String escape(@Nonnull String sequence, int ignored)
...
+                        builder.append(current);
+                    }
+                }
+                else
+                {
+                    builder.append(current);
+                    escaped = false;
+                }
+
+            }
```
MinnDevelopment, line 299: I'm proposing a cleanup:

```suggestion
            for (int i = 0; i < sequence.length(); i++)
            {
                char current = sequence.charAt(i);
                if (newline)
                {
                    newline = Character.isWhitespace(current); // might still be a quote if prefixed by whitespace
                    if (current == '>')
                    {
                        // Check for quote if line starts with angle bracket
                        if (i + 1 < sequence.length() && Character.isWhitespace(sequence.charAt(i+1)))
                        {
                            builder.append("\\>"); // simple quote
                        }
                        else if (i + 3 < sequence.length() && sequence.startsWith(">>>", i) && Character.isWhitespace(sequence.charAt(i+3)))
                        {
                            builder.append("\\>\\>\\>").append(sequence.charAt(i+3)); // block quote
                            i += 3; // since we include 3 angle brackets AND whitespace
                        }
                        else
                        {
                            builder.append(current); // just a normal angle bracket
                        }
                        continue;
                    }
                }

                if (escaped)
                {
                    builder.append(current);
                    escaped = false;
                    continue;
                }
                // Handle average case
                switch (current)
                {
                case '*': // simple markdown escapes for single characters
                case '_':
                case '`':
                    builder.append('\\').append(current);
                    break;
                case '|': // cases that require at least 2 characters in sequence
                case '~':
                    if (i + 1 < sequence.length() && sequence.charAt(i+1) == current)
                    {
                        builder.append('\\').append(current)
                               .append('\\').append(current);
                        i++;
                    }
                    else
                        builder.append(current);
                    break;
                case '\\': // escape character
                    builder.append(current);
                    escaped = true;
                    break;
                case '\n': // linefeed is a special case for quotes
                    builder.append(current);
                    newline = true;
                    break;
                default:
                    builder.append(current);
                }
            }
```

The logic should be identical, I just rewrote it a bit and added some comments. Also, I replaced your check for `== ' '` with an `isWhitespace` which covers all possible space characters. Let me know what you think, or if I made any mistake.

MinnDevelopment:
You need to fix your commit history, you added commits from other PRs.

MinnDevelopment (approved):

PR #1759, opened 2021-08-03
Add new message types for Threads

oliver276:
## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds the new message types for threads of THREAD_CREATED and THREAD_STARTER_MESSAGE, as well as documenting the old types that have been repurposed with the release of Threads.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageHistory.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageHistory.java
@@ -32,7 +32,9 @@ import net.dv8tion.jda.internal.entities.EntityBuilder;
 import net.dv8tion.jda.internal.requests.RestActionImpl;
 import net.dv8tion.jda.internal.requests.Route;
 import net.dv8tion.jda.internal.utils.Checks;
+import net.dv8tion.jda.internal.utils.JDALogger;
 import org.apache.commons.collections4.map.ListOrderedMap;
+import org.slf4j.Logger;
 
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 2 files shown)


MinnDevelopment:
Can you also change the catch here to just catch all exceptions: https://github.com/DV8FromTheWorld/JDA/blob/196f31f25155afdbca3d7e9724aa418437af6125/src/main/java/net/dv8tion/jda/api/entities/MessageHistory.java#L584

And add similar try/catch blocks in these:

https://github.com/DV8FromTheWorld/JDA/blob/196f31f25155afdbca3d7e9724aa418437af6125/src/main/java/net/dv8tion/jda/api/entities/MessageHistory.java#L199

https://github.com/DV8FromTheWorld/JDA/blob/196f31f25155afdbca3d7e9724aa418437af6125/src/main/java/net/dv8tion/jda/api/entities/MessageHistory.java#L268

MinnDevelopment:
There also seems to be `GUILD_INVITE_REMINDER` as `22`, do you have any idea what this is? The docs obviously don't explain anything.

oliver276:
Nope, and as far as I know I haven't encountered it either. I don't have any large bots, so I guess that's to be expected.

Should I add it anyway with the same unhelpful docs as the official docs or should I try to find out its purpose first by trying to ask in ddevs or something?

MinnDevelopment:
For now, I guess just add it with useless docs and we can improve docs once someone knows what it is.

fbrettnich:
In [DiscordDelta](https://github.com/sesaborg/DiscordDelta/wiki/Reference#guild_invite_reminder) it is described as follows:
`GUILD_INVITE_REMINDER` => "A message that serves to remind users that they can invite other users to the guild."

This then looks like this
image

MinnDevelopment (approved):

PR #1764, opened 2021-08-05
Ensure permission in PermissionException is not null

ScoreUnder:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

A null `Permission` is semantic nonsense: it is `Permission.UNKNOWN` but with extra steps (i.e. null checks). As such, it should not be allowed to propagate through a `PermissionException`. This PR enforces and encodes that.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/exceptions/PermissionException.java
+++ b/src/main/java/net/dv8tion/jda/api/exceptions/PermissionException.java
@@ -16,6 +16,9 @@
 package net.dv8tion.jda.api.exceptions;
 
 import net.dv8tion.jda.api.Permission;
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
Permission exceptions are not only used for missing permissions, that is what InsufficientPermissionException is used for. [HierarchyException](https://github.com/DV8FromTheWorld/JDA/blob/development/src/main/java/net/dv8tion/jda/api/exceptions/HierarchyException.java) is also a permission error but is not caused by missing permissions.

MinnDevelopment (approved):

PR #1769, opened 2021-08-08
Introduce Context Menus / User Commands / Message Commands

aasmart:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR introduces Discord's upcoming feature: Context Menu Commands (Note: This should probably wait until after the docs and such release, which I assume is already known).

### How to Create a Context Menu Command
Creating a context menu command is as simple as creating a normal slash command. Instead, using the constructor CommandData(type, name) allows you to create a command of the type CommandType#USER_COMMAND or CommandType#MESSAGE_COMMAND with a name. Attempting to add choices, subcommands, descriptions, etc. does not work on context menu commands.


aasmart:
So after I did some further testing, I realized editing the Context Command doesn't work properly. This is because the action creates a command data with "undefined" as both values, which I don't know how to fix with this current implementation. The only solution I can think of is making a CommandEditAction for each command, which just seems cumbersome.

MinnDevelopment:
Unfortunately, this change requires a lot of breaking changes, which makes it unsuitable for a JDA 4 release. As such, this will have to be on hold until we start work on JDA 5 in the coming months. If Discord had made this a different interaction type as I had requested in discord/discord-api-docs#3590, then we could have included it in JDA 4.

aasmart:
Makes sense. Kind of a good thing because I realized I don't really have a good idea on how to implement editing and command limit checks. Hopefully they separate the two, despite the convenience for Discord

DV8FromTheWorld:
fyi this is going to have to be based off of `v5` as it isn't going to land in `v4`

MinnDevelopment:
Superseded by #1844

Issue #1781, opened 2021-08-18
Suppressed Bot cannot unmute itself


DeveloperTK:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

The bot sometimes cannot unmute itself when it is being suppressed in an AFK channel.

### Expected Behavior

When queueing `Member#mute(false)`, the bot unmutes itself

### Code Example or Reproduction Steps

1. Let the bot join the AFK Channel
2. Try unmuting the Bot inside of the `GuildVoiceJoinEvent`

### Code for JDABuilder or DefaultShardManagerBuilder Used

```java
JDABuilder.createDefault(token).build()
```

### Exception or Error

N/A

Some things I have experiencing while trying to solve this problem: Every time the bot was already server muted (not suppressed), it could unmute itself just fine. I have also got this working by manually calling the API like this: (using OkHttp)
```java
Request request = new Request.Builder()
                .url("https://discord.com/api/v9/guilds/" + guildId + "/members/" + memberId)
                .patch(RequestBody.create(MediaType.get("application/json"), "{\"mute\": false}"))
                .header("Authorization", "Bot " + token)
                .build();
```

Could this maybe have something to do with JDA caching the muted value and refusing to update the muted state? I have got a callback while queueing the `Member#mute()` action so I suppose that is not the problem here...

Andre601:
I think that is normal?
Last time I entered an afk channel as a owner was I muted too and couldn't unmute myself, so I would assume it's the same for bots

arynxd:
Just tested Andre601's theory, can confirm the owner cant speak in the AFK channel.

DeveloperTK:
Yes exactly that is why my plan is to unmute the bot inside the `GuildVoiceJoinEvent` which does not work.

DManstrator:
I guess this is some kind of intended behavior.

Yes, you would need to remove the server mute first but the self-mute still remains and I don't think bots can remove that mute if they have it like normal user can. Perhaps this is a Discord Bug that this self-mute also applies to Bots when they join the AFK channel, not sure (I didn't test it).

[...]

PR #1783, opened 2021-08-19
Enable unmuting suppressed users

DeveloperTK:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1781

## Description

Discord handles unmuting suppressed users with no talk permissions by sending an API call that would otherwise unmute the member guild-wide. This check inside of the mute action prevents this:
```Java
if (voiceState.isGuildMuted() == mute)
    return new CompletedRestAction<>(getJDA(), null);
```
For my thought process debugging this, see: #1781

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
@@ -1436,8 +1436,9 @@ public class GuildImpl implements Guild
         {
             if (voiceState.getChannel() == null)
                 throw new IllegalStateException("Can only mute members who are currently in a voice channel");
-            if (voiceState.isGuildMuted() == mute && !voiceState.isSuppressed())
-                return new CompletedRestAction<>(getJDA(), null);
+            if (voiceState.isGuildMuted() == mute)
+                if (mute || !voiceState.isSuppressed())
+                    return new CompletedRestAction<>(getJDA(), null);
         }
 
         DataObject body = DataObject.empty().put("mute", mute);
--- a/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
@@ -1436,9 +1436,8 @@ public class GuildImpl implements Guild
         {
             if (voiceState.getChannel() == null)
                 throw new IllegalStateException("Can only mute members who are currently in a voice channel");
-            if (voiceState.isGuildMuted() == mute)
-                if (mute || !voiceState.isSuppressed())
-                    return new CompletedRestAction<>(getJDA(), null);
+            if (voiceState.isGuildMuted() == mute && (mute || !voiceState.isSuppressed()))
+                return new CompletedRestAction<>(getJDA(), null);
         }
 
         DataObject body = DataObject.empty().put("mute", mute);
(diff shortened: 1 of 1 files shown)


DV8FromTheWorld:
While this definitely would "fix" the issue, I'm wondering if this is actually a discord bug that allows this functionality and was introduced with the introduction of StageChannels.

When you join a StageChannel you are auto-suppressed, however to become a speaker you get unsuppressed/unmuted.

I wonder if this functionality has somehow bled over into VoiceChannels on accident. I'd like to see this functionality being shown to be intended before we solidify it in JDA.

DeveloperTK:
Good point but I have experienced this behaviour long before StageChannels. When joining a channel with no talking permission (such as the AFK channel), you are suppressed but not muted. Unmuting the user also unsupresses them all while the `mute` field of the member object is still `false`. A few weeks ago I have read something on the discord-developers guild regarding that for not, this behaviour is intended but I cannot seem to find it anymore.

However this was always the way of unmuting people when they are suppressed. Since the client doesn't show the user being suppressed in the context menu, muting and then unmuting them is the way to go (using the client). WIth the API we can just "force un-mute" them which also changes the suppression state and not only guild mutes.

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1436,7 +1436,7 @@ public long getIdLong()
         {
             if (voiceState.getChannel() == null)
                 throw new IllegalStateException("Can only mute members who are currently in a voice channel");
-            if (voiceState.isGuildMuted() == mute)
+            if (voiceState.isGuildMuted() == mute && !voiceState.isSuppressed())
```
MinnDevelopment, line 1439: I'm unsure if this check makes sense for `mute == true` vs `mute == false`

DeveloperTK: Since the `isGuildMuted` value does not represent whether the member can actually talk (the member cannot talk if he is not muted but suppressed) we have to ignore the local voice state and always tell the api to unmute the member. Telling the API `mute: false` does not only change the guild muted state, but also un-suppresses the member.

MinnDevelopment: My point is that the `mute` parameter can be true, in which case the suppressed state is irrelevant.

MinnDevelopment: If `mute` is true, then the suppressed state shouldn't be checked at all. Since you don't unsuppress or suppress this way anyway.

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1437,7 +1437,8 @@ public long getIdLong()
             if (voiceState.getChannel() == null)
                 throw new IllegalStateException("Can only mute members who are currently in a voice channel");
             if (voiceState.isGuildMuted() == mute)
-                return new CompletedRestAction<>(getJDA(), null);
+                if (mute || !voiceState.isSuppressed())
```
MinnDevelopment, line 1440:
```java
if (voiceState.isGuildMuted() == mute && (mute || !voiceState.isSuppressed()))
```

MinnDevelopment (approved):

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
DeveloperTK: done

PR #1785, opened 2021-08-21
Fix a typo in Invite

markozajc:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fix a typo in Invite#isTemporary().

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Invite.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Invite.java
@@ -310,12 +310,17 @@ public interface Invite
     boolean isExpanded();
 
     /**
-     * Whether this Invite grants only temporary access or not
+     * Whether this Invite grants only temporary access or not.
+     *
+     * <p>This works only for expanded invites and will throw a {@link IllegalStateException} otherwise!
      *
      * @throws IllegalStateException
      *         if this invite is not expanded
      *
-     * @return Whether is invite expanded or not
+     * @return Whether is invite temporary or not
+     *
+     * @see    #expand()
+     * @see    #isExpanded()
      */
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -310,12 +310,17 @@ default String getUrl()
     boolean isExpanded();
 
     /**
-     * Whether this Invite grants only temporary access or not
+     * Whether this Invite grants only temporary access or not.
+     *
+     * <p>This works only for expanded invites and will throw a {@link IllegalStateException} otherwise!
```
DV8FromTheWorld, line 315: This line is pointless. The entire point of the `@throws` block is to explain possible exceptions which it already does

markozajc: I did that for consistency's sake, as other methods have that too.

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -310,12 +310,17 @@ default String getUrl()
...
      *
      * @throws IllegalStateException
      *         if this invite is not expanded
      *
-     * @return Whether is invite expanded or not
+     * @return Whether is invite temporary or not
```
MinnDevelopment, line 320: Let's use this chance to fix the grammar as well:

```suggestion
     * @return Whether this invite temporary or not
```

Feel free to fix it everywhere else in the Invite docs.

MinnDevelopment (approved):

PR #1786, opened 2021-08-22
Add missing type conversion

lion0738:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

java.lang.IllegalArgumentException: Cannot add double choice for OptionType.INTEGER
at net.dv8tion.jda.api.interactions.commands.build.OptionData.addChoice(OptionData.java:295)
at net.dv8tion.jda.api.interactions.commands.build.OptionData.lambda$fromData$2(OptionData.java:477)

Added missing type conversion to fix upper issue

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
@@ -474,7 +474,7 @@ public class OptionData implements SerializableData
                     if (o.isType("value", DataType.FLOAT))
                         option.addChoice(o.getString("name"), o.getDouble("value"));
                     else if (o.isType("value", DataType.INT))
-                        option.addChoice(o.getString("name"), o.getLong("value"));
+                        option.addChoice(o.getString("name"), (int) o.getLong("value"));
                     else
                         option.addChoice(o.getString("name"), o.get("value").toString());
                 })


oliver276:
I'm a little confused on what's going on here.

Can you share the code that was causing that exception?

lion0738:
There are multiple addChoice functions for various data types.
All int type choices should call addChoice function with int parameter, but they are parsed as double data type without this code.

lion0738:
> I'm a little confused on what's going on here.
> 
> Can you share the code that was causing that exception?

This is a simple code which reproduces the issue.

CommandData command = new CommandData("cmd", "Simple command")
		.addOptions(new OptionData(OptionType.INTEGER, "value", "int value")
				.addChoice("0", 0));
command.getOptions();

DV8FromTheWorld:
So the reason this is happening is because the `addChoice` function accepts `double`, `int`, and `String`, however we're trying to give it a `long`.

The compiler picks `double` here because that's the only compatible data type because `long` fits into `double` but not into `int`. 

I'm not sure that compressing the `long` into an `int` is the correct thing to do here. I feel ike the `addChoice` function should instead accept a `long` instead.

Indeed, the [docs](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-type) appear to state that the maximum numbers allow for `INTEGER` are the same as those allowed for `NUMBER` which is clearly outside the bounds of 32bit ints, so we should probably re-evaluate our representation of `INTEGER` in the system.

DV8FromTheWorld:
The proper fix for this is included in this PR
https://github.com/DV8FromTheWorld/JDA/pull/1807

PR #1801, opened 2021-09-02
Change choice name length check to 100 instead of 32

oliver276:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

A commit that moved many constants to the MAX_NAME_LENGTH mistakenly also moved a few values that were 100 into a constant with value 32.

This commit will add a new constant, MAX_NAME_OPTION_LENGTH, to fix those values back to their intended values of 100.

The commit title is meant to say Option Choice, not OptionData.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
@@ -51,7 +51,7 @@ public class OptionData implements SerializableData
     /**
      * The maximum length of the name of some OptionData names
      */
-    public static final int MAX_NAME_OPTION_LENGTH = 100;
+    public static final int MAX_CHOICE_NAME_LENGTH = 100;
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -47,6 +47,11 @@
      * The maximum length the name of an option can be.
      */
     public static final int MAX_NAME_LENGTH = 32;
+    
+    /**
+     * The maximum length of the name of some OptionData names
+     */
+    public static final int MAX_NAME_OPTION_LENGTH = 100;
```
MinnDevelopment, line 54: This seems to only apply to choices, so maybe it should be `MAX_CHOICE_NAME_LENGTH`?

MinnDevelopment (approved):

PR #1817, opened 2021-09-14
Add webhook id methods to Message

Vankka:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds `getWebhookId` and `getWebhookIdLong` methods to the Message class, which makes it possible to identify what webhook sent the message.


MinnDevelopment:
I'm fairly certain you can just use `getAuthor().getId()`

Vankka:
Didn't think of that one, but does appear to be the case

PR #1863, opened 2021-10-19
add support for min/max option value

caneleex:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this PR adds support for min/max option value. see https://github.com/discord/discord-api-docs/pull/3967/

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
@@ -552,8 +552,8 @@ public class Command implements ISnowflake
         private final boolean required;
         private final Set<ChannelType> channelTypes;
         private final List<Choice> choices;
-        private Double minValue;
-        private Double maxValue;
+        private Number minValue;
+        private Number maxValue;
 
         public Option(@Nonnull DataObject json)
         {
@@ -646,7 +646,7 @@ public class Command implements ISnowflake
          * @return The minimum value for this option or {@code null}
          */
         @Nullable
-        public Double getMinValue()
+        public Number getMinValue()
         {
             return minValue;
         }
@@ -659,7 +659,7 @@ public class Command implements ISnowflake
          * @return The maximum value for this option or {@code null}
          */
         @Nullable
-        public Double getMaxValue()
+        public Number getMaxValue()
         {
             return maxValue;
         }
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
@@ -636,7 +636,7 @@ public class Command implements ISnowflake
         }
 
         /**
-         * The minimal value which can be provided for this option.
+         * The minimum value which can be provided for this option.
          * <br>This returns {@link OptionData#MIN_NEGATIVE_NUMBER OptionData#MIN_NEGATIVE_NUMBER} if the option is not of type
          * {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}.
          *
@@ -648,7 +648,7 @@ public class Command implements ISnowflake
         }
 
         /**
-         * The maximal value which can be provided for this option.
+         * The maximum value which can be provided for this option.
          * <br>This returns {@link OptionData#MAX_POSITIVE_NUMBER OptionData#MAX_POSITIVE_NUMBER} if the option is not of type
          * {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}.
          *
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
@@ -550,9 +550,9 @@ public class Command implements ISnowflake
         private final int type;
         private final boolean required;
         private final Set<ChannelType> channelTypes;
-        private final double minValue;
-        private final double maxValue;
         private final List<Choice> choices;
+        private Double minValue;
+        private Double maxValue;
 
         public Option(@Nonnull DataObject json)
         {
@@ -560,14 +560,16 @@ public class Command implements ISnowflake
             this.description = json.getString("description");
             this.type = json.getInt("type");
             this.required = json.getBoolean("required");
-            this.minValue = json.getDouble("min_value", OptionData.MIN_NEGATIVE_NUMBER);
-            this.maxValue = json.getDouble("max_value", OptionData.MAX_POSITIVE_NUMBER);
             this.channelTypes = Collections.unmodifiableSet(json.optArray("channel_types")
                     .map(it -> it.stream(DataArray::getInt).map(ChannelType::fromId).collect(Collectors.toSet()))
                     .orElse(Collections.emptySet()));
             this.choices = json.optArray("choices")
                 .map(it -> it.stream(DataArray::getObject).map(Choice::new).collect(Collectors.toList()))
                 .orElse(Collections.emptyList());
+            if (!json.isNull("min_value"))
+                this.minValue = json.getDouble("min_value");
+            if (!json.isNull("max_value"))
+                this.maxValue = json.getDouble("max_value");
         }
 
         /**
@@ -637,24 +639,24 @@ public class Command implements ISnowflake
 
         /**
          * The minimum value which can be provided for this option.
-         * <br>This returns {@link OptionData#MIN_NEGATIVE_NUMBER OptionData#MIN_NEGATIVE_NUMBER} if the option is not of type
-         * {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}.
+         * <br>This returns {@code null} if the value is not set or if the option
+         * is not of type {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}.
          *
-         * @return The minimal value for this option
+         * @return The minimum value for this option or {@code null}
          */
-        public double getMinValue()
+        public Double getMinValue()
         {
             return minValue;
         }
 
         /**
          * The maximum value which can be provided for this option.
-         * <br>This returns {@link OptionData#MAX_POSITIVE_NUMBER OptionData#MAX_POSITIVE_NUMBER} if the option is not of type
-         * {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}.
+         * <br>This returns {@code null} if the value is not set or if the option
+         * is not of type {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}.
          *
-         * @return The maximal value for this option
+         * @return The maximum value for this option or {@code null}
          */
-        public double getMaxValue()
+        public Double getMaxValue()
         {
             return maxValue;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -339,6 +365,54 @@ public OptionData setChannelTypes(@Nonnull Collection<ChannelType> channelTypes)
...
+     *         If any of the following checks fail
+     *         <ul>
+     *             <li>{@link OptionType type of this option} is {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}</li>
+     *             <li>{@code value} is larger than or equal to {@link OptionData#MIN_NEGATIVE_NUMBER MIN_NEGATIVE_NUMBER}</li>
+     *         </ul>
+     *
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData setMinValue(double value)
```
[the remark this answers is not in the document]
caneleex: it gets rounded down. not ideal I suppose. I'll add another overload which takes an int and add checks

src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
```diff
@@ -630,6 +637,30 @@ public OptionType getType()
...
         }
 
+        /**
+         * The minimum value which can be provided for this option.
+         * <br>This returns {@code null} if the value is not set or if the option
+         * is not of type {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}.
+         *
+         * @return The minimum value for this option or {@code null}
+         */
+        public Double getMinValue()
```
MinnDevelopment, line 647:
```suggestion
        @Nullable
        public Double getMinValue()
```

src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
```diff
@@ -630,6 +637,30 @@ public OptionType getType()
...
+        }
+
+        /**
+         * The maximum value which can be provided for this option.
+         * <br>This returns {@code null} if the value is not set or if the option
+         * is not of type {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}.
+         *
+         * @return The maximum value for this option or {@code null}
+         */
+        public Double getMaxValue()
```
MinnDevelopment, line 659:
```suggestion
        @Nullable
        public Double getMaxValue()
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -191,6 +193,30 @@ public boolean isRequired()
...
     }
 
+    /**
+     * The minimum value which can be provided for this option.
+     * <br>This returns {@code null} if the value is not set or if the option
+     * is not of type {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}.
+     *
+     * @return The minimum value for this option
+     */
+    public Number getMinValue()
```
MinnDevelopment, line 203:
```suggestion
    @Nullable
    public Number getMinValue()
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -191,6 +193,30 @@ public boolean isRequired()
...
+    }
+
+    /**
+     * The maximum value which can be provided for this option.
+     * <br>This returns {@code null} if the value is not set or if the option
+     * is not of type {@link OptionType#INTEGER INTEGER} or {@link OptionType#NUMBER NUMBER}.
+     *
+     * @return The maximum value for this option
+     */
+    public Number getMaxValue()
```
MinnDevelopment, line 215:
```suggestion
    @Nullable
    public Number getMaxValue()
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -339,6 +365,102 @@ public OptionData setChannelTypes(@Nonnull Collection<ChannelType> channelTypes)
...
+     *         If any of the following checks fail
+     *         <ul>
+     *             <li>{@link OptionType type of this option} is {@link OptionType#NUMBER NUMBER}</li>
+     *             <li>{@code value} is lower than or equal to {@link OptionData#MAX_POSITIVE_NUMBER MAX_POSITIVE_NUMBER}</li>
+     *         </ul>
+     *
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData setMaxValue(double value)
```
MinnDevelopment, line 455: Is it possible to set only max and only min? Can we also have range setter to set both in one?

src/main/java/net/dv8tion/jda/api/interactions/commands/Command.java
```diff
@@ -550,6 +551,8 @@ private void setStringValue(@Nonnull String value)
         private final boolean required;
         private final Set<ChannelType> channelTypes;
         private final List<Choice> choices;
+        private Double minValue;
+        private Double maxValue;
```
MinnDevelopment, line 555: Any reason to not have these final?

caneleex: well, the values are reassigned in the constructor only if their json keys are not missing.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
caneleex: yes, it is. range method added.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -339,6 +368,160 @@ public OptionData setChannelTypes(@Nonnull Collection<ChannelType> channelTypes)
...
+     *             <li>{@link OptionType type of this option} is {@link OptionType#INTEGER INTEGER}</li>
+     *             <li>{@code value} is larger than or equal to {@link OptionData#MIN_NEGATIVE_NUMBER MIN_NEGATIVE_NUMBER}</li>
+     *         </ul>
+     *
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData setMinValue(long value)
+    {
+        if (type != OptionType.INTEGER)
```
MinnDevelopment, line 388: I mean, a long value is also a valid double value in our case since its restricted to 2^53 regardless and we should allow people to write code like `new OptionData(DOUBLE, "volume", "Set the music voluem").setMaxValue(1)`

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -339,6 +368,160 @@ public OptionData setChannelTypes(@Nonnull Collection<ChannelType> channelTypes)
...
+     *             <li>{@link OptionType type of this option} is {@link OptionType#INTEGER INTEGER}</li>
+     *             <li>{@code value} is lower than or equal to {@link OptionData#MAX_POSITIVE_NUMBER MAX_POSITIVE_NUMBER}</li>
+     *         </ul>
+     *
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData setMaxValue(long value)
+    {
+        if (type != OptionType.INTEGER)
```
MinnDevelopment, line 436: See my other comment

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -339,6 +368,160 @@ public OptionData setChannelTypes(@Nonnull Collection<ChannelType> channelTypes)
...
+     *             <li>{@code minValue} is larger than or equal to {@link OptionData#MIN_NEGATIVE_NUMBER MIN_NEGATIVE_NUMBER}</li>
+     *             <li>{@code maxValue} is lower than or equal to {@link OptionData#MAX_POSITIVE_NUMBER MAX_POSITIVE_NUMBER}</li>
+     *         </ul>
+     *
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData setRequiredRange(long minValue, long maxValue)
+    {
+        if (type != OptionType.INTEGER)
```
MinnDevelopment, line 487: Same here

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -339,6 +368,154 @@ public OptionData setChannelTypes(@Nonnull Collection<ChannelType> channelTypes)
...
 
+    /**
+     * Configure the minimal value which can be provided for this option.
+     *
+     * @param  value
+     *         The minimal value which can be provided for this option.
+     * @throws IllegalArgumentException
+     *         If any of the following checks fail
+     *         <ul>
+     *             <li>{@link OptionType type of this option} is {@link OptionType#INTEGER INTEGER}</li>
```
MinnDevelopment, line 379: We should check whether the type is `INTEGER` or `NUMBER`

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -581,6 +771,10 @@ public static OptionData fromData(@Nonnull DataObject json)
         OptionType type = OptionType.fromKey(json.getInt("type"));
         OptionData option = new OptionData(type, name, description);
         option.setRequired(json.getBoolean("required"));
+        if (!json.isNull("min_value"))
+            option.setMinValue(json.getDouble("min_value"));
+        if (!json.isNull("max_value"))
+            option.setMaxValue(json.getDouble("max_value"));
```
MinnDevelopment, line 777: This will throw if type is INTEGER, probably need to check that

caneleex: added, though I think there's a prettier way than a ternary here.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -581,6 +771,13 @@ public static OptionData fromData(@Nonnull DataObject json)
         OptionType type = OptionType.fromKey(json.getInt("type"));
         OptionData option = new OptionData(type, name, description);
         option.setRequired(json.getBoolean("required"));
+        if (type == OptionType.INTEGER || type == OptionType.NUMBER)
+        {
+            if (!json.isNull("min_value"))
+                option.setMinValue(type == OptionType.INTEGER ? json.getLong("min_value") : json.getDouble("min_value"));
+            if (!json.isNull("max_value"))
+                option.setMaxValue(type == OptionType.INTEGER ? json.getLong("max_value") : json.getDouble("max_value"));
```
MinnDevelopment, line 779: I'm pretty sure this will not call the double/long overload as you would expect, you will have to split it up to get the correct method.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -581,6 +771,23 @@ public static OptionData fromData(@Nonnull DataObject json)
...
+                if (json.isType("min_value", DataType.INT))
+                    option.setMinValue(json.getLong("min_value"));
+                else if (json.isType("min_value", DataType.FLOAT))
+                    option.setMinValue(json.getDouble("min_value"));
+            }
+            if (!json.isNull("max_value"))
+            {
+                if (json.isType("max_value", DataType.INT))
+                    option.setMaxValue(json.getLong("max_value"));
+                else if (json.isType("max_value", DataType.FLOAT))
```
caneleex, line 787: not sure if the type of the field should be checked because, lets say, the user provides some invalid type as value so the call will be ignored without any information.

MinnDevelopment: Users putting in invalid json should be none of our concern, we should expect the json to have the correct types and format.

MinnDevelopment (approved):

PR #1868, opened 2021-10-25
Implement Autocomplete

aasmart:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This introduces Discord's upcoming **autocomplete** feature for Slash Commands. This also features a refactor of the interaction hierarchy to accommodate for the new interaction type. 

### Using the Interaction
`OptionData` can now have autocomplete enabled with `option.setAutoComplete(true)`. This can **only** be enabled on options with types, Integer, Number, or String. Once autocomplete is enabled for an option, you may not add choices (It will not enable if choices are already present on the option).

To respond with choices, you can use the following:
```java
@Override
public void onCommandAutoComplete(@NotNull CommandAutoCompleteEvent event)
{
    event.respondChoices(
            new Command.Choice("choice-1", "one"),
            new Command.Choice("choice-2", "two"),
            new Command.Choice("choice-3", "three")
    ).queue();
}
```
You can also get the option the user is currently focused on with `event.getFocusedOption()`

### Discussion Points
* This implementation adds the `focused` value to OptionMapping, which I'm not sure should be present there or elsewhere
* `InteractionHook` was split into `InteractionHook` and `ChannelInteractionHook`. However `ChannelInteractionHook` cannot extend `InteractionHook`
* `InteractionCallbackAction` and `ChannelInteractionCallbackAction` are kind of janky because of the former
* `CommandAutoCompleteInteraction` and `CommandInteraction` share a large amount of identical code


AfonsoHackers:
Is this ever going to be implemented?

MinnDevelopment:
@AfonsoHackers yes.

AfonsoHackers:
> 

The question is when I suppose?

MinnDevelopment:
Superseded by #1971

Issue #1872, opened 2021-11-02
Calling `override(true)` on message edit does no longer remove embeds from original message


Andre601:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report
When sending a message with at least one MessageEmbed and later on edit said message with `override(true)` will the message still keep the embeds.

Only when calling `setEmbeds(Collections.emptyList())` will the embeds get removed by the edit.

I have encountered this while making slash commands. However, the message itself that is send and/or edited is a default message send by the bot account itself and not a webhook message or similar.

### Expected Behavior
`override(true)` should remove any existing embeds from the message, unless a new one was provided.

### Code Example or Reproduction Steps
```java
public class MessageTest extends ListenerAdapter
{
    @Override
    public void onMessageReceived(MessageReceivedEvent event)
    {
        if(event.getMessage().getContentRaw().equals("!test"))
        {
            MessageEmbed embed = new EmbedBuilder()
                .setDescription("Test embed")
                .build();
            
            event.getTextChannel().sendMessageEmbeds(embed).queue(
                m -> m.editMessage("Test without embed").override(true).queue()
            );
        }
    }
}
```

### Code for JDABuilder or DefaultShardManagerBuilder Used
```java
JDA jda = JDABuilder.createDefault("token_here")
    .addEventListeners(new MessageTest())
    .build()
```

### Exception or Error
N/A

PR #1873, opened 2021-11-02
Fix overriding messages with embeds and components

duncte123:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1872 

## Description

Fixes `override(true)` not clearing the embeds on a message. My assumption is that discord treats `null` the same way as the field not being present, hence sending an empty array fixes it.


Andre601:
From my testing should this work as providing a `Collections.emptyList()` does the trick.

duncte123:
> From my testing should this work as providing a `Collections.emptyList()` does the trick.

I had the same thought, just wanted to make sure that it did.

PR #1874, opened 2021-11-03
Add support for custom rate limiters

Fabricio20:
## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation

Closes Issue: NaN

## Description

This pull request adds support for implementing a custom rate limiter in JDA, as well as the ability to change the URL to which JDA sends requests to. Both of these features are meant to allow advanced users who may want to use a proxy that already handles rate limits with JDA.

As of right now, JDA is one of the few libraries that still isn't compatible with [Twilight](https://github.com/twilight-rs/http-proxy), this pull request fixes that by allowing the user to change the API endpoint to a custom one and to create a no-op rate limiter implementation.

This is doable in both the JDABuilder and ShardManagerBuilder, with the new `setRateLimiter` and `setEndpoint` methods. As requested by DV8, if a custom rate limiter is defined, the User-Agent used by the library is modified to include that as twilight does not modify headers.


MinnDevelopment:
Superseded by #2307

PR #1878, opened 2021-11-04
Removes unnecessary recalculation

koply:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation

## Description

Removed recalculation of all listeners on registration at AnnotatedEventManager. Also added the varargs for listener registrar method. A usable edit with no negative effects.

The token cannot be empty. Because pre-checked field. Removed unnecessary check.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
+++ b/src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
@@ -57,8 +57,10 @@ public class AnnotatedEventManager implements IEventManager
     @Override
     public void register(@Nonnull Object listener)
     {
-        listeners.add(listener);
-        addMethod(listener);
+        if (listeners.add(listener))
+        {
+            addMethod(listener);
+        }
     }
 
     @Override
--- a/src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
+++ b/src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
@@ -55,13 +55,10 @@ public class AnnotatedEventManager implements IEventManager
     private final Map<Class<?>, Map<Object, List<Method>>> methods = new ConcurrentHashMap<>();
 
     @Override
-    public void register(@Nonnull Object...pListeners)
+    public void register(@Nonnull Object listener)
     {
-        for (Object listenerObj : pListeners)
-        {
-            listeners.add(listenerObj);
-            addMethod(listenerObj);
-        }
+        listeners.add(listener);
+        addMethod(listener);
     }
 
     @Override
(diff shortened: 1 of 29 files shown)


src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
```diff
@@ -55,11 +55,12 @@
     private final Map<Class<?>, Map<Object, List<Method>>> methods = new ConcurrentHashMap<>();
 
     @Override
-    public void register(@Nonnull Object listener)
+    public void register(@Nonnull Object...pListeners)
```
MinnDevelopment, line 58: This changes the method signature in a breaking way. The ABI method signature becomes an array which means all code referencing this method must be recompiled, as such this is breaking.

MusaBrt: I'll edit it so that it doesn't change the method signature.

MinnDevelopment:
Should we add unit tests for this?

MusaBrt:
I don't think it should be. My attempts works well. The MessageListenerExample.java works fine.

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -278,7 +278,7 @@ public int login(String gatewayUrl, ShardInfo shardInfo, Compression compression
 
         String token = authConfig.getToken();
         setStatus(Status.LOGGING_IN);
-        if (token == null || token.isEmpty())
+        if (token.isEmpty())
```
Sanduhr32, line 281: THIS WILL NPE.

Sanduhr32 (changes requested):
I think this chance should be reverted.

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
MusaBrt: Check this class. Nothing like that will happen. 

https://github.com/DV8FromTheWorld/JDA/blob/development/src/main/java/net/dv8tion/jda/internal/utils/config/AuthorizationConfig.java

MusaBrt:
Nothing like that will happen.

https://github.com/DV8FromTheWorld/JDA/blob/development/src/main/java/net/dv8tion/jda/internal/utils/config/AuthorizationConfig.java

bump.

Sanduhr32:
I guess, but well

DV8FromTheWorld:
This needs to be rebased against `master`

src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
```diff
@@ -114,31 +114,36 @@ private void updateMethods()
...
         {
-            boolean isClass = listener instanceof Class;
-            Class<?> c = isClass ? (Class) listener : listener.getClass();
-            Method[] allMethods = c.getDeclaredMethods();
-            for (Method m : allMethods)
+            addMethod(listener);
+        }
+    }
+
+    private void addMethod(@Nonnull Object listener)
```
DV8FromTheWorld, line 121: Lets rename this `scanListenerForMethods` or something similar as `addMethod` doesn't really do it for me.

MinnDevelopment:
Superseded by #2454

PR #1879, opened 2021-11-05
Added replyEphemeral method

RealYusufIsmail:

## Pull Request Etiquette

### Changes
I have added a replyEphemeral method which makes it easer for the programmer to create an ephemeral message
instead of always doing .setEphemeral(true) since its quite repetitive.

- [x] Internal code
- [ ] Documentation

Closes Issue: NaN

## Description
Instead if always doing `event.reply("Test).setEphemeral(true)` it makes it easer for the programmer just to do  `event.replyEphemeral("")` which makes it easer for a programmer to create an ephemeral message.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/Interaction.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/Interaction.java
@@ -255,6 +255,27 @@ public interface Interaction extends ISnowflake
 
     /**
      * Reply to this interaction and acknowledge it.
+     * <br>This will send an ephemeral reply message for this interaction.
+     * <p><b>You only have 3 seconds to acknowledge an interaction!</b>
+     *
+     *<br>When the acknowledgement is sent after the interaction expired, you will receive {@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_INTERACTION ErrorResponse.UNKNOWN_INTERACTION}.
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/interactions/Interaction.java
```diff
@@ -253,6 +253,27 @@ default ReplyAction reply(@Nonnull String content)
...
+     * @throws IllegalArgumentException
+     *         If null is provided or the content is empty or longer than {@link Message#MAX_CONTENT_LENGTH}
+     *
+     *@return {@link ReplyAction}
+     */
+    @Nonnull
+    @CheckReturnValue
+    default ReplyAction replyEphemeral(@Nonnull String content) {
+        Checks.notNull(content, "Content");
+        return deferReply().setContent(content).setEphemeral(true);
```
illuminator3, line 274:
```suggestion
        return reply(content).setEphemeral(true);
```

MinnDevelopment:
We don't want to add an overload for every reply method just so you don't need to call a setter. This is unnecessary and a maintenance burden. Calling a setter isn't a big burden for the programmer, and this only makes it marginally less code to write.

CanIGetaPR:
> Instead if always doing `event.reply("Test).setEphemeral(true)` it makes it easer for the programmer just to do `event.replyEphemeral("")` which makes it easer for a programmer to create an ephemeral message.

You could wrap the event class in a custom class and make custom methods like the one you describe, among other conveniences.

PR #1883, opened 2021-11-06
fix firing SelfUpdateDiscriminatorEvent

caneleex:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this PR fixes handling of the SelfUpdateDiscriminatorEvent.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
+++ b/src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
@@ -163,7 +163,7 @@ public abstract class ListenerAdapter implements EventListener
 
     //Self Events. Fires only in relation to the currently logged in account.
     public void onSelfUpdateAvatar(@Nonnull SelfUpdateAvatarEvent event) {}
-    public void onSelfUpdateDisciminator(@Nonnull SelfUpdateDiscriminatorEvent event) {}
+    public void onSelfUpdateDiscriminator(@Nonnull SelfUpdateDiscriminatorEvent event) {}
     public void onSelfUpdateMFA(@Nonnull SelfUpdateMFAEvent event) {}
     public void onSelfUpdateName(@Nonnull SelfUpdateNameEvent event) {}
     public void onSelfUpdateVerified(@Nonnull SelfUpdateVerifiedEvent event) {}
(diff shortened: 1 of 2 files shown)


MinnDevelopment (approved):

DV8FromTheWorld:
Last time this was discussed in chat I brought up the fact that it would be an awkward experience for the developer to need to track two separate events to know the full informational change of a name + discriminator change.

You'd need to write a system that depended on the internals of JDA's event firing order to fully capture this related data:
```java
ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
Map<String, String[]> updatedNames = new HashMap<>;
Map<String, ScheduledFuture> futures = new HashMap<>;

public void onUserUpdateName(UserUpdateNameEvent event) {
  String userId = event.getUser().getId();
  updatedNames.put(userId, new String[] { event.getOldName(), event.getNewName() });
  
  Runnable runnable = new Runnable() {
    @Override
    public void run() {
      String currentDiscrim = event.getUser().getDiscriminator();
      handleUserUpdate(userId, currentDiscrim, currentDiscrim);
    }
  }
  
  futures.put(userId, executor.schedule(runnable, 5, TimeUnit.SECONDS));
}

public void onUserDiscriminatorUpdate(UserDistriminatorUpdateEvent event) {
  String userId = event.getUser().getId();
  if (updatedNames.contains(userId) {
    Future future = futures.remove(userId);
    future.cancel();
    
    handleUserUpdate(userId, event.getOldDiscriminator(), event.getNewDiscriminator());
  }
}

private void handleUserUpdate(String userId, String oldDiscriminator, String newDiscriminator) {
  String[] names = updatedNames.remove(userId);
  String oldName = names[0];
  String newName = names[1];
  //Do logic here
}
```

I really feel like if the name or discriminator changes we should be dispatching the same event with both changed pieces of data given that they're super intertwined. I also recognize that this PR is for the Self event, but this problem exists for the User events as well and we should probably consider how we want the behavior of both to behave as they're basically the same event.

DV8FromTheWorld:
Bump. We need to make a decision on this.

caneleex:
I'm fine with closing the PR and letting you/others decide how to do this properly

DV8FromTheWorld:
I think this PR is probably a good place to discuss how we want JDA to handle this.

caneleex:
It will be better if you discuss this in #lib-dev, I don't have time to deal with this. once you decide how to do this, you can just make a PR/push on your own without having to wait for me.

caneleex:
closing this since I'm not interested in Java/JDA anymore. don't want to keep this stale/open forever.

PR #1888, opened 2021-11-10
Don't throw exception when providing empty channel types and choices in OptionData

CanIGetaPR:
## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation

## Description

Don't throw IllegalAccessException for empty channel types and choices collections in option data. 

If the collection is empty the caller probably wasn't trying to provide channel types and choices in the first place and the empty collection may be a place value.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
@@ -342,7 +342,7 @@ public class OptionData implements SerializableData
      * @throws IllegalArgumentException
      *         If any of the following checks fail
      *         <ul>
-     *             <li>{@link OptionType type of this option} is {@link OptionType#CHANNEL CHANNEL}</li>
+     *             <li>{@link OptionType type of this option} is {@link OptionType#CHANNEL CHANNEL} and the collection is not empty</li>
      *             <li>{@code channelTypes} is not null</li>
      *             <li>{@code channelTypes} doesn't contain {@code null}</li>
      *             <li>{@code channelTypes} only contains guild channels</li>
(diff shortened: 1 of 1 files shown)


freya022 (changes requested):

MinnDevelopment:
Setting empty lists is usually a logic error, so throwing an exception makes sense.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -677,6 +682,8 @@ public OptionData addChoice(@Nonnull String name, @Nonnull String value)
     @Nonnull
     public OptionData addChoices(@Nonnull Command.Choice... choices)
     {
+        if (choices.length == 0)
+            return this;
         if (this.choices == null)
             throw new IllegalStateException("Cannot add choices for an option of type " + type);
         Checks.noneNull(choices, "Choices");
```
Tais993, line 689:
```suggestion
        if (this.choices == null)
            throw new IllegalStateException("Cannot add choices for an option of type " + type);
        Checks.noneNull(choices, "Choices");
        if (choices.length == 0)
            return this;
```

Current code is vulnerable to NPE if the array is null

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -342,7 +342,7 @@ public OptionData setChannelTypes(@Nonnull ChannelType... channelTypes)
      * @throws IllegalArgumentException
      *         If any of the following checks fail
      *         <ul>
-     *             <li>{@link OptionType type of this option} is {@link OptionType#CHANNEL CHANNEL}</li>
+     *             <li>{@link OptionType type of this option} is {@link OptionType#CHANNEL CHANNEL} and the collection is not empty</li>
```
Tais993, line 345:
```suggestion
     *             <li>{@link OptionType type of this option} is {@link OptionType#CHANNEL CHANNEL} and the collection is not empty</li>
```
Docs and code are different.

Docs say "collection is empty? IllegalArgumentException"
While the code ignores empty collections.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -666,7 +671,7 @@ public OptionData addChoice(@Nonnull String name, @Nonnull String value)
      * @throws IllegalArgumentException
      *         If any of the following checks fail
      *         <ul>
-     *             <li>The {@link OptionType} does {@link OptionType#canSupportChoices() support choices}</li>
+     *             <li>The {@link OptionType} does {@link OptionType#canSupportChoices() support choices} and some are provided.</li>
```
Tais993, line 674:
```suggestion
     *             <li>The {@link OptionType} does {@link OptionType#canSupportChoices() support choices} and some are provided.</li>
```
Imo this sounds really vague, not sure how to phrase it differently though

Tais993 (changes requested):

MinnDevelopment:
Closing because this changes the runtime checks in breaking ways, the current behavior makes more sense in my opinion.

CanIGetaPR:
Thanks for the feedback everyone!

Issue #1887, opened 2021-11-09
SEND_MESSAGES_IN_THREADS permission


ToxicMushroom:
It would be useful to have an enum in [Permission.java](https://github.com/DV8FromTheWorld/JDA/blob/development/src/main/java/net/dv8tion/jda/api/Permission.java) for this permission 
`SEND_MESSAGES_IN_THREADS | 0x4000000000 (1 << 38) | Allows for sending messages in threads | T`

[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features.

### Example Use-Case
consistency with the discord docs, being able to use for loops over the enum values to do something for all permissions.

DV8FromTheWorld:
This is done in https://github.com/DV8FromTheWorld/JDA/releases/tag/v5.0.0-alpha.1

PR #1889, opened 2021-11-10
Added Send Messages in Threads to Permission enum

TheSuperGamer20578:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1887 

## Description

Adds Send Messages in Threads to the Permision Enum

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/Permission.java
+++ b/src/main/java/net/dv8tion/jda/api/Permission.java
@@ -54,6 +54,7 @@ public enum Permission
     MESSAGE_EXT_EMOJI(       18, true, true, "Use External Emojis"),
     MESSAGE_EXT_STICKER(     37, true, true, "Use External Stickers"),
     USE_SLASH_COMMANDS(      31, true, true, "Use Slash Commands"),
+    SEND_MESSAGES_IN_THREADS(38, true, true, "Send Messages in Threads"),
 
     MANAGE_THREADS(     34, true, true, "Manage Threads"),
     USE_PUBLIC_THREADS( 35, true, true, "Use Public Threads"),


Xirado:
Already implemented in [v5](https://github.com/DV8FromTheWorld/JDA/blob/3e0d2f868e58af733eb50591ed19f601fdb20541/src/main/java/net/dv8tion/jda/api/Permission.java#L66)

ToxicMushroom:
> Already implemented in [v5](https://github.com/DV8FromTheWorld/JDA/blob/3e0d2f868e58af733eb50591ed19f601fdb20541/src/main/java/net/dv8tion/jda/api/Permission.java#L66)
I don't feel like this is an addition that should be in some other feature's PR

Tais993:
> 
> 
> > Already implemented in [v5](https://github.com/DV8FromTheWorld/JDA/blob/3e0d2f868e58af733eb50591ed19f601fdb20541/src/main/java/net/dv8tion/jda/api/Permission.java#L66)
> 
> I don't feel like this is an addition that should be in some other feature's PR

What do you mean with "other feature's PR"?
It makes sense to be in v5, v5 also adds full Thread support and more.

Chew:
I still think permissions should be added here, even if not used internally. Knowing what permissions a user has regardless of the library's native support for them should be included in the LTS, as otherwise, they would not show up when querying a user or role's permissions, however they may be used.

TheSuperGamer20578:
> By this logic, all the missing permissions should be here (manage events, play activities, etc)

I beleve the activity permission is already implemented
https://github.com/DV8FromTheWorld/JDA/blob/a35789442ba022c81ce9b5f62d5b6b9968e3895d/src/main/java/net/dv8tion/jda/api/Permission.java#L70

DV8FromTheWorld:
I understand the argument, but we considered this to be a feature request. `v4` is closed to feature requests and will only be handling bugs.
`v5` has thread support, and if you wish to control permissions for threads then you need to update to it.

PR #1892, opened 2021-11-10
Add missing ErrorResponse values

Xirado:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This pull requests adds all missing ErrorResponse Enum values.


Andre601:
I spotted two things that confuse me a bit:

1. The longest enum name had no spaces after the `(`, so it was `NAME(1234, "text")` but your change adds spaces where they aren't really needed.
2. Numbers are padded on the left, so that they line up on the right, which your change is not following:
   ```
   # Original
    123
   1234
   
   # Your change
   123
   1234
   ```

Those are obviously only visual changes, but I still believe they should be something important to consider. But let's wait for the input of Minn or DV8.

src/main/java/net/dv8tion/jda/api/requests/ErrorResponse.java
```diff
@@ -37,83 +37,148 @@
...
+    BEFORE_VALUE_EARLIER_THAN_THREAD_CREATION(50085, "\"before\" value is earlier than the thread creation date"),
+    SERVER_NOT_AVAILABLE_IN_YOUR_LOCATION(    50095, "This server is not available in your location"),
+    SERVER_MONETIZATION_DISABLED(             50097, "This server needs monetization enabled in order to perform this action"),
+    SERVER_NOT_ENOUGH_BOOSTS(                 50101, "This server needs more boosts to perform this action"),
+    MFA_NOT_ENABLED(                          60003, "MFA auth required but not enabled"),
+    NO_USER_WITH_TAG_EXISTS(                  80004, "No users with DiscordTag exist"),
+    REACTION_BLOCKED(                         90001, "Reaction Blocked"),
+    RESOURCES_OVERLOADED(                    130000, "Resource overloaded"),
+    STAGE_ALREADY_OPEN(                      150006, "The Stage is already open"),
+    MISSING_READ_MESSAGE_HISTORY_PERM(       160002, "Cannot reply without permission to read message history"),
```
DV8FromTheWorld, line 168: This name seems to be in conflict with its description. This naming could result it in it implying that it applies to many things.

DV8FromTheWorld (approved):

PR #1906, opened 2021-11-28
Change publishing logic

MinnDevelopment:
With this new logic, we can publish to maven central and jitpack.

If the version is bumped (e.g. 5.0.0 -> 5.0.1), we release the artifact to maven central and upload the jars to jenkins as usual.

If the version is not bumped and we only have a new commit, we only build the jars and javadoc on jenkins and append the commit hash (abbreviated to 7 chars) to the artifact version.

For people who are interested in using snapshot builds between version bumps, you can either use jitpack to get specific commit versions or download jars from jenkins. We do not intend to publish every single commit to maven central.

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,13 +437,6 @@ Be sure to replace the **VERSION** key below with the one of the versions shown 
     <version>VERSION</version>
 </dependency>
 ```
-```xml
-<repository>
-    <id>dv8tion</id>
-    <name>m2-dv8tion</name>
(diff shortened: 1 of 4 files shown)


DV8FromTheWorld (approved):
Fantastic work

MinnDevelopment:
I've moved opus-java to maven central, so we no longer need the s3 bucket for our dependencies.

PR #1907, opened 2021-11-28
Add missing region

Xavinlol:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds a new region "Rotterdam" to the Region enum and removes old regions which aren't available to use anymore.
Possible voice channel regions updated (addition of "Rotterdam" and "Hong Kong" and removal of "Europe")


Chew:
Wouldn't removing regions be a breaking change? If so they should target v5, but adding to v4 should be fine?

Xavinlol:
> Wouldn't removing regions be a breaking change? If so they should target v5, but adding to v4 should be fine?

Hey, yeah true that would be a breaking change actually.
I'll revert the removal of regions again, so that it's a non breaking change for v4 but a correct reflection of current possible voice server regions.
Thank you

DV8FromTheWorld:
The `development` branch will be removed soon. `v4` now exists on `legacy/v4`.
However, development on v4 has come to an end. Only critical bug fixes will be backported to `v4`.

An addition of this mission region can be accepted into `v4` as a bug fix, however, I'd also suggest you make a PR to `master` which is `v5` and is where the future of JDA lives.

PR #1909, opened 2021-11-29
Remove all deprecated code from v5

DV8FromTheWorld:
Removes all deprecated from v5. Made as a PR so that contributes can help confirm that all code that supported these deprecated features is removed.


src/main/java/net/dv8tion/jda/api/MessageBuilder.java
```diff
@@ -854,112 +833,6 @@ public MessageBuilder mentionRoles(@Nonnull long... roles)
...
-     *
-     * @deprecated This is not a reliable way to remove mentions from the content,
-     *             you should use {@link #setAllowedMentions(Collection)} instead.
-     */
-    @Nonnull
-    @Deprecated
-    @ForRemoval(deadline = "4.4.0")
-    @ReplaceWith("denyMentions(types)")
-    @DeprecatedSince("4.2.0")
-    public MessageBuilder stripMentions(@Nonnull JDA jda, @Nonnull Message.MentionType... types)
```
MinnDevelopment, line 958: I think there is a bit related code here that also needs to be removed like SplitPolicy and the private helper functions

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -443,40 +443,6 @@ default CommandEditAction editCommandById(long id)
...
-     *
-     * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type {@link java.util.EnumSet EnumSet}
-     *
-     * @deprecated Guilds no longer have the {@link net.dv8tion.jda.api.Region Region} option. Use {@link VoiceChannel#getRegion()} instead.
-     */
-    @Nonnull
-    @CheckReturnValue
-    @Deprecated
-    @ForRemoval(deadline = "5.0.0")
-    RestAction<EnumSet<Region>> retrieveRegions(boolean includeDeprecated);
```
MinnDevelopment, line 478: I think this endpoint should still be supported, not sure why it was deprecated?

Issue #1921, opened 2021-11-30
Casting error in Message methods for permission checks


MinnDevelopment:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

Calling `Message#delete` on a message in a thread, causes a casting error.

### Expected Behavior

No error.

### Code Example or Reproduction Steps

```java
public void onMessageReceived(MessageReceivedEvent event) {
  if (event.getChannelType().isThread()) {
    event.getMessage().delete(); // throws 
  }
}
```

### Code for JDABuilder or DefaultShardManagerBuilder Used

N/A

### Exception or Error

```
java.lang.ClassCastException: class net.dv8tion.jda.internal.entities.ThreadChannelImpl cannot be cast to class net.dv8tion.jda.api.entities.IPermissionContainer (net.dv8tion.jda.internal.entities.ThreadChannelImpl and net.dv8tion.jda.api.entities.IPermissionContainer are in unnamed module of loader 'app')
	at net.dv8tion.jda.internal.entities.ReceivedMessage.delete(ReceivedMessage.java:922)
	at FooKt$main$$inlined$listener$1.onEvent(events.kt:348)
	at dev.minn.jda.ktx.CoroutineEventManager$handle$1.invokeSuspend(CoroutineEventManager.kt:41)
```

TheCoolOne123:
I can confirm this bug in my own bot. 
https://gitlab.com/pinewood-builders/discord/xeus/-/blob/feature/jda-5/src/main/java/com/pinewoodbuilders/handlers/adapter/MessageEventAdapter.java#L304 

When trying to delete a message from a thread, it can't. And trows an error missing that casting

PR #1925, opened 2021-12-01
Implement parent channel fetching when checking thread permissions

rtm516:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1921

## Description

When trying to get the IPermissionContainer for a thread it now gets it from the parent to stop a ClassCastException.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
@@ -919,7 +919,7 @@ public class ReceivedMessage extends AbstractMessage
             if (isFromType(ChannelType.PRIVATE))
                 throw new IllegalStateException("Cannot delete another User's messages in a PrivateChannel.");
 
-            IPermissionContainer permChan = (IPermissionContainer) channel;
+            IPermissionContainer permChan = getPermissionContainer();
             if (!getGuild().getSelfMember().hasAccess(permChan))
                 throw new MissingAccessException(permChan, Permission.VIEW_CHANNEL);
             else if (!getGuild().getSelfMember().hasPermission(permChan, Permission.MESSAGE_MANAGE))
@@ -940,7 +940,7 @@ public class ReceivedMessage extends AbstractMessage
             if (isFromType(ChannelType.PRIVATE))
                 throw new PermissionException("Cannot suppress embeds of others in a PrivateChannel.");
 
-            IPermissionContainer permChan = (IPermissionContainer) channel;
+            IPermissionContainer permChan = getPermissionContainer();
             if (!getGuild().getSelfMember().hasPermission(permChan, Permission.MESSAGE_MANAGE))
                 throw new InsufficientPermissionException(permChan, Permission.MESSAGE_MANAGE);
         }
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -919,7 +919,7 @@ private void checkUser()
             if (isFromType(ChannelType.PRIVATE))
                 throw new IllegalStateException("Cannot delete another User's messages in a PrivateChannel.");
 
-            IPermissionContainer permChan = (IPermissionContainer) channel;
+            IPermissionContainer permChan = getPermissionContainer();
```
MinnDevelopment, line 922: I think you can just directly use `getGuildChannel()` instead, or no?

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -1087,6 +1087,24 @@ public void setMentions(List<User> users, List<Member> members)
         return collection;
     }
 
+    private IPermissionContainer getPermissionContainer()
+    {
+        IPermissionContainer permChan;
+
+        // Check if the channel is a thread
+        if (isFromType(ChannelType.GUILD_NEWS_THREAD) || isFromType(ChannelType.GUILD_PUBLIC_THREAD) || isFromType(ChannelType.GUILD_PRIVATE_THREAD))
```
MinnDevelopment: You should use `ChannelType#isThread` for these checks usually, but using guild channel makes this helper method unnecessary

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
rtm516: No, ends up with the same error.

MinnDevelopment: I do not get an error trying to do this:

```kt
val channel = event.channel as GuildChannel // in a thread
val bool = event.member!!.hasPermission(channel, Permission.MESSAGE_SEND)
println(bool) // true
```

DV8FromTheWorld: `hasPermission` accepts `GuildChannel` and knows to call `GuildChannel#getPermissionContainer()`

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -940,7 +940,7 @@ else if (!getGuild().getSelfMember().hasPermission(permChan, Permission.MESSAGE_
             if (isFromType(ChannelType.PRIVATE))
                 throw new PermissionException("Cannot suppress embeds of others in a PrivateChannel.");
 
-            IPermissionContainer permChan = (IPermissionContainer) channel;
+            IPermissionContainer permChan = getPermissionContainer();
             if (!getGuild().getSelfMember().hasPermission(permChan, Permission.MESSAGE_MANAGE))
```
DV8FromTheWorld, line 944: Same statement as above. We don't need a concrete instance of `IPermissionContainer` here.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -1087,6 +1087,24 @@ public void setMentions(List<User> users, List<Member> members)
...
+            // Get the parent channel for permissions if we are using a thread
+            permChan = ((ThreadChannel) channel).getParentChannel();
+        }
+        else
+        {
+            permChan = (IPermissionContainer) channel;
+        }
+
+        return permChan;
+    }
```
DV8FromTheWorld, line 1106: This is entirely unneeded. `GuildChannel#getPermissionContainer() : IPermissionContainer` exists and should be used when needing to get an `IPermissionContainer` in unknown concrete channel situations. 

However, we don't actually need an `IPermissionContainer` as stated above.

DV8FromTheWorld (changes requested):

MinnDevelopment (approved):

DV8FromTheWorld (approved):

Issue #1932, opened 2021-12-02
GuildManager#setVanityCode doesn't work


MinnDevelopment:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting
  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

You can't set a vanity code through the PATCH guild endpoint. There is a dedicated endpoint for this: `PATCH /guilds/:guild_id/vanity-url` with body `{"code": "..."}` (set `code` to empty string to remove it). This seems to be undocumented, though.

### Expected Behavior

The setter should be removed.

### Exception or Error

N/A

PR #1933, opened 2021-12-03
Remove GuildManager#setVanityCode

caneleex:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1932

## Description

this PR removes `GuildManager#setVanityCode` as the vanity code can't be modified using the `Modify Guild` endpoint.


DV8FromTheWorld:
If we're going to update the bitflags to shift them down in this PR, we might want to update the gap left by the loss of `REGION`. @MinnDevelopment Thoughts?

MinnDevelopment:
Since we are still in alpha, we can probably remove the gap for now. In the future, this would be treated as a breaking change.

DV8FromTheWorld:
Okay. @caneleex can you address the `REGION` gap as well in the bitmasks?

DV8FromTheWorld (approved):

PR #1938, opened 2021-12-05
First steps towards a wiki synchronization

Andre601:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: New GitHub Actions 

Closes Issue: NaN

## Description

This has been discussed on the Discord already, but it would be a good idea to have a two-way wiki synchronization that would allow outsiders to suggest changes to the wiki, without the risk of giving everyone write access to it.

This is still a draft since I need to find a somewhat elegant way to also have a `wiki -> wiki-folder` synchronization... Probably with a scheduled workflow and some git actions, but some assistance here is welcome.

I also need to copy over the wiki's content to a wiki folder (Unless you want the scheduled workflow do this on its own) too...

### DIFF
--- a/.github/workflows/wiki_push.yml
+++ b/.github/workflows/wiki_push.yml
@@ -1,13 +1,10 @@
-#
 # Workflow to push changes from the "wiki" folder in the repository towards the actual wiki itself.
-#
+
 name: 'Push to wiki'
 
 on:
   push:
-    #
     # Only push when changed files are in the wiki/ folder and on the master branch.
-    #
     paths:
     - 'wiki/**'
     branches:
--- a/.github/workflows/wiki_push.yml
+++ b/.github/workflows/wiki_push.yml
@@ -1,0 +1,34 @@
+#
+# Workflow to push changes from the "wiki" folder in the repository towards the actual wiki itself.
+#
+name: 'Push to wiki'
+
+on:
+  push:
+    #
+    # Only push when changed files are in the wiki/ folder and on the master branch.
+    #
+    paths:
+    - 'wiki/**'
+    branches:
+    - 'master'
+    #
+    # Ignore any release to avoid accidental trigger of this action.
+    #
+    tags-ignore:
+    - '**'
+
+jobs:
+  push:
+    runs-on: ubuntu-latest
+    steps:
+    - name: 'Checkout Code'
+      uses: 'actions/checkout@v2'
+    - name: 'Push changes to the wiki'
+      uses: 'Andre-Chen-Wang/github-wiki-action@v2'
+      env:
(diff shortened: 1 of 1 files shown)


.github/workflows/wiki_push.yml
```diff
@@ -0,0 +1,34 @@
+#
+# Workflow to push changes from the "wiki" folder in the repository towards the actual wiki itself.
+#
```
MinnDevelopment, line 3:
```suggestion
# Workflow to push changes from the "wiki" folder in the repository towards the actual wiki itself.

```

.github/workflows/wiki_push.yml
```diff
@@ -0,0 +1,34 @@
...
+name: 'Push to wiki'
+
+on:
+  push:
+    #
+    # Only push when changed files are in the wiki/ folder and on the master branch.
+    #
```
MinnDevelopment, line 10:
```suggestion
    # Only push when changed files are in the wiki/ folder and on the master branch.
```

.github/workflows/wiki_push.yml
```diff
@@ -0,0 +1,34 @@
...
+    paths:
+    - 'wiki/**'
+    branches:
+    - 'master'
+    #
+    # Ignore any release to avoid accidental trigger of this action.
+    #
+    tags-ignore:
+    - '**'
```
MinnDevelopment, line 19: Is this even needed if you check for `wiki/**` anyway?

.github/workflows/wiki_push.yml
```diff
@@ -0,0 +1,34 @@
+#
+# Workflow to push changes from the "wiki" folder in the repository towards the actual wiki itself.
+#
+name: 'Push to wiki'
```
MinnDevelopment, line 4: You don't need quotes, yaml automatically understands its a string.

.github/workflows/wiki_push.yml
Andre601: yes, because a release causes a push event of the entire repository. Had this many times before.

.github/workflows/wiki_push.yml
Andre601: Still have it for consistency.

DV8FromTheWorld:
We will probably have a `wiki` branch explicitly for the wiki. Unlike normal documentation which gets built into the code and javadocs, the wiki is completely separate, so we will treat it like the `assets` branch.

DV8FromTheWorld:
This effort has been superseded by https://github.com/dv8fromtheworld/jda-website

Andre601:
👍

PR #1942, opened 2021-12-09
add more functions for when message edit

OMAR-3OOV:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

I just realize that there is no way to delete the buttons when I need to edit the message only way is to use `setActionRow`, I know there is a lot of codes that can clear it or remove it but I still think that if there is **clearActionRow** or **removeActionRow** and there are a specific type of Action for examples: `ActionRow.BUTTONS` / `ActionRow.SelectMenu`, it will be easier for developers to handle everything on one line code. hopefully, it's clear enough!


Xirado:
Why are you merging v5 into master

OMAR-3OOV:
> Why are you merging master into v5

tbh, I don't have any idea this is my first time creating this thing, just suggest adding this think

MinnDevelopment:
There already exist clear action row methods, they are just not named like that. [`Message#editMessageComponents`](https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/entities/Message.html#editMessageComponents(java.util.Collection)) and [`MessageAction#setActionRows`](https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/requests/restaction/MessageAction.html#setActionRows(java.util.Collection)) can both accept empty lists, which effectively removes the action rows/buttons from the message.

Andre601:
Alternatively could the `override(true)` method be called while not setting any action rows, which has the same result in the end, which is removing the action rows.

PR #1949, opened 2021-12-14
Add IGuildChannelContainer and getChannelById(Class, id)

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This moves all the guild channel cache getters to a dedicated interface. With this interface, we can easily update documentation and maintain the cache getters in the future.

I've also added `getChannelById(Class, id)` to allow easy access to channels by types such as `MessageChannel`: `guild.getChannelById(MessageChannel.class, id)`

With this change, I also extended the docs to be more detailed. I added a section to *all* cache getters in this interface to explain cache scoping and cache initialization.


MinnDevelopment:
This PR could also address the idea to move all channel types to a dedicated subpackage of entities, thoughts?

DV8FromTheWorld:
We know that we definitely want to move channels to their own package but this PR currently sets as a `proposal` PR, so combining them might not be smart unless we want to remove the `proposal` marker.

I'm fine with accepting this proposal after review though, so combining them would probably be fine.

DV8FromTheWorld (approved):
Just needs class level docs and then this is mergeable.

DV8FromTheWorld (approved):

Issue #1920, opened 2021-11-30
Move canTalk from GuildMessageChannel to MessageChannel


MinnDevelopment:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request

It would make life a lot easier to just quickly check if a member can send messages in a channel. For private channel this could probably check the channel user.

### Example Use-Case

When handling an interaction, I want to be able to simply check this with a util method.

DV8FromTheWorld:
Completed in #1956

PR #1956, opened 2021-12-19
Move canTalk() from GuildMessageChannel to MessageChannel

V-Play-Games:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____

Closes Issue: #1920

## Description

Move the canTalk() method from GuildMessageChannel to MessageChannel, which calls for an implementation in PrivateChannel since GuildMessageChannel already has a default one.
For that, I simply use the condition used in the private checkBot() method
This is my first ever pull request for this repository, so please tell me if I am doing anything wrong here, ty

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
@@ -85,7 +85,7 @@ public class PrivateChannelImpl extends AbstractChannelImpl<PrivateChannelImpl> 
     @Override
     public boolean canTalk()
     {
-        return !getUser().isBot() || getJDA().getAccountType() != AccountType.BOT;
+        return true;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -82,6 +82,12 @@ public long getLatestMessageIdLong()
         return latestMessageId;
     }
 
+    @Override
+    public boolean canTalk()
+    {
+        return !getUser().isBot() || getJDA().getAccountType() != AccountType.BOT;
```
Tais993, line 88: Bots can talk in dms?

V-Play-Games: Changed it with true since it's possible

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -82,6 +82,12 @@ public long getLatestMessageIdLong()
         return latestMessageId;
     }
 
+    @Override
+    public boolean canTalk()
+    {
+        return true;
```
DV8FromTheWorld, line 88: @MinnDevelopment correct me if I'm wrong, but this should probably be `!user.isBot()`, right?

MinnDevelopment: I guess we could, as long as the docs outline that we can't check for blocked users

DV8FromTheWorld (approved):

Issue #1961, opened 2021-12-21
Voice regions modifications errors


JumperBoost:
[download]: https://bintray.com/dv8fromtheworld/maven/JDA/_latestVersion
[guild]: https://discord.gg/0hMr4ce0tIk3pSjp
[stack overflow]: https://stackoverflow.com/questions/tagged/java

## General Troubleshooting

  
- [X] I have checked for similar issues.
- [X] I have updated to the [latest JDA version][download].
- [X] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

## Bug Report

When we changed the region of a voice channel (rtc_region), an error appears.
It's appears only with EUROPE (not exists from discord) and HONG_KONG (not in VOICE_CHANNEL_REGIONS list [from Region.java]).

### Expected Behavior

Should be able to change voice channel's region with Europe (Rotterdam instead currently) and Hong-Kong.

### Exception or Error + Code Example or Reproduction Steps

Europe:
![image](https://user-images.githubusercontent.com/38703849/146970859-47120ffd-1337-4f9d-9601-e591f5da5293.png)
Code:  `JDA.getVoiceChannelById("ID").getManager().setRegion(Region.EUROPE).queue();`

Hong-Kong:
![image](https://user-images.githubusercontent.com/38703849/146970643-9c0de7a0-a8fc-4f7e-ad02-495cadd5e466.png)
Code:  `JDA.getVoiceChannelById("ID").getManager().setRegion(Region.HONG_KONG).queue();`

To fix that, i think we should add Rotterdam, replace Europe by Rotterdam (apparently the only European discord server) & add HONG_KONG from VOICE_CHANNEL_REGIONS in Region.java.

MinnDevelopment:
I was told there might be a new EU region in milan soon

JumperBoost:
You're right, I was able to find a server in Milan, but it would be nice to have Milan and Rotterdam, since I have more ping in Milan. It's not possible to select Milan actualy but soon i think.

Milan:
![image](https://user-images.githubusercontent.com/38703849/146981853-a50547e8-6cfd-4cdd-8f2e-2de5890ea113.png)

Rotterdam:
![image](https://user-images.githubusercontent.com/38703849/146981902-bec4355f-4e26-4785-8855-14f1f1a28c7b.png)

PR #1962, opened 2021-12-21
Change some voice Regions and remove VOICE_CHANNEL_REGIONS set

caneleex:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1961 

## Description

this PR removes some unused regions and adds a new `MILAN` region, which is not usable at the moment, but it's possible to get it in the client. the `SOUTH_KOREA` region is not listed in the regions endpoint (or in the client for that matter), but it's still applicable. this PR also removes the `VOICE_CHANNEL_REGIONS` collection, which is useless as of v5.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/Region.java
+++ b/src/main/java/net/dv8tion/jda/api/Region.java
@@ -17,9 +17,6 @@ package net.dv8tion.jda.api;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.Set;
 
 /**
  * Represents the Regions used for Audio connections.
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/Region.java
```diff
@@ -68,12 +61,6 @@
 
     AUTOMATIC("automatic", "Automatic", null, false);
 
-    /**
-     * This {@link java.util.Set Set} represents all regions that can be used for VoiceChannel region overrides.
-     */
-    public static final Set<Region> VOICE_CHANNEL_REGIONS =
```
DV8FromTheWorld, line 74: Why is this useless now?

DV8FromTheWorld (approved):

src/main/java/net/dv8tion/jda/api/Region.java
caneleex: Guilds can't have regions anymore

DV8FromTheWorld: sure, but the channels can. Wouldn't this Set still be important for that?

caneleex: kind of, the set was only used to check whether a region is applicable to voice channels. if someone wants to grab available regions, they can just get them from the enum honestly, depends on the point of view

DV8FromTheWorld: @MinnDevelopment opinion on this?

MinnDevelopment: I'm fine with removing it

PR #1963, opened 2021-12-21
Only update CommandData limiting booleans for non-empty additions

Sanduhr32:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fixes that you can't "mix" options & subcommand (groups) after adding an empty set of these complementary types.
Only updates booleans for non-empty sets.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/build/CommandData.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/build/CommandData.java
@@ -140,6 +140,8 @@ public class CommandData extends BaseCommand<CommandData> implements Serializabl
     public CommandData addOptions(@Nonnull OptionData... options)
     {
         Checks.noneNull(options, "Option");
+        if (options.length == 0)
+            return this;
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/interactions/commands/build/CommandData.java
```diff
@@ -142,7 +142,8 @@ public CommandData addOptions(@Nonnull OptionData... options)
         Checks.noneNull(options, "Option");
         Checks.check(options.length + this.options.length() <= 25, "Cannot have more than 25 options for a command!");
         Checks.check(allowOption, "You cannot mix options with subcommands/groups.");
-        allowSubcommands = allowGroups = false;
+        if (options.length > 0)
+            allowSubcommands = allowGroups = false;
```
MinnDevelopment, line 146: You should just return if length = 0, also this check should be before the mixing of options check is done.

MinnDevelopment (approved):

PR #1965, opened 2021-12-22
Improve entity toString methods

averen:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Every channel previously overrode `Object#toString` individually to differentiate between the channel types. This PR moves this override to the `AbstractChannelImpl` interface and updates the format to `ChannelType:Name(Id)`. This also has the side effect of adding this handling to `StageChannel` and `VoiceChannel` which used the default method.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDA.java
+++ b/src/main/java/net/dv8tion/jda/api/JDA.java
@@ -181,7 +181,7 @@ public interface JDA
         @Override
         public String toString()
         {
-            return "ShardInfo" + getShardString();
+            return "ShardInfo " + getShardString();
         }
 
         @Override
--- a/src/main/java/net/dv8tion/jda/api/JDA.java
+++ b/src/main/java/net/dv8tion/jda/api/JDA.java
@@ -181,7 +181,7 @@ public interface JDA
         @Override
         public String toString()
         {
-            return "Shard " + getShardString();
+            return "ShardInfo" + getShardString();
         }
 
         @Override
--- a/src/main/java/net/dv8tion/jda/api/utils/WidgetUtil.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/WidgetUtil.java
@@ -533,7 +533,7 @@ public class WidgetUtil
         @Override
         public String toString()
         {
-            return "Widget[" + (isAvailable() ? getName() : "") + "](" + id + ')';
+            return "Widget:" + (isAvailable() ? getName() : "") + '(' + id + ')';
         }
 
         private void checkAvailable()
@@ -772,7 +772,7 @@ public class WidgetUtil
             @Override
             public String toString()
             {
-                return "WidgetMember:" + getName() + '(' + id + ')';
+                return "Widget.Member:" + getName() + '(' + id + ')';
             }
         }
 
@@ -863,7 +863,7 @@ public class WidgetUtil
             @Override
             public String toString()
             {
-                return "WidgetVoiceChannel:" + getName() + '(' + id + ')';
+                return "Widget.VoiceChannel:" + getName() + '(' + id + ')';
             }
         }
         
@@ -1014,7 +1014,7 @@ public class WidgetUtil
             
             @Override
             public String toString() {
-                return "WidgetVoiceState:" + widget.getName() + '(' + member.getEffectiveName() + ')';
+                return "Widget.VoiceState:" + widget.getName() + '(' + member.getEffectiveName() + ')';
             }
         }
     }
--- a/src/main/java/net/dv8tion/jda/api/utils/WidgetUtil.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/WidgetUtil.java
@@ -533,7 +533,7 @@ public class WidgetUtil
         @Override
         public String toString()
         {
-            return "W:" + (isAvailable() ? getName() : "") + '(' + id + ')';
+            return "Widget[" + (isAvailable() ? getName() : "") + "](" + id + ')';
         }
 
         private void checkAvailable()
@@ -772,7 +772,7 @@ public class WidgetUtil
             @Override
             public String toString()
             {
-                return "W.M:" + getName() + '(' + id + ')';
+                return "WidgetMember:" + getName() + '(' + id + ')';
             }
         }
 
@@ -863,7 +863,7 @@ public class WidgetUtil
             @Override
             public String toString()
             {
-                return "W.VC:" + getName() + '(' + id + ')';
+                return "WidgetVoiceChannel:" + getName() + '(' + id + ')';
             }
         }
         
@@ -1014,7 +1014,7 @@ public class WidgetUtil
             
             @Override
             public String toString() {
-                return "VS:" + widget.getName() + ':' + member.getEffectiveName();
+                return "WidgetVoiceState:" + widget.getName() + '(' + member.getEffectiveName() + ')';
             }
         }
     }
--- a/src/main/java/net/dv8tion/jda/internal/audio/ConnectionRequest.java
+++ b/src/main/java/net/dv8tion/jda/internal/audio/ConnectionRequest.java
@@ -84,6 +84,6 @@ public class ConnectionRequest
     @Override
     public String toString()
     {
-        return "ConnectionRequest[" + stage + "](" + guildId + '#' + channelId + ')';
+        return "ConnectionRequest[" + stage + "](" + Long.toUnsignedString(guildId) + '#' + Long.toUnsignedString(channelId) + ')';
     }
 }
--- a/src/main/java/net/dv8tion/jda/internal/audio/ConnectionRequest.java
+++ b/src/main/java/net/dv8tion/jda/internal/audio/ConnectionRequest.java
@@ -84,6 +84,6 @@ public class ConnectionRequest
     @Override
     public String toString()
     {
-        return stage + "(" + Long.toUnsignedString(guildId) + "#" + Long.toUnsignedString(channelId) + ")";
+        return "ConnectionRequest[" + stage + "](" + guildId + '#' + channelId + ')';
     }
 }
--- a/src/main/java/net/dv8tion/jda/internal/entities/ActivityImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/ActivityImpl.java
@@ -123,9 +123,6 @@ public class ActivityImpl implements Activity
     @Override
     public String toString()
     {
-        if (url != null)
-            return String.format("Activity(%s | %s)", name, url);
-        else
-            return String.format("Activity(%s)", name);
+        return "Activity[" + name + (url != null ? "](" + url + ')' : ']');
     }
 }
--- a/src/main/java/net/dv8tion/jda/internal/entities/ActivityImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/ActivityImpl.java
@@ -123,6 +123,9 @@ public class ActivityImpl implements Activity
     @Override
     public String toString()
     {
-        return "Activity[" + name + (url != null ? "](" + url + ')' : ']');
+        if (url != null)
+            return String.format("Activity[%s](%s)", name, url);
+        else
+            return String.format("Activity(%s)", name);
     }
 }
--- a/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
@@ -1031,8 +1031,8 @@ public class ReceivedMessage extends AbstractMessage
     public String toString()
     {
         return author != null
-            ? String.format("M:%#s:%.20s(%s)", author, this, getId())
-            : String.format("M:%.20s", this); // this message was made using MessageBuilder
+                ? String.format("Message:%s:%#s(%.20s)", getId(), author, this)
+                : String.format("Message(%.20s)", this); // this message was made using MessageBuilder
     }
 
     @Override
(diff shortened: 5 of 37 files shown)


Chew:
StoreChannels are going to be removed soon anyway, you can probably use "SC", though I suppose that may delay the PR until they're removed?

src/main/java/net/dv8tion/jda/internal/entities/VoiceChannelImpl.java
```diff
@@ -176,4 +176,12 @@ public VoiceChannelImpl setUserLimit(int userLimit)
         this.userLimit = userLimit;
         return this;
     }
+
+    // -- Object overrides --
+
+    @Override
+    public String toString()
+    {
+        return "VC:" + getName() + '(' + id + ')';
```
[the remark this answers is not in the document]
averen: The `id` field was used since every other channel also uses the field here. Would it be within the scope to update this in each channel?

averen:
> StoreChannels are going to be removed soon anyway, you can probably use "SC", though I suppose that may delay the PR until they're removed?

If they are in the process of being removed, I don't think it will be an issue to share the prefix in the meantime.

src/main/java/net/dv8tion/jda/internal/entities/VoiceChannelImpl.java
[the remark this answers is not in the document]
MinnDevelopment: Yes.

src/main/java/net/dv8tion/jda/internal/entities/StageChannelImpl.java
```diff
@@ -197,4 +197,12 @@ public StageChannelImpl setStageInstance(StageInstance instance)
         this.instance = instance;
         return this;
     }
+
+    // -- Object Overrides --
+
+    @Override
+    public String toString()
+    {
+        return "StaC:" + getName() + '(' + getId() + ')';
```
DV8FromTheWorld, line 206: I'm beginning to wonder if we should lose the weird abbreviation system we have and instead do `StageChannel` or `STAGE` (which models after ChannelType) to better represent channels in `toString` instead of `SC`, `StaC`, or whatever.
@MinnDevelopment Is there a strong reason _not_ to go a bit more explicit?

MinnDevelopment: No, i would also prefer a longer toString that shows the full type

DV8FromTheWorld:
Lets go with the actual class (TextChannel) instead of the type (TEXT). The type gets weird for threads and the class is more explicit and understood by users.
Going with the class also means we can update other entities like Member and Guild to use the longer naming.

src/main/java/net/dv8tion/jda/api/JDA.java
```diff
@@ -181,7 +181,7 @@ public String getShardString()
         @Override
         public String toString()
         {
-            return "Shard " + getShardString();
+            return "ShardInfo" + getShardString();
```
DV8FromTheWorld, line 184:
```suggestion
            return "ShardInfo " + getShardString();
```

src/main/java/net/dv8tion/jda/api/utils/WidgetUtil.java
```diff
@@ -533,7 +533,7 @@ public boolean equals(Object obj) {
         @Override
         public String toString()
         {
-            return "W:" + (isAvailable() ? getName() : "") + '(' + id + ')';
+            return "Widget[" + (isAvailable() ? getName() : "") + "](" + id + ')';
```
DV8FromTheWorld, line 536: When `isAvailable` is `false`, this would build `Widget[](id)`. Not desired.

src/main/java/net/dv8tion/jda/api/utils/WidgetUtil.java
```diff
@@ -772,7 +772,7 @@ public boolean equals(Object obj) {
             @Override
             public String toString()
             {
-                return "W.M:" + getName() + '(' + id + ')';
+                return "WidgetMember:" + getName() + '(' + id + ')';
```
DV8FromTheWorld, line 775:
```suggestion
                return "Widget.Member:" + getName() + '(' + id + ')';
```

DV8FromTheWorld: The others should follow this `.` pattern

src/main/java/net/dv8tion/jda/internal/audio/ConnectionRequest.java
```diff
@@ -84,6 +84,6 @@ public long getGuildIdLong()
     @Override
     public String toString()
     {
-        return stage + "(" + Long.toUnsignedString(guildId) + "#" + Long.toUnsignedString(channelId) + ")";
+        return "ConnectionRequest[" + stage + "](" + guildId + '#' + channelId + ')';
```
DV8FromTheWorld, line 87: Please use `Long.toUnsignedString` for the `guildId` and `channelId`

src/main/java/net/dv8tion/jda/internal/entities/ActivityImpl.java
```diff
@@ -123,9 +123,6 @@ public int hashCode()
     @Override
     public String toString()
     {
-        if (url != null)
-            return String.format("Activity(%s | %s)", name, url);
-        else
-            return String.format("Activity(%s)", name);
+        return "Activity[" + name + (url != null ? "](" + url + ')' : ']');
```
DV8FromTheWorld, line 126:
```suggestion
        if (url != null)
            return String.format("Activity[%s](%s)", name, url);
        else
            return String.format("Activity(%s), name);
```
The original form was easier to read.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -1031,8 +1031,8 @@ public int hashCode()
     public String toString()
     {
         return author != null
-            ? String.format("M:%#s:%.20s(%s)", author, this, getId())
-            : String.format("M:%.20s", this); // this message was made using MessageBuilder
+                ? String.format("Message:%s:%#s(%.20s)", getId(), author, this)
+                : String.format("Message(%.20s)", this); // this message was made using MessageBuilder
```
DV8FromTheWorld, line 1035:
```suggestion
            ? String.format("Message:%s:%#s(%.20s)", getId(), author, this)
            : String.format("Message(%.20s)", this); // this message was made using MessageBuilder
```

DV8FromTheWorld:
I feel like we need to define a consistent format for these `toString` implementations. right now they are kinda.. all over the place.
We need to define:
 - Name
 - what `:` defines
 - what `[...]` means
 - what `(...)` means

Basically, I'm wondering if we need a format we can follow. Something like:
`Name:[Type]:Name(Id / Other info)`

So something like Message might become:
`Message(123515896521, User:DV8FromTheWorld(2321892513583), Length trimmed messa)`

PermissionOverride:
`PermissionOverride[Member](1232185621822, 156218362189857)`

I'm actually not sold on the above format, but I do think we should have a discussion about what our format should be since we're touching everything here.

averen:
I agree that the implementations are kind of a mess. I tried my best to follow somewhat of a pattern while retaining the original meaning in my first pass.

- I think it would be best to use `:` for IDs or "internal string" separators wherever possible.
- We could then use `(...)` to hold primary values like parent objects or user-provided content.
- And finally, the square brackets `[...]` could be used for an enum or type, as well as secondary values.

I would also argue to avoid using commas, slashes, and spaces wherever possible unless the string is representing a collection.

While reusing your examples, I was thinking something like:
`ClassName:[Type]:Id:ChannelId(Username)[Other info]`

Where message would become:
`Message:2321892513583(User:2321892513583(DV8FromTheWorld))[Length trimmed messa]`

And PermissionOverride would be:
`PermissionOverride:[Member]:1232185621822:156218362189857`
I'm not completely sold on the colons here between the brackets either.

Lastly, Member could look like:
`Member:1232185621822:156218362189857(effectiveName)`

DV8FromTheWorld:
Honestly, the strings start getting longer than what I'd like to see at that point. I'm not quite sure what the answer here is yet.

averen:
To be fair, we haven't added any information that wasn't already available in the current toString implementations, we've only updated the formatting so far. If they are getting too long, maybe we will need to consolidate which information is included.

DV8FromTheWorld:
# Snowflakes
- Current Implementations:
	  - `"User(" + getId() + ')';`
	  - `"Guild:" + getName() + '(' + id + ')';`
	  - `"Member:" + getEffectiveName() + '[' + getUser() + "](" + getGuild() + ')';`
	  - `"Role:" + getName() + '(' + id + ')';`
	  - `"Emote:" + getName() + '(' + getIdLong() + ')';`
	  - `"Emoji:" + name + '(' + id + ')'`
	  - `"ReactionEmote:" + getName() + '(' + getId() + ')'`
	  - `"Widget:" + (isAvailable() ? getName() : "") + '(' + id + ')';`
	  - `"Webhook:" + getName() + '(' + id + ')';`
	  
- Possible Format: `Class:name(id, internalInfo)`
  - Changed Implementations
	  - `"User:' + getName() +"(" + getId() + ')';`
	  - `"Member:" + getEffectiveName() + '(' + getUser() + ", " + getGuild() + ')';`
	    - This one technically violates the pattern as the `(...)` doesn't start with the id. It uses `getUser` to provide the Id. Is that a problem?

### Channels
- Current Implementations:
	   - `"Category:" + getName() + '(' + getId() + ')';`
	   - `"TextChannel:" + getName() + '(' + getId() + ')';`
	   - `"ThreadChannel:" + getName() + '(' + getId() + ')';`
	   - `"NewsChannel:" + getName() + '(' + getId() + ')';`
	   - `"PermissionOverride:" + (isMemberOverride() ? "MEMBER" : "ROLE") + ':' + channel.getId() + '(' + getId() + ')';`
	   - `"PrivateChannel:" + getName() + '(' + getId() + ')';`
	   - `"StageChannel:" + getName() + '(' + getId() + ')';`
	   - `"StoreChannel:" + getName() + '(' + getId() + ')';`
	   - `"VoiceChannel:" + getName() + '(' + getId() + ')';`
	   
- Possible Format: `Class[type]:name(id, internalInfo)`
  - Changed Implementations
	   - `"PermissionOverride[" + (isMemberOverride() ? "Member" : "Role") + "](" getId() + ", " + channel + ')';`

### Messages
- Current Implementations:
  - `String.format("Message:%s:%#s(%.20s)", getId(), author, this) : String.format("Message(%.20s)", this)`
  - `"SystemMessage[" + type + ']' + author + '(' + id + ')';`
  
- Possible Format: `Class[type](id, author, partialMessage)`
  - Changed Implementations
    - `String.format("Message(%s, %#s, \"%.20s ...\")", getId(), author, this)`
    - `String.format("DataMessage('%.20s ...')", this)`
    - `"SystemMessage[" + type + "](" + id + ", " + author + ')';`

# Events
- Current Implementations:
  - `GenericGuildMemberUpdateEvent[" + getPropertyIdentifier() + "](" + getOldValue() + "->" + getNewValue() + ')'`
  - `"RoleUpdate[" + getPropertyIdentifier() + "](" + getOldValue() + "->" + getNewValue() + ')'`
  
- Possible Format: `Class[propIdentifer](oldVal -> newVal)`
  - Changed Implementations
    - None from this PR. Maybe other events.  
    
    
# Initial Conclusion
So far, it looks like a pattern of `Class[Type]:name(id, additionalInfo, ...)` seems to work fairly well

DV8FromTheWorld:
## The Rest - I've not edited any of these yet.

### Constructs
- `"ShardInfo " + getShardString();`
- `"Command:" + getName() + '(' + getId() + ')';`
- `"Choice:" + name + '(' + stringValue + ')';`
- `"Option[" + getType() + "](" + name + ')';` **(Command.Option)**
- `"Option[" + getType() + "](" + getName() + '=' + getAsString() + ')';` **(OptionMapping)**
- `"Subcommand(" + name + ')';`
- `"SubcommandGroup(" + name + ')';`
- `String.format("Activity[%s](%s)", name, url) :  String.format("Activity(%s)", name)`
- `"VanityInvite(" + code + ')'`
- `"VoiceState:" + getGuild().getName() + '(' + getId() + ')';`
- `"RichPresence[" + name + "](" + applicationId + ')'`
- `"MessageReaction:" + messageId + '(' + emote + ')';`

# Utility
 - `"ConnectionRequest[" + stage + "](" + Long.toUnsignedString(guildId) + '#' + Long.toUnsignedString(channelId) + ')';`
 - `"CompiledRoute[" + method + "](" + compiledRoute + ')';`
 - `"Result(success=" + value + ')' : "Result(error=" + error + ')';`

# Unsure what to do with these
 - `"Widget.Member:" + getName() + '(' + id + ')';`
 - `"Widget.VoiceChannel:" + getName() + '(' + id + ')';`
 - `"Widget.VoiceState:" + widget.getName() + '(' + member.getEffectiveName() + ')';`
 - `"ApplicationInfo(" + this.id + ')';`
 - `"RoleTags(bot=" + getBotId() + ",integration=" + getIntegrationId() + ",boost=" + isBoost() + ')';`
 - `"TeamMember:" + getTeamId() + '(' + user + ')';`
 - `"Button:" + label + '(' + id + ')';`

MinnDevelopment:
What is the status on this?

averen:
I haven't had the time to look into this too closely, but I will try to get this going again this week.

PR #1972, opened 2021-12-24
Add blocking awaitShutdown

ArchdukeLiamus:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds an `awaitShutdown()` method to JDA for those who wish to block instead of waiting on an async ShutdownEvent for taking actions after JDA shuts down. Internally refactors `setStatus`/`getStatus` to be properly threadsafe and sync on a shared object (which I believe should not break existing semantics) which is used to implement the shutdown-await behavior.


freya022:
This may also permit refactoring of awaitStatus using the new monitors

MinnDevelopment:
This doesn't really make sense because even when this `awaitShutdown` returns, JDA might still send out events, do requests, etc. This isn't really going to do what you expect it to do. I also dislike that you introduce some arbitrary object to exploit as object monitor, we already have a working system for this in `awaitStatus` (which is currently limited to login statuses).

ArchdukeLiamus:
Understandable. I'm hesitant to consider overloading `awaitStatus` since I'm rather uncertain of why the `get`/`set`/`awaitStatus` manage to work despite not having paired synched read/writes or any obvious threading primitive attached. I'll have to grok the shutdown sequence more to see if/how I can salvage this.

src/main/java/net/dv8tion/jda/api/JDA.java
```diff
@@ -2374,6 +2375,18 @@ default Emote getEmoteById(long id)
      * @see #shutdown()
      */
     void shutdownNow();
+    
+    /**
+     * Blocks until this JDA instance fully shuts down. <b>This method does not shut down JDA.</b>
+     * 
+     * <p>If you want to run code after shutting down async, use the {@link ShutdownEvent} in preference to this method.
```
V-Play-Games, line 2382: I don't think you should add an import just for a javadoc link, try doing:
```java
{@link net.dv8tion.jda.api.events.ShutdownEvent}
```

ArchdukeLiamus:
Used the volatile modifier for the status flag and went for awaiting the executors and the WS client firing the shutdown event instead, so this should provide semantics more like what I think users would expect it to do.

src/main/java/net/dv8tion/jda/internal/utils/config/ThreadingConfig.java
```diff
@@ -131,6 +131,31 @@ public void shutdownNow()
...
+        if (callbackPool == ForkJoinPool.commonPool())
+        {
+            ForkJoinPool forkJoinPool = (ForkJoinPool) callbackPool;
+            forkJoinPool.awaitQuiescence(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
+        }
+        else
+        {
+            callbackPool.awaitTermination(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
+        }
+        gatewayPool.awaitTermination(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
```
MinnDevelopment, line 148: These will wait forever if the pools aren't shutdown by jda

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -754,6 +752,20 @@ public synchronized void shutdown()
             client.shutdown();
         }
     }
+    
+    @Override
+    public void awaitShutdown() throws InterruptedException
+    {
+        threadConfig.awaitTermination();
```
MinnDevelopment, line 759: This should be at the end

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -754,6 +752,20 @@ public synchronized void shutdown()
...
+        WebSocketClient client = getClient();
+        if (client != null)
+        {
+            while (!client.isShutdownEventFired())
```
MinnDevelopment, line 763: Why aren't we checking for status instead? If its because `shutdownInternals` sets it, we could change it to only be set by the websocket instead. Maybe make a `JDAImpl#signalShutdown` which updates the status and fires the event.

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -754,6 +752,20 @@ public synchronized void shutdown()
             client.shutdown();
         }
     }
+    
+    @Override
+    public void awaitShutdown() throws InterruptedException
```
MinnDevelopment, line 757: There should be a timeout parameter

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -346,12 +346,10 @@ public void setToken(String token)
 
     public void setStatus(Status status)
     {
-        //noinspection SynchronizeOnNonFinalField
-        synchronized (this.status)
+        synchronized (Status.class) // only for use as a mutex, status is volatile and no counterpart is needed
```
MinnDevelopment, line 349: I don't think this makes much sense

ArchdukeLiamus: The object being locked originally depended on the current status which is about to be changed, so the original sync block would effectively not have done anything since multiple calls would have acquired locks in different objects. Status.class is consistent and seemed the most appropriate direct substitution. It need not be so as long as it is a consistent object.

MinnDevelopment: You should use `synchronized(this)` instead then

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
ArchdukeLiamus: The original intent was to sync the return on firing of the ShutdownEvent, so that flag was used wherever the event would have been fired. `shutdownInternals` doesn't directly fire the event, but does set the status; websocket `handleDisconnect` does internals but fires the event; websocket `reconnect`, `queueReconnect`, and constructor both set status and fire.

It seems all of these cases at least set the status flag somewhere along the track, so in hindsight I agree that should be a status check.

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
ArchdukeLiamus: Seems the only reason the block is needed so StatusChangeEvent isn't potentially getting status == oldStatus, status is volatile in any case
A defensive `if (status != oldStatus) handleEvent` check could work to remove the sync block entirely but would that affect event consumers downstream?

MinnDevelopment: The goal is to make sure that status changes aren't skipped when 2 threads try to change it to a different status at once

ArchdukeLiamus: `sync (this)` it is, then

src/main/java/net/dv8tion/jda/api/JDA.java
Tais993:
```java
{@link net.dv8tion.jda.api.events.ShutdownEvent ShutdownEvent}
```
Is the correct way

src/main/java/net/dv8tion/jda/api/JDA.java
```diff
@@ -2374,6 +2375,26 @@ default List<Emote> getEmotesByName(@Nonnull String name, boolean ignoreCase)
...
+     *        
+     * @param unit
+     *        {@link TimeUnit Unit} for the timeout value.
+     *        
+     * @throws InterruptedException
+     *          if the current thread is interrupted while waiting.
+     *          
+     * @return true if shutdown completed, false if the timeout expires.
+     * 
+     * @see {@link net.dv8tion.jda.api.events.ShutdownEvent ShutdownEvent}
```
Tais993, line 2395:
```suggestion
     * @param  timeout
     *         The timeout to wait for shutdown before returning.
     *        
     * @param  unit
     *         {@link TimeUnit Unit} for the timeout value.
     *        
     * @throws InterruptedException
     *          if the current thread is interrupted while waiting.
     *          
     * @return true if shutdown completed, false if the timeout expires.
     * 
     * @see    net.dv8tion.jda.api.events.ShutdownEvent ShutdownEvent
```
These should all follow the same indentation, follow the [Structure-Guide](https://github.com/DV8FromTheWorld/JDA/wiki/6%29-JDA-Structure-Guide)!

@see shouldn't use @link (if you tried you saw the see didn't work)

Tais993 (changes requested):

src/main/java/net/dv8tion/jda/api/JDA.java
```diff
@@ -2374,6 +2375,27 @@ default List<Emote> getEmotesByName(@Nonnull String name, boolean ignoreCase)
...
+     * in preference to this method.
+     * 
+     * @param  timeout
+     *          The timeout to wait for shutdown before returning.
+     *        
+     * @param  unit
+     *          {@link TimeUnit Unit} for the timeout value.
+     *        
+     * @throws InterruptedException
+     *          if the current thread is interrupted while waiting.
```
Tais993, line 2392:
```suggestion
     * @param  timeout
     *         The timeout to wait for shutdown before returning.
     *        
     * @param  unit
     *         {@link TimeUnit Unit} for the timeout value.
     *        
     * @throws InterruptedException
     *         if the current thread is interrupted while waiting.
```

I had 1 space too much on the throws description, and you copied that over + added a space to the other descriptions which is incorrect

src/main/java/net/dv8tion/jda/api/JDA.java
```diff
@@ -2374,6 +2375,27 @@ default List<Emote> getEmotesByName(@Nonnull String name, boolean ignoreCase)
...
+     *          
+     * @return true if shutdown completed, false if the timeout expires.
+     * 
+     * @see    {@link net.dv8tion.jda.api.events.ShutdownEvent ShutdownEvent}
```
Tais993, line 2396: Like I said, `@see` shouldn't use `@link`, and just like I said, you should've tried your code.

You would've seen it doesn't work

Tais993 (changes requested):

ArchdukeLiamus:
I am choosing to abandon this PR due to lack of continued interest from myself, the fact that spin-on-getStatus has become the de-facto convention for implementing similar behaviour rendering this PR somewhat redundant, and a lack of a good way to obtain the clean-shutdown semantics that I originally hoped for.

Given that the use-case I originally envisioned has been worked around, this PR is not really needed anymore, though anyone who expresses interest and is willing to save it is welcome to do so on their own.

PR #1978, opened 2022-01-01
Move requestToSpeak to StageChannel

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

See [Card: Request to Speak](https://github.com/DV8FromTheWorld/JDA/projects/5#card-69504918)

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/StageInstance.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/StageInstance.java
@@ -16,7 +16,6 @@
 
 package net.dv8tion.jda.api.entities;
 
-import net.dv8tion.jda.api.Permission;
 import net.dv8tion.jda.api.managers.StageInstanceManager;
 import net.dv8tion.jda.api.requests.RestAction;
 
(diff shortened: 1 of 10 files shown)


MinnDevelopment:
This PR also deprecates stage discovery as suggested by discord/discord-api-docs#4296

src/main/java/net/dv8tion/jda/api/entities/StageInstance.java
```diff
@@ -190,7 +164,14 @@
...
         UNKNOWN(-1),
-        /** This stage instance can be accessed by lurkers, meaning users that are not active members of the guild */
+        /**
+         * This stage instance can be accessed by lurkers, meaning users that are not active members of the guild
+         *
+         * @deprecated Public stage instances are no longer supported by discord
+         */
+        @Deprecated
+        @ForRemoval
+        @DeprecatedSince("5.0.0")
```
DV8FromTheWorld, line 174: I wonder if they'll kill off the entire "privacy level" concept

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -936,11 +935,13 @@ public AudioManager getAudioManager()
...
         AudioChannel channel = getSelfMember().getVoiceState().getChannel();
-        StageInstance instance = channel instanceof StageChannel ? ((StageChannel) channel).getStageInstance() : null;
-        if (instance == null)
-            return new GatewayTask<>(CompletableFuture.completedFuture(null), () -> {});
-        CompletableFuture<Void> future = instance.cancelRequestToSpeak().submit();
-        return new GatewayTask<>(future, () -> future.cancel(false));
+        if (channel instanceof StageChannel)
+        {
+            CompletableFuture<Void> future = ((StageChannel) channel).cancelRequestToSpeak().submit();
+            return new GatewayTask<>(future, () -> future.cancel(false));
```
DV8FromTheWorld, line 941: Why are these `GatewayTasks` when the actual request is a `RestAction`?

DV8FromTheWorld (commented):
What is the motivation for keeping the the `requestToSpeak` and `cancelRequestToSpeak` on the `Guild` object?
It feels like `requestToSpeak`, at a minimum, should be removed from `Guild` and the `updateRequestToSpeak` should become an exposed, but internal, method that _must_ accept a `StageChannel` instead of doing implicit stuff based on the currently connected channel. 🤔

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
DV8FromTheWorld: Is it just so that `requestToSpeak` and `cancelRequestToSpeak` have a mirrored return type?

MinnDevelopment: Symmetry and my hopes that this will in the future be a request you can do through the gateway

MinnDevelopment:
I want to keep  `Guild#requestToSpeak` because code like this would be unavoidable otherwise:

```java
guild.getJDA().addEventListener(new ListenerAdapter() {
  public void onGuildVoiceJoin(GuildVoiceJoinEvent event) {
    if (event.getChannel() instanceof StageChannel) {
      ((StageChannel) event.getChannel()).requestToSpeak().queue();
    }
  }
});
guild.getAudioManager().openAudioConnection(channel); // no callbacks available
```

DV8FromTheWorld (approved):

Issue #1987, opened 2022-01-13
Guild Welcome Screen Modification


Aseeef:
## General Troubleshooting
  
- [X] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [X] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request

I am requesting API support for modifying the guild welcome screen as documented here:
https://discord.com/developers/docs/resources/guild#modify-guild-welcome-screen

### Example Use-Case

I personally wanted to use this for a discord "clone" / "backup" bot.

The API may go something like this for retrieving welcome screen info:
boolean b = guild.getWelcomeScreen().getEnabled();
List<WelcomeChannel> wc = guild.getWelcomeScreen().getWelcomeChannels();
String desc = guild.getWelcomeScreen().getDescription();

And for updating, maybe something like:
guild.getWelcomeScreen().updateEnabled(boolean);
guild.getWelcomeScreen().updateWelcomeChannels(List<WelcomeChannel>);
guild.getWelcomeScreen().updateDescription(String);

WelcomeChannel might be a subclass TextChannel adding the addition parameter's of their structure as specified in the discord API: https://discord.com/developers/docs/resources/guild#welcome-screen-object-welcome-screen-channel-structure

MinnDevelopment:
Support for this has been added in [5.0.0-beta.2](https://github.com/DV8FromTheWorld/JDA/releases/tag/v5.0.0-beta.2)

PR #1989, opened 2022-01-15
Add support for guild welcome screens

caneleex:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1987 

## Description

this PR adds support for Guild welcome screens. **this currently doesn't support modifying the welcome screen as I'm still figuring out a good design.**

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2977,6 +2977,9 @@ public interface Guild extends ISnowflake
      * <ul>
      *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_GUILD_WELCOME_SCREEN Unknown Guild Welcome Screen}
      *     <br>The guild has no welcome screen</li>
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#MISSING_PERMISSIONS Missing Permissions}
+     *     <br>The guild's welcome screen is disabled
+     *     and the currently logged in account doesn't have the {@link net.dv8tion.jda.api.Permission#MANAGE_SERVER MANAGE_SERVER} permission</li>
      * </ul>
      *
      * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.GuildWelcomeScreen GuildWelcomeScreen}
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2979,6 +2979,9 @@ public interface Guild extends ISnowflake
      *     <br>The guild has no welcome screen</li>
      * </ul>
      *
+     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
+     *         if the account does not have {@link net.dv8tion.jda.api.Permission#MANAGE_SERVER MANAGE_SERVER} in this Guild.
+     *
      * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.GuildWelcomeScreen GuildWelcomeScreen}
      *         <br>The welcome screen for this Guild.
      */
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2979,9 +2979,6 @@ public interface Guild extends ISnowflake
      *     <br>The guild has no welcome screen</li>
      * </ul>
      *
-     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
-     *         if the account does not have {@link net.dv8tion.jda.api.Permission#MANAGE_SERVER MANAGE_SERVER} in this Guild.
-     *
      * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.GuildWelcomeScreen GuildWelcomeScreen}
      *         <br>The welcome screen for this Guild.
      */
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2970,6 +2970,23 @@ public interface Guild extends ISnowflake
     RestAction<List<Webhook>> retrieveWebhooks();
 
     /**
+     * Retrieves the {@link GuildWelcomeScreen Welcome screen} for this Guild.
+     * <br>The welcome screen is shown to all members after joining the Guild.
+     *
+     * <p>Possible {@link net.dv8tion.jda.api.requests.ErrorResponse ErrorResponses} include:
+     * <ul>
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_GUILD_WELCOME_SCREEN Unknown Guild Welcome Screen}
+     *     <br>The guild has no welcome screen</li>
+     * </ul>
+     *
+     * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.GuildWelcomeScreen GuildWelcomeScreen}
+     *         <br>The welcome screen for this Guild.
+     */
+    @Nonnull
+    @CheckReturnValue
+    RestAction<GuildWelcomeScreen> retrieveWelcomeScreen();
+
+    /**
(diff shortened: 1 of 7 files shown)


src/main/java/net/dv8tion/jda/api/entities/GuildWelcomeScreen.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        /**
+         * The id of the emote that is used for this recommended channel.
+         * <br><b>This will return {@code null} if the emoji is a unicode emoji.</b>
+         *
+         * @return The id of the emote that is used for this recommended channel or {@code null}
+         */
+        @Nullable
+        public String getEmoteId()
```
caneleex, line 105: I'm not sure whether I should wrap this into an `Emoji` object instead. The class constructor accepts the `animated` field which we can't know, while a recommended channel's emote can be animated.

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -2969,6 +2969,26 @@ default Emote getEmoteById(long id)
     @CheckReturnValue
     RestAction<List<Webhook>> retrieveWebhooks();
 
+    /**
+     * Retrieves the {@link GuildWelcomeScreen Welcome screen} for this Guild.
+     * <br>The welcome screen is shown to all members after joining the Guild.
+     *
+     * <p>Possible {@link net.dv8tion.jda.api.requests.ErrorResponse ErrorResponses} include:
+     * <ul>
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_GUILD_WELCOME_SCREEN Unknown Guild Welcome Screen}
```
DV8FromTheWorld, line 2978: I wonder if we should convert this error into a `null` instead of forcing users to handle it as a user.

src/main/java/net/dv8tion/jda/api/entities/GuildWelcomeScreen.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.List;
+
+/**
+ * The welcome screen of a {@link net.dv8tion.jda.api.entities.Guild}.
+ * This welcome screen will be shown to all members after joining the Guild.
+ *
+ * @see Guild#retrieveWelcomeScreen()
+ * @see Invite.Guild#getWelcomeScreen()
+ */
+public class GuildWelcomeScreen
```
DV8FromTheWorld, line 32: This should have a `getGuild` getter

src/main/java/net/dv8tion/jda/api/entities/GuildWelcomeScreen.java
```diff
@@ -0,0 +1,147 @@
...
+        /**
+         * Returns the {@link net.dv8tion.jda.api.entities.BaseGuildMessageChannel} that is linked to this recommended channel.
+         * <br>This will be {@code null} if the linked channel was deleted.
+         *
+         * @return The {@link net.dv8tion.jda.api.entities.BaseGuildMessageChannel} that is linked to this recommended channel or {@code null}
+         */
+        @Nullable
+        public BaseGuildMessageChannel getChannel()
+        {
+            return (BaseGuildMessageChannel) api.getGuildChannelById(id);
```
DV8FromTheWorld, line 108: This is questionable as I could see ForumChannels being a recommendable channel in the future.

src/main/java/net/dv8tion/jda/api/entities/GuildWelcomeScreen.java
DV8FromTheWorld: We can't tell by the emote's id via something like `a_` prefix or whatever?

src/main/java/net/dv8tion/jda/api/entities/GuildWelcomeScreen.java
```diff
@@ -0,0 +1,147 @@
...
+
+        /**
+         * The name of the emoji that is used for this recommended channel.
+         * <br>This will return the emote name if {@link #getEmoteId()} is not null, otherwise the unicode emoji or {@code null}
+         * is returned if no emoji was set.
+         *
+         * @return The name of the emoji that is used for this recommended channel or {@code null}
+         */
+        @Nullable
+        public String getEmojiName()
```
DV8FromTheWorld, line 142: So this method can return a name or a unicode emoji? Yeah, not sure how to feel about that representation given the method naming..

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -446,6 +446,15 @@ default String getUrl()
...
         Set<String> getFeatures();
+
+        /**
+         * The welcome screen of the {@link net.dv8tion.jda.api.entities.Invite.Guild Guild}.
+         * <br>This will be {@code null} if the Guild has no welcome screen.
+         *
+         * @return The welcome screen of this Guild or {@code null}
+         */
+        @Nullable
+        GuildWelcomeScreen getWelcomeScreen();
```
DV8FromTheWorld, line 457: So all resolved invites have this information, but the Guild itself doesnt? Curious

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1895,6 +1895,19 @@ public WebhookImpl createWebhook(DataObject object, boolean allowMissingChannel)
...
 
+    public GuildWelcomeScreen createWelcomeScreen(DataObject object)
+    {
+        final DataArray welcomeChannelsArray = object.getArray("welcome_channels");
+        final List<GuildWelcomeScreen.Channel> welcomeChannels = new ArrayList<>(welcomeChannelsArray.length());
+        for (int i = 0; i < welcomeChannelsArray.length(); i++)
+        {
+            final DataObject welcomeChannelObj = welcomeChannelsArray.getObject(i);
+            welcomeChannels.add(new GuildWelcomeScreen.Channel(api, welcomeChannelObj.getLong("channel_id"), welcomeChannelObj.getString("description"),
+                    welcomeChannelObj.getString("emoji_id", null), welcomeChannelObj.getString("emoji_name", null)));
```
DV8FromTheWorld, line 1906:
```suggestion
            welcomeChannels.add(new GuildWelcomeScreen.Channel(
                api,
                welcomeChannelObj.getLong("channel_id"),
                welcomeChannelObj.getString("description"),
                welcomeChannelObj.getString("emoji_id", null),
                welcomeChannelObj.getString("emoji_name", null)
            ));
```

DV8FromTheWorld:
We should probably include creation/mutation of GuildWelcomeScreens in this PR instead of providing only read-only access.

We can likely use a simple builder system like we do with EntityBuilder to build it up.

src/main/java/net/dv8tion/jda/api/entities/Invite.java
caneleex: exactly, you have to retrieve it yourself.

src/main/java/net/dv8tion/jda/api/entities/GuildWelcomeScreen.java
caneleex: no, the object only contains the emote id and name, not the hash

DV8FromTheWorld:
Something else that I forgot to mention is that `GuildWelcomeScreen` should be inside `Guild` as `Guild.WelcomeScreen`.
This follows the pattern established by other constructs that make up the `Guild` itself.

caneleex:
> 

it's intentionally in its own file because it has an inner `Channel` class.

DV8FromTheWorld:
I don't personally see a problem with `Guild.WelconeScreen.Channel`.
The `Guild.X` classes represent constructs which are partials of a Guild's metadata. That metadata could be represented as `Guild#getWelcomeScreenChannels`, `Guild#getWelcomeScreenDescription`. We separate them only so that things logically make sense.

In contrast, something like `GuildVoiceState` or `TextChannel` are entities unto themselves. They are not metadata for the `Guild`.

Aseeef:
Hey any updates on this one?

caneleex:
> Hey any updates on this one?

not right now, no. I won't have time for the next 2-3 weeks, so if someone wants to tackle this earlier, let me know.

Yashar256:
Yeah choosing a good design for modifying the welcome screen seems tricky. One possible design could be to have a welcome screen manager, with methods such as `SortedSet<GuildWelcomeScreen.Channel> getWelcomeChannels()` to see the order of the welcome screen channels while updating them, and methods such as `void putChannelBeforeIndex(int index, GuildWelcomeScreen.Channel channel)` and `void putChannelAfterIndex(int index, GuildWelcomeScreen.Channel channel)` to emulate the drag and drop ordering that people are used to in the client. And of course, more conventional methods for replacing a certain channel, removing a certain channel, etc.

MinnDevelopment:
Closing due to inactivity. If anyone wants to continue this based on the current master branch, feel free to open a new PR.

PR #1996, opened 2022-01-19
Override ThreadChannel#getTimeCreated

caneleex:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this PR overrides `ISnowflake#getTimeCreated` for `ThreadChannel` with a field that is present in a Thread payload.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/ThreadChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/ThreadChannel.java
@@ -475,6 +475,8 @@ public interface ThreadChannel extends GuildMessageChannel, IMemberContainer
 
     /**
      * The timestamp when this thread was created.
+     * <br><b>This will only be valid for threads created after 2022-01-09.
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/api/entities/ThreadChannel.java
```diff
@@ -473,6 +473,13 @@ default ThreadMember getOwnerThreadMember()
     @Nonnull
     AutoArchiveDuration getAutoArchiveDuration();
 
+    /**
+     * The timestamp when this thread was created.
+     *
+     * @return The timestamp when this thread was created
+     */
```
DV8FromTheWorld, line 480: These docs need to mention that this will only be valid for ThreadChannels created after 2022-01-09. Otherwise, it will default to using the id of the thread which if it is based on a root message will be the creation time of the message, not the thread.

DV8FromTheWorld (approved):

PR #2006, opened 2022-02-03
Feature: Command choices from enum

PrimeAPI:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: SlashBotExample

Closes Issue: NaN

## Description

Adds `OptionData#addChoices(Class)` to easily add predefined choices a user can choose in a slash-command.
Enum-Class can implement `OptionData.ChoiceNameTransformer` to map the constants name that will be displayed in the Discord client
In the `SlashCommandInteractionEvent` the enum can be easily used by `OptionMapping#getAsEnum`.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
@@ -248,10 +248,10 @@ public class OptionMapping
      * @return The enum-constant corresponding with this option value
      */
     @Nonnull
-    public <T> T getAsEnum(Class<T> enumClass)
+    public <T extends Enum<T>> T getAsEnum(Class<T> enumClass)
     {
         Checks.check(enumClass.isEnum(), "enumClass");
-        T t = Enum.valueOf(enumClass, getAsString())
+        T t = Enum.valueOf(enumClass, getAsString());
         Checks.notNull(t, String.format("Enum constant for %s not found", getAsString()));
         return t;
     }
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
@@ -19,6 +19,7 @@ package net.dv8tion.jda.api.interactions.commands;
 import gnu.trove.map.TLongObjectMap;
 import net.dv8tion.jda.api.entities.*;
 import net.dv8tion.jda.api.utils.data.DataObject;
+import net.dv8tion.jda.internal.utils.Checks;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
```diff
@@ -233,6 +234,31 @@ public String getAsString()
...
+     */
+    @Nonnull
+    public <T> T getAsEnum(Class<T> enumClass)
+    {
+        Checks.check(enumClass.isEnum(), "enumClass");
+        T t = Arrays.stream(enumClass.getEnumConstants())
+                .filter(t1 -> t1.toString().equals(getAsString()))
+                .findAny()
+                .orElse(null);
+        Checks.notNull(t, String.format("Enum konstant for %s not found", getAsString()));
```
freya022, line 258: Enum constant*

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -704,6 +709,67 @@ public OptionData addChoice(@Nonnull String name, @Nonnull String value)
         return this;
     }
 
+    /**
+     * Add predefined Choices by Enum-Constants for this OptionData
+     * <br> The user can only provide one of the choices, provided by the enum
+     *
+     * <p> An Enum-Constant can be annotated by {@link OptionChoice} to determine the displayname in the Discord client
```
freya022, line 716: Should `displayname` be `displayed name` ?

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -704,6 +709,67 @@ public OptionData addChoice(@Nonnull String name, @Nonnull String value)
...
+     *
+     * @return The OptionData instance, for chaining
+     *
+     * @see net.dv8tion.jda.api.interactions.commands.OptionMapping#getAsEnum(Class)
+     * @see OptionChoice
+     */
+    @Nonnull
+    public OptionData addChoices(@Nonnull Class<?> enumClass)
+    {
+        if (this.choices == null)
```
freya022, line 738: This is unclear code, I would replace it with whether the `OptionType` supports choices, it should be equivalent

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -704,6 +709,67 @@ public OptionData addChoice(@Nonnull String name, @Nonnull String value)
...
+            throw new IllegalStateException("Cannot add choices for an option of type " + type);
+        Checks.check(enumClass.isEnum(), "enumClass has to be an enum!");
+        Checks.check(type == OptionType.STRING, "Cannot add enum choice for OptionType." + type);
+        Object[] enumConstants = enumClass.getEnumConstants();
+        Checks.check(enumConstants.length + this.choices.size() <= MAX_CHOICES, "Cannot have more than 25 choices for one option!");
+        for (Field field : enumClass.getFields())
```
freya022, line 744: Imo you should use a mapping function instead of annotations, this will reduce boilerplate in enums, for example if the transforming function is something like "All lowercase except first letters"

src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
```diff
@@ -233,6 +234,31 @@ public String getAsString()
...
+     *             <li>{@code} contains a constant corresponding with this option value</li>
+     *         </ul>
+     *
+     * @return The enum-constant corresponding with this option value
+     */
+    @Nonnull
+    public <T> T getAsEnum(Class<T> enumClass)
+    {
+        Checks.check(enumClass.isEnum(), "enumClass");
+        T t = Arrays.stream(enumClass.getEnumConstants())
```
freya022, line 254: This could be replaced with `Enum.valueOf(enumClass, getAsString())`

freya022 (changes requested):

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
PrimeAPI: I struggled with finding a good solution for changing the displayed name. How exactly do you suggest implementing a mapping function? I cant really imagine a good way doing this...

DV8FromTheWorld:
I appreciate the work put into this PR and I recognize the gap you are attempting to fill, however we won't be accepting this PR at this time. 

Adding choices to an option via an enum can easily live inside of Userland without additions to JDA like this. While this addition may suite some usecases, it is generally more restrictive than the `addChoices(Choice...)` method and enforces certain restrictions, like `string` type only.

PR #2012, opened 2022-02-04
Change PrivateChannel#getUser to handle API nullability

oliver276:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Project Card: https://github.com/DV8FromTheWorld/JDA/projects/5#card-68160406

## Description

Changes:
 + PrivateChannels may not always have a User
 + PrivateChannels contain a convenience retrieveUser method
 + JDA#retrieveChannelById
 + Reaction events no longer wait for the channel to be cached
 + The GET_CHANNEL route, for retrieving channels

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDA.java
+++ b/src/main/java/net/dv8tion/jda/api/JDA.java
@@ -1118,60 +1118,6 @@ public interface JDA
     RestAction<User> retrieveUserById(long id, boolean update);
 
     /**
-     * Retrieves a {@link Channel} object based on the provided id.
-     *
-     * <p>The returned {@link net.dv8tion.jda.api.requests.RestAction RestAction} can encounter the following Discord errors:
-     * <ul>
-     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_CHANNEL ErrorResponse.UNKNOWN_CHANNEL}
-     *     <br>Occurs when the provided id does not refer to a {@link Channel Channel} known by Discord.
-     *     known by Discord. Typically occurs when developers provide an incomplete id (cut short).</li>
-     *
-     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#MISSING_ACCESS MISSING_ACCESS}
-     *     <br>Occurs when the provided id does refers to a {@link Channel Channel} that this account does not have access to.
-     *     </li>
-     * </ul>
-     *
-     * @param  id
-     *         The id of the requested {@link Channel Channel}.
-     *
-     * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.Channel Channel}
-     *         <br>On request, gets the User with id matching provided id from Discord.
-     */
-    @Nonnull
-    @CheckReturnValue
-    RestAction<Channel> retrieveChannelById(long id);
-
-    /**
-     * Retrieves a {@link Channel} object based on the provided id.
-     *
-     * <p>The returned {@link net.dv8tion.jda.api.requests.RestAction RestAction} can encounter the following Discord errors:
-     * <ul>
-     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_CHANNEL ErrorResponse.UNKNOWN_CHANNEL}
-     *     <br>Occurs when the provided id does not refer to a {@link Channel Channel} known by Discord.
-     *     known by Discord. Typically occurs when developers provide an incomplete id (cut short).</li>
-     *
-     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#MISSING_ACCESS MISSING_ACCESS}
-     *     <br>Occurs when the provided id does refers to a {@link Channel Channel} that this account does not have access to.
-     *     </li>
-     * </ul>
-     *
-     * @param  id
-     *         The id of the requested {@link Channel Channel}.
-     *
-     * @throws IllegalArgumentException
-     *         If the provided id is null or is not a valid Snowflake.
-     *
-     * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.Channel Channel}
-     *         <br>On request, gets the User with id matching provided id from Discord.
-     */
-    @Nonnull
-    @CheckReturnValue
-    default RestAction<Channel> retrieveChannelById(String id)
-    {
-        return retrieveChannelById(MiscUtil.parseSnowflake(id));
-    }
-
-    /**
      * {@link net.dv8tion.jda.api.utils.cache.SnowflakeCacheView SnowflakeCacheView} of
      * all cached {@link Guild Guilds} visible to this JDA session.
      *
--- a/src/main/java/net/dv8tion/jda/api/JDA.java
+++ b/src/main/java/net/dv8tion/jda/api/JDA.java
@@ -1118,6 +1118,59 @@ public interface JDA
     RestAction<User> retrieveUserById(long id, boolean update);
 
     /**
+     * Retrieves a {@link Channel} object based on the provided id.
+     *
+     * <p>The returned {@link net.dv8tion.jda.api.requests.RestAction RestAction} can encounter the following Discord errors:
+     * <ul>
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_CHANNEL ErrorResponse.UNKNOWN_CHANNEL}
+     *     <br>Occurs when the provided id does not refer to a {@link Channel Channel} known by Discord.
+     *     known by Discord. Typically occurs when developers provide an incomplete id (cut short).</li>
+     *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#MISSING_ACCESS MISSING_ACCESS}
+     *     <br>Occurs when the provided id does refers to a {@link Channel Channel} that this account does not have access to.
+     *     </li>
+     * </ul>
+     *
+     * @param  id
+     *         The id of the requested {@link Channel Channel}.
+     *
+     * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.Channel Channel}
+     *         <br>On request, gets the User with id matching provided id from Discord.
+     */
+    @Nonnull
+    @CheckReturnValue
+    RestAction<Channel> retrieveChannelById(long id);
+
+    /**
+     * Retrieves a {@link Channel} object based on the provided id.
+     *
+     * <p>The returned {@link net.dv8tion.jda.api.requests.RestAction RestAction} can encounter the following Discord errors:
+     * <ul>
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_CHANNEL ErrorResponse.UNKNOWN_CHANNEL}
+     *     <br>Occurs when the provided id does not refer to a {@link Channel Channel} known by Discord.
+     *     known by Discord. Typically occurs when developers provide an incomplete id (cut short).</li>
+     *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#MISSING_ACCESS MISSING_ACCESS}
+     *     <br>Occurs when the provided id does refers to a {@link Channel Channel} that this account does not have access to.
+     *     </li>
+     * </ul>
+     *
+     * @param  id
+     *         The id of the requested {@link Channel Channel}.
+     *
+     * @throws IllegalArgumentException
+     *         If the provided id is null or is not a valid Snowflake.
+     *
+     * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.Channel Channel}
+     *         <br>On request, gets the User with id matching provided id from Discord.
+     */
+    @Nonnull
+    @CheckReturnValue
+    default RestAction<Channel> retrieveChannelById(String id){
+        return retrieveChannelById(MiscUtil.parseSnowflake(id));
+    }
+
+    /**
      * {@link net.dv8tion.jda.api.utils.cache.SnowflakeCacheView SnowflakeCacheView} of
      * all cached {@link Guild Guilds} visible to this JDA session.
      *
--- a/src/main/java/net/dv8tion/jda/api/JDA.java
+++ b/src/main/java/net/dv8tion/jda/api/JDA.java
@@ -1166,7 +1166,8 @@ public interface JDA
      */
     @Nonnull
     @CheckReturnValue
-    default RestAction<Channel> retrieveChannelById(String id){
+    default RestAction<Channel> retrieveChannelById(String id)
+    {
         return retrieveChannelById(MiscUtil.parseSnowflake(id));
     }
 
--- a/src/main/java/net/dv8tion/jda/api/entities/PrivateChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/PrivateChannel.java
@@ -41,7 +41,7 @@ public interface PrivateChannel extends MessageChannel
      * The consequence of this is that for any message this bot receives from a guild or from other users will not be null.
      *
      *
-     * @return A potentially null {@link net.dv8tion.jda.api.entities.User User}.
+     * @return Possibly-null {@link net.dv8tion.jda.api.entities.User User}.
      */
     @Nullable
     User getUser();
--- a/src/main/java/net/dv8tion/jda/api/entities/PrivateChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/PrivateChannel.java
@@ -16,7 +16,6 @@
 package net.dv8tion.jda.api.entities;
 
 import net.dv8tion.jda.api.requests.RestAction;
-import org.jetbrains.annotations.NotNull;
 
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
@@ -64,12 +63,19 @@ public interface PrivateChannel extends MessageChannel
     /**
      * The human-readable name of this channel.
      *
-     * If getUser returns null, this method will return the empty String.
-     * This happens when JDA does not have enough information to populate the channel name
+     * If getUser returns null, this method will return an empty String.
+     * This happens when JDA does not have enough information to populate the channel name.
+     *
+     * This will occur only when {@link #getUser()} is null, and the reasons are given in {@link #getUser()}
+     *
+     * If the channel name is important, {@link #retrieveUser()} should be used, instead.
      *
      * @return The name of this channel
+     *
+     * @see #retrieveUser()
+     * @see #getUser()
      */
-    @NotNull
+    @Nonnull
     @Override
     String getName();
 }
--- a/src/main/java/net/dv8tion/jda/api/entities/PrivateChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/PrivateChannel.java
@@ -15,7 +15,11 @@
  */
 package net.dv8tion.jda.api.entities;
 
+import net.dv8tion.jda.api.requests.RestAction;
+
+import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 /**
  * Represents the connection used for direct messaging.
@@ -27,8 +31,29 @@ public interface PrivateChannel extends MessageChannel
     /**
      * The {@link net.dv8tion.jda.api.entities.User User} that this {@link net.dv8tion.jda.api.entities.PrivateChannel PrivateChannel} communicates with.
      *
-     * @return A non-null {@link net.dv8tion.jda.api.entities.User User}.
+     * <p>This user is only null if this channel is currently uncached, and one the following occur:
+     * <ul>
+     *     <li>A reaction is removed</li>
+     *     <li>A reaction is added</li>
+     *     <li>A message is deleted</li>
+     *     <li>This account sends a message to a user from another shard (not shard 0)</li>
+     * </ul>
+     * The consequence of this is that for any message this bot receives from a guild or from other users will not be null.
+     *
+     *
+     * @return A potentially null {@link net.dv8tion.jda.api.entities.User User}.
+     */
+    @Nullable
+    User getUser();
+
+    /**
+     * Retrieves the {@link User User} that this {@link net.dv8tion.jda.api.entities.PrivateChannel PrivateChannel} communicates with.
+     *
+     * <br>This method fetches the channel from the API and retrieves the User from that.
+     *
+     * @return A {@link RestAction RestAction} to retrieve the {@link User User} that this {@link PrivateChannel PrivateChannel} communicates with.
      */
     @Nonnull
-    User getUser();
+    @CheckReturnValue
+    RestAction<User> retrieveUser();
 }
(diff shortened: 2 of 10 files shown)


src/main/java/net/dv8tion/jda/api/JDA.java
```diff
@@ -1117,6 +1117,60 @@ default User getUserByTag(@Nonnull String username, @Nonnull String discriminato
...
+     *
+     * @param  id
+     *         The id of the requested {@link Channel Channel}.
+     *
+     * @return {@link net.dv8tion.jda.api.requests.RestAction RestAction} - Type: {@link net.dv8tion.jda.api.entities.Channel Channel}
+     *         <br>On request, gets the User with id matching provided id from Discord.
+     */
+    @Nonnull
+    @CheckReturnValue
+    RestAction<Channel> retrieveChannelById(long id);
```
DV8FromTheWorld, line 1142: We might not actually want to publicly expose this yet. If a developer attempts to retrieve a channel that is part of a guild but that guild isn't loaded in this shard, then the EntityBuilder would likely explode trying to craft the channel.

We probably wanna hold off on public access to this functionality until we work on the RestOnly Module.

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -16,56 +16,76 @@
...
 
     @Nonnull
     @Override
-    public JDA getJDA()
+    public String getName()
     {
-        return user.getJDA();
+        if (getUser() == null)
+            //don't break or override the contract of @NotNull
+            return "";
```
DV8FromTheWorld, line 87: if you have 2 lines in an if-statement, then use `{ }`

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -85,7 +105,7 @@ public long getLatestMessageIdLong()
     @Override
     public boolean canTalk()
     {
-        return !user.isBot();
+        return user == null || !user.isBot();
```
DV8FromTheWorld, line 108: Comment on why `user == null` makes this ok

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -162,7 +182,7 @@ private void updateUser()
 
     private void checkBot()
     {
-        if (getUser().isBot() && getJDA().getAccountType() == AccountType.BOT)
+        if (getUser() != null && getUser().isBot())
```
DV8FromTheWorld, line 185: Comment on why this looks for null.

src/main/java/net/dv8tion/jda/internal/handle/MessageReactionHandler.java
```diff
@@ -105,16 +105,6 @@ protected Long handleInternally(DataObject content)
...
-        if (user == null)
-        {
-            if (add && guild != null)
-            {
-                api.getEventCache().cache(EventCache.Type.USER, userId, responseNumber, allContent, this::handle);
-                EventCache.LOG.debug("Received a reaction for a user that JDA does not currently have cached. " +
-                        "UserID: {} ChannelId: {} MessageId: {}", userId, channelId, messageId);
-                return null;
-            }
-        }
```
DV8FromTheWorld, line 117: Why are we removing this? This only triggers for Guild reactions

src/main/java/net/dv8tion/jda/internal/handle/MessageReactionHandler.java
```diff
@@ -126,11 +116,12 @@ protected Long handleInternally(DataObject content)
...
         if (channel == null)
         {
-            api.getEventCache().cache(EventCache.Type.CHANNEL, channelId, responseNumber, allContent, this::handle);
-            EventCache.LOG.debug("Received a reaction for a channel that JDA does not currently have cached");
-            return null;
+            //create a new private channel with minimal information for this event
+            channel = getJDA().getEntityBuilder().createPrivateChannel(
+                    DataObject.empty()
+                            .put("id", channelId)
+            );
```
DV8FromTheWorld, line 123: We definitely cannot just assume this here.
This could be from a channel type we don't support yet, or it could be from a channel we just haven't received yet.
Not sure that I understand why it'd be implemented this way 😕

DV8FromTheWorld: We should at a minimum be able to see if `guild_id` is populated to make a decision about whether to try and create a private channel or use the event cache.

src/main/java/net/dv8tion/jda/api/events/message/react/GenericMessageReactionEvent.java
```diff
@@ -60,17 +60,23 @@ public GenericMessageReactionEvent(@Nonnull JDA api, long responseNumber, @Nulla
...
      * @return The user id
      */
-    @Nonnull
+    @Nullable
     public String getUserId()
     {
+        if (userId == 0)
+            return null;
         return Long.toUnsignedString(userId);
     }
```
DV8FromTheWorld, line 73: afaict, this isn't true.
The `userId` is never `0` here because it is provided by the gateway event and passed in:
https://github.com/DV8FromTheWorld/JDA/blob/12d4a0af25cc6bad3a8f4a37bd5230d861e66268/src/main/java/net/dv8tion/jda/internal/handle/MessageReactionHandler.java#L59
https://github.com/DV8FromTheWorld/JDA/blob/12d4a0af25cc6bad3a8f4a37bd5230d861e66268/src/main/java/net/dv8tion/jda/internal/handle/MessageReactionHandler.java#L155-L158

src/main/java/net/dv8tion/jda/api/events/message/react/GenericMessageReactionEvent.java
```diff
@@ -60,17 +60,23 @@ public GenericMessageReactionEvent(@Nonnull JDA api, long responseNumber, @Nulla
...
 
     /**
      * The id for the user who owns reaction.
      *
+     * This is {@code 0} if the user was not cached.
```
DV8FromTheWorld, line 78: Same as above

src/main/java/net/dv8tion/jda/api/events/message/react/GenericMessageReactionEvent.java
```diff
@@ -151,6 +157,11 @@ public MessageReaction getReaction()
         User user = getUser();
         if (user != null)
             return new CompletedRestAction<>(getJDA(), user);
+        if (userId == 0)
+        {
+            return getJDA().retrieveChannelById(getChannel().getIdLong())
+                    .map(channel -> ((PrivateChannel) channel).getUser());
+        }
```
DV8FromTheWorld, line 164: This isn't needed. See statements above

src/main/java/net/dv8tion/jda/api/events/message/react/GenericMessageReactionEvent.java
```diff
@@ -175,6 +186,8 @@ public MessageReaction getReaction()
     {
         if (member != null)
             return new CompletedRestAction<>(getJDA(), member);
+        if (userId == 0 || !getChannel().getType().isGuild())
+            throw new IllegalStateException("Cannot retrieve member for a private reaction not from a guild");
```
DV8FromTheWorld, line 190: This doesn't need to check for `userId == 0`

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -577,6 +577,58 @@ public DirectAudioControllerImpl getDirectAudioController()
...
+        if (channel == null)
+            channel = getThreadChannelById(channelId);
+        if (channel == null)
+            channel = getPrivateChannelById(channelId);
+        if (channel == null)
+            channel = getStageChannelById(channelId);
+        if (channel == null)
+            channel = getVoiceChannelById(channelId);
+        return channel;
+    }
```
DV8FromTheWorld, line 630: Am I missing something? Doesn't this mostly already exist in `getGuildChannelById`?

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -577,6 +577,58 @@ public DirectAudioControllerImpl getDirectAudioController()
                 });
     }
 
+    @Nonnull
+    @Override
+    public RestAction<Channel> retrieveChannelById(long channelId)
```
DV8FromTheWorld, line 582: Given that we are likely not going to expose this publicly yet, we can kill off the handling of anything but private channel and move the impl into PrivateChannelImpl.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -311,6 +311,33 @@ public GuildImpl createGuild(long guildId, DataObject guildJson, TLongObjectMap<
         return guildObj;
     }
 
+    public Channel createChannel(DataObject dataObject){
```
DV8FromTheWorld, line 314: Since we are not going to expose the method publicly, we can remove this for now.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1233,6 +1260,14 @@ public PrivateChannel createPrivateChannel(DataObject json)
         if (channel != null)
             return channel;
 
+        //if we don't have enough information to construct the channel properly, the rest of the properties can be null
+        if (!json.hasKey("recipients") && !json.hasKey("recipient"))
+        {
+            PrivateChannelImpl priv = new PrivateChannelImpl(channelId, getJDA());
+            cachePrivateChannel(priv);
```
DV8FromTheWorld, line 1267: This isn't playing back the cache

DV8FromTheWorld: I'd recommend adding the cache playback to run at the end of the `cachePrivateChannel` method

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1387,11 +1427,13 @@ public ReceivedMessage createMessage(DataObject jsonObject, @Nullable MessageCha
...
         MemberImpl member = null;
 
-        if (channel == null && jsonObject.isNull("guild_id") && authorId != getJDA().getSelfUser().getIdLong())
+        if (channel == null && jsonObject.isNull("guild_id"))
         {
             DataObject channelData = DataObject.empty()
-                    .put("id", channelId)
-                    .put("recipient", author);
+                    .put("id", channelId);
+            if (authorId != getJDA().getSelfUser().getIdLong())
```
DV8FromTheWorld, line 1434: This needs a comment to explain why we conditionally place the author as the recipient.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1387,11 +1427,13 @@ public ReceivedMessage createMessage(DataObject jsonObject, @Nullable MessageCha
...
+                    channelData.put("recipient", author);
+            //this will construct a minimal private channel if we are the author
             channel = createPrivateChannel(channelData);
```
DV8FromTheWorld, line 1437: This comment should be retooled to differently explain that while this creates a private channel, it may only be a partial channel if we are the author as we don't have the recipient info.

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -16,56 +16,76 @@
...
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 public class PrivateChannelImpl extends AbstractChannelImpl<PrivateChannelImpl> implements PrivateChannel, MessageChannelMixin<PrivateChannelImpl>
 {
     private User user;
     private long latestMessageId;
 
-    public PrivateChannelImpl(long id, User user)
+    private PrivateChannelImpl(JDA api, long id, User user)
```
DV8FromTheWorld, line 38: This constructors are internal. You could just have `public PrivateChannelImpl(JDA api, long id, User user)` and let `user` be `nullable`

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -46,26 +48,43 @@ public ChannelType getType()
...
         return getUser().getName();
     }
 
     @Nonnull
-    @Override
-    public JDA getJDA()
+    private RestAction<PrivateChannel> retrievePrivateChannel()
     {
-        return user.getJDA();
+        Route.CompiledRoute route = Route.Channels.GET_CHANNEL.compile(Long.toUnsignedString(id));
```
MinnDevelopment, line 86:
```suggestion
        Route.CompiledRoute route = Route.Channels.GET_CHANNEL.compile(getId());
```

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -46,26 +48,43 @@ public ChannelType getType()
...
         return user;
     }
 
+    @Nonnull
+    @Override
+    public RestAction<User> retrieveUser()
+    {
+        if (user != null){
+            return new CompletedRestAction<>(getJDA(), user);
+        }
```
MinnDevelopment, line 65:
```suggestion
        if (user != null)
            return new CompletedRestAction<>(getJDA(), user);
```

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1250,19 +1258,23 @@ public PrivateChannel createPrivateChannel(DataObject json)
...
                 .setLatestMessageIdLong(json.getLong("last_message_id", 0));
         user.setPrivateChannel(priv);
 
-        // only add channels to cache when they come from an event, otherwise we would never remove the channel
+        // only add channels to the cache when they come from an event, otherwise we would never remove the channel
+        cachePrivateChannel(priv);
+        return priv;
+    }
+
+    private void cachePrivateChannel(PrivateChannelImpl priv){
```
MinnDevelopment, line 1270:
```suggestion
    private void cachePrivateChannel(PrivateChannelImpl priv)
    {
```

src/main/java/net/dv8tion/jda/api/entities/PrivateChannel.java
```diff
@@ -27,8 +31,29 @@
...
+     * <ul>
+     *     <li>A reaction is removed</li>
+     *     <li>A reaction is added</li>
+     *     <li>A message is deleted</li>
+     *     <li>This account sends a message to a user from another shard (not shard 0)</li>
+     * </ul>
+     * The consequence of this is that for any message this bot receives from a guild or from other users will not be null.
+     *
+     *
+     * @return A potentially null {@link net.dv8tion.jda.api.entities.User User}.
```
MinnDevelopment, line 44:
```suggestion
     * @return Possibly-null {@link net.dv8tion.jda.api.entities.User User}.
```

src/main/java/net/dv8tion/jda/api/entities/PrivateChannel.java
```diff
@@ -27,8 +31,29 @@
...
+     *     <li>This account sends a message to a user from another shard (not shard 0)</li>
+     * </ul>
+     * The consequence of this is that for any message this bot receives from a guild or from other users will not be null.
+     *
+     *
+     * @return Possibly-null {@link net.dv8tion.jda.api.entities.User User}.
      */
-    @Nonnull
+    @Nullable
     User getUser();
```
MinnDevelopment, line 47: Add a reference to `retrieveUser()` on this method, it isn't easy to discover otherwise.

src/main/java/net/dv8tion/jda/api/events/message/react/GenericMessageReactionEvent.java
```diff
@@ -62,7 +62,7 @@ public GenericMessageReactionEvent(@Nonnull JDA api, long responseNumber, @Nulla
      *
      * @return The user id
      */
-    @Nonnull
+    @Nullable
     public String getUserId()
```
MinnDevelopment, line 66: This seems wrong to me

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -46,26 +48,42 @@ public ChannelType getType()
...
+        //even if the user blocks the bot, this does not fail.
+        return retrievePrivateChannel()
+                .map(PrivateChannel::getUser);
+    }
+
     @Nonnull
     @Override
     public String getName()
     {
+        if (getUser() == null)
```
MinnDevelopment, line 74: I would make this getUser() save to a local variable, since every call to getUser has to acquire the cache read-lock.

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -46,26 +48,42 @@ public ChannelType getType()
...
         updateUser();
         return user;
     }
 
+    @Nonnull
+    @Override
+    public RestAction<User> retrieveUser()
+    {
+        if (user != null)
+            return new CompletedRestAction<>(getJDA(), user);
```
MinnDevelopment, line 64:
```suggestion
            return new CompletedRestAction<>(getJDA(), getUser());
```

src/main/java/net/dv8tion/jda/api/events/message/react/GenericMessageReactionEvent.java
oliver276: Yep, it was a leftover change from when it could return null.

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -46,26 +48,42 @@ public ChannelType getType()
...
     @Nonnull
     @Override
     public String getName()
     {
+        if (user == null)
+        {
+            //don't break or override the contract of @NotNull
+            return "";
+        }
         return getUser().getName();
```
MinnDevelopment, line 79:
```suggestion
        User user = getUser();
        if (user == null)
        {
            //don't break or override the contract of @NotNull
            return "";
        }
        return user.getName();
```

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1233,6 +1233,14 @@ public PrivateChannel createPrivateChannel(DataObject json)
         if (channel != null)
             return channel;
 
+        //if we don't have enough information to construct the channel properly, the rest of the properties can be null
+        if (!json.hasKey("recipients") && !json.hasKey("recipient"))
```
MinnDevelopment, line 1237: Above this, we return without checking if the channel already has a user attached, maybe we should add a check and populate the user there if provided?

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -56,16 +58,33 @@ public User getUser()
...
     @Override
-    public String getName()
+    public RestAction<User> retrieveUser()
     {
-        return getUser().getName();
+        if (user != null)
+            return new CompletedRestAction<>(getJDA(), getUser());
+        //even if the user blocks the bot, this does not fail.
+        return retrievePrivateChannel()
+                .map(PrivateChannel::getUser);
```
MinnDevelopment, line 67:
```suggestion
        User user = getUser();
        if (user != null)
            return new CompletedRestAction<>(getJDA(), user);
        //even if the user blocks the bot, this does not fail.
        return retrievePrivateChannel()
                .map(PrivateChannel::getUser);
```

MinnDevelopment (approved):

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -162,7 +189,9 @@ private void updateUser()
 
     private void checkBot()
     {
-        if (getUser().isBot() && getJDA().getAccountType() == AccountType.BOT)
+        //the only way user is null, is when an event is dispatched that doesn't give us enough information
+        //since events will only happen in this channel if it is between the bot and the user, a null user is a valid channel state
```
DV8FromTheWorld, line 193:
```suggestion
        //The only way user is null is when an event is dispatched that doesn't give us enough information to build the recipient user,
        // which only happens if this bot sends a message (or otherwise triggers an event) from a shard other than shard 0.
        // The event will be received on shard 0 and not have enough information to build the recipient user.
        //As such, since events will only happen in this channel if it is between the bot and the user, a null user is a valid channel state. 
        // Events cannot happen between a bot and another bot, so the user would never be null in that case.
```

src/main/java/net/dv8tion/jda/internal/entities/PrivateChannelImpl.java
```diff
@@ -85,7 +105,9 @@ public long getLatestMessageIdLong()
     @Override
     public boolean canTalk()
     {
-        return !user.isBot();
+        //the only way user is null, is when an event is dispatched that doesn't give us enough information
+        //since events will only happen in this channel if it is between the bot and the user, a null user is a valid channel state
+        return user == null || !user.isBot();
```
DV8FromTheWorld, line 110:
```suggestion
        //The only way user is null is when an event is dispatched that doesn't give us enough information to build the recipient user,
        // which only happens if this bot sends a message (or otherwise triggers an event) from a shard other than shard 0.
        // The event will be received on shard 0 and not have enough information to build the recipient user.
        //As such, since events will only happen in this channel if it is between the bot and the user, a null user is a valid channel state. 
        // Events cannot happen between a bot and another bot, so the user would never be null in that case.
        return user == null || !user.isBot();
```

src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java
```diff
@@ -137,7 +137,7 @@ public PrivateChannel getPrivateChannel()
         if (!hasPrivateChannel())
             return null;
         PrivateChannel channel = getJDA().getPrivateChannelById(privateChannel);
-        return channel != null ? channel : new PrivateChannelImpl(privateChannel, this);
+        return channel != null ? channel : new PrivateChannelImpl(getJDA(), privateChannel, this);
```
DV8FromTheWorld, line 140: I didn't realize we were doing this. This technically is creating a private channel when we are missing info like `setLastMessageId`. Oh well I guess.

DV8FromTheWorld: Also, we should probably rename `privateChannel` here to `privateChannelId` as the variable isn't fantastically named tbh.

src/main/java/net/dv8tion/jda/internal/handle/MessageReactionHandler.java
```diff
@@ -105,6 +105,7 @@ protected Long handleInternally(DataObject content)
         User user = api.getUserById(userId);
         if (user == null && member != null)
             user = member.getUser(); // this happens when we have guild subscriptions disabled
+
         if (user == null)
         {
             if (add && guild != null)
```
DV8FromTheWorld, line 111: Lets add a comment to this line like:
```
//We expect there to be a user object already cached when we are in a guild and adding a new reaction as the user should be a member cached in the guild. 
// The only time we can receive a reaction add but not have the user cached would be if we receive the event in an uncached or partially built PrivateChannel. 
```

DV8FromTheWorld (approved):

Issue #2009, opened 2022-02-04
Change behavior of MessageChannel#getLatestMessageId


DV8FromTheWorld:
Currently, JDA sets the latestMessageId to `null` if the latest message is deleted, however should not be reset upon message delete because this ID is useful to check if a database is dirty. 

#### Expected Change
Update JDA to stop resetting the `latestMessageId` for MessageChannel entities when receiving a MESSAGE_DELETE

PR #2013, opened 2022-02-04
Dont override latest message id on deletion event

ishwi:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #2009

## Description

The internal handle of the Message Deletion event no longer sets the LatestMessageIdLong to 0.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
@@ -89,7 +89,7 @@ public interface MessageChannel extends Channel, Formattable
      *
      * <p>This value is updated on each {@link net.dv8tion.jda.api.events.message.MessageReceivedEvent MessageReceivedEvent}
      * and <u><b>the value might point to an already deleted message since the ID is not cleared when the message is deleted,
-     * so calling {@link #retrieveMessageById(long)} with this id can throw {@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}</b></u>
+     * so calling {@link #retrieveMessageById(long)} with this id can result in an {@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE} error</b></u>
      *
      * @return The most recent message's id
      */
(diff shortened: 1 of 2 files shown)


DV8FromTheWorld (approved):

DV8FromTheWorld:
Actually, I forgot: We need to update the docs for `getLatestMessageId` to no longer mention that the id gets reset if the message gets deleted.
We need to instead mention that the id could point to a deleted message if `MessageChannel#retrieveMessageById` is used

MinnDevelopment (changes requested):
Please also update the relevant docs on the getters for latest message id

ishwi (commented):
Is the text added clear or should it be expressed in any other way?

MinnDevelopment:
We need to make a decision on the TODO for line 100/101:

> //TODO-v5: Revisit this. Surely this should be Nullable instead of throw an exception...

@DV8FromTheWorld do you have any final thoughts on this? If we decide not to change the behavior, we should remove the comment in this PR.

DV8FromTheWorld:
Let's remove the exception from this getter. That means we no longer need the hasLastMessageId getter anymore either.

MinnDevelopment:
@ishwi can you remove the exception and `hasLatestMessage` getter? After that, it should be ready to merge.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -86,18 +86,14 @@
...
      * in this current MessageChannel.
-     * <br>This should only be used if {@link #hasLatestMessage()} returns {@code true}!
      *
      * <p>This value is updated on each {@link net.dv8tion.jda.api.events.message.MessageReceivedEvent MessageReceivedEvent}
-     * and <u><b>will be reset to {@code null} if the message associated with this ID gets deleted</b></u>
-     *
-     * @throws java.lang.IllegalStateException
-     *         If no message id is available
+     * and <u><b>the value might point to an already deleted message since the ID is not cleared when the message is deleted,
+     * so calling {@link #retrieveMessageById(long)} with this id can throw {@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}</b></u>
```
DV8FromTheWorld, line 92:
```suggestion
     * so calling {@link #retrieveMessageById(long)} with this id can result in an {@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE} error</b></u>
```

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -106,36 +102,15 @@ default String getLatestMessageId()
...
      * in this current MessageChannel.
-     * <br>This should only be used if {@link #hasLatestMessage()} returns {@code true}!
      *
      * <p>This value is updated on each {@link net.dv8tion.jda.api.events.message.MessageReceivedEvent MessageReceivedEvent}
-     * and <u><b>will be reset to {@code null} if the message associated with this ID gets deleted</b></u>
-     *
-     * @throws java.lang.IllegalStateException
-     *         If no message id is available
+     * and <u><b>the value might point to an already deleted message since the value is not cleared when the message is deleted,
+     * so calling {@link #retrieveMessageById(long)} with this id can throw {@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}</b></u>
```
DV8FromTheWorld, line 108:
```suggestion
     * so calling {@link #retrieveMessageById(long)} with this id can result in an {@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE} error</b></u>
```

DV8FromTheWorld (changes requested):

DV8FromTheWorld (approved):

Issue #2008, opened 2022-02-04
Introduce Event#getRawData


DV8FromTheWorld:
We can add a getter on events to access the passthrough event data in a `DataObject`. This should be off by default and configurable on the JDABuilder. Something akin to `JDABuilder#setEventPassthrough(true)`.

Some events do not have any data to pass through, in which case we can either throw an exception or return null.

MinnDevelopment:
This is out with 5.0.0-alpha.14

PR #2014, opened 2022-02-04
Add JDABuilder#setEventPassthrough

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #2008 

## Description

Adds `JDABuilder#setEventPassthrough`, which allows JDA to pass the `DataObject` Discord sends on each event, this might be useful if Discord added a property to an object and JDA has not been updated yet


src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -520,6 +520,24 @@ public JDABuilder setRawEventsEnabled(boolean enable)
         return setFlag(ConfigFlag.RAW_EVENTS, enable);
     }
 
+    /**
+     * Whether JDA should store the raw {@link net.dv8tion.jda.api.utils.data.DataObject} for every discord event.
+     * <br>Warning: be aware that enabling this could consume a lot of memory if your event objects have a long lifetime
+     * <br>Default: {@code false}
+     *
+     * @param  enable
+     *         True, if JDA should add the raw {@link net.dv8tion.jda.api.utils.data.DataObject} to every discord events.
```
caneleex, line 529:
```suggestion
     *         True, if JDA should add the raw {@link net.dv8tion.jda.api.utils.data.DataObject} to every discord event.
```

src/main/java/net/dv8tion/jda/internal/handle/GuildRoleUpdateHandler.java
```diff
@@ -74,7 +74,7 @@ protected Long handleInternally(DataObject content)
             role.setName(name);
             getJDA().handleEvent(
                     new RoleUpdateNameEvent(
-                            getJDA(), responseNumber,
+                            getJDA(), responseNumber, getJDA().isEventPassthrough() ? content : null,
```
MinnDevelopment, line 77: It makes more sense to just handle this inside the event constructor, instead of repeating the same check everywhere

src/main/java/net/dv8tion/jda/api/events/GenericEvent.java
```diff
@@ -37,4 +38,24 @@
...
+     * @see #hasRawData()
+     */
+    @Nonnull
+    DataObject getRawData();
+
+    /**
+     * Whether this event contains its raw event data
+     *
+     * @return <code>True</code> if the event has raw data, <code>false</code> if not
+     * @see #getRawData()
```
MinnDevelopment, line 58:
```suggestion
     * @return True. if the event has raw data
     *
     * @see    #getRawData()
```

MinnDevelopment (changes requested):

src/main/java/net/dv8tion/jda/api/events/GenericEvent.java
```diff
@@ -37,4 +38,28 @@
...
+     */
+    @Nonnull
+    DataObject getRawData();
+
+    /**
+     * Whether this event contains its raw event data
+     *
+     * @return True if the event has raw data
+     *
+     * @see #getRawData()
```
Tais993, line 62:
```suggestion
     * @see    #getRawData()
```

Sanduhr32:
any update on this?

src/main/java/net/dv8tion/jda/api/events/thread/member/GenericThreadMemberEvent.java
```diff
@@ -33,7 +33,7 @@ public class GenericThreadMemberEvent extends GenericThreadEvent
 
     public GenericThreadMemberEvent(@Nonnull JDA api, long responseNumber, ThreadChannel thread, long threadMemberId, ThreadMember threadMember)
     {
-        super(api, responseNumber, thread);
+        super(api, responseNumber, null, thread);
```
DV8FromTheWorld, line 36: Why is this null?

src/main/java/net/dv8tion/jda/api/events/user/update/UserUpdateActivitiesEvent.java
```diff
@@ -57,7 +57,7 @@ public class UserUpdateActivitiesEvent extends GenericUserUpdateEvent<List<Activ
 
     public UserUpdateActivitiesEvent(@Nonnull JDA api, long responseNumber, @Nonnull Member member, @Nullable List<Activity> previous)
     {
-        super(api, responseNumber, member.getUser(), previous, member.getActivities(), IDENTIFIER);
+        super(api, responseNumber, null, member.getUser(), previous, member.getActivities(), IDENTIFIER);
```
DV8FromTheWorld, line 60: why is this null?

src/main/java/net/dv8tion/jda/internal/handle/ChannelUpdateHandler.java
```diff
@@ -444,7 +444,7 @@ private void applyPermissions(IPermissionContainerMixin<?> channel, DataArray pe
             addPermissionHolder(changed, guild, override.getIdLong());
             api.handleEvent(
                 new PermissionOverrideDeleteEvent(
-                    api, responseNumber,
+                    api, responseNumber, null,
```
DV8FromTheWorld, line 447: Why is this null?

DV8FromTheWorld (commented):
This looks pretty good. My only questions come down to "why are certain events getting `null` for their raw data and others aren't. It seems seemingly at random.

I've left comments on a few, but there are a bunch, whether via directly passing in `null` when the event is fired, or it is provide as a default value to the `super`. I'd like to better understand why, along with some comments in the code in the various places explaining why.

src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -520,6 +520,22 @@ public JDABuilder setRawEventsEnabled(boolean enable)
         return setFlag(ConfigFlag.RAW_EVENTS, enable);
     }
 
+    /**
+     * Whether JDA should store the raw {@link net.dv8tion.jda.api.utils.data.DataObject} for every discord event.
```
Tais993, line 524:
```suggestion
     * Whether JDA should store the raw {@link net.dv8tion.jda.api.utils.data.DataObject DataObject} for every discord event.
```

src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -520,6 +520,22 @@ public JDABuilder setRawEventsEnabled(boolean enable)
...
+     * <br>Warning: be aware that enabling this could consume a lot of memory if your event objects have a long lifetime
+     * <br>Default: {@code false}
+     *
+     * @param  enable
+     *         True, if JDA should add the raw {@link net.dv8tion.jda.api.utils.data.DataObject} to every discord event.
```
Tais993, line 529:
```suggestion
     *         True, if JDA should add the raw {@link net.dv8tion.jda.api.utils.data.DataObject DataObject} to every discord event.
```

Tais993 (changes requested):

src/main/java/net/dv8tion/jda/api/events/Event.java
```diff
@@ -45,6 +49,7 @@ public Event(@Nonnull JDA api, long responseNumber)
     {
         this.api = api;
         this.responseNumber = responseNumber;
+        this.rawData = ((JDAImpl) api).isEventPassthrough() ? SocketHandler.CURRENT_EVENT.get() : null;
```
MinnDevelopment, line 52:
```suggestion
        this.rawData = api instanceof JDAImpl && ((JDAImpl) api).isEventPassthrough() ? SocketHandler.CURRENT_EVENT.get() : null;
```

src/main/java/net/dv8tion/jda/api/events/Event.java
```diff
@@ -71,4 +76,20 @@ public long getResponseNumber()
     {
         return responseNumber;
     }
+
+    @Nonnull
+    @Override
+    public DataObject getRawData()
+    {
+        if (rawData == null)
+            throw new IllegalStateException("This event does not contain any raw data, or event passthrough is not enabled, see JDABuilder#setEventPassthrough(boolean)");
```
MinnDevelopment, line 85: Is this needed?

src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -519,6 +519,22 @@ public JDABuilder setRawEventsEnabled(boolean enable)
         return setFlag(ConfigFlag.RAW_EVENTS, enable);
     }
 
+    /**
+     * Whether JDA should store the raw {@link net.dv8tion.jda.api.utils.data.DataObject DataObject} for every discord event.
+     * <br>Warning: be aware that enabling this could consume a lot of memory if your event objects have a long lifetime
```
MinnDevelopment, line 524:
```suggestion
     * <br>Warning: be aware that enabling this could consume a lot of memory if your event objects have a long lifetime.
```

src/main/java/net/dv8tion/jda/api/events/Event.java
```diff
@@ -45,6 +49,7 @@ public Event(@Nonnull JDA api, long responseNumber)
     {
         this.api = api;
         this.responseNumber = responseNumber;
+        this.rawData = api instanceof JDAImpl && ((JDAImpl) api).isEventPassthrough() ? SocketHandler.CURRENT_EVENT.get() : null;
```
DV8FromTheWorld, line 52: You don't need this instanceof check. JDA will always be JDAImpl

src/main/java/net/dv8tion/jda/api/events/Event.java
```diff
@@ -71,4 +76,20 @@ public long getResponseNumber()
...
     }
+
+    @Nonnull
+    @Override
+    public DataObject getRawData()
+    {
+        if (rawData == null)
+            throw new IllegalStateException("This event does not contain any raw data, or event passthrough is not enabled, see JDABuilder#setEventPassthrough(boolean)");
+
+        return rawData;
```
DV8FromTheWorld, line 87: I'm not sure about this `hasRawData` and `getRawData` w/ exception pattern.

I think we should remove `hasRawData` and instead return `null` if there is no data.
We should only throw here if `isEventPassthrough` is false.

src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
```diff
@@ -518,6 +518,22 @@ public DefaultShardManagerBuilder setRawEventsEnabled(boolean enable)
...
+     * @param  enable
+     *         True, if JDA should add the raw {@link net.dv8tion.jda.api.utils.data.DataObject DataObject} to every discord event.
+     *
+     * @return The DefaultShardManagerBuilder instance. Useful for chaining.
+     */
+    @Nonnull
+    public DefaultShardManagerBuilder setEventPassthrough(boolean enable)
+    {
+        return setFlag(ConfigFlag.EVENT_PASSTHROUGH, enable);
+    }
```
DV8FromTheWorld, line 535: Add an `@see` pointed at the new method on Event.
Also, we should probably add information to point to what _type_ of data they might see. Specifically, we should point them at the gateway docs

https://discord.com/developers/docs/topics/gateway#commands-and-events-gateway-events

src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -519,6 +519,22 @@ public JDABuilder setRawEventsEnabled(boolean enable)
...
+     * @param  enable
+     *         True, if JDA should add the raw {@link net.dv8tion.jda.api.utils.data.DataObject DataObject} to every discord event.
+     *
+     * @return The JDABuilder instance. Useful for chaining.
+     */
+    @Nonnull
+    public JDABuilder setEventPassthrough(boolean enable)
+    {
+        return setFlag(ConfigFlag.EVENT_PASSTHROUGH, enable);
+    }
```
DV8FromTheWorld, line 536: Add an `@see` pointed at the new method on Event.
Also, we should probably add information to point to what _type_ of data they might see. Specifically, we should point them at the gateway docs

https://discord.com/developers/docs/topics/gateway#commands-and-events-gateway-events

src/main/java/net/dv8tion/jda/api/events/GenericEvent.java
```diff
@@ -37,4 +38,28 @@
...
+     *
+     * @throws IllegalStateException 
+     *         If the event has no raw data, see {@link #hasRawData()}
+     *
+     * @return The corresponding {@link DataObject}
+     *
+     * @see    #hasRawData()
+     */
+    @Nonnull
+    DataObject getRawData();
```
DV8FromTheWorld, line 55: Lets update the docs link to this https://discord.com/developers/docs/topics/gateway#commands-and-events-gateway-events

src/main/java/net/dv8tion/jda/internal/handle/SocketHandler.java
```diff
@@ -33,10 +35,13 @@ public final synchronized void handle(long responseTotal, DataObject o)
     {
         this.allContent = o;
         this.responseNumber = responseTotal;
-        final Long guildId = handleInternally(o.getObject("d"));
+        final DataObject data = o.getObject("d");
+        if (getJDA().isEventPassthrough()) CURRENT_EVENT.set(data);
```
DV8FromTheWorld, line 39: Not sure I agree with this.
Providing access to the `t` as well would allow users to look at the documentation to know what event the data is for when looking at the Discord Documentation.

I'd rather we just send down the `o` here than pulling out the `d`

DV8FromTheWorld: It would also match the format of what we send in `RawGatewayEvent`

DV8FromTheWorld (changes requested):

src/main/java/net/dv8tion/jda/api/events/Event.java
MinnDevelopment: This is for the purpose of mocking.

src/main/java/net/dv8tion/jda/api/events/Event.java
MinnDevelopment: If we throw on `isEventPassthrough` false, we have to provide an easy way to check for this too.

src/main/java/net/dv8tion/jda/internal/handle/SocketHandler.java
MinnDevelopment: I don't think it is worth doing that, since if you want to handle that level you should be using `RawGatewayEvent`. The point of the passthrough is to use it on specific events, where the type is already known.

src/main/java/net/dv8tion/jda/api/events/Event.java
DV8FromTheWorld: Do we? They set it in the JDABuilder from the start. We throw an exception stating "if you want to use this feature, enable it in the builder".

src/main/java/net/dv8tion/jda/api/events/Event.java
DV8FromTheWorld: We do this cast blindly throughout all of JDA. What makes you think this situation is different?

DV8FromTheWorld (approved):

PR #2016, opened 2022-02-05
Make getDefaultChannel return BaseGuildMessageChannel

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Changes the return type to `BaseGuildMessageChannel` because it can also be a news channel.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2019,17 +2019,17 @@ public interface Guild extends IGuildChannelContainer, ISnowflake
     Role getPublicRole();
 
     /**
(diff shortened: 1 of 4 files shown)


DV8FromTheWorld:
We might actually need to make this a GuildChannel because I bet ForumChannels will able to be defaults.

Maybe we can cross that bridge when we get to it though?

DV8FromTheWorld:
Until we learn more about the full functionality available to ForumChannels, we will allow this to be a BaseGuildMessageChannel.
If ForumChannels can be used as the default channel then we will revisit this functionality at that point.

DV8FromTheWorld (approved):

PR #2018, opened 2022-02-05
Add CommandInteractionPayload#getOption fallback overloads

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds fallback methods which can be used to more easily handle optional command arguments without needing 3 lines of code:

```java
public void onSlashCommandInteraction(SlashCommandInteractionEvent event)
{
    //                              name      fallback          resolver
    
    // null if user is not a member of the guild (fallback not used, since option is mapped to a User instance)
    Member member = event.getOption("target", event::getMember, OptionMapping::getAsMember);
    // null if fallback is used (getAsUser cannot be null)
    User user =     event.getOption("target", null            , OptionMapping::getAsUser);
    // null fallback implied
    User nullable = event.getOption("target",                   OptionMapping::getAsUser); // <- implicit fallback null

    // real use-cases:
    int days = event.getOption("days", 7, OptionMapping::getAsInt); // optional ban days
    String reason = event.getOption("reason", "banned by mod", OptionMapping::getAsString); // optional ban reason
}
```

It may seem strange to put the resolver function as the last argument, but this goes back to kotlin interop:

```kt
// if it is the second argument you have to write this
val user = event.getOption("user", { it.asUser }, event.user)
// with it as last argument it becomes this
val user = event.getOption("user", event.user) { it.asUser }
```

An easy way to interpret the fallback syntax is to think of it as a ternary, like `option == null ? fallback : option.value`

### DIFF
--- a/src/examples/java/SlashBotExample.java
+++ b/src/examples/java/SlashBotExample.java
@@ -52,7 +52,9 @@ public class SlashBotExample extends ListenerAdapter
             Commands.slash("ban", "Ban a user from this server. Requires permission to ban users.")
                 .addOptions(new OptionData(USER, "user", "The user to ban") // USER type allows to include members of the server or other users by id
                     .setRequired(true)) // This command requires a parameter
-                .addOptions(new OptionData(INTEGER, "del_days", "Delete messages from the past days.")) // This is optional
+                .addOptions(new OptionData(INTEGER, "del_days", "Delete messages from the past days.") // This is optional
+                        .setRequiredRange(0, 7)) // Only allow values between 0 and 7 (inclusive)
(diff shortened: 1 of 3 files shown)


MinnDevelopment:
I added `OptionMapping#getAsInt` which uses `Math.toIntExact(value)` to assert int compatibility. This is useful because you can't cast `java.lang.Long` to `int`, which is a problem when using generics for `getOption(...)`.

DV8FromTheWorld (approved):

PR #2022, opened 2022-02-07
Add Booster to memberCachePolicy

SweetSofiMC:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

## Description

- Introduces method Member.isBoosting() without allocation of OffsetDateTime, using class attribute boostDate.
- Adds MemberCachePolicy.BOOSTER featuring the new method isBoosting.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Member.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Member.java
@@ -122,7 +122,7 @@ public interface Member extends IMentionable, IPermissionHolder
     /**
      * Returns whether a member is boosting the guild or not.
      *
-     * @return true if it is boosting, false otherwise.
+     * @return True, if it is boosting
      */
     boolean isBoosting();
 
--- a/src/main/java/net/dv8tion/jda/api/entities/Member.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Member.java
@@ -120,6 +120,13 @@ public interface Member extends IMentionable, IPermissionHolder
     OffsetDateTime getTimeBoosted();
 
     /**
+     * Returns whether a member is boosting the guild or not.
+     *
+     * @return true if it is boosting, false otherwise.
+     */
+    boolean isBoosting();
+
+    /**
      * The time this Member will be released from time out.
      * <br>If this Member is not in time out, this returns {@code null}.
      * This may also return dates in the past, in which case the time out has expired.
--- a/src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
@@ -117,7 +117,8 @@ public class MemberImpl implements Member
     }
 
     @Override
-    public boolean isBoosting(){
+    public boolean isBoosting()
+    {
         return boostDate != 0;
     }
 
--- a/src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
@@ -116,6 +116,11 @@ public class MemberImpl implements Member
         return boostDate != 0 ? Helpers.toOffset(boostDate) : null;
     }
 
+    @Override
+    public boolean isBoosting(){
+        return boostDate != 0;
+    }
+
     @Nullable
     @Override
     public OffsetDateTime getTimeOutEnd()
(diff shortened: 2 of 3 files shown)


src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
```diff
@@ -116,6 +116,11 @@ public OffsetDateTime getTimeBoosted()
         return boostDate != 0 ? Helpers.toOffset(boostDate) : null;
     }
 
+    @Override
+    public boolean isBoosting(){
```
RedDaedalus, line 120:
```suggestion
    public boolean isBoosting()
    {
```

SweetSofiMC: Thank you! Forgot about the convention

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
```diff
@@ -116,6 +116,12 @@ public OffsetDateTime getTimeBoosted()
         return boostDate != 0 ? Helpers.toOffset(boostDate) : null;
```
java-coding-prodigy, line 116:
```suggestion
        return isBoosting() ? Helpers.toOffset(boostDate) : null;
```
This is better IMO, would like to hear your opinions on this.

SweetSofiMC: Hmm, I'm not sure about this, both approaches are certainly valid, it reuses the new code but I feel it adds an unnecessary step in the stack.

java-coding-prodigy: If we decide to change what `boostDate` contains if the member is not boosting, then this will be easier. Also, it is also more readable
About the extra method call, I think it should be fine, I doubt we should try to over optimize our code like that.

src/main/java/net/dv8tion/jda/api/entities/Member.java
```diff
@@ -119,6 +119,13 @@
     @Nullable
     OffsetDateTime getTimeBoosted();
 
+    /**
+     * Returns whether a member is boosting the guild or not.
+     *
+     * @return true if it is boosting, false otherwise.
```
DManstrator, line 125: Not sure if that breaks consistency or not but I would prefer `{@code true|false}` over true|false.

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
```diff
@@ -116,6 +116,12 @@ public OffsetDateTime getTimeBoosted()
...
     }
 
+    @Override
+    public boolean isBoosting()
+    {
+        return boostDate != 0;
```
DManstrator, line 122: This is perhaps out of scope for this PR but why is a that variable called `boostDate` when it is a long? I think it contains a timestamp so `boostTimestamp` would fit more and is less confusing to read.

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
java-coding-prodigy: So @SweetSofiMC , are you going to implement this suggestion?

src/main/java/net/dv8tion/jda/api/entities/Member.java
SweetSofiMC: It wouldn't be consistent, but I found it should be True and not true, so I'll change it

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
SweetSofiMC: Well... yeah, but I'm not sure if I should change it. It's private and internal anyways, but then we should also change `joinDate`, which is a long.

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
SweetSofiMC: Yes, it shouldn't be much of a performance problem. Will change it!

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
SweetSofiMC: IMO it would best fit in another PR, since it would change a lot of the method names and usages.

MinnDevelopment (approved):

PR #2029, opened 2022-02-19
Add missing cache cleanup for news/stage/thread channels

Mitmocc:
## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

There is a bug where if you kick a bot from a guild and then re-invite it, news channels would return null if fetched from the jda instance. I don't understand it fully but there might be objects of the newsChannel left when the bot is kicked because the GuildDeleteHandler wasn't cleaning the global cache from the news channels. This change fixes that bug.

Edit: Same thing applied to stage channels and thread channels, so I added those too.


V-Play-Games:
With that logic, store channels should behave the same way too

Mitmocc:
> With that logic, stage channels should behave the same way too

You are right, didn't notice that

MinnDevelopment (approved):

DV8FromTheWorld (approved):

PR #2038, opened 2022-02-21
Make Messages for Interactions properly reference their channels instead of PrivateChannel

oliver276:
Put `guild_id` into MessageContextInteractionImpl's messages as Discord doesn't provide them

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Discord message context menu events do not put `guild_id` into the message payload.  This PR will put that into the payload so that the EntityBuilder constructs the correct channel type surrounding it


DV8FromTheWorld:
This is a bandaid fix for now. After discussing with Discord devs it has been made apparently that we will only ever receive a `guild_id` with a message object when it is received over the websocket gateway.

As such, we need a more fundamental restructure for how we build Message objects to handle this lack of information as right now the system is open to these kinds of problems in future additions.

PR #2053, opened 2022-03-09
Do not ignore member cache policy when chunking is enabled

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

With this change, the `ChunkingFilter` no longer overwrites `MemberCachePolicy`. The reason this was originally implemented this way was due to the way chunking worked in the past. Back when it was implemented, there was no reliable way to check which chunk is the last in the sequence. Discord has since improved this API and we can now reliably chunk without checking the member count.

This change improves user-experience by disentangling the member loading from member caching.

I tested this locally and could not notice any issues with this change. However, it would be nice if others could test as well.

### Migration

To recover the previous behavior, you must use `setMemberCachePolicy(MemberCachePolicy.ALL)` in *addition* to your chunking configuration.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/JDABuilder.java
@@ -55,7 +55,6 @@ import java.util.stream.Collectors;
  */
 public class JDABuilder
 {
-    public static final int GUILD_SUBSCRIPTIONS = GatewayIntent.getRaw(GatewayIntent.GUILD_MEMBERS, GatewayIntent.GUILD_PRESENCES, GatewayIntent.GUILD_MESSAGE_TYPING);
     protected final List<Object> listeners = new LinkedList<>();
     protected final EnumSet<CacheFlag> automaticallyDisabled = EnumSet.noneOf(CacheFlag.class);
 
@@ -1410,7 +1409,7 @@ public class JDABuilder
     /**
      * The {@link ChunkingFilter} to filter which guilds should use member chunking.
      *
-     * <p>If a guild is configured for chunking the {@link #setMemberCachePolicy(MemberCachePolicy)} will be ignored.
+     * <p>Use {@link #setMemberCachePolicy(MemberCachePolicy)} to configure which members to keep in cache from chunking.
      *
      * @param  filter
      *         The filter to apply
--- a/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
@@ -1812,9 +1812,8 @@ public class  DefaultShardManagerBuilder
 
     /**
      * The {@link ChunkingFilter} to filter which guilds should use member chunking.
-     * <br>By default this uses {@link ChunkingFilter#ALL}.
      *
-     * <p>If a guild is configured for chunking the {@link #setMemberCachePolicy(MemberCachePolicy)} will be ignored.
+     * <p>Use {@link #setMemberCachePolicy(MemberCachePolicy)} to configure which members to keep in cache from chunking.
      *
      * @param  filter
      *         The filter to apply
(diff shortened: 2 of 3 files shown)


src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -1410,7 +1409,7 @@ public JDABuilder setVoiceDispatchInterceptor(@Nullable VoiceDispatchInterceptor
     /**
      * The {@link ChunkingFilter} to filter which guilds should use member chunking.
      *
-     * <p>If a guild is configured for chunking the {@link #setMemberCachePolicy(MemberCachePolicy)} will be ignored.
+     * <p>Use {@link #setMemberCachePolicy(MemberCachePolicy)} to configure which members to keep in cache from chunking.
```
DV8FromTheWorld, line 1412:
```suggestion
     * <p>Use {@link #setMemberCachePolicy(MemberCachePolicy)} to configure which members to keep in cache from chunking. By default, {@link ChunkingFilter.ALL all members will be retained} when a Guild is chunked.
```

src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
```diff
@@ -1812,9 +1812,8 @@ public DefaultShardManagerBuilder setWebsocketFactory(@Nullable WebSocketFactory
 
     /**
      * The {@link ChunkingFilter} to filter which guilds should use member chunking.
-     * <br>By default this uses {@link ChunkingFilter#ALL}.
      *
-     * <p>If a guild is configured for chunking the {@link #setMemberCachePolicy(MemberCachePolicy)} will be ignored.
+     * <p>Use {@link #setMemberCachePolicy(MemberCachePolicy)} to configure which members to keep in cache from chunking.
```
DV8FromTheWorld, line 1816: I'm noticing that the `setChunkingFilter` for DefaultShardManagerBuilder is different in behavior to that of JDABuilder. JDABuilder sets ChunkingFilter.ALL. Is there a reason the shard manager doesn't?

DV8FromTheWorld:
> Discord has since improved this API and we can now reliably chunk without checking the member count.

This PR doesn't include a functionality change in regards to actually chunking, so are we using this new API? How does it work?

src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManagerBuilder.java
MinnDevelopment: The behavior is the same since you can't create one without setting a chunking filter

src/main/java/net/dv8tion/jda/api/JDABuilder.java
MinnDevelopment: That is not true, the defaults depend on the JDABuilder factory you used. For instance, `JDABuilder.createDefault` sets it to `NONE` and `JDABuilder.create` only sets it to `ALL`, if the `GUILD_MEMBERS` intent is enabled.

MinnDevelopment:
> This PR doesn't include a functionality change in regards to actually chunking, so are we using this new API? How does it work?

We have been using the new functionality since it was introduced in the API, the behavior of cache was just never adjusted because it is semi-breaking.

https://github.com/DV8FromTheWorld/JDA/blob/555adcbd76a9baf46ee19b30b1a0dc25fe7312a5/src/main/java/net/dv8tion/jda/internal/handle/GuildSetupController.java#L226

src/main/java/net/dv8tion/jda/api/JDABuilder.java
DV8FromTheWorld: Ah, appologies. I read the code in this method incorrectly. I incorrectly read "ChunkingFilter.All" to mean "MemberCachePolicy.All" basically.

DV8FromTheWorld (approved):

PR #2054, opened 2022-03-12
Add missing method overrides for `GuildImpl#ban` & `GuildImpl#kick`

DynxstyGIT:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds a missing override for the `GuildImpl#ban` method, which uses a `long` user id.

Closed in favor of #2065

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
@@ -1327,6 +1327,13 @@ public class GuildImpl implements Guild
     }
 
     @Nonnull
+    @Override
+    public AuditableRestAction<Void> kick(long userId, @Nullable String reason)
+    {
+        return kick(Long.toUnsignedString(userId), reason);
+    }
+
+    @Nonnull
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1365,6 +1365,20 @@ public long getIdLong()
         return ban0(userId, delDays, reason);
     }
 
+    @Nonnull
+    @Override
+    public AuditableRestAction<Void> ban(long userId, int delDays, String reason)
```
caneleex, line 1370: why not just use the string overload with `Long.toUnsignedString` instead of copying the whole impl?

DynxstyGIT: yep, you're right

MoonTM-GIT:
Pretty sure this is also missing for `GuildImpl#kick`

caneleex:
you can make both new overloads default methods that just call the string overloads (`Long.toUnsignedString`) within the interface.

PR #2062, opened 2022-03-20
Added getters for EmbedBuilder

VigusWG:
## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Added to EmbedBuilder
- getTitle()
- getFooter()
- getTimestamp()
- getColor()
- getThumbnail()
- getImage()
- getAuthor()

Currently there is no other way to get them.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -237,7 +237,7 @@ public class EmbedBuilder
     /**
      * The title that will be used for the builder, this will return null when no title was found.
      *
-     * @return String with current title.
+     * @return Nullable String containing the title.
      */
     @Nonnull
     public String getTitle()
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -378,9 +378,9 @@ public class EmbedBuilder
     }
 
     /**
-     * The {@link java.time.OffsetDateTime Timestamp} of the embed
+     * The Timestamp of the builder
      *
-     * @return Timestamp the builder has been set with.
+     * @return Nullable {@link java.time.OffsetDateTime OffsetDateTime} object representing the timestamp
      */
     @Nonnull
     public OffsetDateTime getTimestamp()
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -237,9 +237,9 @@ public class EmbedBuilder
     /**
      * The title that will be used for the builder, this will return null when no title was found.
      *
-     * @return Nullable String containing the title.
+     * @return Possibly-null String containing the title.
      */
-    @Nonnull
+    @Nullable
     public String getTitle()
     {
         return title;
@@ -380,9 +380,9 @@ public class EmbedBuilder
     /**
      * The Timestamp of the builder
      *
-     * @return Nullable {@link java.time.OffsetDateTime OffsetDateTime} object representing the timestamp
+     * @return Possibly-null {@link java.time.OffsetDateTime OffsetDateTime} object representing the timestamp
      */
-    @Nonnull
+    @Nullable
     public OffsetDateTime getTimestamp()
     {
         return timestamp;
@@ -412,14 +412,14 @@ public class EmbedBuilder
      * The color of the stripe on the side of the embed.
      * <br>If the color is 0, this will return null.
      *
-     * @return Nullable {@link java.awt.Color Color}
+     * @return Possibly-null {@link java.awt.Color Color}
      *
      * @see    #getColorRaw()
      */
-    @Nonnull
+    @Nullable
     public Color getColor()
     {
-        return new Color(color);
+        return color == 0 ? null : new Color(color);
     }
 
     /**
@@ -478,7 +478,7 @@ public class EmbedBuilder
      * @return Possibly-null {@link net.dv8tion.jda.api.entities.MessageEmbed.Thumbnail Thumbnail} instance
      *         containing general information on the displayable thumbnail.
      */
-    @Nonnull
+    @Nullable
     public MessageEmbed.Thumbnail getThumbnail()
     {
         return thumbnail; 
@@ -599,7 +599,7 @@ public class EmbedBuilder
      * @return Possibly-null {@link net.dv8tion.jda.api.entities.MessageEmbed.AuthorInfo AuthorInfo}
      *         containing author information.
      */
-    @Nonnull
+    @Nullable
     public MessageEmbed.AuthorInfo getAuthor()
     {
         return author; 
@@ -724,7 +724,7 @@ public class EmbedBuilder
      * @return Possibly-null {@link net.dv8tion.jda.api.entities.MessageEmbed.Footer Footer}
      *         containing the embed footer content.
      */
-    @Nonnull
+    @Nullable
     public MessageEmbed.Footer getFooter()
     {
         return footer;
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -529,10 +529,10 @@ public class EmbedBuilder
     }
 
     /**
-     * The {@link net.dv8tion.jda.api.entities.MessageEmbed.ImageInfo ImageInfo} used to
-     * build the image for the embed.
+     * The information about the image in builder, or null when not set
      *
-     * @return ImageInfo with current image context.
+     * @return Possibly-null {@link net.dv8tion.jda.api.entities.MessageEmbed.ImageInfo ImageInfo}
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -386,6 +409,17 @@ public EmbedBuilder setTimestamp(@Nullable TemporalAccessor temporal)
         return this;
     }
 
+    /**
+     * The {@link java.awt.Color Color} of the embed
+     *
+     * @return Color the builder has been set with.
+     */
+    @Nonnull
+    public java.awt.Color getColor()
```
duncte123, line 418: You should also add `getColorRaw`, keep it in line with the rest of JDA

VigusWG:
ps sorry if it didnt do everything right, this is my first PR ever :)

duncte123 (changes requested):

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
VigusWG: thanks

Chew: We already import java.awt.Color so we don't need to use FQDN

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -521,6 +590,18 @@ public EmbedBuilder setImage(@Nullable String url)
         return this;
     }
 
+    /**
+     * The {@link net.dv8tion.jda.api.entities.MessageEmbed.AuthorInfo AuthorInfo} used to
+     * build the footer for the embed.
```
Chew, line 595:
```suggestion
     * build the author for the embed.
```

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -386,6 +409,28 @@ public EmbedBuilder setTimestamp(@Nullable TemporalAccessor temporal)
...
+    /**
+     * The {@link Int int} of the color as a raw integer
+     *
+     * @return Raw color the builder has been set with.
+     */
+    @Nonnull
+    public int getColorRaw()
+    {
+        return color;
+    }
```
Xirado, line 432: You shouldn't annotate primitives with nullability annotations

VigusWG: Alright will fix, thanks.

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
VigusWG: All fixed, thanks

Xirado:
You should also check what those getters actually return. I don't think they should be `@Nonnull`

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -425,6 +469,18 @@ public EmbedBuilder setColor(int color)
         return this;
     }
 
+    /**
+     * The {@link net.dv8tion.jda.api.entities.MessageEmbed.Thumbnail Thumbnail} used to
+     * build the thumbnail for the embed.
+     *
+     * @return Thumbnail with current thumbnail context.
+     */
```
Tais993, line 477:
```suggestion
    /**
     * The information about the {@link net.dv8tion.jda.api.entities.MessageEmbed.Thumbnail Thumbnail} image to be displayed with the embed.
     * <br>If a {@link net.dv8tion.jda.api.entities.MessageEmbed.Thumbnail Thumbnail} hasn't been set on this builder, this returns null.
     *
     * @return Possibly-null {@link net.dv8tion.jda.api.entities.MessageEmbed.Thumbnail Thumbnail} instance
     *         containing general information on the displayable thumbnail.
     */
```

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -472,6 +528,18 @@ public EmbedBuilder setThumbnail(@Nullable String url)
         return this;
     }
 
+    /**
+     * The {@link net.dv8tion.jda.api.entities.MessageEmbed.ImageInfo ImageInfo} used to
+     * build the image for the embed.
+     *
+     * @return ImageInfo with current image context.
```
Tais993, line 535:
```suggestion
     * The information about the image in builder, or null when not set
     *
     * @return Possibly-null {@link net.dv8tion.jda.api.entities.MessageEmbed.ImageInfo ImageInfo}
     *         containing image information.
```

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -521,6 +589,18 @@ public EmbedBuilder setImage(@Nullable String url)
         return this;
     }
 
+    /**
+     * The {@link net.dv8tion.jda.api.entities.MessageEmbed.AuthorInfo AuthorInfo} used to
+     * build the author for the embed.
+     *
+     * @return AuthorInfo with current author context
```
Tais993, line 596:
```suggestion
     * The information on the author of builder.
     * <br>This is typically used to represent the account on the providing site.
     *
     * @return Possibly-null {@link net.dv8tion.jda.api.entities.MessageEmbed.AuthorInfo AuthorInfo}
     *         containing author information.
```

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -633,6 +713,18 @@ public EmbedBuilder setAuthor(@Nullable String name, @Nullable String url, @Null
         return this;
     }
 
+    /**
+     * The {@link net.dv8tion.jda.api.entities.MessageEmbed.Footer Footer} used to
+     * build the footer for the embed.
+     *
+     * @return Footer with current footer context.
```
Tais993, line 720:
```suggestion
     * The footer (bottom) of the embedded content.
     * <br>This is typically used for timestamps or site icons.
     *
     * @return Possibly-null {@link net.dv8tion.jda.api.entities.MessageEmbed.Footer Footer}
     *         containing the embed footer content.
```

Tais993 (commented):
Missing getter for fields, description and the title URL, annotations should be changed to `@Nullable` and I'd personally keep the javadoc almost equal to the MessageEmbed documentation.

No need for differences, can be the same

VigusWG:
Thanks everyone for your critiques, I apologies for not having everything perfect the first time.

Andre601:
There are still some inconsistencies imo.

For example are there lines like `@return Nullable <something>` while other lines have `@return Possibly-null <something>`

I think you should stick with the second one as that is afaik more used in the docs and the first one makes more sense for `@param` descriptions of nullable ones.

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -366,6 +377,17 @@ public EmbedBuilder appendDescription(@Nonnull CharSequence description)
         return this;
     }
 
+    /**
+     * The Timestamp of the builder
+     *
+     * @return Nullable {@link java.time.OffsetDateTime OffsetDateTime} object representing the timestamp
+     */
+    @Nonnull
```
freya022, line 385: Should be Nullable

freya022 (changes requested):

VigusWG:
> 

These should now all been fixed.

freya022 (changes requested):

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -472,6 +531,18 @@ public EmbedBuilder setThumbnail(@Nullable String url)
         return this;
     }
 
+    /**
+     * The information about the image in builder, or null when not set
+     *
+     * @return Possibly-null {@link net.dv8tion.jda.api.entities.MessageEmbed.ImageInfo ImageInfo}
+     *         containing image information.
+     */
+    @Nonnull
```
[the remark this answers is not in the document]
VigusWG: idk how i missed that lol.
all fixed

MinnDevelopment:
> Currently there is no other way to get them.

You can use `build().getX()`, I don't think this is needed. Builders in JDA usually do not offer getters because the setters are asymmetric to the getters. For instance, when you do `setTitle(name, url)`, you cannot make a getter that returns both unless its wrapped into a specific type, which is not symmetric and kind of defeats the idea of setter/getter relationships.

VigusWG:
> > Currently there is no other way to get them.
> 
> You can use `build().getX()`, I don't think this is needed. Builders in JDA usually do not offer getters because the setters are asymmetric to the getters. For instance, when you do `setTitle(name, url)`, you cannot make a getter that returns both unless its wrapped into a specific type, which is not symmetric and kind of defeats the idea of setter/getter relationships.

Fair enough just sounds a little inefficient, but I understand the main point of it. Apologies for my java inexperience, I'll close.

PR #2085, opened 2022-04-10
Improve handling of Role#getPosition and canInteract

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This improves the user experience for RoleDeleteEvent and allows `Role#getPosition` to still be usable once it's detached from cache. This position value is naturally not accurate for long, but still good enough for informational purposes at the time of deletion. Like any other entity value, this drifts over time after detaching from cache.

Additionally, this optimizes `Role#canInteract(Role)` by not re-calculating positions for both roles each time. We can completely rely on `compareTo` as this is only a relative comparison, rather than absolute positions.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/RoleImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/RoleImpl.java
@@ -60,6 +60,7 @@ public class RoleImpl implements Role
     private long rawPermissions;
     private int color;
     private int rawPosition;
+    private int realPosition = Integer.MIN_VALUE; // this is used exclusively for delete events
     private RoleIcon icon;
 
     public RoleImpl(long id, Guild guild)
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/internal/entities/RoleImpl.java
```diff
@@ -60,6 +60,7 @@ public class RoleImpl implements Role
     private long rawPermissions;
     private int color;
     private int rawPosition;
+    private int realPosition = Integer.MIN_VALUE; // this is used exclusively for delete events
```
DV8FromTheWorld, line 63: Let's call this `frozenPosition` instead, or something similar

PR #2087, opened 2022-04-11
transform hardcoded jump url format into a constant

caneleex:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this PR transforms the hardcoded jump url format into a constant in the `Message` interface. this can be useful for users who need it for something else.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -117,6 +117,9 @@ import java.util.stream.Collectors;
  */
 public interface Message extends ISnowflake, Formattable
 {
+    /** Template for {@link #getJumpUrl()}.*/
+    String JUMP_URL = "https://discord.com/channels/%s/%s/%s";
+
     /**
      * The maximum sendable file size (8 MiB)
      *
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -117,6 +117,9 @@
  */
 public interface Message extends ISnowflake, Formattable
 {
+    /** Template for {@link #getJumpUrl()}.*/
+    String JUMP_URL = "https://discord.com/channels/%s/%s/%s";
```
DV8FromTheWorld, line 121: If this is to be used by developers outside of the method it was intended for them we need to document the replacement parameters

caneleex: if that were to be done, we'd have to change it for all url templates.

and also, how would you want the docs for it to look like?

DV8FromTheWorld: Yeah, you make a good point.
I'd be okay not including it in this PR as it is an expansion of scope.

However, something like this might be good:
```
[0] - {guildId} or "@me"
[1] - {channelId}
[2] - {messageId}
```

DV8FromTheWorld (approved):

PR #1955, opened 2021-12-18
Add ImageProxy & FileProxy

freya022:
[contributing]: https://github.com/DV8FromTheWorld/JDA/wiki/5%29-Contributing

## Pull Request Etiquette

- [X] I have checked the PRs for upcoming features/bug fixes.
- [X] I have read the [contributing guidelines][contributing].

### Changes

- [X] Internal code
- [X] Library interface (affecting end-user code) 
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds ImageProxy, AttachmentProxy and FileProxy to allow easier image and file downloading. This also adds convenience methods where URLs are found

PR #2089, opened 2022-04-14
Option to change size off your pfp

PhoenixImpulse:
Added option to change size of your avatar from avatar url.


freya022:
Unfortunately, this PR is already handled in #1955, and image sizes aren't really indicated by Discord so I'd not check them

MinnDevelopment:
Thank you, but this is already something that will be added in #1955 with `user.getAvatar().getUrl(size)`

You also did not handle some of the special cases like size 100.

PR #2104, opened 2022-04-22
Rework stickers

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This completely overhauls the current implementation of stickers.

It looks like you can only send stickers through normal POST messages routes, not webhooks (including interactions). You can also not edit stickers on messages: `50080: Cannot edit a message with stickers`.

Bots can only use stickers from the same guild, cross-guild is unsupported.

### TODO

- [x] Documentation
- [x] Checks
- [x] Send sticker message
- [x] Create Guild Sticker
- [x] Modify Guild Sticker
- [x] Delete Guild Sticker
- [x] List Nitro Sticker Packs
- [x] Get Sticker
- [x] Sticker Pack Object
- [x] Guild Sticker Cache
- [x] Guild Sticker Events
- [x] Owner is null on events

### Depends on

- [x] Addition of `FileUpload` for sticker creation #2120 
- [x] Manager synchronization decision (Whether to keep the manager cached in entities)

### Relevant Discord Feature Requests

- https://github.com/discord/discord-api-docs/discussions/3422
- https://github.com/discord/discord-api-docs/discussions/4441
- https://github.com/discord/discord-api-docs/discussions/4817


src/main/java/net/dv8tion/jda/internal/entities/sticker/StickerItemImpl.java
```diff
@@ -0,0 +1,88 @@
...
+    {
+        return name;
+    }
+
+    public String setName(String name)
+    {
+        String old = this.name;
+        this.name = name;
+        return old;
+    }
```
DV8FromTheWorld, line 62: This breaks our pattern of set + return `this`. Why?

src/main/java/net/dv8tion/jda/internal/entities/sticker/StickerPackImpl.java
```diff
@@ -0,0 +1,113 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.internal.entities.sticker;
+
+import net.dv8tion.jda.api.entities.sticker.StandardSticker;
+import net.dv8tion.jda.api.entities.sticker.StickerPack;
+import org.jetbrains.annotations.NotNull;
```
DV8FromTheWorld, line 21: Wrong annotation

src/main/java/net/dv8tion/jda/internal/entities/sticker/StickerSnowflakeImpl.java
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.internal.entities.sticker;
+
+import net.dv8tion.jda.api.entities.sticker.StickerSnowflake;
+
+public class StickerSnowflakeImpl implements StickerSnowflake
```
DV8FromTheWorld, line 21: Needs a `toString` impl

src/main/java/net/dv8tion/jda/internal/entities/sticker/StickerUnionImpl.java
```diff
@@ -0,0 +1,100 @@
...
+
+import net.dv8tion.jda.api.entities.sticker.GuildSticker;
+import net.dv8tion.jda.api.entities.sticker.RichSticker;
+import net.dv8tion.jda.api.entities.sticker.Sticker;
+import net.dv8tion.jda.api.entities.sticker.StickerUnion;
+
+import javax.annotation.Nonnull;
+import java.util.Set;
+
+public class StickerUnionImpl implements StickerUnion
```
DV8FromTheWorld, line 27: I expect this class is gonna die based on our convo

src/main/java/net/dv8tion/jda/internal/handle/GuildStickersUpdateHandler.java
```diff
@@ -0,0 +1,165 @@
...
+                if (sticker == null)
+                {
+                    sticker = (GuildStickerImpl) builder.createRichSticker(current);
+                    newStickers.add(sticker);
+                }
+                else
+                {
+                    // sticker is in our cache which is why we don't want to remove it in cleanup later
+                    oldStickers.remove(sticker);
+                    oldSticker = sticker.copy();
```
DV8FromTheWorld, line 92: What triggers this event? Any add/update/remove? So do we get then entire list again on any change?
I'm assuming only one sticker can reasonable be added/updated/removed at a single time, so if we _do_ get the entire list on any change it means any change will end up cloning all existing stickers even if they didn't change.

Seems like an expensive pattern to me honestly. What value does this provide to us compared to the "if-statement + update value + dispatch event" pattern we have for other entities?

src/main/java/net/dv8tion/jda/internal/requests/Route.java
```diff
@@ -143,6 +143,8 @@
         public static final Route GET_GUILD_EMBED =    new Route(GET,    "guilds/{guild_id}/embed");
         public static final Route MODIFY_GUILD_EMBED = new Route(PATCH,  "guilds/{guild_id}/embed");
         public static final Route GET_GUILD_EMOTES =   new Route(GET,    "guilds/{guild_id}/emojis");
+        public static final Route GET_GUILD_STICKERS = new Route(GET,    "guilds/{guild_id}/stickers");
+        public static final Route GET_GUILD_STICKER =  new Route(GET,    "guilds/{guild_id}/stickers/{sticker_id}");
```
DV8FromTheWorld, line 147: These should be in `Route.Stickers` if we're breaking the others out.

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageActionImpl.java
```diff
@@ -398,6 +408,41 @@ public MessageActionImpl setActionRows(@Nonnull ActionRow... rows)
...
+            this.stickers = new ArrayList<>();
+            return this;
+        }
+
+        if (!(channel instanceof GuildChannel))
+            throw new IllegalStateException("Cannot send stickers in direct messages!");
+        GuildChannel guildChannel = (GuildChannel) channel;
+
+        Checks.noneNull(stickers, "Stickers");
+        Checks.check(stickers.size() <= 3, "Cannot send more than 3 stickers in a message!"); // TODO: Make this a constant
```
DV8FromTheWorld, line 428: Don't forget the TODO. Code and docs could make use of it.

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageActionImpl.java
```diff
@@ -398,6 +408,41 @@ public MessageActionImpl setActionRows(@Nonnull ActionRow... rows)
...
+    @Override
+    public MessageAction setStickers(@Nullable Collection<? extends StickerSnowflake> stickers)
+    {
+        if (isEdit())
+            throw new IllegalStateException("Cannot edit stickers on messages!");
+        if (stickers == null || stickers.isEmpty())
+        {
+            this.stickers = new ArrayList<>();
+            return this;
+        }
```
DV8FromTheWorld, line 421: Not sure if it matters, but setting this to an empty array list instead of `null` will cause `sticker_ids: []` to be sent.

DV8FromTheWorld (changes requested):
Nice rework. Lots of great stuff here.

src/main/java/net/dv8tion/jda/api/MessageBuilder.java
```diff
@@ -221,6 +234,57 @@ public MessageBuilder setActionRows(@Nullable ActionRow... rows)
...
+    {
+        this.stickers.clear();
+        if (stickers == null || stickers.isEmpty())
+            return this;
+        Checks.noneNull(stickers, "Stickers");
+
+        stickers.stream()
+                .map(StickerSnowflake::getId)
+                .map(StickerSnowflake::fromId)
+                .forEach(this.stickers::add);
```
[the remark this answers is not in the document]
MinnDevelopment: To minimize memory usage of the MessageBuilder, since it can be re-used. It also ensures there aren't any strange custom implementations present.

src/main/java/net/dv8tion/jda/api/entities/sticker/StandardSticker.java
```diff
@@ -0,0 +1,52 @@
...
+    {
+        return Long.toUnsignedString(getPackIdLong());
+    }
+
+    /**
+     * The sticker's sort order within its pack
+     *
+     * @return The sort order value
+     */
+    int getSortValue();
```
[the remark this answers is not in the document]
MinnDevelopment: I have no idea, the docs don't explain this at all. I'm guessing its just the index in the pack

src/main/java/net/dv8tion/jda/api/entities/sticker/StickerItem.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package net.dv8tion.jda.api.entities.sticker;
+
+/**
+ * Stickers used primarily for message objects.
+ *
+ * <p>These types contain the minimal required information to display the sticker in the discord client.
+ */
+public interface StickerItem extends Sticker
```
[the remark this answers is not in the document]
MinnDevelopment: Consistency with discord naming: https://discord.com/developers/docs/resources/sticker#sticker-item-object

src/main/java/net/dv8tion/jda/api/entities/sticker/StickerPack.java
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * A pack of {@link StandardSticker StandardStickers} used for nitro.
+ */
+public interface StickerPack extends ISnowflake
+{
+    /**
+     * Format string used for {@link #getBannerUrl()}.
+     * <br>The parameters of the format string are the {@link #getBannerId()} and the file extension (png).
+     */
+    String BANNER_URL = "https://cdn.discordapp.com/app-assets/710982414301790216/store/%s.%s";
```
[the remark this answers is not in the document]
MinnDevelopment: Because thats how discord designed it

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1761,25 +1797,72 @@ public static MessageEmbed createMessageEmbed(String url, String title, String d
...
-        final String name = content.getString("name");
-        final String description = content.getString("description", "");
-        final long packId = content.getLong("pack_id", content.getLong("guild_id", 0L));
-        final MessageSticker.StickerFormat format = MessageSticker.StickerFormat.fromId(content.getInt("format_type"));
-        final Set<String> tags;
-        if (content.isNull("tags"))
-        {
-            tags = Collections.emptySet();
+        long id = content.getLong("id");
+        String name = content.getString("name");
```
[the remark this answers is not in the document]
MinnDevelopment: Yes.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1761,25 +1797,72 @@ public static MessageEmbed createMessageEmbed(String url, String title, String d
...
+            boolean available = content.getBoolean("available");
+            long guildId = content.getUnsignedLong("guild_id", 0L);
+            User owner = content.isNull("user") ? null : createUser(content.getObject("user"));
+            return new GuildStickerImpl(id, format, name, tags, description, available, guildId, api, owner);
+        case STANDARD:
+            long packId = content.getUnsignedLong("pack_id", 0L);
+            int sortValue = content.getInt("sort_value", -1);
+            return new StandardStickerImpl(id, format, name, tags, description, packId, sortValue);
+        default:
+            throw new IllegalArgumentException("Unknown sticker type. Type: " + type  +" JSON: " + content);
```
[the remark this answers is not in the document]
MinnDevelopment: That should be prevented here

https://github.com/DV8FromTheWorld/JDA/blob/73df1e5b700331818cc15c5a549d9f5515fb0087/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java#L188

src/main/java/net/dv8tion/jda/internal/entities/sticker/GuildStickerImpl.java
```diff
@@ -0,0 +1,173 @@
...
+        this.guild = jda.getGuildById(guildId);
+        this.owner = owner;
+    }
+
+    @Nonnull
+    @Override
+    public Type getType()
+    {
+        return Type.GUILD;
+    }
```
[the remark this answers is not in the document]
MinnDevelopment: Why?

src/main/java/net/dv8tion/jda/internal/handle/GuildStickersUpdateHandler.java
MinnDevelopment: This implementation is entirely copied from the handling of emoji updates. We copy the list to compare which ones were added/removed and firing respective events. I'm unsure why discord designed it this way.

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageActionImpl.java
MinnDevelopment: Yes, that is intended. This would be the logical way to handle it if PATCH was supported for stickers. In the future, if they allow edit we only need to remove the check and it works. For the POST endpoint, empty, null, and omitted are treated the same way by discord.

src/main/java/net/dv8tion/jda/api/managers/GuildStickerManager.java
```diff
@@ -0,0 +1,185 @@
...
+     */
+    @Nonnull
+    Guild getGuild();
+
+    /**
+     * Sets the <b><u>name</u></b> of the sticker.
+     *
+     * <p>A sticker name <b>must</b> be between 2-30 characters long!
+     *
+     * <p><b>Example</b>: {@code catDance} or {@code dogWave}
```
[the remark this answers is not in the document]
MinnDevelopment: Yes they can have spaces but it breaks the suggestions by name:

(the ratJam is a different sticker, from another guild)
image

image

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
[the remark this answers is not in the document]
DV8FromTheWorld: Yup, that tracks.

src/main/java/net/dv8tion/jda/internal/entities/sticker/GuildStickerImpl.java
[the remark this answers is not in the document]
DV8FromTheWorld: Why not? Isn't that how we've done it for things like components?

src/main/java/net/dv8tion/jda/internal/entities/sticker/RichStickerImpl.java
```diff
@@ -90,28 +88,6 @@ public String setDescription(String description)
...
-        if (!(obj instanceof RichStickerImpl))
-            return false;
-        RichStickerImpl other = (RichStickerImpl) obj;
-        return id == other.id
-                && format == other.format
-                && getType() == other.getType()
-                && Objects.equals(name, other.name)
-                && Objects.equals(description, other.description)
-                && Helpers.deepEqualsUnordered(tags, other.tags);
+        return "RichSticker[" + getType() + "]:" + name + '(' + getId() + ')';
```
DV8FromTheWorld, line 89: Is it actually possible to get this considering the implementers override it?

src/main/java/net/dv8tion/jda/internal/requests/Route.java
```diff
@@ -181,6 +179,9 @@
 
     public static class Stickers
     {
+        public static final Route GET_GUILD_STICKERS = new Route(GET,    "guilds/{guild_id}/stickers");
+        public static final Route GET_GUILD_STICKER =  new Route(GET,    "guilds/{guild_id}/stickers/{sticker_id}");
+
         public static final Route MODIFY_STICKER = new Route(PATCH,  "guilds/{guild_id}/stickers/{sticker_id}");
         public static final Route DELETE_STICKER = new Route(DELETE, "guilds/{guild_id}/stickers/{sticker_id}");
         public static final Route CREATE_STICKER = new Route(POST,   "guilds/{guild_id}/stickers");
```
DV8FromTheWorld, line 187: For consistency, lets make these X_GUILD_STICKER

src/main/java/net/dv8tion/jda/api/managers/GuildStickerManager.java
```diff
@@ -0,0 +1,210 @@
...
+     *
+     * @see    #getGuildId()
+     */
+    @Nullable
+    Guild getGuild();
+
+    long getGuildIdLong();
+
+    @Nonnull
+    default String getGuildId()
```
DV8FromTheWorld, line 113: Missing Docs

src/main/java/net/dv8tion/jda/api/managers/GuildStickerManager.java
```diff
@@ -0,0 +1,210 @@
...
+     * @return GuildStickerManager for chaining convenience
+     */
+    @Nonnull
+    @Override
+    GuildStickerManager reset(long... fields);
+
+    /**
+     * The {@link Guild} this Manager's {@link GuildSticker} is in.
+     *
+     * <p>This is null if {@link GuildSticker#getManager()} is used no a sticker with an uncached guild.
```
DV8FromTheWorld, line 101:
```suggestion
     * <p>This is null if {@link GuildSticker#getManager()} is used on a sticker with an uncached guild.
```

src/main/java/net/dv8tion/jda/api/managers/GuildStickerManager.java
```diff
@@ -0,0 +1,210 @@
...
+     *         The new tags for the sticker (up to 200 characters)
+     *
+     * @throws IllegalArgumentException
+     *         Thrown if
+     *         <ul>
+     *             <li>{@code tags} is {@code null}</li>
+     *             <li>{@code tags} is empty</li>
+     *             <li>{@code tags} contains {@code null} or empty strings</li>
+     *             <li>The concatenated tags are more than 200 characters long (including commas between tags)</li>
+     *         </ul>
```
DV8FromTheWorld, line 173: Lets put the `If X` on each line as that is the pattern we're starting to see appear after we've discussed in other PRs. It reads better
```suggestion
     *         <ul>
     *             <li>If {@code tags} is {@code null}</li>
     *             <li>If {@code tags} is empty</li>
     *             <li>If {@code tags} contains {@code null} or empty strings</li>
     *             <li>If the concatenated tags are more than 200 characters long (including commas between tags)</li>
     *         </ul>
```

Applies to multiple places in the code.

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -4433,6 +4590,65 @@ default RoleAction createCopyOfRole(@Nonnull Role role)
...
+     *             <li>The description is more than 100 characters long or exactly 1 character long</li>
+     *             <li>The asset file is null or of an invalid format (must be PNG or LOTTIE)</li>
+     *             <li>Anything is {@code null}</li>
+     *         </ul>
+     *
+     * @return {@link AuditableRestAction} - Type: {@link GuildSticker}
+     */
+    @Nonnull
+    @CheckReturnValue
+    AuditableRestAction<GuildSticker> createSticker(@Nonnull String name, @Nonnull String description, @Nonnull FileUpload file, @Nonnull String tag, @Nonnull String... tags);
```
DV8FromTheWorld, line 4630: We need a Collection<String> option for setting tags.
As a note, do you _have_ to add a tag to a sticker?

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1781,6 +1784,62 @@ public AuditableRestAction<Emote> createEmote(@Nonnull String name, @Nonnull Ico
...
+        // Extract file extension and map to media type
+        int index = file.getName().lastIndexOf('.');
+        if (index < 0)
+            throw new IllegalArgumentException("Missing file extension for image. Must be PNG or JSON.");
+
+        String extension = file.getName().substring(index + 1).toLowerCase(Locale.ROOT);
+        MediaType mediaType;
+        switch (extension)
+        {
+            case "apng":
```
DV8FromTheWorld, line 1822: I thought that APNG actually used `.png`? is `.apng` an actual extension?

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1781,6 +1784,62 @@ public AuditableRestAction<Emote> createEmote(@Nonnull String name, @Nonnull Ico
...
+        // Add sticker meta data as form parts (because payload_json is broken)
+        MultipartBody.Builder builder = new MultipartBody.Builder().setType(MultipartBody.FORM);
+        builder.addFormDataPart("name", name);
+        builder.addFormDataPart("description", description);
+        builder.addFormDataPart("tags", csv);
+
+        // Extract file extension and map to media type
+        int index = file.getName().lastIndexOf('.');
+        if (index < 0)
+            throw new IllegalArgumentException("Missing file extension for image. Must be PNG or JSON.");
```
DV8FromTheWorld, line 1816: So if someone uses the `byte[]` or `InputStream` ways of building `FileUpload` it is expected that they'll need to provide a `name` with an extension in it?

I guess that makes sense as that's what discord would require to render a preview of the file in the client.

DV8FromTheWorld (commented):
Love the CachedRestAction

src/main/java/net/dv8tion/jda/internal/handle/GuildStickersUpdateHandler.java
DV8FromTheWorld: To be clear, making a shallow clone of the `valueCollection` makes sense for before/after comparison of add/removals.
I was talking about the `.copy()` to do attribute comparisons. Seems heavy handed compared to iteration and attribute checking, no?

src/main/java/net/dv8tion/jda/api/entities/sticker/StandardSticker.java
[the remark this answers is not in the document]
DV8FromTheWorld: Gonna double check with the DDevs. I expect the same.
I don't think it is overly impactful to this PR though tbh.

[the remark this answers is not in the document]
DV8FromTheWorld: image
So not actually an index, so we should definitely leave it named as `sortValue` instead of trying to convey it as an index 👍

src/main/java/net/dv8tion/jda/api/entities/Guild.java
MinnDevelopment: Yes, tags are required:

```
net.dv8tion.jda.api.exceptions.ErrorResponseException: 50035: Invalid Form Body
tags
	- BASE_TYPE_REQUIRED: This field is required
```

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
MinnDevelopment: Yes, it is a valid extension, but not very common. Discord doesn't really care about the file extension, we only need it to parse the media type.

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
MinnDevelopment: The extension is only required because the sticker endpoint expects a valid media type, and we don't want to try and parse it from the data.

src/main/java/net/dv8tion/jda/internal/entities/sticker/RichStickerImpl.java
MinnDevelopment: If future sticker types are added and we forget to override toString, this could be used. I think its fine to just keep it.

src/main/java/net/dv8tion/jda/internal/handle/GuildStickersUpdateHandler.java
MinnDevelopment: It allows us to handle remove/add/update in a simple method by just checking if oldSticker is null or not. Since sticker updates are quite rare and stickers themselves are very tiny, I think thats a reasonable thing to do.

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
DV8FromTheWorld: Maybe we should change the error to be something more descriptive of how to fix it then:
`FileUpload.getName() is missing file extension for image. Must be PNG or JSON.` or something similar?

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1786,24 +1786,21 @@ public AuditableRestAction<Emote> createEmote(@Nonnull String name, @Nonnull Ico
...
     @Override
-    public AuditableRestAction<GuildSticker> createSticker(@Nonnull String name, @Nonnull String description, @Nonnull FileUpload file, @Nonnull String tag, @Nonnull String... tags)
+    public AuditableRestAction<GuildSticker> createSticker(@Nonnull String name, @Nonnull String description, @Nonnull FileUpload file, @Nonnull Collection<String> tags)
     {
         checkPermission(Permission.MANAGE_EMOTES_AND_STICKERS);
         Checks.inRange(name, 2, 30, "Name");
         Checks.notNull(file, "File");
-        Checks.notEmpty(tag, "Tags");
         Checks.notNull(description, "Description");
+//        Checks.notEmpty(tags, "Tags");
```
DV8FromTheWorld: Why the loss of checking for empty in the collection?

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1715,6 +1784,67 @@ public AuditableRestAction<Emote> createEmote(@Nonnull String name, @Nonnull Ico
...
+    public AuditableRestAction<GuildSticker> createSticker(@Nonnull String name, @Nonnull String description, @Nonnull FileUpload file, @Nonnull Collection<String> tags)
+    {
+        checkPermission(Permission.MANAGE_EMOTES_AND_STICKERS);
+        Checks.inRange(name, 2, 30, "Name");
+        Checks.notNull(file, "File");
+        Checks.notNull(description, "Description");
+        if (!description.isEmpty())
+            Checks.inRange(description, 2, 100, "Description");
+        for (String t : tags)
+            Checks.notEmpty(t, "Tags");
```
DV8FromTheWorld, line 1798: I think we lost the `notEmpty` check on the Collection?

DV8FromTheWorld (approved):

PR #2115, opened 2022-04-29
[Bugfix] retrieving public archived threads

Sanduhr32:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation

Closes Issue: NaN

## Description

Follow-up requests of ThreadChannelPaginationAction had duplicated `before` request parameters,
while discord only read the first (invalid) one provided by PaginationAction route finalization.

```
net.dv8tion.jda.api.exceptions.ErrorResponseException: 50035: Invalid Form Body
    - DATE_TIME_TYPE_PARSE: Could not parse 968844240228585543. Should be ISO8601.
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/PaginationActionImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/PaginationActionImpl.java
@@ -34,8 +34,8 @@ import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 
 public abstract class PaginationActionImpl<T, M extends PaginationAction<T, M>>
-    extends RestActionImpl<List<T>>
-    implements PaginationAction<T, M>
+        extends RestActionImpl<List<T>>
+        implements PaginationAction<T, M>
 {
     protected final List<T> cached = new CopyOnWriteArrayList<>();
     protected final int maxLimit;
@@ -245,7 +245,8 @@ public abstract class PaginationActionImpl<T, M extends PaginationAction<T, M>>
     @Override
     public CompletableFuture<List<T>> takeAsync(int amount)
     {
-        return takeAsync0(amount, (task, list) -> forEachAsync(val -> {
+        return takeAsync0(amount, (task, list) -> forEachAsync(val ->
+        {
             list.add(val);
             return list.size() < amount;
         }, task::completeExceptionally));
@@ -255,7 +256,8 @@ public abstract class PaginationActionImpl<T, M extends PaginationAction<T, M>>
     @Override
     public CompletableFuture<List<T>> takeRemainingAsync(int amount)
     {
-        return takeAsync0(amount, (task, list) -> forEachRemainingAsync(val -> {
+        return takeAsync0(amount, (task, list) -> forEachRemainingAsync(val ->
+        {
             list.add(val);
             return list.size() < amount;
         }, task::completeExceptionally));
@@ -347,8 +349,21 @@ public abstract class PaginationActionImpl<T, M extends PaginationAction<T, M>>
         }
     }
 
+    // Introduced for paginating archived threads, because two endpoints require a different request parameter value format.
+    // May become more useful if discord introduces more pagination endpoints not using ids.
+    // Check ThreadChannelPaginationActionImpl
+    // Background of #formatKeyValue:
+    //     Archived thread channel pagination (example: TextChannel#retrieveArchivedPublicThreadChannels) would throw an exception,
+    //     where Discord complained about receiving a snowflake instead of an ISO8601 date.
+    //     The snowflake originated from this class (creating initial & subsequent requests),
+    //     while the correct value was set in ThreadChannelPaginationActionImpl for the initial request
+    //     and appended as a second value for subsequent requests.
+    //     However, withQueryParams is a simple string append and Discord only reads the first parameter.
+    //     If you debugged, you would see some duplicated fields on the compiled route.
+    //     The fix here is to let the implementor supply the "last" string value, which could be anything,
+    //     while the default implementation still would provide snowflakes
     @Nonnull
-    protected String getLastKeyValue(long last)
+    protected String formatKeyValue(long last)
     {
         return Long.toUnsignedString(last);
     }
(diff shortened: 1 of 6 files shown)


Sanduhr32:
I've not tested this yet, however i'm working on a patch to the code i submitted

Sanduhr32:
Thanks @freya022 for figuring the issue & and submitting a working fix.

freya022:
Note that this fix has been only tested against [the original bug report](https://discord.com/channels/125227483518861312/869965768245260358/969669618195124224)

src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/PaginationActionImpl.java
```diff
@@ -347,6 +347,12 @@ public void forEachRemaining(@Nonnull final Procedure<? super T> action)
         }
     }
 
+    @Nonnull
+    protected String getLastKeyValue(long last)
```
Sanduhr32, line 351: I feel like `formatKey` would be a better name for this method, because the initial forwards order might break, if we just pass a simple `"0"` instead of a proper format for given pagination endpoints relying on different formats such as ISO8601.

Sanduhr32:
i'm currently having weird issues paginating over 100 threads, however accumulating 100 elements with a custom limit requiring several requests works fine.

src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/ThreadChannelPaginationActionImpl.java
```diff
@@ -46,23 +49,27 @@ public EnumSet<PaginationOrder> getSupportedOrders()
...
     {
-        Route.CompiledRoute route = super.finalizeRoute();
-
-        final String limit = String.valueOf(this.limit.get());
-        final long last = this.lastKey;
+        if (useID)
+            return Long.toUnsignedString(lastId);
 
-        route = route.withQueryParams("limit", limit);
+        if (order == PaginationOrder.FORWARD && lastId == 0)
```
MinnDevelopment, line 60: This checks if the order is FORWARD, but the only supported order according to `getSupportedOrder` is `BACKWARDS` and infact if you manually change it to `FORWARD` the code doesn't work because it keeps fetching the same channels.

MinnDevelopment (approved):

src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/PaginationActionImpl.java
```diff
@@ -358,9 +379,9 @@ protected Route.CompiledRoute finalizeRoute()
         route = route.withQueryParams("limit", limit);
 
         if (last != 0)
-            route = route.withQueryParams(order.getKey(), Long.toUnsignedString(last));
+            route = route.withQueryParams(order.getKey(), getPaginationLastEvaluatedKey(last, this.getLast()));
         else if (order == PaginationOrder.FORWARD)
-            route = route.withQueryParams("after", "0");
+            route = route.withQueryParams("after", getPaginationLastEvaluatedKey(0, this.last));
```
DV8FromTheWorld, line 384: Why is one using `getLast` and the other is using `this.last`?

Also, this code is kinad confusing given that there is a localVariable called `last` which is a `long` and a class variable of `last` that is of type `T`.

If the point of making the local variable is for it to be `final` (?) then lets at least call the local variable `lastKey` or `localLastKey` to help break the confusion.

src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/ThreadChannelPaginationActionImpl.java
```diff
@@ -46,23 +50,27 @@ public EnumSet<PaginationOrder> getSupportedOrders()
...
+            // first second of 2015 aka discords epoch, hard coding something older makes no sense to me
+            return "2015-01-01T00:00:00.000";
+        }
 
-        if (last == 0)
-            return route;
+        // this should be redundant, due to calling this with PaginationAction#getLast() as last param,
+        // but let's have this here.
+        if (last == null)
+            throw new NoSuchElementException("No entities have been retrieved yet.");
```
DV8FromTheWorld, line 70: This could be replaced with the current ISO8601 timestamp instead of throwing an exception

src/main/java/net/dv8tion/jda/internal/utils/Helpers.java
```diff
@@ -225,6 +231,18 @@ public static <E extends Enum<E>> EnumSet<E> copyEnumSet(Class<E> clazz, Collect
...
     }
 
+    public static TLongObjectMap<DataObject> convertToMap(ToLongFunction<DataObject> getId, DataArray array)
+    {
+        TLongObjectMap<DataObject> map = new TLongObjectHashMap<>();
+        for (int i = 0; i < array.length(); i++)
+        {
+            DataObject obj = array.getObject(i);
+            long userId = getId.applyAsLong(obj);
+            map.put(userId, obj);
```
DV8FromTheWorld, line 241:
```suggestion
            DataObject obj = array.getObject(i);
            long objId = getId.applyAsLong(obj);
            map.put(objId, obj);
```

DV8FromTheWorld (commented):
Lookin' good. just a few things i noticed

src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/ThreadChannelPaginationActionImpl.java
```diff
@@ -9,20 +9,27 @@
...
+import net.dv8tion.jda.internal.utils.Helpers;
 
 import javax.annotation.Nonnull;
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.NoSuchElementException;
```
DV8FromTheWorld, line 25: Kill

DV8FromTheWorld (approved):
LGTM

src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/ThreadChannelPaginationActionImpl.java
Sanduhr32:
```suggestion
```

PR #2127, opened 2022-05-09
Remove checks for archive durations

DerBanko:
Discord removed the unlockable duration of three / seven days of auto archiving.
image


MinnDevelopment:
Please also update the documentation [here](https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/api/managers/channel/concrete/ThreadChannelManager.java#L76-L77)

MinnDevelopment:
You should also remove them in [ThreadChannelAction](https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/internal/requests/restaction/ThreadChannelActionImpl.java#L120-L126)

DerBanko:
Solved

Issue #2132, opened 2022-05-16
When one embed field value is set null it results in an errror


RealYusufIsmail:
### General Troubleshooting

- [X] I have checked for similar issues on the Issue-tracker.
- [X] I have updated to the [latest JDA version](https://ci.dv8tion.net/job/JDA/lastSuccessfulBuild/)
- [X] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

### Expected Behaviour

Not to give an error as both values are marked as Nullable.

This can be shown here
<img width="270" alt="Screenshot 2022-05-16 at 18 17 58" src="https://user-images.githubusercontent.com/67903886/168647856-a648e8c6-9f12-42ca-bbe6-6cb59f96efcb.png">

### Code Example for Reproduction Steps

```java
EmbedBuilder builder = new EmbedBuilder()
                       .addField("one", null)
                       .build();
```

### Code for JDABuilder or DefaultShardManagerBuilder used

```java
Someone else code
```

### Exception or Error

```yesyes
java.lang.IlligalArgumentException: Both name and value must be set!
```

RealYusufIsmail:
If this error is valid can I make a pr on it.

Tais993:
Both name & value have to be set, or neither

There's no name OR value possible, pretty sure this is intentional

RealYusufIsmail:
Then why are they Nullable

RealYusufIsmail:
If you want have Nullable just have addfield() and addfield(NotNull name, Notnull value)

RealYusufIsmail:
so this problem does not occour.

Tais993:
Read what I said.

Either **BOTH** the name AND value are set

OR

The name and value are NOT set

You CANNOT have the name OR value

MinnDevelopment:
Since this confusion has come up multiple times now, I think we can just make them non-null only. The case of having both null doesn't really do anything in the first place.

RealYusufIsmail:
ok, can i be assigned for this.

PR #2133, opened 2022-05-16
Changed addField String name and String Value to Nonnull

RealYusufIsmail:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #2132

## Description

As stated in issue #2132 the Nullable has been changed to Nonnull to avoid confusion.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -740,7 +740,7 @@ public class EmbedBuilder
      *
      * @throws java.lang.IllegalArgumentException
      *         <ul>
-     *             <li>If only {@code name} or {@code value} is set. Both must be set.</li>
+     *             <li>If {@code name} or {@code value} are not set. They must be be both set.</li>
      *             <li>If the character limit for {@code name}, defined by {@link net.dv8tion.jda.api.entities.MessageEmbed#TITLE_MAX_LENGTH} as {@value net.dv8tion.jda.api.entities.MessageEmbed#TITLE_MAX_LENGTH},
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -750,7 +750,7 @@ public EmbedBuilder addField(@Nullable MessageEmbed.Field field)
      * @return the builder after the field has been added
      */
     @Nonnull
-    public EmbedBuilder addField(@Nullable String name, @Nullable String value, boolean inline)
+    public EmbedBuilder addField(@Nonnull String name, @Nonnull String value, boolean inline)
```
MinnDevelopment, line 753: You need to actually add null checks for these too. And please also update the exception docs

RealYusufIsmail: Will do so.

src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -740,7 +740,7 @@ public EmbedBuilder addField(@Nullable MessageEmbed.Field field)
      *
      * @throws java.lang.IllegalArgumentException
      *         <ul>
-     *             <li>If only {@code name} or {@code value} is set. Both must be set.</li>
+     *             <li>If {@code name} or {@code value} are not set. They must be be both set.</li>
```
MinnDevelopment, line 743:
```suggestion
     *             <li>If {@code null} is provided</li>
```

PR #2138, opened 2022-05-22
Introduce Channel Unions

DV8FromTheWorld:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description
The goal of this PR is to introduce Unions to JDA for channel types while also reworking the type hierarchy slightly to make way for ForumChannels and Messages in VoiceChannels (aka TextInVoice (TiV))

This PR changes the return types of a _lot_ of method (basically all of them) that are the type `getX : XChannel` (e.g: `Message#getChannel() : MessageChannel`) to instead return union types (`MessageChannelUnion` in this case). **Also, before you panic, anywhere that changed from XChannel -> XChannelUnion is completely compatible with 0 changes to existing bot/developer code because unions _are extensions of the interface their representing_**.

Additionally, a variety of channel getters have been changed to return more limited channel classes so that they can properly represent the channel types returned. For example, the `Webhook#getChannel` used to return `BaseGuildMessageChannel` but now instead returns `IWebhookContainerUnion`. Read on for more info!

## Unions
In JDA, a Union will represent a type that has getters on it to provide a signal to the developer of what types are available in this type. Technically this information is _typically_ available via documentation, but we've found that this is also a great way to introduce additional discoverability.

These getters are in the format `asX` like `asTextChannel() : TextChannel`. The are effectively just casts, but they provide a method of introducing the available interfaces and types that are represented by the union. They also provide a way to introduce developers to "in-between" or "bridge" interfaces that may be useful like `StandardGuildChannel` or direct them to known useful casts like to `IThreadContainer`.

For example, the `GuildMessageChannelUnion`. The `GuildMessageChannelUnion` interface _extends_ the `GuildMessageChannel` interface and differs _only_ by providing the getters that can implement `GuildMessageChannel`:
- `asTextChannel() : TextChannel`
- `asNewsChannel() : NewsChannel`
- `asThreadChannel() : ThreadChannel`
- `asStandardGuildChannel() : StandardGuildChannel`
- `asStandardGuildMessageChannel() : StandardGuildMessageChannel`

### Clean Interfaces
Another thing to note is that Unions are not directly included in the normal type hierarchy, so their `asX` getters are _not_ present on the end concrete interface. For example, when looking at a `TextChannel tchan = jda.getTextChannelById("...")` the `tchan` variable will  **not** have the `asNewsChannel` (nor any other `asXChannel` getter) on it even though `TextChannel` implements `GuildMessageChannel`. This is because the type hierarchy for `TextChannel` _does not_ contain the `GuildMessageChannelUnion` interface, only the `GuildMessageChannel` interface. This means devs aren't going to have weird `asX` getters when they aren't needed, and this is an expectation we place on all unions going forward.

### Unions are NOT wrapper objects
(This is mostly for the curious)
While the concrete type _interfaces_ like `TextChannel` do not inherit from the Union classes, their `XImpl` (like `TextChannelImpl`) _do_ implement them.

The advantage of this is 2 fold. Like mentioned before, the interfaces presented to developers are clean of these `asX` getters when they aren't pertinent anymore (after you've left the union type for a concrete type) however since the Impl implements it it means that when a method calls for a union (like `GuildChannelMessageUnion`) and we have the non union type (like `GuildChannelMessage`) we can _confidently_ just cast from the normal type to the union type because we know the underlying Impl that represents that channel supports the union.

This is nice because it means we don't need special union wrappers to wrap objects to be returned by `asX`. The object itself _is_ the union.

#### For the _really_ curious
Given the above, here is an example
`TextChannel < StandardGuildMessageChannel < StandardGuildChannel + GuildMessageChannel + ...`
`TextChannelImpl < StandardGuildMessageChannelImpl < StandardGuildChannelImpl + GuildChannelMessageUnion + ...`

Note how the impl uses the union when building up `StandardGuildMessageChannelImpl` but the interface uses `GuildMessageChannel` when building up `StandardGuildMessageChannel`.

(_technically the impl uses `GuildMessageChannelMixin` which includes the union, but I digress_) 

## Removed Types
- **`BaseGuildMessageChannel`**
  The BGMC served as a great accumulator of all of the "attribute interfaces" (the IXChannel / IXContainer interfaces, ex: `IThreadContainer`) and acted as a _base_ between `TextChannel` and `NewsChannel`. While this is really useful it cannot be reused for other messaging situations like `TiV` (because TiV doesn't support threads!) and we were returning BGMC from some methods that no longer can fully model the functionalities promised by BGMC like `Guild#getDefaultChannel` as `ForumChannels` can be a default channel in the future and they don't support `MessageChannel`! 
  
  As such, the functionality present in this type technically lives on in a form in `StandardGuildMessageChannel`, but it would be wrong to say it is simply renamed. A large amount of the "base" functionality present in this interface has now been spread out across `StandardGuildChannel` _and_ `StandardGuildMessageChannel`. Additionally, most (if not all) of the `getX : BGMC` getters have been changed to return a super type with a more limited set of functionalities compared to the previous usages of BGMC due needing to support future Discord changes. Going back to our `Guild#getDefaultChannel` it now returns `DefaultGuildChannelUnion` to better represent the functionalities expected by a default channel.

### Channel Types
- **`IWebhookContainer`**
  Represents any channel that can add/edit/remove webhooks
  
- **`StandardGuildChannel`**
  Represents any of the "normal" Guild channels. These are the channels that are in the sidebar and have the "standard" functionalities expected by channels. 
  
- **`StandardGuildMessageChannel`**
  Represents any of the "normal" message channels in a Guild. This is the bridge interface between `TextChannels` and `NewsChannels` and generally supersedes the responsibility that `BaseGuildMessageChannel` had previously.
  
### Union Types
- **`GuildChannelUnion`**
  Represents the union of all `GuildChannel` channels.  

- **`MessageChannelUnion`**
  Represents the union of all `MessageChannel` channels.

- **`GuildMessageChannelUnion`**
  Represents the union of all `GuildMessageChannel` channels.

- **`IWebhookContainerUnion`**
  Represents the union of all `IWebhookContainer` channels.

- **`IThreadContainerUnion`**
  Represents the union of all `IThreadContainer` channels.

- **`DefaultGuildChannelUnion`**
   This is a specialized Union as it doesn't follow the expected pattern defined by the "normal" unions above. This Union extends `StandardGuildChannel` just like `StandardGuildChannelUnion` but it is specialized to only have `asXChannel` methods for the channels that are actually viable for the `Guild#getDefaultChannel` and `Member#getDefaultChannel` methods. This is in contrast to the `StandardGuildChannelUnion` which has an `asXChannel` on it for every channel type that implements `StandardGuildChannel`
   
   In the future we may have more specialized unions that follow this pattern, but for now this is the only specialized union.
## Examples
```java
@Override
public void onMessageReceived(MessageReceivedEvent event) {
    if (!event.isFromGuild())
        return;

    GuildMessageChannelUnion union = event.getGuildChannel();

    //Methods available on this union.
    //Of course you couldn't use these all together as one of them would throw since it can only be one!
    TextChannel tChan = union.asTextChannel();
    NewsChannel nChan = union.asNewsChannel();
    ThreadChannel thChan = union.asThreadChannel();
    //etc

    //This would throw if the channel was actually a ThreadChannel,
    //  but if it isn't a ThreadChannel then it provides a lot of shared functionality to work with!
    StandardGuildMessageChannel sChan = union.asStandardGuildMessageChannel();


    // And, if you want to check the actual type, you can just use the `instanceof` checks like you already do today
    //  before doing any asX getters. Remember, they're only there to help you know what classes are available
    //  and the union is still just a normal channel at the end of the day. Typecheck and do whatever you want!
    if (!(union instanceof StandardGuildMessageChannel))
        return;

    //I don't need no stinkin' asX methods. I can cast all by myself thank-you-very-much!
    StandardGuildMessageChannel ssChan = (StandardGuildMessageChannel) union;

    //Also, like previously mentioned, the union itself _is_ the channel type it is representing
    //  which in this case is `GuildMessageChannel`, so you don't have to cast or
    //  use asX if you just need access to the methods on the interface itself
    union.sendMessage("See, just using methods on GuildMessageChannel interface!").queue();

    //Lastly, if you don't care about unions, you can stick to the types you've used up until now.
    //The union returns are just there to make things more discoverable, especially for attribute interfaces
    GuildMessageChannel channel = event.getGuildChannel();
}
```


src/main/java/net/dv8tion/jda/api/entities/IWebhookContainer.java
```diff
@@ -102,7 +89,7 @@ public interface BaseGuildMessageChannel extends GuildMessageChannel, IThreadCon
      * <p>Possible ErrorResponses include:
      * <ul>
      *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_WEBHOOK}
-     *     <br>The provided id does not refer to a WebHook present in this TextChannel, either due
+     *     <br>The provided id does not refer to a WebHook present in this chanel, either due
```
DV8FromTheWorld, line 92:
```suggestion
     *     <br>The provided id does not refer to a WebHook present in this channel, either due
```

src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
```diff
@@ -203,21 +205,20 @@ public PrivateChannel getPrivateChannel()
      * @return The channel this Reaction was used in
      */
     @Nonnull
-    public MessageChannel getChannel()
+    public MessageChannelUnion getChannel()
     {
-        return channel;
+        return (MessageChannelUnion) channel;
```
MinnDevelopment, line 210: Should `channel` not be this type instead of just casting every time its used?

src/main/java/net/dv8tion/jda/api/entities/Member.java
```diff
@@ -422,18 +423,18 @@ default ImageProxy getEffectiveAvatar()
...
-     * The default {@link net.dv8tion.jda.api.entities.BaseGuildMessageChannel BaseGuildMessageChannel} for a {@link net.dv8tion.jda.api.entities.Member Member}.
+     * The default {@link net.dv8tion.jda.api.entities.StandardGuildChannel channel} for a {@link net.dv8tion.jda.api.entities.Member Member}.
      * <br>This is the channel that the Discord client will default to opening when a Guild is opened for the first time
      * after joining the guild.
      * <br>The default channel is the channel with the highest position in which the member has
      * {@link net.dv8tion.jda.api.Permission#VIEW_CHANNEL Permission.VIEW_CHANNEL} permissions. If this requirement doesn't apply for
      * any channel in the guild, this method returns {@code null}.
      *
-     * @return The {@link net.dv8tion.jda.api.entities.BaseGuildMessageChannel BaseGuildMessageChannel} representing the default channel for this member
+     * @return The {@link StandardGuildChannel channel} representing the default channel for this member
```
MinnDevelopment, line 433:
```suggestion
     * @return The {@link DefaultGuildChannelUnion channel} representing the default channel for this member
```

src/main/java/net/dv8tion/jda/api/entities/Member.java
```diff
@@ -422,18 +423,18 @@ default ImageProxy getEffectiveAvatar()
     boolean isPending();
 
     /**
-     * The default {@link net.dv8tion.jda.api.entities.BaseGuildMessageChannel BaseGuildMessageChannel} for a {@link net.dv8tion.jda.api.entities.Member Member}.
+     * The default {@link net.dv8tion.jda.api.entities.StandardGuildChannel channel} for a {@link net.dv8tion.jda.api.entities.Member Member}.
```
MinnDevelopment, line 426:
```suggestion
     * The {@link DefaultGuildChannelUnion default channel} for a Member.
```

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * <br><br><b>Not all {@link StandardGuildChannel} channels can be used as a default channel!</b>
+ *
+ * <br><br>This interface represents the follow concrete channel types:
+ * <ul>
+ *     <li>{@link TextChannel}</li>
+ *     <li>{@link NewsChannel}</li>
+ * </ul>
+ */
+public interface DefaultGuildChannelUnion extends StandardGuildChannel
```
MinnDevelopment, line 37: Maybe also explain what default means

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
```diff
@@ -0,0 +1,101 @@
...
+
+    //TODO: add asForumChannel() (I think ForumChannels can be the default?)
+
+    /**
+     * Casts this union to a {@link StandardGuildMessageChannel}.
+     * <br>This only works for the following channel types represented by this union:
+     * <ul>
+     *     <li>{@link TextChannel}</li>
+     *     <li>{@link NewsChannel}</li>
+     * </ul>
```
MinnDevelopment, line 85: We shouldn't keep lists like this in getters, at most in the docs of the interface. But we don't really need these kinds of lists anyway since javadoc already provides it as "All Known Subinterfaces".

image

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
```diff
@@ -0,0 +1,101 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * A specialized union representing all channel types that can be used for the "default" channel for
+ * {@link Guild#getDefaultChannel()} or {@link Member#getDefaultChannel()}.
+ * <br>This class extends {@link StandardGuildChannel} and primarily acts as a discovery tool for
+ * developers to understand which channels might be returned as default channels.
+ *
+ * <br><br><b>Not all {@link StandardGuildChannel} channels can be used as a default channel!</b>
```
MinnDevelopment, line 29: Instead of `<br><br>`, use `<p>`

src/main/java/net/dv8tion/jda/api/entities/channel/unions/IWebhookContainerUnion.java
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * Note: This is effectively equivalent to using the cast operator:
+     * <pre><code>
+     * //These are the same!
+     * TextChannel channel = union.asTextChannel();
+     * TextChannel channel2 = (TextChannel) union;
+     * </code></pre>
+     *
+     * @throws IllegalStateException
+     *         If the channel represented by this union is not actually a {@link TextChannel}.
```
MinnDevelopment, line 48: Maybe add a note about type checking with `getType()`?

src/main/java/net/dv8tion/jda/api/entities/channel/unions/MessageChannelUnion.java
```diff
@@ -0,0 +1,117 @@
...
+     *         If the channel represented by this union is not actually a {@link ThreadChannel}.
+     *
+     * @return The channel as a {@link ThreadChannel}
+     */
+    @Nonnull
+    ThreadChannel asThreadChannel();
+
+    //TODO: Add asVoiceChannel once TiV is launched
+
+    //TODO: should we also expose the getThreadContainer/getStandardX getters like in GuildMessageChannelUnion?
```
MinnDevelopment, line 116: I think we should at least offer `asGuildMessageChannel()`

src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
```diff
@@ -342,10 +343,10 @@ public GuildChannel getAsGuildChannel()
      * @return The resolved {@link GuildMessageChannel}, or null if this was not a message channel
      */
     @Nullable
-    public GuildMessageChannel getAsMessageChannel()
+    public GuildMessageChannelUnion getAsMessageChannel()
```
MinnDevelopment, line 346: Should we get rid of all these specialized getters and just have `getChannel()`?

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
```diff
@@ -62,4 +64,69 @@ public T setName(String name)
...
+    {
+        return transformChannel(StandardGuildMessageChannel.class);
+    }
+
+    private <TOut extends Channel> TOut transformChannel(Class<TOut> toObjectClass)
+    {
+        if (toObjectClass.isInstance(this))
+            return toObjectClass.cast(this);
+
+        String cleanedClassName = this.getClass().getSimpleName().replace("Impl", "");
```
MinnDevelopment, line 129: Use `this.getType()` instead?

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
DV8FromTheWorld: The reasons these exist is because once ForumChannels and ViT exist, these lists wont be 1:1 with the implementers. Some channels represented by the union will be able to use the method and others wont.

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
DV8FromTheWorld: This gives a "TextChannel" or "VoiceChannel".
`getType()` would give TEXT or VOICE, so this was an intentional choice given how it is used for user messaging.

MinnDevelopment:
This PR should also move the channel interfaces into a dedicated package

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
MinnDevelopment: I still think we should not document the list on every getter, since that is a lot of things to remember to update.

src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
DV8FromTheWorld: So changing this to store the channel as the union instead of the supertype has 2 effects:
1) We have to update all the code passing channels _in_ to do the casts. This ends up touching a _lot_ more places than casting at getter site
2) For some situations, like in MessageReaction, storing it as `MessageChannelUnion` would still result in a cast because `getGuildChannel` needs a cast to `GuildMessageChannelUnion`.

Where possible I've taken the mentality of accepting the "base" interface (`MessageChannel` in this case) and handling the messiness of union upcasting in the getters as it seems like the easiest way to prevent passing unions around everywhere and forcing casts everywhere.

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
DV8FromTheWorld: The `<br><br>` pattern results in more vertical space separation. What's the motivation behind using `<p>` instead?

image
image

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
DV8FromTheWorld: As a note, the All Known Subinterfaces wont work for this because the channel interfaces are not children to the union interfaces. i,e, this doesn't exist:
`TextChannel -> StandardChanneUnion -> Channel`

Unions are implemented only by the `impl` classes to avoid things like `TextChannel#asVoiceChannel()` being present.

DV8FromTheWorld: My assumption here is that when we add a new channel type to a union we're going to have to add the getter. At that point we'd update the list in the interface methods.

src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
DV8FromTheWorld: I think maybe yes. We might want to consider having a `ChannelUnion` in that case though.

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
MinnDevelopment: The primary target for javadoc is the hosted html, which should prefer `<p>` over `<br><br>`. The spacing in intellij is honestly something they should change, since it seems like they treat it like a `<br>` for some reason.

src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
MinnDevelopment: Well in this case GuildChannelUnion, since you can't mention a private channel.

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
MinnDevelopment: I wouldn't even know which methods to look for when updating these lists because there are so many. I'd definitely say documenting on type and linking there is better for maintainability.

```
See {@link MessageChannelUnion} for a list of supported types. You can use {@link #getType} to check the type at runtime.
```

src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
MinnDevelopment: Couldn't most of the casts be done in the constructor though?

src/main/java/net/dv8tion/jda/internal/entities/AbstractChannelImpl.java
MinnDevelopment: The problem is that this locks us into the concrete types we have now, if we ever make some kind of more specialized type, we would end up breaking this name. Isn't the channel type clear enough?

DV8FromTheWorld: We've had the same naming scheme for our Interface -> Impl classes since 2015. I'm not overly concerned that we're gonna get this wrong.

I don't think the channel type is clear enough, especially since threads are collapsed into ThreadChannel whereas their type is actually 3 different types.

Generally, the point of the unions is to get people comfortable knowing what interfaces can be casted to what other interfaces, so I want these errors to convey that same thing.

src/main/java/net/dv8tion/jda/api/interactions/commands/OptionMapping.java
DV8FromTheWorld: Sure

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
DV8FromTheWorld: Gotcha. Can update.

src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
DV8FromTheWorld: I've elected to go forward with return site casting so that unions don't need to be distributed throughout the system.

src/main/java/net/dv8tion/jda/api/entities/channel/unions/DefaultGuildChannelUnion.java
DV8FromTheWorld: Removed the lists

src/main/java/net/dv8tion/jda/api/interactions/Interaction.java
```diff
@@ -177,11 +178,11 @@ default GuildChannel getGuildChannel()
      * @return The {@link MessageChannel}
      */
     @Nonnull
-    default MessageChannel getMessageChannel()
+    default MessageChannelUnion getMessageChannel()
     {
         Channel channel = getChannel();
         if (channel instanceof MessageChannel)
```
MinnDevelopment, line 184:
```suggestion
        if (channel instanceof MessageChannelUnion)
```

src/main/java/net/dv8tion/jda/internal/entities/WebhookImpl.java
```diff
@@ -252,7 +253,12 @@ public WebhookMessageActionImpl<Void> sendRequest()
     {
         checkToken();
         Route.CompiledRoute route = Route.Webhooks.EXECUTE_WEBHOOK.compile(getId(), token);
-        WebhookMessageActionImpl<Void> action = new WebhookMessageActionImpl<>(api, channel, route, (json) -> null);
+
+        //TODO: This MessageChannel cast seems problematic, especially once ForumChannels come along.
+        // Note for Minn: Can we make this cast here because the WebhookClient is only used when responding to interactions
+        //                actually present in a MessageChannel?
+        //                Also might not actually be needed as it doesn't look like WebhookMessageActionImpl actually uses the channel variable?
+        WebhookMessageActionImpl<Void> action = new WebhookMessageActionImpl<>(api, (MessageChannel) channel, route, (json) -> null);
```
MinnDevelopment, line 261: Yeah, I don't think this channel is used for anything and can be removed.

Also, as far as I know, you can only use interactions in message channels currently anyway. It wouldn't make sense to send a followup if you can't see it in the channel. So the entire concept of InteractionHook only makes sense in message channels in the first place.

src/main/java/net/dv8tion/jda/api/events/message/GenericMessageEvent.java
```diff
@@ -71,11 +73,11 @@ public MessageChannel getChannel()
      * @return The GuildMessageChannel
      */
     @Nonnull
-    public GuildMessageChannel getGuildChannel()
+    public GuildMessageChannelUnion getGuildChannel()
```
MinnDevelopment, line 76: Should we get rid of all the other `getXChannel()` getters on these events?

src/main/java/net/dv8tion/jda/api/events/user/UserTypingEvent.java
```diff
@@ -65,9 +66,9 @@ public OffsetDateTime getTimestamp()
      * @return The channel
      */
     @Nonnull
-    public MessageChannel getChannel()
+    public MessageChannelUnion getChannel()
```
MinnDevelopment, line 69: Same here

src/main/java/net/dv8tion/jda/api/interactions/Interaction.java
MinnDevelopment: And here

MinnDevelopment:
This PR should also address [GenericChannelEvent](https://github.com/DV8FromTheWorld/JDA/blob/feature/channel-unions/src/main/java/net/dv8tion/jda/api/events/channel/GenericChannelEvent.java#L63)

src/main/java/net/dv8tion/jda/api/events/message/GenericMessageEvent.java
DV8FromTheWorld: Done

src/main/java/net/dv8tion/jda/internal/entities/WebhookImpl.java
DV8FromTheWorld: Done

DV8FromTheWorld:
PR now addresses GenericChannelEvent

src/main/java/net/dv8tion/jda/api/entities/MessageReaction.java
```diff
@@ -170,52 +172,30 @@ public Guild getGuild()
         return channel != null ? channel.getGuild() : null;
```
MinnDevelopment, line 172: This null check can be removed then, and getGuild is still nullable

MinnDevelopment (approved):
After addressing the `MessageReaction#getGuild` and the javadoc error, this should be good to go.

PR #2139, opened 2022-05-23
Fix Message.Attachment#getProxy()

Xirado:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fixes `Message.Attachment#getProxy()`, the proxy url returns status code 415 (unsupported media type) for non-image attachments.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -2696,7 +2696,7 @@ public interface Message extends ISnowflake, Formattable
         @Nonnull
         public String getProxyUrl()
         {
-            return proxyUrl;
+            return isImage() ? proxyUrl : url;
         }
 
         /**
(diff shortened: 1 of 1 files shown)


DV8FromTheWorld:
Both URLs point at a discord CDN, correct? it's just the image one is the media-proxy?

MinnDevelopment (approved):

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2709,7 +2709,7 @@ public String getProxyUrl()
         @Nonnull
         public AttachmentProxy getProxy()
         {
-            return new AttachmentProxy(getProxyUrl());
+            return new AttachmentProxy(isImage() ? proxyUrl : url);
```
MinnDevelopment, line 2712: Media proxy also works for videos, or any content that has a width and height.

```suggestion
            return new AttachmentProxy(width > 0 && height > 0 ? proxyUrl : url);
```

PR #2148, opened 2022-06-05
Add newly_created field to ThreadChannel

Shea4:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: N/A

## Description

Adds a new method `isNewlyCreated` to ThreadChannel so you can differenciate when a ThreadChannel is created or if the self member joined the thread

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/ThreadChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/ThreadChannel.java
@@ -110,6 +110,15 @@ public interface ThreadChannel extends GuildMessageChannel, IMemberContainer
     boolean isInvitable();
 
     /**
+     * Whether this thread is newly created.
+     * <br>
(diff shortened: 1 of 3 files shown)


MinnDevelopment:
Isn't this field only applicable to `CHANNEL_CREATE` events? I don't think it makes much sense to put that on the actual channel entity, when its primarily used to distinguish events.

Shea4:
It is yes. Would a seperate event for `THREAD_CREATE` events be better or a method on the ChannelCreateEvent which throws if it's not for a thread, or any other suggestions you may have

DV8FromTheWorld:
I'm gonna need to read up more on how we'd use this, but this definitely should not be on the entity itself.

DV8FromTheWorld:
Superseded by https://github.com/DV8FromTheWorld/JDA/pull/2254

PR #2160, opened 2022-06-19
fix: Make MessageType#deletable return an accurate value.

RedDaedalus:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fixes the `MessageType#deletable` getter.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageType.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageType.java
@@ -131,7 +131,7 @@ public enum MessageType
      * A new message sent as the first message in threads that are started from an existing message in the parent channel.
      * It only contains a message reference field that points to the message from which the thread was started.
      */
-    THREAD_STARTER_MESSAGE(21),
+    THREAD_STARTER_MESSAGE(21, false, true),
 
     /**
      * The "Invite your friends" messages that are sent to guild owners in new servers.
--- a/src/main/java/net/dv8tion/jda/api/entities/MessageType.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/MessageType.java
@@ -26,35 +26,35 @@ public enum MessageType
     /**
      * The normal text messages received when a user or bot sends a Message.
      */
-    DEFAULT(0, false),
+    DEFAULT(0, false, true),
 
     /**
      * Specialized messages used for Groups as a System-Message showing that a new User has been added to the Group.
      * Also used in message threads to indicate a member has joined that thread.
      */
-    RECIPIENT_ADD(1),
+    RECIPIENT_ADD(1, true, false),
 
     /**
      * Specialized messages used for Groups as a System-Message showing that a new User has been removed from the Group.
      * Also used in message threads to indicate a member has left that thread.
      */
-    RECIPIENT_REMOVE(2),
+    RECIPIENT_REMOVE(2, true, false),
 
     /**
      * Specialized message used for Groups as a System-Message showing that a Call was started.
      */
-    CALL(3),
+    CALL(3, true, false),
 
     /**
      * Specialized message used for Groups as a System-Message showing that the name of the Group was changed.
      * Also used in message threads to indicate the name of that thread has changed.
      */
-    CHANNEL_NAME_CHANGE(4),
+    CHANNEL_NAME_CHANGE(4, true, false),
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/entities/MessageType.java
```diff
@@ -148,25 +148,31 @@
...
-    MessageType(int id, boolean system)
-    {
+    MessageType(int id, boolean system, boolean deletable) {
         this.id = id;
         this.system = system;
+        this.deletable = deletable;
+    }
+
+    MessageType(int id, boolean system) {
+        // most system messages are not deletable, most non-system messages are
```
MinnDevelopment, line 169: Pretty much every system message is deletable, the only one i know that prevents it is automod. Do you have actual exampels? I'm not a fan of having to say "true, true" on every enum constant here.

V-Play-Games: Outside of Automod, Message sent in group DMs when changing the icon or the name, making calls, addition or removal of members are not deletable. But that's that

MinnDevelopment: Group DMs are not even accessible to bots normally, so we shouldn't really worry about those types at all. We could comment them out instead. From what i can tell the only actually undeletable messages accessible to bots, are thread start messages and automod, am i wrong?

RedDaedalus: Automod messages can be deleted. The two that can't be deleted are channel name and thread starter.

MinnDevelopment: Then they changed it, they were not deletable when the type was added.

RedDaedalus: Yes, that is what this PR is primarily addressing. 
image

src/main/java/net/dv8tion/jda/api/entities/MessageType.java
```diff
@@ -148,25 +148,31 @@
...
+        this(id, system, !system);
+    }
+
+    MessageType(int id) {
```
MinnDevelopment, line 173: Remove these overload constructors and only have one with both booleans. I don't like having this implicit true and false here, its confusing.

src/main/java/net/dv8tion/jda/api/entities/MessageType.java
```diff
@@ -148,25 +148,31 @@
...
+    private final boolean deletable;
 
-    MessageType(int id)
-    {
-        this(id, true); // true as default because the majority are system
-    }
-
-    MessageType(int id, boolean system)
-    {
+    MessageType(int id, boolean system, boolean deletable) {
```
MinnDevelopment, line 162:
```suggestion
    MessageType(int id, boolean system, boolean deletable)
    {
```

src/main/java/net/dv8tion/jda/api/entities/MessageType.java
```diff
@@ -131,12 +131,12 @@
      * A new message sent as the first message in threads that are started from an existing message in the parent channel.
      * It only contains a message reference field that points to the message from which the thread was started.
      */
-    THREAD_STARTER_MESSAGE(21),
+    THREAD_STARTER_MESSAGE(21, false, true),
```
MinnDevelopment, line 134: I thought this one is not deletable?

src/main/java/net/dv8tion/jda/api/entities/MessageType.java
```diff
@@ -193,13 +199,13 @@ public boolean isSystem()
 
     /**
      * Whether messages of this type can be deleted.
-     * <br>Currently the only type that cannot be deleted is {@link #AUTO_MODERATION_ACTION}.
+     * <br>This includes most system messages.
```
MinnDevelopment, line 202:
```suggestion
     * <br>These are messages which are required to stay such as thread starter messages.
```

MinnDevelopment (changes requested):

MinnDevelopment (approved):

DV8FromTheWorld (approved):

PR #2165, opened 2022-06-26
Update to API version 10

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This implements the required changes for version 10.

See [Changelog](https://discord.com/developers/docs/change-log#api-v10)

~~TODO: Check behavior of `attachments` in all message edit endpoints. This wasn't mentioned in the changelog but was supposed to change in v10.~~ See https://github.com/discord/discord-api-docs/issues/5170


MinnDevelopment:
I've added a warning for the new content intent to raise awareness. Whenever you use `message.getContentRaw()` (and other related methods) without this intent, you will now receive this warning:

```
Attempting to access message content without GatewayIntent.MESSAGE_CONTENT.
Discord now requires to explicitly enable access to this using the MESSAGE_CONTENT intent.
Useful resources to learn more:
	- https://support-dev.discord.com/hc/en-us/articles/4404772028055-Message-Content-Privileged-Intent-FAQ
	- https://jda.wiki/using-jda/gateway-intents-and-member-cache-policy/
	- https://jda.wiki/using-jda/troubleshooting/#im-getting-closecode4014-disallowed-intents
Or suppress this warning if this is intentional with Message.suppressContentIntentWarning()
```

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1444,12 +1447,14 @@ public AuditableRestAction<Void> kick(@Nonnull UserSnowflake user, String reason
             checkPosition(member);
 
         Route.CompiledRoute route = Route.Guilds.KICK_MEMBER.compile(getId(), user.getId());
+
+        AuditableRestAction<Void> action = new AuditableRestActionImpl<>(getJDA(), route);
         if (!Helpers.isBlank(reason))
         {
             Checks.check(reason.length() <= 512, "Reason cannot be longer than 512 characters.");
-            route = route.withQueryParams("reason", EncodingUtil.encodeUTF8(reason));
+            return action.reason(reason);
```
DV8FromTheWorld, line 1455: I thought this didn't work for the kick and ban endpoints. was it updated to work?

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -120,6 +122,28 @@ public ReceivedMessage withHook(InteractionHook hook)
         return this;
     }
 
+    private void checkIntent()
+    {
+        // Checks whether access to content is limited and the message content intent is not enabled
+        if (api != null && !contentIntentWarning && !api.isIntent(GatewayIntent.MESSAGE_CONTENT))
```
DV8FromTheWorld, line 128: When would `api` be null here?

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -52,6 +53,7 @@
 
 public class ReceivedMessage extends AbstractMessage
 {
+    public static boolean contentIntentWarning = false;
```
DV8FromTheWorld, line 56: can we make this a more boolean-y name?
`shouldShowContentIntentWarning` (which ofc would flip the actual meaning, but still)

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -120,6 +122,28 @@ public ReceivedMessage withHook(InteractionHook hook)
         return this;
     }
 
+    private void checkIntent()
+    {
+        // Checks whether access to content is limited and the message content intent is not enabled
+        if (api != null && !contentIntentWarning && !api.isIntent(GatewayIntent.MESSAGE_CONTENT))
+        {
+            SelfUser selfUser = api.getSelfUser();
+            if (!Objects.equals(selfUser, author) && !mentions.getUsers().contains(selfUser) && isFromGuild())
```
DV8FromTheWorld, line 131: Something I'm curious about here:
What about messages from the REST api? Don't they also require the Intent to be turned on in the admin dashboard?

So it could be that we received a message via REST and the dashboard didn't have the intent turned on but we used it in the gateway, so this warning wouldn't show, right?

Or does discord not let a WS connection establish with MESSAGE_CONTENT if it isn't enabled for REST too in the dashboard?

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
MinnDevelopment: This is because of the [changelog](https://discord.com/developers/docs/change-log#api-v10):

> To specify a reason for an administrative action in audit logs, apps must now pass the X-Audit-Log-Reason header rather than the reason parameter for all endpoints. Read more in the [Audit Logs documentation](https://discord.com/developers/docs/resources/audit-log).

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
MinnDevelopment: I would hope it can only be enabled in the gateway if its enabled in the application dashboard, similar to all other privileged intents.

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
DV8FromTheWorld: Wow the v10 release post is surprisingly easy to miss in that changelog

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
DV8FromTheWorld: We can and should remove this from the constructor.
`new ReceivedMessage` only happens in one place, the EntityBuilder
https://github.com/DV8FromTheWorld/JDA/blob/a1c3ecdb43700139e9474f3531c1b590019a25e5/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java#L1554

And it does so in a method that requires that `channel` be non-null.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
DV8FromTheWorld: Works for me.

DV8FromTheWorld (approved):

MinnDevelopment:
Converting back to draft for now because of the breaking changes to the `attachments` field in v10, which we still have to handle in the message edit rewrite.

PR #2167, opened 2022-06-28
Add EmojiUnion

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Oversight from the emoji rework, this adds a way to easily convert the emoji types in events like reaction events.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/emoji/Emoji.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/emoji/Emoji.java
@@ -147,14 +147,14 @@ public interface Emoji extends SerializableData, Formattable
      * @return The emoji instance
      */
     @Nonnull
-    static Emoji fromFormatted(@Nonnull String code)
+    static EmojiUnion fromFormatted(@Nonnull String code)
     {
         Checks.notEmpty(code, "Formatting Code");
         Matcher matcher = Message.MentionType.EMOJI.getPattern().matcher(code);
         if (matcher.matches())
-            return fromCustom(matcher.group(1), Long.parseUnsignedLong(matcher.group(2)), code.startsWith("<a"));
+            return (EmojiUnion) fromCustom(matcher.group(1), Long.parseUnsignedLong(matcher.group(2)), code.startsWith("<a"));
         else
-            return fromUnicode(code);
+            return (EmojiUnion) fromUnicode(code);
     }
 
     /**
@@ -168,14 +168,17 @@ public interface Emoji extends SerializableData, Formattable
      * @return The emoji instance
      */
     @Nonnull
-    static Emoji fromData(@Nonnull DataObject emoji)
+    static EmojiUnion fromData(@Nonnull DataObject emoji)
     {
         Checks.notNull(emoji, "Emoji Data");
         if (emoji.isNull("id"))
-            return fromUnicode(emoji.getString("name"));
+            return (EmojiUnion) fromUnicode(emoji.getString("name"));
         else
-            return fromCustom(emoji.getString("name"), emoji.getUnsignedLong("id"), emoji.getBoolean("animated"));
+            return (EmojiUnion) fromCustom(emoji.getString("name"), emoji.getUnsignedLong("id"), emoji.getBoolean("animated"));
     }
+
+    @Nonnull
+    EmojiUnion toUnion();
 
     /**
      * The {@link Type} of this emoji.
--- a/src/main/java/net/dv8tion/jda/api/entities/emoji/Emoji.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/emoji/Emoji.java
@@ -178,14 +178,6 @@ public interface Emoji extends SerializableData, Formattable
     }
 
     /**
-     * Converts this to a {@link EmojiUnion}, mostly used for internal conversion.
-     *
-     * @return {@link EmojiUnion}
-     */
-    @Nonnull
-    EmojiUnion toUnion();
-
-    /**
      * The {@link Type} of this emoji.
      *
      * @return The {@link Type}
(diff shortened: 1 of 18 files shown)


src/main/java/net/dv8tion/jda/api/entities/emoji/Emoji.java
```diff
@@ -168,15 +168,23 @@ static Emoji fromFormatted(@Nonnull String code)
...
     }
 
+    /**
+     * Converts this to a {@link EmojiUnion}, mostly used for internal conversion.
+     *
+     * @return {@link EmojiUnion}
+     */
+    @Nonnull
+    EmojiUnion toUnion();
+
```
DV8FromTheWorld, line 187: This seems useless honestly? Why would we expose this?

src/main/java/net/dv8tion/jda/api/interactions/components/selections/SelectOption.java
```diff
@@ -208,7 +210,12 @@ public SelectOption withDefault(boolean isDefault)
     @CheckReturnValue
     public SelectOption withEmoji(@Nullable Emoji emoji)
     {
-        return new SelectOption(label, value, description, isDefault, emoji);
+        EmojiUnion union = null;
+        if (emoji instanceof EmojiUnion)
+            union = (EmojiUnion) emoji;
+        else if (emoji != null)
+            union = (EmojiUnion) Emoji.fromCustom((CustomEmoji) emoji);
```
DV8FromTheWorld, line 217: This is very confusing and I don't understand what is happening.
EmojiUnion is an extension of Emoji that every EmojiImpl implements. We should be able to blindly cast any `Emoji` to `EmojiUnion`, right? What is it doing with `fromCustom` here?

DV8FromTheWorld:
Worst case we could cast in constructor and store as a variable, but I definitely don't enjoy having to pipe that interface through to everything. It really should just be a discovery tool exposed as a return, not something that pervades the internals of the lib.

src/main/java/net/dv8tion/jda/api/interactions/components/selections/SelectOption.java
MinnDevelopment: Because Emoji is now an interface, users can implement their own types, which don't necessarily have the union implemented.

DV8FromTheWorld: We do not support external implementation of JDA interfaces.
Interfaces of entities are to hide internal implementation.

src/main/java/net/dv8tion/jda/api/interactions/components/buttons/Button.java
```diff
@@ -156,7 +157,7 @@ default Button withDisabled(boolean disabled)
     @CheckReturnValue
     default Button withEmoji(@Nullable Emoji emoji)
     {
-        return new ButtonImpl(getId(), getLabel(), getStyle(), getUrl(), isDisabled(), emoji);
+        return new ButtonImpl(getId(), getLabel(), getStyle(), getUrl(), isDisabled(), (EmojiUnion) emoji);
```
DV8FromTheWorld, line 160: Don't cast _into_ these constructors

src/main/java/net/dv8tion/jda/api/interactions/components/selections/SelectMenu.java
```diff
@@ -465,7 +466,7 @@ public Builder addOption(@Nonnull String label, @Nonnull String value, @Nonnull
         @Nonnull
         public Builder addOption(@Nonnull String label, @Nonnull String value, @Nullable String description, @Nullable Emoji emoji)
         {
-            return addOptions(new SelectOption(label, value, description, false, emoji));
+            return addOptions(new SelectOption(label, value, description, false, (EmojiUnion) emoji));
```
DV8FromTheWorld, line 469: Same for here.

src/main/java/net/dv8tion/jda/api/interactions/components/selections/SelectOption.java
```diff
@@ -208,7 +209,7 @@ public SelectOption withDefault(boolean isDefault)
     @CheckReturnValue
     public SelectOption withEmoji(@Nullable Emoji emoji)
     {
-        return new SelectOption(label, value, description, isDefault, emoji);
+        return new SelectOption(label, value, description, isDefault, (EmojiUnion) emoji);
```
DV8FromTheWorld, line 212: Same for here

src/main/java/net/dv8tion/jda/api/interactions/components/selections/SelectOption.java
```diff
@@ -260,7 +261,7 @@ public boolean isDefault()
      * @return The attached emoji
      */
     @Nullable
-    public Emoji getEmoji()
+    public EmojiUnion getEmoji()
     {
         return emoji;
```
DV8FromTheWorld, line 266: This would be the only place that needs a cast. All the others could be removed.

src/main/java/net/dv8tion/jda/internal/entities/ActivityImpl.java
```diff
@@ -95,7 +95,7 @@ public RichPresence.Timestamps getTimestamps()
 
     @Nullable
     @Override
-    public Emoji getEmoji()
+    public EmojiUnion getEmoji()
     {
         return emoji;
```
DV8FromTheWorld, line 100: This is only place that would need the cast if we removed the union requirement on the constructor.

MinnDevelopment: We only ever create the `ActivityImpl` using `Emoji.fromData` which already returns `EmojiUnion`, adding any casts at all to this is unnecessary.

DV8FromTheWorld (approved):

DV8FromTheWorld (approved):

DV8FromTheWorld:
Lol Github with the double approval

PR #2169, opened 2022-07-02
add support for string option bounds

caneleex:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this PR adds support for setting the minimum and maximum length for `STRING` options: https://github.com/discord/discord-api-docs/pull/5143

it's been a while since I touched JDA/Java so there may be things I forgot to add/change.


MinnDevelopment:
Could we allow to use `setRequiredRange` for these as well?

https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java#L559

caneleex:
> Could we allow to use `setRequiredRange` for these as well?
> 
> [`master`/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java#L559](https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java?rgh-link-date=2022-07-02T10%3A06%3A36Z#L559)

something like `OptionData#setRequiredLengthRange`?

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -596,6 +633,27 @@ public OptionData setRequiredRange(double minValue, double maxValue)
         return this;
     }
 
+    @Nonnull
+    public OptionData setMinLength(int minLength)
+    {
+        if (type != OptionType.STRING)
+            throw new IllegalArgumentException("Can only set min length for options of type STRING");
+        Checks.check(minLength >= 0, "Min length must be greater than or equal to 0");
```
caneleex, line 641: would it make sense to add a bound check for the maximum value here as well?

MinnDevelopment:
> something like OptionData#setRequiredLengthRange?

no, just the same method with a conditional check for type.

caneleex:
> > something like OptionData#setRequiredLengthRange?
> 
> no, just the same method with a conditional check for type.

I feel like that could be a bit confusing. introducing a new method would get rid of having to explain 2 behaviors of that method in the docs + long -> int conversions. making an overload which'd accept ints instead would probably confuse/annoy users as well - if they wanted to set the range for an integer option, they'd have to use **L**ongs specifically otherwise it'd clash with the ints overload and throw, or am I misunderstanding here?

MinnDevelopment:
> I feel like that could be a bit confusing

I think it's more confusing for it not to work, which is why I'm suggesting it.

> making an overload which'd accept ints instead 

I don't think we need one?

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -596,6 +633,87 @@ public OptionData setRequiredRange(double minValue, double maxValue)
...
+     * Configure the minimum and maximum length for strings which can be provided for this option.
+     *
+     * @param  minLength
+     *         The minimum length for strings which can be provided for this option.
+     * @param  maxLength
+     *         The maximum length for strings which can be provided for this option.
+     * @throws IllegalArgumentException
+     *         If any of the following checks fail
+     *         <ul>
+     *             <li>{@link OptionType type of this option} is {@link OptionType#STRING STRING}</li>
```
Xirado, line 696:
```suggestion
     *             <li>{@link OptionType type} of this option is {@link OptionType#STRING STRING}</li>
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -596,6 +633,87 @@ public OptionData setRequiredRange(double minValue, double maxValue)
...
+    {
+        if (type != OptionType.STRING)
+            throw new IllegalArgumentException("Can only set min and max length for options of type STRING");
+        Checks.check(minLength >= 0, "Min length must be greater than or equal to 0");
+        Checks.check(maxLength >= MAX_STRING_OPTION_LENGTH_MINIMUM && maxLength <= MAX_STRING_OPTION_LENGTH,
+                "Max length must be greater than or equal to %d and lower than or equal to %d", MAX_STRING_OPTION_LENGTH_MINIMUM, MAX_STRING_OPTION_LENGTH);
+        this.minLength = minLength;
+        this.maxLength = maxLength;
+        return this;
+    }
```
Xirado, line 715: This should probably also check if `minLength` is greater than `maxLength`

Xirado (changes requested):

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
caneleex: this is correct.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
caneleex: this isn't done in the range methods either, so I'd much rather leave this as is and maybe make a followup PR if you really wanted to and change it for both range and length methods.

DV8FromTheWorld: Currently this is done during the serialization for min and max value.

DV8FromTheWorld: Well, that's interesting. I definitely remember reading code that was doing the min/max comparison at serialization time because I remember deleting a comment i left on the min/max PR about adding that check in on the setX methods themselves but realized we were doing it just before serialization.

We definitely need to add these in. Them not being there for the others doesn't mean we shouldn't add it for these. I see no reason why we wouldn't add them tbh.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
caneleex: I'll add the check. however, according to docs, quote:

> The value of `min_length` must be greater or equal to `0`

this check is correct.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -71,12 +71,23 @@ public class OptionData implements SerializableData
      */
     public static final int MAX_CHOICES = 25;
 
+    /**
+     * The minimum value for the maximum length for a {@link OptionType#STRING String option}.
+     */
+    public static final int MAX_STRING_OPTION_LENGTH_MINIMUM = 1;
```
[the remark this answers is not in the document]
caneleex: the name you proposed doesn't reflect what the field is for. the minimum length for the *maximum* length (`max_length`) should be greater than or equal to this value.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
caneleex: where do I put it? putting the check into the `setXLength` methods doesn't really make sense because you can't know the order of these calls, and I doubt it's good practice to throw in `toData()`.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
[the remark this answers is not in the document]
MinnDevelopment: Do we really need a constant for this?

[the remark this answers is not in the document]
caneleex: it's your call really.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
Xirado: That's how i did it in `TextInput`:
https://github.com/DV8FromTheWorld/JDA/blob/6e8c030fbe000c16224f658b1e37dffabe5a9383/src/main/java/net/dv8tion/jda/api/interactions/components/text/TextInput.java#L341-L350

In the cases where users use `setMaxLength()` and `setMinLength()`, it's checked in the `build()` method.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -724,6 +761,87 @@ public OptionData setRequiredRange(double minValue, double maxValue)
...
+     *         </ul>
+     *
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData setMinLength(int minLength)
+    {
+        if (type != OptionType.STRING)
+            throw new IllegalArgumentException("Can only set min length for options of type STRING");
+        Checks.check(minLength >= 0, "Min length must be greater than or equal to 0");
```
MinnDevelopment, line 783:
```suggestion
        Checks.notNegative(minLength, "Min length");
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -724,6 +761,87 @@ public OptionData setRequiredRange(double minValue, double maxValue)
...
+     *
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData setMaxLength(int maxLength)
+    {
+        if (type != OptionType.STRING)
+            throw new IllegalArgumentException("Can only set max length for options of type STRING");
+        Checks.check(maxLength >= MAX_STRING_OPTION_LENGTH_MINIMUM && maxLength <= MAX_STRING_OPTION_LENGTH,
+                "Max length must be greater than or equal to %d and lower than or equal to %d", MAX_STRING_OPTION_LENGTH_MINIMUM, MAX_STRING_OPTION_LENGTH);
```
MinnDevelopment, line 809:
```suggestion
        Checks.positive(maxLength, "Max length");
        Checks.check(maxLength <= MAX_STRING_OPTION_LENGTH, "Max length must not be greater than %d. Provided: %d", MAX_STRING_OPTION_LENGTH, maxLength);
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -724,6 +761,87 @@ public OptionData setRequiredRange(double minValue, double maxValue)
...
+     *         </ul>
+     *
+     * @return The OptionData instance, for chaining
+     */
+    @Nonnull
+    public OptionData setRequiredLength(int minLength, int maxLength)
+    {
+        if (type != OptionType.STRING)
+            throw new IllegalArgumentException("Can only set min and max length for options of type STRING");
+        Checks.check(minLength >= 0, "Min length must be greater than or equal to 0");
```
MinnDevelopment, line 837:
```suggestion
        Checks.notNegative(minLength, "Min length");
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -724,6 +761,87 @@ public OptionData setRequiredRange(double minValue, double maxValue)
...
+        Checks.check(maxLength >= MAX_STRING_OPTION_LENGTH_MINIMUM && maxLength <= MAX_STRING_OPTION_LENGTH,
+                "Max length must be greater than or equal to %d and lower than or equal to %d", MAX_STRING_OPTION_LENGTH_MINIMUM, MAX_STRING_OPTION_LENGTH);
```
MinnDevelopment, line 839:
```suggestion
        Checks.check(maxLength <= MAX_STRING_OPTION_LENGTH,
                "Max length must not be greater than %d. Provided: %d", MAX_STRING_OPTION_LENGTH, maxLength);
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -724,6 +761,87 @@ public OptionData setRequiredRange(double minValue, double maxValue)
...
+        this.minLength = minLength;
```
MinnDevelopment, line 840:
```suggestion
        Checks.check(minLength <= maxLength, "Min length must not be greater than max length. Provided: %d > %d", minLength, maxLength);
        this.minLength = minLength;
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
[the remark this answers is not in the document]
MinnDevelopment: I think we can remove this now

MinnDevelopment (approved):

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -724,6 +756,87 @@ public OptionData setRequiredRange(double minValue, double maxValue)
...
+     * Configure the minimum length for strings which can be provided for this option.
+     *
+     * @param  minLength
+     *         The minimum length for strings which can be provided for this option.
+     * @throws IllegalArgumentException
+     *         If any of the following checks fail
+     *         <ul>
+     *             <li>{@link OptionType type of this option} is {@link OptionType#STRING STRING}</li>
+     *             <li>{@code minLength} is greater than or equal to {@code 0}</li>
+     *         </ul>
```
DV8FromTheWorld, line 769: Oh I thought we killed this way of defining `IllegalArgumentException`.
It really should be more like
```
@throws IllegalArgumentException
     *         <ul>
     *             <li>If the {@link OptionType type of this option} is not {@link OptionType#STRING STRING}</li>
     *             <li>If {@code minLength} is less than to {@code 0}</li>
     *         </ul>
``` 

I guess we don't _have_ to do it in this PR, but this is the pattern we've moved to in JDA in every other place.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
```diff
@@ -724,6 +756,87 @@ public OptionData setRequiredRange(double minValue, double maxValue)
...
+    public OptionData setRequiredLength(int minLength, int maxLength)
+    {
+        if (type != OptionType.STRING)
+            throw new IllegalArgumentException("Can only set min and max length for options of type STRING");
+        Checks.notNegative(minLength, "Min length");
+        Checks.check(maxLength <= MAX_STRING_OPTION_LENGTH,
+                "Max length must not be greater than %d. Provided: %d", MAX_STRING_OPTION_LENGTH, maxLength);
+        Checks.check(minLength <= maxLength, "Min length must not be greater than max length. Provided: %d > %d", minLength, maxLength);
+        this.minLength = minLength;
+        this.maxLength = maxLength;
```
DV8FromTheWorld, line 836: Is there a reason we re-implement these checks instead of doing:
```suggestion
        Checks.check(minLength <= maxLength, "Min length must not be greater than max length. Provided: %d > %d", minLength, maxLength);
        this.setMinLength(minLength);
        this.setMaxLength(maxLength);
```

src/main/java/net/dv8tion/jda/api/interactions/commands/build/OptionData.java
DV8FromTheWorld: Ah, tracked this down. We changed it in a PR that never got merged.
I updated and merged that PR, so this language should be changed to match the new system before we merge.

DV8FromTheWorld (approved):

PR #2172, opened 2022-07-05
default to an empty string in Emoji#fromData

caneleex:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

this PR fixes possible exceptions when `name` in emoji objects is `null`, as mentioned in [this example in the api docs](https://discord.com/developers/docs/resources/emoji#emoji-object-custom-emoji-examples).

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/emoji/Emoji.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/emoji/Emoji.java
@@ -174,7 +174,7 @@ public interface Emoji extends SerializableData, Formattable
         if (emoji.isNull("id"))
             return (EmojiUnion) fromUnicode(emoji.getString("name"));
         else
-            return (EmojiUnion) fromCustom(emoji.getString("name"), emoji.getUnsignedLong("id"), emoji.getBoolean("animated"));
+            return (EmojiUnion) fromCustom(emoji.getString("name", ""), emoji.getUnsignedLong("id"), emoji.getBoolean("animated"));
     }
 
     /**
--- a/src/main/java/net/dv8tion/jda/internal/handle/MessageReactionHandler.java
+++ b/src/main/java/net/dv8tion/jda/internal/handle/MessageReactionHandler.java
@@ -147,11 +147,7 @@ public class MessageReactionHandler extends SocketHandler
             );
         }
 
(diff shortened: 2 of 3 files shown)


src/main/java/net/dv8tion/jda/internal/handle/MessageReactionHandler.java
```diff
@@ -147,11 +147,7 @@ protected Long handleInternally(DataObject content)
             );
         }
 
-        // reaction remove has null name sometimes
-        if (emoji.isNull("name"))
-            emoji.put("name", "");
         EmojiUnion rEmoji = Emoji.fromData(emoji);
-
```
DManstrator, line 154: For consistency reason I would add that newline back or remove it too from the `MessageReactionClearEmojiHandler`.

src/main/java/net/dv8tion/jda/api/entities/emoji/Emoji.java
```diff
@@ -174,7 +174,7 @@ static EmojiUnion fromData(@Nonnull DataObject emoji)
         if (emoji.isNull("id"))
             return (EmojiUnion) fromUnicode(emoji.getString("name"));
```
DManstrator, line 175: Is there a reason the empty string is not used as the default value here too?

src/main/java/net/dv8tion/jda/internal/handle/MessageReactionHandler.java
caneleex: I don't think anything mentioned matters at all. in `MessageReactionClearEmojiHandler`, I think the variables passed into `new MessageReaction(...)` were just meant to be separated, same with `MessageReactionHandler` and the check.

Mitmocc:
There is no real purpose to this. It skips over the check for null and ends up throwing an IllegalArgumentException because of an empty name. I guess the exception is a little bit better to understand but also misleading as the name is not empty but null

MinnDevelopment:
Thank you, but I would like to keep the checks the way they are right now in the Emoji factory methods. Instead, I'm adding `EntityBuilder.createEmoji` as an unchecked alternative to `Emoji.fromData` in #2176

PR #2187, opened 2022-07-22
Message Rework

### LINKED ISSUE #1675: Add Attachment Support to MessageBuilder
- [x] I have checked for similar issues.
- [x] I have updated to the [latest JDA version][download].
- [x] I have checked the branches or the maintainers' PRs for upcoming features.

## Feature Request

Add support for file attachments to the `MessageBuilder` class.  `MessageBuilder` currently provides most options for construction of a new `Message`, but `Attachment` support is conspicuously absent.

### Example Use-Case

When two or more objects are collaborating on the construction of a `Message`, pass a `MessageBuilder` around as a container.  Use the new functionality to attach files, when necessary.

### Alternative Option

Add getter methods to `MessageAction`.  `MessageAction` supports mutable `Message` construction, similar to `MessageBuilder`, and does provide `Attachment` support.  However, it does not offer any way to read data already inside it.  This can be quite inconvenient if collaborating objects need to read each other's contributions to a `Message` under construction.

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes #1675

## Description

Reworks the message sending and editing capabilities for more consistency and reducing code duplication.

- Split send and edit into separate request types (MessageCreateRequest and MessageEditRequest)
- Split MessageBuilder into MessageEditBuilder and MessageCreateBuilder
- Support allowed mentions on edits
- Support file descriptions
- Major abstraction to cover the 90% overlap between sending and editing messages
- Allow setting files on message building
- Message builders no longer return Message interface, instead returning concrete data classes
- The concept of `override(true)` that was previously used on `MessageAction`, has been moved explicitly to edit requests and is now called `setReplace(true)`, which I believe to be more descriptive.
- `MessageBuilder#buildAll` has moved to a dedicated util class and works for arbitrary strings/boundaries `SplitUtil#split`

### Interface Hierarchy

image

### Example

To send a message to a channel, you can use `MessageCreateBuilder`:

```java
MessageCreateBuilder builder = new MessageCreateBuilder()
      .setContent("Hello Guys")
      .addActionRow(Button.link("https://github.com/DV8FromTheWorld/JDA", "GitHub"))
      .addFiles(FileUpload.fromData(new File("github_logo.png")));
try (MessageCreateData createData = builder.build()) {
  for (MessageChannel channel : targetChannels) {
    channel.sendMessage(createData).queue();
  }
} // if there is an exception, we should close the file
```

For message edit requests, use the `MessageEditBuilder` instead:

```java
Message m = /* ... */ ;
List<AttachedFile> attachments = new ArrayList<>(m.getAttachments()); // keep original attachments
attachments.add(FileUpload.fromData(new File("github_logo.png"))); // add a file to the message
MessageEditBuilder builder = new MessageEditBuilder()
      .setContent("Hello Guys")
      .setActionRow(Button.link("https://github.com/DV8FromTheWorld/JDA", "GitHub"))
      .setAttachments(attachments); // set the attachments to the updated list
try (MessageEditData editData = builder.build()) {
  m.editMessage(editData).queue();
} // if there is an exception, we should close the file
```

Convert between edit and send:

```java
// Allows you to make an edit from a create request.
// This will be a *replace*, causing the entire message to be replaced by this new data.
MessageEditData editData = MessageEditData.fromCreateData(createData);
```

```java
// Only sets the parts that were explicitly edited (if you only called setContent, it will only have content)
MessageCreateData createData = MessageCreateData.fromEditData(editData);
```

Similarly, you can also use `fromMessage(Message)` on each of these. Those factories have the same kind of semantics.

Note: In theory, an extension library could *always* use `MessageEditData` and simply require users to convert it. There is no real downside to that, other than user boilerplate.

## Breaking Changes

| Before | After |
|---|---|
| `retainFiles*` | `setAttachments` |
| `setActionRows` | `setComponents` |
| `addActionRows` | `addComponents` |
| `allowedMentions` | `setAllowedMentions` |
| `MessageAction#content` | `MessageRequest#setContent` |
| `MessageAction#tts` | `MessageCreateAction#setTTS` |
| `MessageBuilder` | `MessageCreateBuilder` and `MessageEditBuilder` |
| `WebhookMessageUpdateAction` | `WebhookMessageEditAction` |
| `WebhookMessageAction` | `WebhookMessageCreateAction` |
| `MessageAction` | `MessageCreateAction` and `MessageEditAction` |
| `sendMessage(Message)` | `sendMessage(MessageCreateData)` |
| `editMessage(Message)` | `editMessage(MessageEditData)` |
| `sendFile(...)` | `sendFiles(FileUpload...)` |
| `addFile(...)` | `addFiles(...)`, `setFiles(...)`, and `setAttachments(...)` |
| `editMessage(InputStream, ...)` | `editMessageAttachments(AttachedFile...)` |

`MessageBuilder#buildAll` has been replaced by `SplitUtil#split`:

Before:

```java
List<Message> messages = new MessageBuilder().setContent(input).buildAll();
```

After:

```java
List<String> messages = SplitUtil.split(
    input,            // input string of arbitrary length
    2000,             // the split limit, can be arbitrary (>0)
    true,             // whether to trim the strings (empty will be discarded)
    Strategy.NEWLINE, // split on '\n' characters if possible
    Strategy.ANYWHERE // otherwise split on the limit
);
```


src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
```diff
@@ -0,0 +1,168 @@
...
+        return this;
+    }
+
+    @Nonnull
+    @Override
+    public MessageCreateBuilder setFiles(@Nullable Collection<? extends FileUpload> files)
+    {
+        if (files != null)
+            Checks.noneNull(files, "Files");
+        this.files.forEach(IOUtil::silentClose);
```
MinnDevelopment, line 97: This might not be a good idea

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return new MessageCreateBuilder().applyData(data);
+    }
+    
+    @Nonnull
+    @Override
+    public MessageCreateBuilder addContent(@Nonnull String content)
+    {
+        Checks.notNull(content, "Content");
+        this.content.append(content.trim());
```
MinnDevelopment, line 53:
```suggestion
        this.content.append(content);
```

MinnDevelopment:
I've added [SplitUtil](https://github.com/DV8FromTheWorld/JDA/blob/feature/message-rework/src/main/java/net/dv8tion/jda/api/utils/SplitUtil.java) as a replacement for [MessageBuilder#buildAll](https://github.com/DV8FromTheWorld/JDA/blob/master/src/main/java/net/dv8tion/jda/api/MessageBuilder.java#L1114).

Example:

```java
String longtext = ...;
for (String part : SplitUtil.split(longtext, 2000, true, SplitUtil.Strategy.NEWLINE, SplitUtil.Strategy.ANYWHERE) {
    channel.sendMessage(part).queue();
}
```

This will try to first find a suitable substring using `'\n'` as a delimiter, and if it cannot find it anywhere in the next 2000 characters, it will split exactly at the limit.

MinnDevelopment:
I downgraded the gateway API to v9 again, since v10 seems to unnecessarily enforce the content intent before the deadline. You will still get a warning if you disable the intent. Also, if you do not have access to the intent due to missing approval by discord, you can still enable it in v9. This allows bots to smoothly transition from content to no content, but will also break every bot that enables it without access on the deadline.

This a really strange enforcement and makes it very unintuitive for bots to transition, but that's just what we gotta deal with downstream.

src/main/java/net/dv8tion/jda/api/utils/SplitUtil.java
```diff
@@ -0,0 +1,225 @@
...
+import net.dv8tion.jda.internal.utils.Checks;
+
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.function.Predicate;
+
+/**
+ * Utility to strategically split strings.
```
DV8FromTheWorld, line 28: We might want to show a usecase for message building here.

We probably also want to show a before/after in the PR description of MessageBuilder.buildAll vs impl using this now.

src/main/java/net/dv8tion/jda/api/utils/messages/AbstractMessageBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+ *        The return type used for method chaining
+ *
+ * @see   MessageCreateBuilder
+ * @see   MessageEditBuilder
+ */
+@SuppressWarnings({"ResultOfMethodCallIgnored", "unchecked"})
+public abstract class AbstractMessageBuilder<T, R extends AbstractMessageBuilder<T, R>> implements MessageRequest<R>
+{
+    protected final List<MessageEmbed> embeds = new ArrayList<>(10);
+    protected final List<LayoutComponent> components = new ArrayList<>(5);
```
DV8FromTheWorld, line 50: These initial sizes are based on the max that can be added right? Lets use the constants

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
```diff
@@ -0,0 +1,265 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Builder specialized for building a {@link MessageCreateData}.
+ *
+ * @see MessageEditBuilder
+ */
```
DV8FromTheWorld, line 39: Probably should mention exactly what `MessageCreateData` is representing / for.
Probably should have an example.
Probably should have some `@see` in here.

DV8FromTheWorld, line 39: As a note, the MessageBuilder used to live at the root of JDA at `net.dv8tion.jda.api.MessageBuilder` but these new builders live deep inside of utils. Not sure what to think about that yet.

DV8FromTheWorld: Actually, it looks like not just Builders, but the Data and Request interfaces live there too. Makes me think we need to move this out of `utils` and into a closer to top-level package. We have a looot of message related stuff
 - `MessageType`
 - `MessageHistory`
 - `MessageReference`
 - `MessageEmbed`
 - `MessageReaction`
 - `Message`

Maybe it would be better to introduce
 - `net.dv8tion.jda.api.entities.message`
 - `net.dv8tion.jda.api.entities.message.data`
 - `net.dv8tion.jda.api.entities.message.builder`
 
 Or something similar?

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
```diff
@@ -0,0 +1,265 @@
...
+        this.tts = tts;
+        return this;
+    }
+
+    @Nonnull
+    @Override
+    public MessageCreateBuilder setSuppressEmbeds(boolean suppress)
+    {
+        if (suppress)
+            this.flags |= 4;
```
DV8FromTheWorld, line 197: Can we move this magic number to a constant or a set of flags?

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditBuilder.java
```diff
@@ -0,0 +1,322 @@
...
+    protected static final int COMPONENTS    = 1 << 2;
+    protected static final int ATTACHMENTS   = 1 << 3;
+    protected static final int MENTIONS      = 1 << 4;
+
+    protected static final int ALL = CONTENT | EMBEDS | COMPONENTS | ATTACHMENTS | MENTIONS;
+
+    private boolean replace = false;
+    private int set = 0;
+
+    private final List<AttachedFile> attachments = new ArrayList<>(10);
```
DV8FromTheWorld, line 58: If this 10 comes from a known constant, lets use that.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditBuilder.java
```diff
@@ -0,0 +1,322 @@
...
+    public boolean isValid()
+    {
+        if (isSet(EMBEDS) && embeds.size() > Message.MAX_EMBED_COUNT)
+            return false;
+        if (isSet(COMPONENTS) && components.size() > Message.MAX_COMPONENT_COUNT)
+            return false;
+        return !isSet(CONTENT) || Helpers.codePointLength(content) <= Message.MAX_CONTENT_LENGTH;
+    }
+
+    private boolean isSet(int flag)
```
DV8FromTheWorld, line 276: can we make this package protected to allow it to be used in `MessageCreateBuilder`?

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditBuilder.java
```diff
@@ -0,0 +1,322 @@
...
+        set |= COMPONENTS;
+        return this;
+    }
+
+    @Nonnull
+    @Override
+    public MessageEditBuilder applyData(@Nonnull MessageEditData data)
+    {
+        Checks.notNull(data, "Data");
+        this.set = data.getFlags();
```
DV8FromTheWorld, line 235: I understand what this is mutating the flags so that `isSet` resolves correctly, but honestly this is confusing.
I wish it was doing `data.isSet(CONTENT)` instead of `this.isSet(CONTENT)` because it semantically doesn't follow.

We can package protected expose `isSet` on `MessageEditData` to solve this problem.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditData.java
```diff
@@ -0,0 +1,343 @@
...
+        );
+    }
+
+    @Override
+    public synchronized void close()
+    {
+        files.forEach(IOUtil::silentClose);
+    }
+
+    private boolean isSet(int flag)
```
DV8FromTheWorld, line 339: Lets make this package protected so that siblings can call it.

DV8FromTheWorld: Honestly though, I do wonder if it would be useful for users to have access to this too 🤔

DV8FromTheWorld (changes requested):
Okay, stopping at a halfway point.
Looking good!

src/main/java/net/dv8tion/jda/api/JDAInfo.java
```diff
@@ -20,7 +20,7 @@
  */
 public class JDAInfo
 {
-    public static final int DISCORD_GATEWAY_VERSION = 10;
+    public static final int DISCORD_GATEWAY_VERSION = 9;
```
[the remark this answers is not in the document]
MinnDevelopment: Explained in my comment here: https://github.com/DV8FromTheWorld/JDA/pull/2187#issuecomment-1195376998

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditBuilder.java
MinnDevelopment: What for?

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
MinnDevelopment: Builders make more sense here in this util package, they aren't really entities or related to those listed types beside the name.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditBuilder.java
MinnDevelopment: The API actually doesn't limit the number of files, 10 is just a reasonable expectation.

src/main/java/net/dv8tion/jda/api/utils/messages/AbstractMessageBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+    @Override
+    public R mentionRepliedUser(boolean mention)
+    {
+        allowedMentions.mentionRepliedUser(mention);
+        return (R) this;
+    }
+
+    @Nonnull
+    @Override
+    public R allowedMentions(@Nullable Collection<Message.MentionType> allowedMentions)
```
DV8FromTheWorld, line 91: Should these mention methods be on the abstract one? Seems like they should only be on the Create builder as you can't retroactively mention ping someone in an edit

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
```diff
@@ -0,0 +1,288 @@
...
+     *
+     * @throws IllegalArgumentException
+     *         If null is provided
+     *
+     * @return A new MessageCreateBuilder instance with the applied data
+     *
+     * @see    #applyEditData(MessageEditData)
+     */
+    @Nonnull
+    public static MessageCreateBuilder fromEdit(@Nonnull MessageEditData data)
```
DV8FromTheWorld, line 109: Question: Why are we doing `from` + `fromEdit` + `fromMessage` instead of just `from`?
We could just overload.

Just for developer discovery?

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
```diff
@@ -0,0 +1,288 @@
...
+            Checks.noneNull(files, "Files");
+        this.files.clear();
+        if (files != null)
+            this.files.addAll(files);
+        return this;
+    }
+
+    @Nonnull
+    @Override
+    public List<? extends FileUpload> getAttachments()
```
DV8FromTheWorld, line 192: shouldn't this be `getFiles`?

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
```diff
@@ -0,0 +1,288 @@
...
+        else
+            this.flags &= ~flag;
+        return this;
+    }
+
+    @Override
+    public boolean isSuppressEmbeds()
+    {
+        return (this.flags & MessageFlag.EMBEDS_SUPPRESSED.getValue()) != 0;
+    }
```
DV8FromTheWorld, line 230: This can be done on edit too. Why isn't this impl in the abstract builder?

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
```diff
@@ -0,0 +1,288 @@
...
+    public MessageCreateData build()
+    {
+        // Copy to prevent modifying data after building
+        String content = this.content.toString().trim();
+        List<MessageEmbed> embeds = new ArrayList<>(this.embeds);
+        List<FileUpload> files = new ArrayList<>(this.files);
+        List<LayoutComponent> components = new ArrayList<>(this.components);
+        AllowedMentionsImpl allowedMentions = this.allowedMentions.copy();
+
+        if (content.isEmpty() && embeds.isEmpty() && files.isEmpty())
```
DV8FromTheWorld, line 256: Is there a reason this doesn't call `isEmpty`?

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageCreateActionImpl.java
```diff
@@ -0,0 +1,186 @@
...
+    {
+        request.onSuccess(api.getEntityBuilder().createMessageWithChannel(response.getObject(), channel, false));
+    }
+
+    @Nonnull
+    @Override
+    public MessageCreateAction setNonce(@Nullable String nonce)
+    {
+        if (nonce != null)
+            Checks.notLonger(nonce, 25, "Nonce");
```
DV8FromTheWorld, line 118: Is 25 a known length?
If so, we should make it a constant somewhere.

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageEditActionImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    {
+        return (MessageEditAction) super.setCheck(checks);
+    }
+
+    @Nonnull
+    @Override
+    public MessageEditAction deadline(long timestamp)
+    {
+        return (MessageEditAction) super.deadline(timestamp);
+    }
```
DV8FromTheWorld, line 88: Can we use Fluent here?

src/main/java/net/dv8tion/jda/internal/requests/restaction/WebhookMessageCreateActionImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    {
+        return (WebhookMessageCreateAction<T>) super.setCheck(checks);
+    }
+
+    @NotNull
+    @Override
+    public WebhookMessageCreateAction<T> deadline(long timestamp)
+    {
+        return (WebhookMessageCreateAction<T>) super.deadline(timestamp);
+    }
```
DV8FromTheWorld, line 107: Can we use fluent here?

src/main/java/net/dv8tion/jda/internal/requests/restaction/WebhookMessageEditActionImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    protected RequestBody finalizeData()
+    {
+        try (MessageEditData data = builder.build())
+        {
+            DataObject json = data.toData();
+            List<FileUpload> files = data.getFiles();
+            if (files.isEmpty())
+                return getRequestBody(json);
+            return AttachedFile.createMultipartBody(files, json).build();
```
DV8FromTheWorld, line 65: Doesn't this need to be nested in `payload_json` like Create?

src/main/java/net/dv8tion/jda/internal/requests/restaction/WebhookMessageEditActionImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    {
+        return (WebhookMessageEditAction<T>) super.setCheck(checks);
+    }
+
+    @Nonnull
+    @Override
+    public WebhookMessageEditAction<T> deadline(long timestamp)
+    {
+        return (WebhookMessageEditAction<T>) super.deadline(timestamp);
+    }
```
DV8FromTheWorld, line 88: Can we use Fluent here?

src/main/java/net/dv8tion/jda/internal/requests/restaction/interactions/AutoCompleteCallbackActionImpl.java
```diff
@@ -96,14 +98,28 @@ public AutoCompleteCallbackAction addChoices(@Nonnull Collection<Command.Choice>
...
+    {
+        return (AutoCompleteCallbackAction) super.setCheck(checks);
+    }
+
+    @Nonnull
+    @Override
+    public AutoCompleteCallbackAction deadline(long timestamp)
+    {
+        return (AutoCompleteCallbackAction) super.deadline(timestamp);
     }
```
DV8FromTheWorld, line 124: Lets use Fluent here

src/main/java/net/dv8tion/jda/internal/requests/restaction/interactions/MessageEditCallbackActionImpl.java
```diff
@@ -71,114 +72,32 @@ public MessageEditCallbackActionImpl deadline(long timestamp)
...
-    {
-        Checks.noneNull(embeds, "MessageEmbed");
-        Checks.check(embeds.size() <= Message.MAX_EMBED_COUNT, "Cannot have more than %d embeds per message!", Message.MAX_EMBED_COUNT);
-        for (MessageEmbed embed : embeds)
+        try (MessageEditData data = builder.build())
         {
-            Checks.check(embed.isSendable(),
-                    "Provided Message contains an empty embed or an embed with a length greater than %d characters, which is the max for bot accounts!",
-                    MessageEmbed.EMBED_MAX_LENGTH_BOT);
+            json.put("data", data);
```
DV8FromTheWorld, line 93: `data` here is `MessageEditData`, not `DataObject`. Is there some magic here or should this be calling `toData`?

DV8FromTheWorld (commented):
Here are some more

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditRequest.java
```diff
@@ -0,0 +1,223 @@
...
+        Checks.noneNull(attachments, "Attachments");
+        return setAttachments(Arrays.asList(attachments));
+    }
+
+    @Nonnull
+    @Override
+    default R setFiles(@Nullable Collection<? extends FileUpload> files)
+    {
+        return setAttachments(files);
+    }
```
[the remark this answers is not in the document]
MinnDevelopment: Yes exactly, the alternative is not having access to files at all on the abstraction

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageCreateActionImpl.java
```diff
@@ -0,0 +1,186 @@
...
+                "Cannot send more than %d stickers in a message!", Message.MAX_STICKER_COUNT);
+        for (StickerSnowflake sticker : stickers)
+        {
+            if (sticker instanceof GuildSticker)
+            {
+                GuildSticker guildSticker = (GuildSticker) sticker;
+                Checks.check(guildSticker.isAvailable(),
+                        "Cannot use unavailable sticker. The guild may have lost the boost level required to use this sticker!");
+                Checks.check(guildSticker.getGuildIdLong() == guildChannel.getGuild().getIdLong(),
+                        "Sticker must be from the same guild. Cross-guild sticker posting is not supported!");
```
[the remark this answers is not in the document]
MinnDevelopment: Yes it is, see https://github.com/discord/discord-api-docs/discussions/3422

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageCreateActionImpl.java
```diff
@@ -0,0 +1,186 @@
...
+
+    @Override
+    protected RequestBody finalizeData()
+    {
+        if (builder.isEmpty())
+        {
+            if (!stickers.isEmpty())
+                return getRequestBody(DataObject.empty().put("sticker_ids", stickers));
+            throw new IllegalStateException("Cannot build empty messages! Must provide at least one of: content, embed, file, or stickers");
+        }
```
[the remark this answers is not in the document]
MinnDevelopment: Because you can only use stickers on the create message endpoint, see https://github.com/discord/discord-api-docs/discussions/4441

src/main/java/net/dv8tion/jda/api/utils/messages/AbstractMessageBuilder.java
MinnDevelopment: You can edit the mentions on edit requests, you cannot add new pings, but you can affect whether it highlights the message.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
MinnDevelopment: Its consistent with other factory methods like on Emoji

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
MinnDevelopment: This is declared on the abstract interface, so it also has to support attachment updates.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
MinnDevelopment: Thread safety, since we copy the data out before this.

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageEditActionImpl.java
MinnDevelopment: No, it clashes with RestActionImpl

src/main/java/net/dv8tion/jda/internal/requests/restaction/WebhookMessageEditActionImpl.java
MinnDevelopment: In this case its handled by createMultipartBody

src/main/java/net/dv8tion/jda/internal/requests/restaction/interactions/MessageEditCallbackActionImpl.java
MinnDevelopment: It implements `SerializableData`, which is special cased by `DataObject#put`.

src/main/java/net/dv8tion/jda/internal/entities/SystemMessage.java
```diff
@@ -57,35 +59,42 @@ public RestAction<Void> unpin()
 
     @Nonnull
     @Override
-    public MessageAction editMessage(@Nonnull CharSequence newContent)
+    public MessageEditAction editMessage(@Nonnull CharSequence newContent)
     {
         throw new UnsupportedOperationException("Cannot edit message of this Message Type. MessageType: " + getType());
     }
```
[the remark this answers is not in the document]
MinnDevelopment: That is out of scope for this PR, I think it should be done in a separate one.

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageEditActionImpl.java
DV8FromTheWorld: Ah i see. We _are_ using fluent in the interface but it conflicts in the impl

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditBuilder.java
```diff
@@ -0,0 +1,332 @@
...
+    public MessageEditBuilder setAllowedMentions(@Nullable Collection<Message.MentionType> allowedMentions)
+    {
+        super.setAllowedMentions(allowedMentions);
+        set |= MENTIONS;
+        return this;
+    }
+
+    @Nonnull
+    @Override
+    public MessageEditBuilder mention(@Nonnull Collection<? extends IMentionable> mentions)
```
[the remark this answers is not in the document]
MinnDevelopment: The varargs is a default method implementation in the interface, which just uses Arrays.asList

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateBuilder.java
```diff
@@ -0,0 +1,267 @@
...
+    @Nonnull
+    public static MessageCreateBuilder fromEdit(@Nonnull MessageEditData data)
+    {
+        return new MessageCreateBuilder().applyEditData(data);
+    }
+
+    /**
+     * Factory method to start a builder from an existing instance of {@link Message}.
+     * <br>Equivalent to {@code new MessageCreateBuilder().applyMessage(data)}.
+     * <br>The {@link net.dv8tion.jda.api.utils.MessageData} are not updated to reflect the provided message, and might mention users that the message did not.
```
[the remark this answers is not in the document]
MinnDevelopment: Because we cannot extract the `allowed_mentions` from it.

src/main/java/net/dv8tion/jda/api/entities/WebhookClient.java
```diff
@@ -423,13 +273,13 @@ default WebhookMessageAction<T> sendFile(@Nonnull byte[] data, @Nonnull String n
      *         The new message content to use
      *
      * @throws IllegalArgumentException
-     *         If the provided content is null, empty, or longer than {@link Message#MAX_CONTENT_LENGTH}
+     *         If the provided content is null or longer than {@link Message#MAX_CONTENT_LENGTH} characters
```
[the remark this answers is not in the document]
MinnDevelopment: Yes, this allows creating a request builder without anything set. This was already possible previously with `sendMessage("tmp").setContent("")`.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -0,0 +1,619 @@
...
+     *         <ul>
+     *             <li>If {@code null} is provided</li>
+     *             <li>If any component is not {@link ItemComponent#isMessageCompatible() message compatible}</li>
+     *             <li>In all the same cases as {@link ActionRow#of(ItemComponent...)} throws an exception</li>
+     *         </ul>
+     *
+     * @return The same instance for chaining
+     */
+    @Nonnull
+    default R setActionRow(@Nonnull Collection<? extends ItemComponent> components)
```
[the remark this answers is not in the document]
MinnDevelopment: It is rather rare for people to use more than one action row, and this is not harmful to keep as it reduces the boilerplate quite a bit.

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -279,511 +285,273 @@ default List<CompletableFuture<Void>> purgeMessagesById(@Nonnull long... message
...
+     *         {@link net.dv8tion.jda.api.Permission#VIEW_CHANNEL Permission.VIEW_CHANNEL} or {@link net.dv8tion.jda.api.Permission#MESSAGE_SEND Permission.MESSAGE_SEND}
      *
-     * @return {@link MessageAction MessageAction}
-     *         <br>Providing the {@link net.dv8tion.jda.api.entities.Message Message} created from this upload.
+     * @return {@link MessageCreateAction}
      */
     @Nonnull
     @CheckReturnValue
-    default MessageAction sendFile(@Nonnull File file, @Nonnull String fileName, @Nonnull AttachmentOption... options)
+    default MessageCreateAction sendMessageEmbeds(@Nonnull Collection<? extends MessageEmbed> embeds)
```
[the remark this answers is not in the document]
MinnDevelopment: The permission checks are done by the mixin interface

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -0,0 +1,619 @@
...
+     *         If null is provided
+     *
+     * @return The same instance for chaining
+     *
+     * @see    #setAllowedMentions(Collection)
+     * @see    #setDefaultMentions(Collection)
+     */
+    @Nonnull
+    @CheckReturnValue
+    default R mentionUsers(@Nonnull long... userIds)
```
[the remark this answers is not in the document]
MinnDevelopment: The UserSnowflake version would be `mention(...)` which accepts any `IMentionable` and automatically puts it into the right array based on the type.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditBuilder.java
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public MessageEditBuilder applyData(@Nonnull MessageEditData data)
+    {
+        Checks.notNull(data, "Data");
+        this.set |= data.getSet();
+        if ((this.set & ALL) == ALL)
+            this.replace = true;
```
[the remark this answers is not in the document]
MinnDevelopment: I changed the code to carry the `isReplace` flag through the data now

src/main/java/net/dv8tion/jda/api/entities/MessageChannel.java
```diff
@@ -279,511 +285,273 @@ default List<CompletableFuture<Void>> purgeMessagesById(@Nonnull long... message
...
+     * @return {@link MessageCreateAction}
      *
-     * @return {@link MessageAction MessageAction}
-     *         <br>Provides the {@link net.dv8tion.jda.api.entities.Message Message} created from this upload.
+     * @see    FileUpload#fromData(InputStream, String)
      */
     @Nonnull
     @CheckReturnValue
-    default MessageAction sendFile(@Nonnull InputStream data, @Nonnull String fileName, @Nonnull AttachmentOption... options)
+    default MessageCreateAction sendFiles(@Nonnull Collection<? extends FileUpload> files)
```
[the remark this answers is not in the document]
MinnDevelopment: Also checked by the mixin interface

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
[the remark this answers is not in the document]
DV8FromTheWorld: Ah. I guess this caught me off guard because it might actually be something like `mentionUsersByIds` with our normal naming.
So, realistically, we could consider dropping `mentionUsers` and ask people to use the UserSnowflake, but i think this is fine to stay

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditData.java
```diff
@@ -353,6 +359,6 @@ public synchronized void close()
 
     protected boolean isSet(int flag)
     {
-        return (set & flag) != 0;
+        return isReplace || (set & flag) != 0;
```
DV8FromTheWorld: So this might have unintended side effects.
```java
MessageEditBuilder builder1 = new MessageEditBuilder();
builder1.setContent("A cool thing!");

MessageEditBuilder builder2 = new MessageEditBuilder();
builder2.setAttachments(FileUpload.fromData(new File("./test.txt")));
builder2.setReplace(true);

builder1.applyData(builder2.build());

System.out.println("content: " + builder1.getContent());
System.out.println("files: " + builder1.getAttachments().size());
```

I'd expect the `builder1.getContent()` here to be `"A cool thing!"` because I didn't actually mutate content on `builder2`, but if I'm understanding this correctly, it would actually be `null` or `""` because of the replace?

DV8FromTheWorld (commented):
Looking great! Just need to get the last conversations resolved and then we can ship

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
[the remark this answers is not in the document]
DV8FromTheWorld: I consider this similarly to the fact that we remove the `addFile(various parts that make up a file as params)` and instead went with `addFile(FileUpload)`. Same type of concept around MessageEmbeds in general.

When there are new layout components I _do not_ want to add another `addXLayout` method.
So eventually this is going to become a situation where we have shortcut support for ActionRows and none for others when they come. Just something to note.

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageCreateActionImpl.java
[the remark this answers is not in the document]
DV8FromTheWorld: Discussed internally bringing this to other endpoints; Sounds like it wasn't done only because it was forgotten; As far as I can tell there are no real reasons to block this.

I've added it to my infinitely long backlog.

src/main/java/net/dv8tion/jda/internal/requests/restaction/MessageCreateActionImpl.java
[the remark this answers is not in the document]
DV8FromTheWorld: Discussed internally; Sounds like there wasn't a technical reason it wasn't done, but to do so will require discussion with other teams.

I've added it to my backlog.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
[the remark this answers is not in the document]
MinnDevelopment: I doubt that new layouts are coming any time soon. I think for now this is fine.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditData.java
MinnDevelopment: Yes, that's correct, and that does make sense to me. Replace always implies the entire state being replaced.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditBuilder.java
```diff
@@ -242,8 +242,7 @@ public MessageEditBuilder applyData(@Nonnull MessageEditData data)
     {
         Checks.notNull(data, "Data");
         this.set |= data.getSet();
-        if ((this.set & ALL) == ALL)
-            this.replace = true;
+        this.replace |= data.isReplace();
```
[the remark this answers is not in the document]
MinnDevelopment: You can also do `true | false` on booleans actually

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditData.java
DV8FromTheWorld: > Replace always implies the entire state being replaced.

Works for me

DV8FromTheWorld (approved):
Nice work. Long time coming; glad to see this one over the finish line

Issue #2188, opened 2022-07-22
ThreadChannel perm checks checking for Manage_Channel


Vrontis:
### General Troubleshooting

- [X] I have checked for similar issues on the Issue-tracker.
- [X] I have updated to the [latest JDA version](https://ci.dv8tion.net/job/JDA/lastSuccessfulBuild/)
- [X] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

### Expected Behaviour

JDA should _only_ check for the Manage_Threads permission, on edit/delete related perm checks on ThreadChannels, not also for Manage_Channel.
Tested needed permissions on regular Discord client & on non-JDA bots.

### Code Example for Reproduction Steps

```java
public void threadPermExample(ThreadChannel threadChannel) {
    // Here the .getManager() throws the the InsufficientPermissionException, effectively calling
    // net.dv8tion.jda.internal.managers.channel.ChannelManagerImpl#checkPermissions
    threadChannel.getManager().setArchived(false).complete();

    // Here it's .delete()
    // net.dv8tion.jda.internal.entities.mixin.channel.middleman.GuildChannelMixin#delete
    threadChannel.delete().queue();
}
```

### Code for JDABuilder or DefaultShardManagerBuilder used

```java
api = JDABuilder.createDefault(config.getAuthToken()).addEventListeners(
                client,
                new GuildListener(),
                new ThreadListener()
        ).enableIntents(GatewayIntent.GUILD_MEMBERS).build();
```

### Exception or Error

```yesyes
net.dv8tion.jda.api.exceptions.InsufficientPermissionException: Cannot perform action due to a lack of Permission. Missing permission: MANAGE_CHANNEL
	at net.dv8tion.jda.internal.managers.channel.ChannelManagerImpl.checkPermissions(ChannelManagerImpl.java:575)
	at net.dv8tion.jda.internal.managers.channel.ChannelManagerImpl.<init>(ChannelManagerImpl.java:79)
	at net.dv8tion.jda.internal.managers.channel.concrete.ThreadChannelManagerImpl.<init>(ThreadChannelManagerImpl.java:27)
	at net.dv8tion.jda.internal.entities.ThreadChannelImpl.getManager(ThreadChannelImpl.java:280)
	at dev.vrontis.quickbot.listener.GuildListener.lambda$onChannelUpdateArchived$0(GuildListener.java:38)
```

MinnDevelopment:
Resolved by 5.0.0-alpha.19

PR #2190, opened 2022-07-22
Fixed permission checks when managing/deleting a thread

Almighty-Satan:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: #2188 

## Description

When calling `ThreadChannel#getManager` or `ThreadChannel#delete` it is currently checked if the user has the `MANAGE_CHANNEL` permission, but the permission actually required to manage or delete a thread is `MANAGE_THREADS`.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/GuildChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/GuildChannel.java
@@ -54,9 +54,7 @@ public interface GuildChannel extends Channel, Comparable<GuildChannel>
      * @return The {@link Permission}
      */
     @Nonnull
-    default Permission getManagePermission() {
-    	return Permission.MANAGE_CHANNEL;
-    }
+    Permission getManagePermission();
 
     /**
      * TODO-v5: this override might not be needed anymore if we remove AuditableRestAction and instead place auditable hooks onto RestAction itself.
--- a/src/main/java/net/dv8tion/jda/api/entities/GuildChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/GuildChannel.java
@@ -1,6 +1,7 @@
 package net.dv8tion.jda.api.entities;
 
 import net.dv8tion.jda.api.JDA;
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/api/entities/GuildChannel.java
```diff
@@ -35,12 +36,23 @@ public interface GuildChannel extends Channel, Comparable<GuildChannel>
...
     ChannelManager<?, ?> getManager();
+    
+    /**
+     * Returns the {@link Permission} needed in order to manage or delete this channel.
+     * This returns {@link Permission#MANAGE_CHANNEL} or {@link Permission#MANAGE_THREADS} for {@link ThreadChannel ThreadChannels}
+     * 
+     * @return The {@link Permission}
+     */
+    @Nonnull
+    default Permission getManagePermission() {
```
MinnDevelopment, line 53: I'm against this design, I think instead we should pass the required permissions to the manager using `EnumSet<Permission>` as a constructor parameter.

Also, please try to follow the style of the project.

Almighty-Satan: Why use a `Set`? Do you think Discord will require multiple permissions for editing channels in the future?

DV8FromTheWorld:
Bump. This feature seems simple, can we get it across the finish line?

DV8FromTheWorld:
TLDR: We aren't going to go with this approach and instead update the manager check only.

An enumset here would provide more future proofing than a single Permission return would. While right now only a single Permission is required, this could change in the future, and since permissions are represented as a Set everywhere else, thatd make the most sense here as well.

However, in reality, this check is now more complicated than just a single Permission check. For threads, we also need to be aware of who owns the thread when determining if it is manageable. You don't need MANAGE_THREADS if you own the thread.

As such, we are going to entirely scrap the getter as it cannot represent that complex logic via a single Permission return, nor an EnumSet return.

PR #2212, opened 2022-08-10
Add DataPath util

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This new util class allows to access deep nested fields with a simple path expression.

### Example

```java
long value = DataPath.getLong(object, "data.options[0].value");
```

This also allows safe-calls similar to kotlin via `?.` delimiters:

```java
long value = DataPath.getLong(object, "data?.options?[0]?.value?", 0L);
```

This would be equivalent to:

```java
long value = 0L;
DataObject data = object.optObject("data").orElse(null);
if (data != null)
{
  DataArray options = data.optArray("options").orElse(null);
  if (options != null)
    value = options.getLong("value", 0L);
}
```

I'm sure you can see the appeal.

### Motivation

This is very useful in various places, and now also relevant in the case of `Event#getRawData` which usually only is used to access a specific part of the json payload. Previously, you would have to exhaustively get each object and array down the path, which is *especially* annoying with optionals.

### TODO

- [x] Array equivalent methods for each type
- [x] Documentation for all the new getters, including examples
- [x] Unit tests

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/data/DataArray.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/data/DataArray.java
@@ -542,6 +542,44 @@ public class DataArray implements Iterable<Object>, SerializableArray
     }
 
     /**
+     * Resolves the value at the specified index to a double.
+     *
+     * @param  index
+     *         The index to resolve
+     *
+     * @throws net.dv8tion.jda.api.exceptions.ParsingException
+     *         If the value is of the wrong type
+     *
+     * @return The resolved double value
+     */
+    public double getDouble(int index)
+    {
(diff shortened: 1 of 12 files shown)


DV8FromTheWorld:
So I like this concept a lot, however, using it inside of JDA itself seems kinda expensive if it were pervasively used.

MinnDevelopment:
> So I like this concept a lot, however, using it inside of JDA itself seems kinda expensive compared to the way we do accessors now. Yeah, the code would look a bit better, but it'd come at the cost of performance.

@DV8FromTheWorld I reverted those changes.

DV8FromTheWorld (approved):

PR #2220, opened 2022-08-19
Return immutable reconstructed interaction object lists

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

I have seen that some interaction related methods were returning mutable lists, causing difficult to diagnose bugs, this makes sure the stream collector returns an unmodifiable list.

While there's a lot of other methods which could fail due to manipulating reconstructed objects, it is at least simple to prevent the users to add options, for example.

Other issues not covered by the PR include adding translations to an OptionData after adding it to the command

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/IPermissionContainer.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/IPermissionContainer.java
@@ -18,6 +18,7 @@ package net.dv8tion.jda.api.entities;
 
 import net.dv8tion.jda.api.managers.channel.attribute.IPermissionContainerManager;
 import net.dv8tion.jda.api.requests.restaction.PermissionOverrideAction;
+import net.dv8tion.jda.api.utils.MiscUtil;
 
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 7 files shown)


src/main/java/net/dv8tion/jda/api/entities/IPermissionContainer.java
```diff
@@ -18,6 +18,7 @@
 
 import net.dv8tion.jda.api.managers.channel.attribute.IPermissionContainerManager;
 import net.dv8tion.jda.api.requests.restaction.PermissionOverrideAction;
+import net.dv8tion.jda.api.utils.MiscUtil;
```
MinnDevelopment, line 21: Why was this added?

src/main/java/net/dv8tion/jda/api/utils/MiscUtil.java
```diff
@@ -209,4 +213,8 @@ public static void appendTo(Formatter formatter, int width, int precision, boole
             throw new UncheckedIOException(e);
         }
     }
+
+    public static <T> Collector<T, ?, List<T>> toUnmodifiableList() {
```
MinnDevelopment, line 217: This should probably be inside of `Helpers` instead, where we already have other collection utils.
```suggestion
    public static <T> Collector<T, ?, List<T>> toUnmodifiableList()
    {
```

MinnDevelopment (approved):

PR #2232, opened 2022-08-31
Use github actions to publish artifacts

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: CI/CD

Closes Issue: NaN

## Description

This adds a new github action workflow to automatically build and upload artifacts on each push to master.

An example run is here: https://github.com/MinnDevelopment/JDA/actions/runs/2966524280

Artifacts are retained for 90 days.

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ master ]
     paths:
-      - 'src/main'
+      - 'src/main/**'
       - 'build.gradle.kts'
       - '.github/workflows/artifacts.yml'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/artifacts.yml
```diff
@@ -0,0 +1,33 @@
+name: Upload Artifacts
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - 'src/main'
```
Andre601, line 7: To my knowledge should this be `src/main/**` as GitHub would otherwise try to check if a file called `main` has been updated. The new pattern would accept any path after `src/main/`

DV8FromTheWorld (approved):

Issue #2234, opened 2022-09-01
Message components cloning bug


Xlkki:
### General Troubleshooting

- [X] I have checked for similar issues on the Issue-tracker.
- [X] I have updated to the [latest JDA version](https://ci.dv8tion.net/job/JDA/lastSuccessfulBuild/)
- [X] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

### Expected Behaviour

I have a message generated using <b>MessageCreateBuilder</b> with components. They should be disabled depending on a number of conditions, but in order to keep these components unchanged in the original message, I create a copy of it using the <code>MessageCreateBuilder.from(original)</code> method. I get a builder with a duplicate of the original message, which I can modify for my needs, but the components of this builder and the original message are linked, and if I change them in the copy using the <code>LayoutComponent.updateComponent(copy.getComponents(), someId, newComponent)</code> method, it will also change in the original message.

I showed an example of this in the screenshot. 
![image](https://user-images.githubusercontent.com/91030266/187877224-08e6114c-99b6-4373-80ab-31beb4f35774.png)

I created a message with a button and sent it. Then I created a copy of this message, replaced the button with its disabled copy, and sent it. After all that, I sent the original message again, but it also replaced the button with a disabled copy

### Code Example for Reproduction Steps

```java
public void someMethod(JDA jda, TextChannel channel) {
        Button originalButton = Button.of(ButtonStyle.SECONDARY, "some-button", "Test");
        MessageCreateData original = new MessageCreateBuilder()
                .setContent("Original message")
                .setActionRow(originalButton)
                .build();

        channel.sendMessage(original).queue();

        MessageCreateData copy = MessageCreateBuilder.from(original).setContent("Copied message").build();
        LayoutComponent.updateComponent(copy.getComponents(), "some-button", originalButton.asDisabled());

        channel.sendMessage(copy).queue();        
        channel.sendMessage(original).queue();
}
```

### Code for JDABuilder or DefaultShardManagerBuilder used

```java
JDABuilder builder = JDABuilder.createDefault("some-token");
```

MinnDevelopment:
Fixed in [5.0.0-beta.1](https://github.com/DV8FromTheWorld/JDA/releases/tag/v5.0.0-beta.1)

PR #2236, opened 2022-09-01
Use deep copy for layout components when applying message data

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: #2234 

## Description

This PR fixes layout components being copied by reference, which allowed modifications to carry over duplicated messages


src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateRequest.java
```diff
@@ -325,11 +325,18 @@ default R addFiles(@Nonnull FileUpload... files)
     default R applyData(@Nonnull MessageCreateData data)
     {
         Checks.notNull(data, "MessageCreateData");
+
+        final List<LayoutComponent> layoutComponents = data.getComponents().stream()
+                .map(LayoutComponent::createCopy)
+                .collect(Collectors.toList());
+        final List<FileUpload> files = data.getFiles().stream()
+                .map(f -> FileUpload.fromData(f.getData(), f.getName()).setDescription(f.getDescription()))
```
MinnDevelopment, line 333: I don't want file uploads to be copied this way because it opens up for issues with concurrent reads to the same input stream.

freya022: How then ? passing existing file uploads ?

DV8FromTheWorld: I think we just have to document that files do not get copied over as a limitation.
If people want to introduce concurrent problems by sharing the same FileUpload, that's one them.

It'd also be good to not try to copy files over anyways because you wouldn't be able to copy over attachments as they'd only be valid if you were editing a the same message.

DV8FromTheWorld:
Getting some kind of fix out that addresses the components part of this problem seems like a small win that we should push forward with. @MinnDevelopment do you have any other suggestions on how to proceed on this? Seems like we could just drop files and mention that they're dropped in the copy process?

MinnDevelopment:
I'm fine moving forward with it as long as we limit it to components.

MinnDevelopment (commented):
This also needs to handle some more cases where components are copied, such as `MessageEditRequest#applyCreateData/applyData` and `MessageCreateRequest#applyEditData`.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateRequest.java
```diff
@@ -325,15 +327,18 @@ default R addFiles(@Nonnull FileUpload... files)
...
+                .collect(Collectors.toList());
         return setContent(data.getContent())
                 .setAllowedMentions(data.getAllowedMentions())
                 .mentionUsers(data.getMentionedUsers())
                 .mentionRoles(data.getMentionedRoles())
                 .mentionRepliedUser(data.isMentionRepliedUser())
                 .setEmbeds(data.getEmbeds())
                 .setTTS(data.isTTS())
-                .setComponents(data.getComponents())
-                .setFiles(data.getFiles());
```
MinnDevelopment, line 336: setFiles can be added back, the concern was about `InputStream` being copied.

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateRequest.java
```diff
@@ -313,6 +313,8 @@ default R addFiles(@Nonnull FileUpload... files)
     /**
      * Applies the provided {@link MessageCreateData} to this request.
      *
+     * <p>Note that files will be copied but the input streams themselves can only be consumed once.
+     *
```
MinnDevelopment, line 317:
```suggestion
```

MinnDevelopment (approved):

PR #2237, opened 2022-09-02
Improve handling of server errors and fix some code duplication

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This changes the retry handling for server errors to only retry on a few specific codes. Additionally, the retries now use an exponential backoff instead of linear.

There also were some methods which apply body and headers, which are now used instead.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/Requester.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/Requester.java
@@ -159,29 +159,8 @@ public class Requester
         String url = DISCORD_API_PREFIX + route.getCompiledRoute();
         builder.url(url);
 
-        String method = apiRequest.getRoute().getMethod().toString();
-        RequestBody body = apiRequest.getBody();
-
-        if (body == null && HttpMethod.requiresRequestBody(method))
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/requests/Requester.java
```diff
@@ -329,4 +309,8 @@ public void shutdown()
         rateLimiter.shutdown();
     }
 
+    private static boolean shouldRetry(int code)
+    {
+        return code == 502 || code == 504 || code == 529;
```
DV8FromTheWorld, line 314: so why aren't we retrying 500s?

DV8FromTheWorld (approved):
LGTM

src/main/java/net/dv8tion/jda/internal/requests/Requester.java
MinnDevelopment: Because it likely never succeeds when a server error happened. And usually just results in double send.

PR #2240, opened 2022-09-05
Change handling of speaking updates in voice connections

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Changes handling of speaking updates, removing all cases of sending a `speaking = 0` on audio end.

Discord now handles speaking updates client-side, meaning we no longer have to set speaking to off explicitly. By removing this handling, we no longer encounter early cutoff when buffering audio packets, like with JDA-NAS.

This PR also removes the initial sending of silence packets when audio starts, discord no longer seems to have issues with that.


src/main/java/net/dv8tion/jda/api/audio/factory/IPacketProvider.java
```diff
@@ -77,53 +77,34 @@
...
      *
-     * @param  changeTalking
-     *         Whether or not to change the talking indicator if the AudioSendHandler cannot provide a new audio packet.
-     *
      * @return Possibly-null {@link java.net.DatagramPacket DatagramPacket} containing an encoded and encrypted packet
      *         of audio data ready to be sent to discord.
      */
     @Nullable
-    DatagramPacket getNextPacket(boolean changeTalking);
+    DatagramPacket getNextPacket(boolean unused);
```
DV8FromTheWorld, line 107: Why are we doing this instead of deprecating this impl and making a breaking change?

DV8FromTheWorld (commented):
Will review the rest later

src/main/java/net/dv8tion/jda/api/audio/factory/IPacketProvider.java
MinnDevelopment: Because I don't expect jda-nas to get updated and its fine to just drop it imo

PR #2250, opened 2022-09-14
feat(api): add `receiveAcknowledged` to SubscribeEvent

danorris709:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation

Closes Issue: NaN

## Description

Adds a value to the `@SubscribeEvent` annotation so that you can disable the method being triggered if the event is an GenericInteractionCreateEvent and has already been acknowledged (reduced the boiler plate of checking if the event has been interacted with in every single method)


MinnDevelopment:
Thank you, but this better belongs in end-user code rather than the library.

PR #2254, opened 2022-09-20
Handle newly_created gracefully

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds handling for `newly_created` on THREAD_CREATE.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/handle/ThreadCreateHandler.java
+++ b/src/main/java/net/dv8tion/jda/internal/handle/ThreadCreateHandler.java
@@ -18,6 +18,7 @@ package net.dv8tion.jda.internal.handle;
 
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import net.dv8tion.jda.api.events.channel.ChannelCreateEvent;
+import net.dv8tion.jda.api.events.thread.ThreadRevealedEvent;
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/handle/ThreadCreateHandler.java
```diff
@@ -35,9 +36,14 @@ protected Long handleInternally(DataObject content)
         if (api.getGuildSetupController().isLocked(guildId))
             return guildId;
 
+        if (api.getThreadChannelById(content.getUnsignedLong("id")) != null)
+            return null;
```
DV8FromTheWorld, line 40: When would this happen?

MinnDevelopment: Without wanting to test every edge case, I'm assuming the event also fires in some cases when the member is added despite already knowing it exists (moderator permissions). I just want to account for any unexpected situations this way.

DV8FromTheWorld: Lets drop a comment on there indicating that this is guess-based protection instead of specifically known situation based protection

PR #2255, opened 2022-09-20
Improve channel update handling

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Replaces the numerous duplicate update handling by using the mixin interfaces.


src/main/java/net/dv8tion/jda/internal/entities/channel/mixin/attribute/ITopicChannelMixin.java
```diff
@@ -0,0 +1,26 @@
...
+
+package net.dv8tion.jda.internal.entities.channel.mixin.attribute;
+
+import net.dv8tion.jda.internal.entities.channel.mixin.middleman.GuildChannelMixin;
+
+public interface ITopicChannelMixin<T extends ITopicChannelMixin<T>> extends GuildChannelMixin<T>
+{
+    T setTopic(String topic);
+
+    String getTopic();
```
DV8FromTheWorld, line 25: Why is the getter here in the mixin?

src/main/java/net/dv8tion/jda/internal/handle/ChannelUpdateHandler.java
```diff
@@ -103,165 +99,62 @@ protected Long handleInternally(DataObject content)
...
                         .map(json -> {
                             json.opt("emoji_id").ifPresent(id -> json.put("id", id));
                             json.opt("emoji_name").ifPresent(n -> json.put("name", n));
                             return EntityBuilder.createEmoji(json);
                         })
                         .orElse(null);
 
-                ForumChannelImpl forumChannel = (ForumChannelImpl) channel;
                 content.optArray("available_tags").ifPresent(array -> handleTagsUpdate(forumChannel, array));
+                int flags = content.getInt("flags", 0);
```
DV8FromTheWorld, line 151: Lets not have a mutation in between all of the variable initializations.

Lets move this handleTagsUpdate call to be after them.

DV8FromTheWorld (approved):
Generally this looks good to me. Had a few Qs though before merge

src/main/java/net/dv8tion/jda/internal/entities/channel/mixin/attribute/ITopicChannelMixin.java
MinnDevelopment: Because there is no other interface for it, I didn't feel it necessary to add an `ITopicableChannel`

DV8FromTheWorld (approved):

DV8FromTheWorld (approved):

PR #2258, opened 2022-09-21
Improve implementation of command data

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This moves the serialization of options into `toData()`, which removes the need to reconstruct them for getters. Additionally, I've added some more constants to replace the magic numbers in checks.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/interactions/CommandDataImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/interactions/CommandDataImpl.java
@@ -20,27 +20,24 @@ import net.dv8tion.jda.api.interactions.DiscordLocale;
 import net.dv8tion.jda.api.interactions.commands.Command;
 import net.dv8tion.jda.api.interactions.commands.DefaultMemberPermissions;
 import net.dv8tion.jda.api.interactions.commands.OptionType;
-import net.dv8tion.jda.api.interactions.commands.build.OptionData;
-import net.dv8tion.jda.api.interactions.commands.build.SlashCommandData;
-import net.dv8tion.jda.api.interactions.commands.build.SubcommandData;
-import net.dv8tion.jda.api.interactions.commands.build.SubcommandGroupData;
+import net.dv8tion.jda.api.interactions.commands.build.*;
 import net.dv8tion.jda.api.interactions.commands.localization.LocalizationFunction;
 import net.dv8tion.jda.api.interactions.commands.localization.LocalizationMap;
 import net.dv8tion.jda.api.utils.data.DataArray;
 import net.dv8tion.jda.api.utils.data.DataObject;
+import net.dv8tion.jda.api.utils.data.SerializableData;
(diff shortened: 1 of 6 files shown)


MinnDevelopment:
I've added methods to allow removing options/subcommands/groups via conditions:

```java
command.removeOptions(o -> o.isRequired());
```

This can be useful when updating commands or using some kind of template system. For example:

```java
SlashCommandData command = SlashCommandData.fromData(json); // json which has options set
command.removeOptions(v -> true); // remove options, we only need name/description
```

src/main/java/net/dv8tion/jda/internal/interactions/CommandDataImpl.java
```diff
@@ -243,18 +244,17 @@ public CommandDataImpl addSubcommandGroups(@Nonnull SubcommandGroupData... group
         if (groups.length == 0)
             return this;
         checkType(Command.Type.SLASH, "add subcommand groups");
-        if (!allowGroups)
+        if (!allowSubcommands)
```
DV8FromTheWorld, line 247: was something changed to allow groups and subcommands to be added at the same time? I'm confused as to why this changed

MinnDevelopment: Yes, according to the [docs](https://discord.com/developers/docs/interactions/application-commands#subcommands-and-subcommand-groups) you can mix subcommands and groups.

```
VALID

command
|
|__ subcommand-group
    |
    |__ subcommand
|
|__ subcommand
```

DV8FromTheWorld: Interesting! We will need to update our documentation then
Currently, variations of this are in the codebase: `Also throws if you try to mix subcommands/options/groups in one command.`

DenuxPlays:
Any updates on this?

RedPanda4552:
Curious - I see the same 25 limit for options being used for subcommands here. Is the 25 subcommand limit a limit of JDA, or Discord? I recall in the past Discord's API docs did mention a limit of 25 subcommand groups and 25 subcommands. Mysteriously, that has disappeared and all that remains are the limits of 25 options and 25 choices per option. I for one naively hope that Discord did scrub the 25 subcommand limit (hard to explain, using config/runtime generated commands and lots of them) - does anyone happen to know if that limit is still in place and perhaps just not written down on their API docs now? Personally I'd like to see the subcommand and subcommand group limits lifted but if it's still there and Discord just removed the mention of it, then that'd be too bad.

MinnDevelopment:
In the API, subcommands are equivalent to options, thus they use the same limit.

RedPanda4552:
> In the API, subcommands are equivalent to options, thus they use the same limit.

Well that's unfortunate. Explains things though, thanks!

DV8FromTheWorld (approved):

PR #2259, opened 2022-09-22
Fix check for ChannelAction#setCategory

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fixes a regression which broke Guild#createCategory due to setting category. Instead of changing the logic for that, I made sure to only throw if the parent is not null.


DV8FromTheWorld (approved):
_Teeeechnically_ we should probably throw on any channel type that doesn't support categorization, like threads.

MinnDevelopment:
@DV8FromTheWorld since threads don't use ChannelAction this seems to be the only type we need to check here anyways.

PR #2260, opened 2022-09-25
Fix some checks not throwing the intended exception

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This corrects the exception thrown for owner checks in a few places. This was unintentionally changed to throw `IllegalArgumentException`, however it should throw `HieararchyException`.

There also was a wrong annotation in `Member#timeoutUntil`.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Member.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Member.java
@@ -675,7 +675,7 @@ public interface Member extends IMentionable, IPermissionHolder, UserSnowflake
      */
     @Nonnull
     @CheckReturnValue
-    default AuditableRestAction<Void> timeoutUntil(@Nullable TemporalAccessor temporal)
+    default AuditableRestAction<Void> timeoutUntil(@Nonnull TemporalAccessor temporal)
     {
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/api/entities/Member.java
```diff
@@ -675,7 +675,7 @@ default AuditableRestAction<Void> timeoutFor(@Nonnull Duration duration)
      */
     @Nonnull
     @CheckReturnValue
-    default AuditableRestAction<Void> timeoutUntil(@Nullable TemporalAccessor temporal)
+    default AuditableRestAction<Void> timeoutUntil(@Nonnull TemporalAccessor temporal)
```
caneleex, line 678: `null` can be passed here (it will remove the timeout). if you want to change the ABI then it should be probably changed in the documentation too

MinnDevelopment: It can't be null since the method throws (and the one in Guild is nonnull)

PR #2261, opened 2022-09-25
Put "options" between command path and options, same for choices

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Puts an "options" key between the command path and the option, as well as "choices" between the option and the choice
This helps reading localization files such as JSON or YAML as option names aren't on the same indent as other command-related strings, as well as making it explicit that the user is modifying an option/choice's localization

Related convo: https://discord.com/channels/125227483518861312/869965829024915466/1023634607360462938

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/interactions/command/localization/LocalizationMapper.java
+++ b/src/main/java/net/dv8tion/jda/internal/interactions/command/localization/LocalizationMapper.java
@@ -119,7 +119,9 @@ public class LocalizationMapper
                 };
 
                 //We need to differentiate subcommands/groups from options before inserting the "options" separator
-                if (OptionType.fromKey(item.getInt("type", -1)).isOption()) { //-1 when the object isn't an option
+                final OptionType type = OptionType.fromKey(item.getInt("type", -1)); //-1 when the object isn't an option
+                final boolean isOption = type != OptionType.SUB_COMMAND && type != OptionType.SUB_COMMAND_GROUP;
+                if (isOption) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/internal/interactions/command/localization/LocalizationMapper.java
```diff
@@ -107,10 +110,25 @@ private void forObjects(DataArray source, Function<DataObject, String> keyExtrac
...
+                final Runnable runnable = () ->
+                {
+                    final String key = keyExtractor.apply(item);
+                    keyComponents.push(key);
+                    consumer.accept(item);
+                    keyComponents.pop();
+                };
+
+                //We need to differentiate subcommands/groups from options before inserting the "options" separator
+                if (OptionType.fromKey(item.getInt("type", -1)).isOption()) { //-1 when the object isn't an option
```
DV8FromTheWorld, line 122: I don't really like this.
I think we should drop the `isOption` method and instead explicitly check here if it is a SUB_COMMAND or SUB_COMMAND_GROUP instead.

```suggestion
                OptionType type = OptionType.fromKey(item.getInt("type", -1));
                bool isOption = type != OptionType.SUB_GROUP_COMMAND && type != OptionType.SUB_COMMAND;
                if (isOption) {
```

DV8FromTheWorld (approved):

PR #2269, opened 2022-09-28
Improve conditional waiting and shutdown handling

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This is a cleaner way to handle status checks in a loop.


src/main/java/net/dv8tion/jda/api/JDA.java
```diff
@@ -378,6 +379,71 @@ default JDA awaitReady() throws InterruptedException
...
+     *
+     * @throws IllegalArgumentException
+     *         If the provided timeout is null
+     * @throws InterruptedException
+     *         If the current thread is interrupted while waiting
+     *
+     * @return False, if the timeout has elapsed before the shutdown has completed, true otherwise.
+     */
+    @CheckReturnValue
+    boolean awaitShutdown(@Nonnull Duration timeout) throws InterruptedException;
```
DV8FromTheWorld, line 412: We should consider accepting `long, TemporalUnit` as it is what a lot of our apis already accept.

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -480,22 +485,56 @@ public long getGatewayPing()
...
+    public boolean awaitShutdown(@Nonnull Duration timeout) throws InterruptedException
+    {
+        long deadline = timeout.isZero() || timeout.isNegative() ? Long.MAX_VALUE : System.currentTimeMillis() + timeout.toMillis();
+        MiscUtil.tryLock(statusLock);
+        try
+        {
+            Status current = getStatus();
+            while (current != Status.SHUTDOWN)
+            {
+                if (!statusCondition.await(deadline - System.currentTimeMillis(), TimeUnit.MILLISECONDS))
```
DV8FromTheWorld, line 526: why does this compute a deadline as timeout + currentTimeMillis just to remove it later?
Couldn't we just do this?:
```suggestion
        long deadline = timeout.isZero() || timeout.isNegative() ? Long.MAX_VALUE : timeout.toMillis();
        MiscUtil.tryLock(statusLock);
        try
        {
            Status current = getStatus();
            while (current != Status.SHUTDOWN)
            {
                if (!statusCondition.await(deadline, TimeUnit.MILLISECONDS))
```

DV8FromTheWorld (commented):
Nice. Didn't know conditions were a thing but make total sense.

src/main/java/net/dv8tion/jda/internal/JDAImpl.java
MinnDevelopment: The condition can wake multiple times and each time it would wait the same timeout, which would result in a total timeout that is longer than the provided one.

DV8FromTheWorld: Yup, makes sense

DV8FromTheWorld (approved):

Issue #2280, opened 2022-10-13
Add support for the new select menu types


seailz:
### General Troubleshooting

- [X] I have checked for similar issues on the Issue-tracker.
- [X] I have updated to the [latest JDA version](https://ci.dv8tion.net/job/JDA/lastSuccessfulBuild/)
- [X] I have checked the branches or the maintainers' PRs for upcoming features.

### Feature Request

Discord have just released new select menu types, role select, user select, channel select, & mentionable select. These would be wonderful additions to JDA.

### Example Use-Case

```java
UserSelectMenu userSelect = …;
RoleSelectMenu roleSelect = …;
MentionableSelectMenu mentionableSelect = …;
ChannelSelectMenu channelSelect = …;
```

seailz:
I will probably open a PR for this shortly.

Eronary:
![image](https://user-images.githubusercontent.com/63160741/196030309-0466b654-5315-4c2f-87d8-3799a568554a.png)
Even without processing the event, I get this error when the user selects the user

PR #2281, opened 2022-10-14
New Select Menu Types

seailz:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: #2280

## Description

Adding support for the new select menu types
I hope it’s not too messy as this is my first time working with the internals of JDA :)

- [x] Implement Channel Select Menus
- [x] Implement User Select Menus
- [x] Implement Role Select Menus
- [x] Implement Mentionable Select Menus
- [x] Implement Channel Select Menu Filters

- [x] Implement new structure
- [x] Start testing
- [x] Finish testing

Finished testing, everything seems to be working well.

### **The SelectMenu class has been renamed to StringSelectMenu, to represent their new name.**
<p>

**How to create a role select menu**
```java
    public void foo() {
        EntitySelectMenu.create("hello!", EnumSet.of(SelectType.ROLE))
                .setPlaceholder("Pick a role you would like to obtain")
                .build();
    }
```

**How to create a channel select menu**
```java
    public void foo() {
        EntitySelectMenu.create("hello!", EnumSet.of(SelectType.CHANNEL))
                .setPlaceholder("Pick a channel you would like to visit to")
                .build();
    }
```

**How to create a user select menu**
```java
    public void foo() {
        EntitySelectMenu.create("hello!", EnumSet.of(SelectType.USER))
                .setPlaceholder("Pick a user you would like to talk to")
                .build();
    }
```

**How to create a mentionable select menu**
```java
    public void foo() {
        EntitySelectMenu.create("hello!", EnumSet.of(SelectType.USER, SelectType.ROLE))
                .setPlaceholder("Pick a role or user you would like to ping")
                .build();
    }
```

**Listening for an entity select interaction**
```java
                    @Override
                    public void onEntitySelectMenuInteraction(@NotNull EntitySelectMenuInteractionEvent event) {
                        event.reply("You selected " + event.getSelectedEntities(getJda()).size() + " options!").queue();
                    }
```


seailz:
I will add the remaining two select types in the morning as it's getting quite late.

src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
```diff
@@ -143,6 +141,9 @@ public void onUserContextInteraction(@Nonnull UserContextInteractionEvent event)
     public void onMessageContextInteraction(@Nonnull MessageContextInteractionEvent event) {}
     public void onButtonInteraction(@Nonnull ButtonInteractionEvent event) {}
     public void onSelectMenuInteraction(@Nonnull SelectMenuInteractionEvent event) {}
```
DV8FromTheWorld, line 143: We might want to consider renaming the original "SelectMenu" to "StringSelectMenu" as that's what it is now considered.
@MinnDevelopment opinion?

DV8FromTheWorld:
We also may want to consider making an abstraction that represents all select types as there is a looooot of duplicate code here now.
Unclear if that'd paint us into a corner, but i think it's something worth considering.

MinnDevelopment:
I think the design I'd go for would be something like this:

```
SelectMenu
|- StringSelectMenu
\- EntitySelectMenu
```

And then have a factory method like `EntitySelectMenu.create(customId, types)` where `types` is an `EnumSet<SelectType>`. The enum would be the currently supported types:

```java
enum SelectType {
  USER, ROLE, CHANNEL
}
```

Naturally, we will have to make a runtime exception if you try to do `EnumSet.of(ROLE, CHANNEL)`, since that combination is not supported.

seailz:
> We also may want to consider making an abstraction that represents all select types as there is a looooot of duplicate code here now.
> 
> Unclear if that'd paint us into a corner, but i think it's something worth considering.

There's something there with the interaction events though, because atm the event just returns a string list of the selected types, which obviously wouldn't work for a channel select menu, since Discord returns a list of channel objects etc.

seailz:
> I think the design I'd go for would be something like this:
> 
> 
> 
> ```
> 
> SelectMenu
> 
> |- StringSelectMenu
> 
> \- EntitySelectMenu
> 
> ```
> 
> 
> 
> And then have a factory method like `EntitySelectMenu.create(customId, types)` where `types` is an `EnumSet<SelectType>`. The enum would be the currently supported types:
> 
> 
> 
> ```java
> 
> enum SelectType {
> 
>   USER, ROLE, CHANNEL
> 
> }
> 
> ```
> 
> 
> 
> Naturally, we will have to make a runtime exception if you try to do `EnumSet.of(ROLE, CHANNEL)`, since that combination is not supported.

This also looks like a great way of going about implementing this, but again how would the event work?

src/main/java/net/dv8tion/jda/api/hooks/ListenerAdapter.java
seailz: Seems like a good idea to me

DV8FromTheWorld:
> There would need to be some kind of new Mentionable class since the existing one also includes channels, which is not the case for selects.
There are channel selects. I don't understand.

seailz:
> > There would need to be some kind of new Mentionable class since the existing one also includes channels, which is not the case for selects.
> 
> 
> 
> 
> 
> There are channel selects. I don't understand.

oh yeah nvm I'm stupid

seailz:
> I think the design I'd go for would be something like this:
> 
> 
> 
> ```
> 
> SelectMenu
> 
> |- StringSelectMenu
> 
> \- EntitySelectMenu
> 
> ```
> 
> 
> 
> And then have a factory method like `EntitySelectMenu.create(customId, types)` where `types` is an `EnumSet<SelectType>`. The enum would be the currently supported types:
> 
> 
> 
> ```java
> 
> enum SelectType {
> 
>   USER, ROLE, CHANNEL
> 
> }
> 
> ```
> 
> 
> 
> Naturally, we will have to make a runtime exception if you try to do `EnumSet.of(ROLE, CHANNEL)`, since that combination is not supported.

Should I implement this then?

Eronary:
image
Even without processing the event, I get this error when the user selects the user

`dependencies {
        implementation("com.github.seailz:JDA:feature~newselecttypes-SNAPSHOT")
}`

freya022:
> I think the design I'd go for would be something like this:
> 
> ```
> SelectMenu
> |- StringSelectMenu
> \- EntitySelectMenu
> ```
> 
> And then have a factory method like `EntitySelectMenu.create(customId, types)` where `types` is an `EnumSet<SelectType>`. The enum would be the currently supported types:
> 
> ```java
> enum SelectType {
>   USER, ROLE, CHANNEL
> }
> ```
> 
> Naturally, we will have to make a runtime exception if you try to do `EnumSet.of(ROLE, CHANNEL)`, since that combination is not supported.

I don't think the `CHANNEL` type should be put here, an `EntitySelectMenu` subclass would be needed for channels, as you can specify channel types, just like in slash command options

> The new select menu components are defined similarly to the existing string select menu—with the exception of not including the options field and, *within channel select menus, having the option to include a channel_types field*.

MinnDevelopment:
> I don't think the CHANNEL type should be put here, an EntitySelectMenu subclass would be needed for channels, as you can specify channel types, just like in slash command options

This is not a problem. We handle it the same way for `OptionData` by providing a setter for any type and throwing at runtime.

seailz:
> image
> 
> Even without processing the event, I get this error when the user selects the user
> 
> 
> 
> `dependencies {
> 
>         implementation("com.github.seailz:JDA:feature~newselecttypes-SNAPSHOT")
> 
> }`

This PR isn't even finished let alone ready for production. You shouldn't be using it in your project.

src/main/java/net/dv8tion/jda/api/interactions/components/selections/EntitySelectMenu.java
```diff
@@ -0,0 +1,469 @@
...
+        @Nonnull
+        public EntitySelectMenu build()
+        {
+            Checks.check(minValues <= maxValues, "Min values cannot be greater than max values!");
+
+            int typeCode;
+
+            if (type.size() == 1)
+                typeCode = type.iterator().next().getKey();
+            else if (type.size() == 3 && type.contains(SelectType.ROLE) && type.contains(SelectType.USER))
```
lolsuffocate, line 461: Should this be `type.size() == 2`? Setting `SelectType.ROLE` and `SelectType.USER` causes this to throw an IllegalArgumentException.

seailz: Yep, thanks for pointing that out. I remembered that the mentionable select menu doesn't include channels about half way through doing this haha.

src/main/java/net/dv8tion/jda/api/interactions/components/selections/EntitySelectMenu.java
```diff
@@ -0,0 +1,469 @@
...
+ * Other users cannot see the choices selected, and they will disappear when the client restarts or the message is reloaded.
+ *
+ * <p>There are two types of select menus within JDA. Entity select menus, which are used for channels, users and roles, and string select menus which are used for custom options.
+ *
+ * <p><b>Examples</b><br>
+ * <pre>{@code
+ * public void onSlashCommandInteraction(SlashCommandInteractionEvent event) {
+ *   if (!event.getName().equals("friend")) return;
+ *
+ *   SelectMenu menu = SelectMenu.create("menu:friend")
```
duncte123, line 46: This documentation seems wrong

duncte123 (commented):
This could be a limitation of discord, but I don't see any way to limit what roles are selectable

seailz:
> This could be a limitation of discord, but I don't see any way to limit what roles are selectable

Yeah, it's a limitation of discord.

MinnDevelopment:
Thank you for your efforts, however I already started my implementation before this PR was opened and would like to finish it there. Feel free to review or PR into #2287, there is still a bit in terms of documentation and testing that is needed for its completion. And I'm sorry for having to close this one, as it is unfortunately is not close enough to what I'm looking for.

seailz:
> Thank you for your efforts, however I already started my implementation before this PR was opened and would like to finish it there. Feel free to review or PR into #2287, there is still a bit in terms of documentation and testing that is needed for its completion. And I'm sorry for having to close this one, as it is unfortunately is not close enough to what I'm looking for.

Would of been nice if you could mention that before...

MinnDevelopment:
@seailz Sorry, I was hoping your implementation would end up as I intended. You were unfortunately too quick at opening this PR and I just didn't really have time to look into this until now.

PR #2292, opened 2022-10-21
Update docs regarding deletion of ephemeral messages

seailz:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Added ability to delete ephemeral messages

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -1390,7 +1390,6 @@ public interface Message extends ISnowflake, Formattable
      *         <ul>
      *              <li>If this Message was not sent by the currently logged in account and it was <b>not</b> sent in a
      *              {@link net.dv8tion.jda.api.entities.channel.middleman.GuildChannel GuildChannel}.</li>
-     *              <li>If this Message is ephemeral</li>
      *              <li>If this message type cannot be deleted. (See {@link MessageType#canDelete()})</li>
      *         </ul>
      *
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -1390,6 +1390,7 @@ public interface Message extends ISnowflake, Formattable
      *         <ul>
      *              <li>If this Message was not sent by the currently logged in account and it was <b>not</b> sent in a
      *              {@link net.dv8tion.jda.api.entities.channel.middleman.GuildChannel GuildChannel}.</li>
+     *              <li>If this Message is ephemeral</li>
      *              <li>If this message type cannot be deleted. (See {@link MessageType#canDelete()})</li>
      *         </ul>
      *
--- a/src/main/java/net/dv8tion/jda/api/interactions/callbacks/IReplyCallback.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/callbacks/IReplyCallback.java
@@ -87,7 +87,6 @@ public interface IReplyCallback extends IDeferrableCallback
      * <br>When a message is ephemeral, it will only be visible to the user that used the interaction.
      * <br>Limitations:
      * <ul>
-     *     <li>Cannot be deleted by the bot</li>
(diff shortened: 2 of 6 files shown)


seailz:
Alright, tested this and it seems to be working.

seailz:
> did you update the documentation for deferReply(true)

Just pushed a new commit that should fix that

jman13378:
> > did you update the documentation for deferReply(true)
> 
> Just pushed a new commit that should fix that

might wanna put the code for the method to delete it

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -579,9 +579,6 @@ private void checkUser()
     @Override
     public AuditableRestAction<Void> delete()
     {
-        if (isEphemeral())
-            throw new IllegalStateException("Cannot delete ephemeral messages.");
```
MinnDevelopment, line 583: This still doesn't work, since this method does not use interactions.

image

This method uses `DELETE /channels/:channel_id/messages/:message_id`.

src/main/java/net/dv8tion/jda/api/interactions/callbacks/IReplyCallback.java
```diff
@@ -87,7 +87,6 @@ public interface IReplyCallback extends IDeferrableCallback
      * <br>When a message is ephemeral, it will only be visible to the user that used the interaction.
      * <br>Limitations:
      * <ul>
-     *     <li>Cannot be deleted by the bot</li>
```
freya022, line 90: This needs to be removed in `InteractionHook#deleteOriginal` too

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1390,7 +1390,6 @@ default MessageCreateAction replyFiles(@Nonnull Collection<? extends FileUpload>
      *         <ul>
      *              <li>If this Message was not sent by the currently logged in account and it was <b>not</b> sent in a
      *              {@link net.dv8tion.jda.api.entities.channel.middleman.GuildChannel GuildChannel}.</li>
-     *              <li>If this Message is ephemeral</li>
```
freya022, line 1393: This needs to be removed in `InteractionHook#deleteOriginal` too

src/main/java/net/dv8tion/jda/api/interactions/callbacks/IReplyCallback.java
freya022, line 90: This needs to be removed in `InteractionHook`, `ReplyCallbackAction` and `WebhookMessageCreateAction`

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
seailz: Wait a sec, why would that check for if it is ephemeral even be there if you can't retrieve ephemeral messages?

MinnDevelopment: You can retrieve them

seailz: Hmm... how do you want me to grab the interaction token then?

MinnDevelopment: Just revert the change. The only needed change is for the docs on `setEphemeral` really.

seailz: Okay

src/main/java/net/dv8tion/jda/api/entities/Message.java
MinnDevelopment, line 1393: Can you add this back?

seailz: > Can you add this back?

done

Issue #2294, opened 2022-10-22
Move Widgets into the .entities packages instead of nesting in WidgetUtil


DV8FromTheWorld:
### General Troubleshooting

- [X] I have checked for similar issues on the Issue-tracker.
- [X] I have updated to the [latest JDA version](https://ci.dv8tion.net/job/JDA/lastSuccessfulBuild/)
- [X] I have checked the branches or the maintainers' PRs for upcoming features.

### Feature Request

Based on conversations, there is an opportunity for a small reorganization of code. 
Context: https://github.com/DV8FromTheWorld/JDA/pull/2273#discussion_r996193457

TLDR: We want to move the `Widget` related classes out of WidgetUtil and instead have them in the `.entities` folder.
Specifically, promote the nested `Widget` class to its own file. The other Widget-subclasses should remain subclasses of Widget.
PRs welcome

RealYusufIsmail:
Will hopefully make a pr for this

MinnDevelopment:
Done in [5.0.0-beta.1](https://github.com/DV8FromTheWorld/JDA/releases/tag/v5.0.0-beta.1)

PR #2295, opened 2022-10-22
Make Widget an interface and move it to it's own file

Almighty-Satan:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #2294

## Description

Moves `Widget` to it's own file.


src/main/java/net/dv8tion/jda/api/entities/Widget.java
```diff
@@ -0,0 +1,806 @@
...
+    private final long id;
+    private final String name;
+    private final String invite;
+    private final TLongObjectMap<VoiceChannel> channels;
+    private final TLongObjectMap<Member> members;
+    
+    /**
+     * Constructs an unavailable Widget
+     */
+    public Widget(long guildId)
```
Sanduhr32, line 46: The constructors probably should not be exposed

src/main/java/net/dv8tion/jda/api/entities/Widget.java
```diff
@@ -0,0 +1,806 @@
...
+        members = new TLongObjectHashMap<>();
+    }
+    
+    /**
+     * Constructs an available Widget
+     *
+     * @param json
+     *        The {@link net.dv8tion.jda.api.utils.data.DataObject DataObject} to construct the Widget from
+     */
+    public Widget(@Nonnull DataObject json)
```
Sanduhr32, line 62: also a constructor

src/main/java/net/dv8tion/jda/api/entities/Widget.java
```diff
@@ -0,0 +1,806 @@
...
+        return "W:" + (isAvailable() ? getName() : "") + '(' + id + ')';
+    }
+
+    private void checkAvailable()
+    {
+        if (!isAvailable)
+            throw new IllegalStateException("The widget for this Guild is unavailable!");
+    }
+
+    public static class Member implements IMentionable
```
Sanduhr32, line 291: nested classes could be lifted out as their own classes

src/main/java/net/dv8tion/jda/api/entities/Widget.java
```diff
@@ -0,0 +1,806 @@
...
+import gnu.trove.map.hash.TLongObjectHashMap;
+import net.dv8tion.jda.api.OnlineStatus;
+import net.dv8tion.jda.api.utils.ImageProxy;
+import net.dv8tion.jda.api.utils.MiscUtil;
+import net.dv8tion.jda.api.utils.data.DataArray;
+import net.dv8tion.jda.api.utils.data.DataObject;
+import net.dv8tion.jda.internal.entities.EntityBuilder;
+
+public class Widget implements ISnowflake
+{
```
Sanduhr32, line 35: why are the static patterns for the html iframe and images still in the util class?
they should be members of the widget class!

src/main/java/net/dv8tion/jda/api/entities/Widget.java
```diff
@@ -0,0 +1,806 @@
...
+    private final boolean isAvailable;
+    private final long id;
+    private final String name;
+    private final String invite;
+    private final TLongObjectMap<VoiceChannel> channels;
+    private final TLongObjectMap<Member> members;
+    
+    /**
```
Sanduhr32, line 43: WidgetUtilgetWidget() probably shoud not build an OKHTTP Client etc each call.

Sanduhr32 (commented):
Hmm, i have these thoughts

src/main/java/net/dv8tion/jda/api/entities/Widget.java
DV8FromTheWorld: The nested classes should not be lifted out as their own.
They are attributes of Widget, just like `Invite.Channel` and `Invite.Guild`

src/main/java/net/dv8tion/jda/api/entities/Widget.java
DV8FromTheWorld: We might want to consider making Widget an interface and moving the underlying implementation into the `internal` classes via `WidgetImpl`

src/main/java/net/dv8tion/jda/api/entities/Widget.java
Almighty-Satan: Because they are not even used by any methods in the Widget class.

src/main/java/net/dv8tion/jda/internal/entities/WidgetImpl.java
```diff
@@ -0,0 +1,601 @@
...
+        return entityString.toString();
+    }
+
+    private void checkAvailable()
+    {
+        if (!isAvailable)
+            throw new IllegalStateException("The widget for this Guild is unavailable!");
+    }
+
+    public static class MemberImpl implements Member
```
DV8FromTheWorld, line 219: Like with the interfaces, I don't think these need to be static.

src/main/java/net/dv8tion/jda/internal/entities/WidgetImpl.java
```diff
@@ -0,0 +1,601 @@
...
+            if (!(obj instanceof Member))
+                return false;
+            Member oMember = (Member) obj;
+            return this == oMember || (this.id == oMember.getIdLong() && this.widget.getIdLong() == oMember.getWidget().getIdLong());
+        }
+
+        @Override
+        public String toString()
+        {
+            return new EntityString("Widget.Member")
```
DV8FromTheWorld, line 400: I don't think we have to pass explicit strings here anymore.

DV8FromTheWorld: or for any of these anymore

DV8FromTheWorld (approved):
Ship it

Issue #2291, opened 2022-10-19
guild.getMembersWithRoles(...) returns empty list for @everyone


TheRisenPhoenix:
### General Troubleshooting

- [X] I have checked for similar issues on the Issue-tracker.
- [X] I have updated to the [latest JDA version](https://ci.dv8tion.net/job/JDA/lastSuccessfulBuild/)
- [X] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

### Expected Behaviour

When using slash commands with a Role-Option, one can mention _@ everyone_ , but not _@ here_ , as this is no valid role and discords checks will reject that mention.
However, if I later want to receive all members that are affected by the mentioned role, using `guild.getMembersWithRoles` , I expect to receive a list of all members if _@ everyone_ is mentioned, as it is, apparently, a valid role-mention (in contrast to _@ here_).

### Code Example for Reproduction Steps

```java
....
slashCommandData.addOptions(new OptionData(OptionType.ROLE, "role", "Mention a role",false));
....

public void onSlashCommandInteraction(SlashCommandInteractionEvent event) {
  var role = event.getOption("role").getAsRole();
  var members = guild.getMembersWithRoles(role);
}
```

### Code for JDABuilder or DefaultShardManagerBuilder used

```java
JDABuilder.createDefault(token)
          .enableIntents(GatewayIntent.GUILD_MEMBERS)
          .setMemberCachePolicy(MemberCachePolicy.ALL)
          .setChunkingFilter(ChunkingFilter.ALL)
          .addEventListeners(new MessageHandler())
          .build();
```

### Exception or Error

```yesyes
guild.getMembersWithRole returns an empty list for @everyone mentions.
```

PR #2297, opened 2022-10-23
Fix MemberCacheViewImpl#getElementsWithRoles and Guild#findMembersWithRoles ignoring the public role

Almighty-Satan:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #2291

## Description

Fixes an issue that caused `MemberCacheViewImpl#getElementsWithRoles` to return an empty list when given the public role.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -79,6 +79,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 /**
  * Represents a Discord {@link net.dv8tion.jda.api.entities.Guild Guild}.
@@ -2508,13 +2509,17 @@ public interface Guild extends IGuildChannelContainer, ISnowflake
         for (Role role : roles)
             Checks.check(this.equals(role.getGuild()), "All roles must be from the same guild!");
 
-        if (isLoaded() || roles.isEmpty() || roles.contains(getPublicRole())) // Member#getRoles never contains the public role
+        if (isLoaded())
         {
             CompletableFuture<List<Member>> future = CompletableFuture.completedFuture(getMembersWithRoles(roles));
             return new GatewayTask<>(future, () -> {});
         }
 
-        return findMembers(member -> member.getRoles().containsAll(roles));
+        List<Role> rolesWithoutPublicRole = roles.stream().filter(role -> !role.isPublicRole()).collect(Collectors.toList());
+        if (rolesWithoutPublicRole.isEmpty())
+            return loadMembers();
+
+        return findMembers(member -> member.getRoles().containsAll(rolesWithoutPublicRole));
     }
 
     /**
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -1234,7 +1234,7 @@ public interface Guild extends IGuildChannelContainer, ISnowflake
     @Nonnull
     default List<Member> getMembersWithRoles(@Nonnull Role... roles)
     {
-        Checks.noneNull(roles, "Roles");
+        Checks.notNull(roles, "Roles");
         return getMembersWithRoles(Arrays.asList(roles));
     }
 
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -1258,6 +1258,7 @@ public interface Guild extends IGuildChannelContainer, ISnowflake
     @Nonnull
     default List<Member> getMembersWithRoles(@Nonnull Collection<Role> roles)
     {
+        Checks.noneNull(roles, "Roles");
         for (Role role : roles)
             Checks.check(this.equals(role.getGuild()), "All roles must be from the same guild!");
         return getMemberCache().getElementsWithRoles(roles);
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -1234,6 +1234,7 @@ public interface Guild extends IGuildChannelContainer, ISnowflake
     @Nonnull
     default List<Member> getMembersWithRoles(@Nonnull Role... roles)
     {
+        Checks.noneNull(roles, "Roles");
         return getMembersWithRoles(Arrays.asList(roles));
     }
 
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -1234,7 +1234,7 @@ public interface Guild extends IGuildChannelContainer, ISnowflake
     @Nonnull
     default List<Member> getMembersWithRoles(@Nonnull Role... roles)
     {
-        return getMemberCache().getElementsWithRoles(roles);
+        return getMembersWithRoles(Arrays.asList(roles));
     }
 
     /**
@@ -1258,6 +1258,8 @@ public interface Guild extends IGuildChannelContainer, ISnowflake
     @Nonnull
     default List<Member> getMembersWithRoles(@Nonnull Collection<Role> roles)
     {
+        for (Role role : roles)
+            Checks.check(this.equals(role.getGuild()), "All roles must be from the same guild!");
         return getMemberCache().getElementsWithRoles(roles);
     }
 
--- a/src/main/java/net/dv8tion/jda/internal/utils/cache/MemberCacheViewImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/utils/cache/MemberCacheViewImpl.java
@@ -24,6 +24,7 @@ import net.dv8tion.jda.internal.utils.Checks;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.*;
+import java.util.stream.Collectors;
 
 public class MemberCacheViewImpl extends SnowflakeCacheViewImpl<Member> implements MemberCacheView
 {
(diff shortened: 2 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/utils/cache/MemberCacheViewImpl.java
```diff
@@ -96,8 +96,10 @@ public List<Member> getElementsWithRoles(@Nonnull Collection<Role> roles)
         List<Member> members = new ArrayList<>();
         forEach(member ->
         {
-            if (member.getRoles().containsAll(roles))
-                members.add(member);
+            for (Role role : roles)
+                if (!role.isPublicRole()  && !member.getRoles().contains(role))
+                    return;
+            members.add(member);
```
DManstrator, line 102: The code seems a bit over complicated imo. I think technically it's correct (process next member if one role is neither the pulic role nor does the member have it) but you need to think about it for a second.

A simplier solution imo would be removing the public role from `roles` (in case it is present). Then the rest of the code still does what is supposed to be.

DV8FromTheWorld: I think, like you mentioned, this might work better:
```java
roles = roles.filter(role -> !role.isPublicRole());
forEach(member -> 
{
  if (member.getRoles().containsAll(roles))
    members.add(member);
})
```
That being said, we need to confirm that `containsAll` returns true when provided an empty list.

DManstrator: > roles = roles.filter(role -> !role.isPublicRole());

Should be `roles.removeIf` and perhaps be a new list so the original passed list doesn't get modified. Or yeah, List -> Stream -> List. Perhaps the latter is better since it could be a one liner since the performance impact shouldn't be high.

> That being said, we need to confirm that containsAll returns true when provided an empty list.

Tested it right now, it does (as expected).

DV8FromTheWorld: don't mutate the passed in list.
Otherwise, go with this approach.

DManstrator: > Edit: For improved performance, you could also check if only the public role (of this guild) is in the list and then return all members directly.

What about this? It might not be needed but it a) shows that the public role is something special and should improve the performance by a lot. From what I saw `elements.valueCollection()` should provide that, shouldn't it?

DV8FromTheWorld: If the only thing in the list is the public role then yeah, you could skip the check.

So I'd remove the public role and see if the length is 0. If it is, return all members. Otherwise, iterate and check

src/main/java/net/dv8tion/jda/internal/utils/cache/MemberCacheViewImpl.java
```diff
@@ -93,10 +94,13 @@ public List<Member> getElementsWithRoles(@Nonnull Collection<Role> roles)
         Checks.noneNull(roles, "Roles");
         if (isEmpty())
             return Collections.emptyList();
+        List<Role> rolesWithoutPublicRole = roles.stream().filter(role -> !role.isPublicRole()).collect(Collectors.toList());
+        if (rolesWithoutPublicRole.isEmpty())
+            return asList();
         List<Member> members = new ArrayList<>();
```
DManstrator, line 100: Looks much better now, thanks.

One nitty thing: I would perfer some new lines in between for a better readability.

```java
if (isEmpty())
    return Collections.emptyList()

List<Role> rolesWithoutPublicRole = roles.stream().filter(role -> !role.isPublicRole()).collect(Collectors.toList());
if (rolesWithoutPublicRole.isEmpty())
    return asList();

List<Member> members = new ArrayList<>();
```

But perhaps that's just me.

DManstrator (commented):
Another thing:

The JavaDocs of [`Guild#getMembersWithRoles`](https://github.com/DV8FromTheWorld/JDA/blob/5d13c14b4b14f776b5e93a148c7749ff3bbdcaa3/src/main/java/net/dv8tion/jda/api/entities/Guild.java#L1250-L1251) states it will be checked if roles are from a different guild. But according to the impl, this is not the case. Could you add that check?

Only [`Guild#findMembersWithRoles`](https://github.com/DV8FromTheWorld/JDA/blob/5d13c14b4b14f776b5e93a148c7749ff3bbdcaa3/src/main/java/net/dv8tion/jda/api/entities/Guild.java#L2508-L2509) checks for that yet.

DManstrator:
One another thing:

I'm not sure if it is a good idea to call [`getMembersWithRoles` in case the public role is present](https://github.com/DV8FromTheWorld/JDA/blob/5d13c14b4b14f776b5e93a148c7749ff3bbdcaa3/src/main/java/net/dv8tion/jda/api/entities/Guild.java#L2511). If it's loaded or empty, it makes sense to me. I don't know enough to be sure that it won't have any negative effect to not make a request in such a situation.

Can somebody confirm it's okay?

Almighty-Satan:
> One another thing:
> 
> I'm not sure if it is a good idea to call [`getMembersWithRoles` in case the public role is present](https://github.com/DV8FromTheWorld/JDA/blob/5d13c14b4b14f776b5e93a148c7749ff3bbdcaa3/src/main/java/net/dv8tion/jda/api/entities/Guild.java#L2511). If it's loaded or empty, it makes sense to me. I don't know enough to be sure that it won't have any negative effect to not make a request in such a situation.
> 
> Can somebody confirm it's okay?

That methods seems to be broken too. `isLoaded()` makes sense but the other two don't

DManstrator:
> That method seems to be broken too. isLoaded() makes sense but the other two don't
I think you are right. I thought doing it for `isEmpty` is now also valid since you added a "roles is empty? return all members" check for `getMembersWithRoles`. But yeah, what if the locally cached member list is outdated? Then wrong members might be returned.

So I think in the end, the same changes you made need to be applied for `findMembersWithRoles` too. And regarding performance, if the list is empty (after removing the everyone role), `loadMembers()` could be returned.

Sadly, this is not really a good performance improvement since this will still iterate each member ([`loadMembers(Consumer<Member>)` is used under the hood which returns a `Task<Void>`](https://github.com/DV8FromTheWorld/JDA/blob/5d13c14b4b14f776b5e93a148c7749ff3bbdcaa3/src/main/java/net/dv8tion/jda/api/entities/Guild.java#L2570-L2571)).

Perhaps in a different MR, this could be adjusted but it's definitely not in the scope of this MR.

DManstrator:
Re commit: 2480ef9b137cb602d5d613c648df0c3a4725cefe

Couldn't you add that role check to `MemberCacheViewImpl` instead? The Varargs -> List transformation is also done there.

DManstrator:
True, didn't saw that, thanks.

Spealing about `Guild#findMembersWithRoles`: Could you add the `Checks.noneNull(roles, "Roles");` for consistency too?

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -1234,6 +1234,7 @@ default List<Member> getMembersByEffectiveName(@Nonnull String name, boolean ign
     @Nonnull
     default List<Member> getMembersWithRoles(@Nonnull Role... roles)
     {
+        Checks.noneNull(roles, "Roles");
```
DManstrator, line 1237: You actually don't need it since you can't pass `null` in here.

The compiler wouldn't know if `getMembersWithRoles(Role...)` or `getMembersWithRoles(List<Role>)` is meant so it won't compile. And if you pass nothing, it will be an empty array.

But I also had to think about that for a second. Because if it would be possible, it would be needed.

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -1234,7 +1234,7 @@ default List<Member> getMembersByEffectiveName(@Nonnull String name, boolean ign
     @Nonnull
     default List<Member> getMembersWithRoles(@Nonnull Role... roles)
     {
-        Checks.noneNull(roles, "Roles");
+        Checks.notNull(roles, "Roles");
         return getMembersWithRoles(Arrays.asList(roles));
     }
```
DManstrator, line 1239: No, you got me wrong. As said before, it can't be null due to two methods having the same name and same number of parameters. The compiler would simply not allow it without explicit casting. And if somebody does that, it's more or less their fault imo.

And no need to check for `nonNull` again since you already do that in `getMembersWithRoles(List<Role>)`.

Almighty-Satan: That method has to check `notNull` otherwise something like `Role[] roles = null; guild.getMembersWithRoles(roles);` would cause `Arrays#asList` to throw an NPE

> And no need to check for `noneNull` again since you already do that in `getMembersWithRoles(List<Role>)`

Yes

DManstrator: > That method has to check notNull otherwise something like Role[] roles = null; guild.getMembersWithRoles(roles); would cause Arrays#asList to throw an NPE

Yes and no. As said, you would explicitly need to cast it / define it like that so it compiles. And as a user you should see this is simply wrong. So imo it's fine to let the user run into this. Because essentially you just trade a `NullPointerException` with an `IllegalArgumentException`. Doesn't help much but needs more code. So I would need to remove it.

From what I see it's also not used in other places and consistency is more important. So imo, it's really fine to remove that check again.

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -2508,13 +2513,17 @@ default Task<List<Member>> findMembersWithRoles(@Nonnull Collection<Role> roles)
...
+        if (isLoaded())
         {
             CompletableFuture<List<Member>> future = CompletableFuture.completedFuture(getMembersWithRoles(roles));
             return new GatewayTask<>(future, () -> {});
         }
 
-        return findMembers(member -> member.getRoles().containsAll(roles));
+        List<Role> rolesWithoutPublicRole = roles.stream().filter(role -> !role.isPublicRole()).collect(Collectors.toList());
+        if (rolesWithoutPublicRole.isEmpty())
+            return loadMembers();
```
DV8FromTheWorld, line 2524: Originally this was returning a completed GatewayTask. Why is this doing loadMembers instead?

src/main/java/net/dv8tion/jda/internal/utils/cache/MemberCacheViewImpl.java
DV8FromTheWorld: agree

src/main/java/net/dv8tion/jda/api/entities/Guild.java
Almighty-Satan: When using `MemberCachePolicy.NONE` `findMembersWithRoles(publicRole)` would previously just return an empty Collection instead of a Collection that contains all Members

see https://github.com/DV8FromTheWorld/JDA/pull/2297#issuecomment-1288691702 and https://github.com/DV8FromTheWorld/JDA/pull/2297#issuecomment-1288754380

DV8FromTheWorld (approved):

PR #2307, opened 2022-10-25
Add rate-limiter customization

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #1874 (keeping it open until feature parity)

## Description

This will allow users to replace the rate-limiter implementation with a custom one. This is useful for something like an [http-proxy](https://github.com/twilight-rs/http-proxy) or to implement parallel handling using a semaphore.

- [x] Introduce new interfaces
- [x] Allow configuration through builders
- [x] Allow replacing endpoint and user-agent
- [x] Move Route into api package
- [x] Documentation


src/main/java/net/dv8tion/jda/api/requests/RestConfig.java
```diff
@@ -0,0 +1,220 @@
...
+     *
+     * @return The current RestConfig for chaining convenience
+     */
+    @Nonnull
+    public RestConfig setUserAgentSuffix(@Nonnull String suffix)
+    {
+        Checks.notNull(suffix, "Suffix");
+        if (Helpers.isBlank(suffix))
+            this.userAgent = USER_AGENT;
+        else
```
DV8FromTheWorld, line 137: Should this instead allow `null` instead of checking explicitly for `""`?

src/main/java/net/dv8tion/jda/api/requests/RestRateLimiter.java
```diff
@@ -0,0 +1,243 @@
...
+        @Nonnull
+        Route.CompiledRoute getRoute();
+
+        /**
+         * The JDA instance which started the request.
+         *
+         * @return The JDA instance
+         */
+        @Nonnull
+        JDA getJDA();
```
DV8FromTheWorld, line 117: Would be an interesting concept to try and consider separating the requester system from relying on the JDA object to better facilitate RestOnly JDA in the future (and allow us to use the requester system in situations where we don't have the obj)

src/main/java/net/dv8tion/jda/api/requests/Route.java
```diff
@@ -433,10 +387,7 @@ public boolean equals(Object o)
     @Override
     public String toString()
     {
-        return new EntityString(this)
-                .setType(method)
-                .addMetadata("route", route)
-                .toString();
+        return method + "/" + route;
```
DV8FromTheWorld, line 390: Looks like a regression. Likely from before the EntityString PR as merged.

src/main/java/net/dv8tion/jda/api/requests/SequentialRestRateLimiter.java
```diff
@@ -14,68 +14,67 @@
...
-import net.dv8tion.jda.api.requests.Request;
 import net.dv8tion.jda.api.utils.MiscUtil;
-import net.dv8tion.jda.internal.requests.RateLimiter;
-import net.dv8tion.jda.internal.requests.Requester;
-import net.dv8tion.jda.internal.requests.Route;
+import net.dv8tion.jda.internal.utils.JDALogger;
 import okhttp3.Headers;
+import okhttp3.Response;
 import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
```
DV8FromTheWorld, line 24: Wrong annotation

src/main/java/net/dv8tion/jda/internal/requests/Requester.java
```diff
@@ -264,17 +275,12 @@ private void applyBody(Request<?> apiRequest, okhttp3.Request.Builder builder)
...
-        builder.header("user-agent", USER_AGENT)
+        builder.header("user-agent", userAgent)
                .header("accept-encoding", "gzip")
                .header("x-ratelimit-precision", "millisecond"); // still sending this in case of regressions
 
-        //adding token to all requests to the discord api or cdn pages
-        //we can check for startsWith(DISCORD_API_PREFIX) because the cdn endpoints don't need any kind of authorization
-        if (authorized)
-            builder.header("authorization", authConfig.getToken());
-
```
DV8FromTheWorld, line 277: Am I missing something? Where did this go?

src/main/java/net/dv8tion/jda/internal/requests/Requester.java
```diff
@@ -299,18 +305,87 @@ public void setRetryOnTimeout(boolean retryOnTimeout)
         this.retryOnTimeout = retryOnTimeout;
     }
 
-    public boolean stop()
+    public void stop(boolean shutdowm, Runnable callback)
```
DV8FromTheWorld, line 308: `shutdowm` -> `shutdown`

src/main/java/net/dv8tion/jda/api/requests/Route.java
MinnDevelopment: This change is intentional, the toString is used for logging and I'd like to keep that one simple.

src/main/java/net/dv8tion/jda/internal/requests/Requester.java
MinnDevelopment: We generally don't use the requester with CDN URLs, which makes this kind of useless. Pretty sure `authorized` is always true.

src/main/java/net/dv8tion/jda/api/requests/RestRateLimiter.java
MinnDevelopment: Eventually that would be ideal, but that is a non-goal of this PR.

src/main/java/net/dv8tion/jda/api/requests/Route.java
DV8FromTheWorld: `Route[Put](route=applications/{application_id}/commands)`
`Put/applications/{application_id}/commands`

Between these two, i definitely think the top one is better.

src/main/java/net/dv8tion/jda/internal/requests/Requester.java
```diff
@@ -264,17 +275,13 @@ private void applyBody(Request<?> apiRequest, okhttp3.Request.Builder builder)
         builder.method(method, body);
     }
 
-    private void applyHeaders(Request<?> apiRequest, okhttp3.Request.Builder builder, boolean authorized)
+    private void applyHeaders(Request<?> apiRequest, okhttp3.Request.Builder builder)
     {
-        builder.header("user-agent", USER_AGENT)
+        builder.header("user-agent", userAgent)
                .header("accept-encoding", "gzip")
+               .header("authorization", authConfig.getToken())
```
DV8FromTheWorld, line 282: This actually makes me nervous. I don't like the idea of accidentally forgetting this for the requester and sending the token to non-discord.
Could we instead check to see if the request is using the `RestConfig.baseUrl` to attach this?

MinnDevelopment: No, since proxies use it too. Making requests to untrusted servers isn't really something we can prevent.

MineKing9534:
I currently get the following error:
```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.util.concurrent.ScheduledExecutorService.scheduleAtFixedRate(java.lang.Runnable, long, long, java.util.concurrent.TimeUnit)" because the return value of "net.dv8tion.jda.api.requests.RestRateLimiter$RateLimitConfig.getPool()" is null
    at net.dv8tion.jda.api.requests.SequentialRestRateLimiter.<init>(SequentialRestRateLimiter.java:92)
    at net.dv8tion.jda.internal.JDAImpl.<init>(JDAImpl.java:165)
    at net.dv8tion.jda.api.JDABuilder.build(JDABuilder.java:1806)
```

MinnDevelopment:
@MineKing9534 should be fixed

MineKing9534:
Yes, working again, thanks

src/main/java/net/dv8tion/jda/api/requests/Route.java
```diff
@@ -14,52 +14,40 @@
...
-        // Bot only
         public static final Route GET_BOT_APPLICATION =             new Route(GET, "oauth2/applications/@me");
         public static final Route GET_ROLE_CONNECTION_METADATA =    new Route(GET, "applications/{application_id}/role-connections/metadata");
         public static final Route UPDATE_ROLE_CONNECTION_METADATA = new Route(PUT, "applications/{application_id}/role-connections/metadata");
-
-        // Client only
-        public static final Route GET_APPLICATIONS =              new Route(GET,    "oauth2/applications");
-        public static final Route CREATE_APPLICATION =            new Route(POST,   "oauth2/applications");
-        public static final Route GET_APPLICATION =               new Route(GET,    "oauth2/applications/{application_id}");
-        public static final Route MODIFY_APPLICATION =            new Route(PUT,    "oauth2/applications/{application_id}");
```
DV8FromTheWorld, line 52: I don't think these are client only anymore as we've opened up the ability of an application to modify itself.
Not sure if that's documented yet, but it's coming.

src/main/java/net/dv8tion/jda/api/requests/Route.java
```diff
@@ -329,100 +315,317 @@ public static Route custom(@Nonnull Method method, @Nonnull String route)
...
+        "guild_id", "channel_id", "webhook_id", "interaction_token"
+    );
+
     private final Method method;
     private final int paramCount;
+    private final String[] template;
+    private final boolean interaction;
 
-    private Route(Method method, String route)
+    private Route(Method method, String route, boolean interaction)
```
DV8FromTheWorld, line 514: isInteraction?

src/main/java/net/dv8tion/jda/api/requests/Route.java
```diff
@@ -329,100 +315,317 @@ public static Route custom(@Nonnull Method method, @Nonnull String route)
...
-            int paramEnd = compiledRoute.indexOf("}");
-            String paramName = compiledRoute.substring(paramStart+1, paramEnd);
-            if (majorParameters.contains(paramName))
+            if (element.charAt(0) == '{')
             {
-                if (params[i].length() > 30) // probably a long interaction_token, hash it to keep logs clean (not useful anyway)
-                    major.add(paramName + "=" + Integer.toUnsignedString(params[i].hashCode()));
-                else
-                    major.add(paramName + "=" + params[i]);
+                String name = element.substring(1, element.length() - 1);
```
DV8FromTheWorld, line 605: Can we do a check somewhere, like an assert, to ensure there _is_ a closing `}`?

DV8FromTheWorld: Since the routes system is getting pulled into API, the Custom routes are more likely to get touched, so checking for mismatched `{`/`}` would be nice

src/main/java/net/dv8tion/jda/api/requests/Route.java
```diff
@@ -329,100 +315,317 @@ public static Route custom(@Nonnull Method method, @Nonnull String route)
...
 
-        return new CompiledRoute(this, compiledRoute.toString(), major.isEmpty() ? "n/a" : String.join(":", major));
+        return new CompiledRoute(this, compiledRoute.toString(), major.toString());
     }
 
     @Override
     public int hashCode()
     {
-        return (route + method.toString()).hashCode();
+        return 31 * Arrays.hashCode(template) + method.hashCode();
```
DV8FromTheWorld, line 628: Can we not use some combination of `Objects.hashCode` or `Objects.hash` here? direct usage of `31 *` feels weird.

src/main/java/net/dv8tion/jda/api/requests/Route.java
```diff
@@ -432,68 +635,146 @@ public boolean equals(Object o)
...
     }
 
     @Override
     public String toString()
     {
-        return new EntityString(this)
-                .setType(method)
-                .addMetadata("route", route)
-                .toString();
+        return method + "/" + getRoute();
```
DV8FromTheWorld, line 644: I still think this reads better as an EntityString toString.

src/main/java/net/dv8tion/jda/api/requests/Route.java
```diff
@@ -432,68 +635,146 @@ public boolean equals(Object o)
...
+         * @return A copy of this CompiledRoute with the provided parameters added as query
+         */
         @Nonnull
         @CheckReturnValue
-        public CompiledRoute withQueryParams(String... params)
+        public CompiledRoute withQueryParams(@Nonnull String... params)
         {
-            Checks.check(params.length >= 2, "params length must be at least 2");
-            Checks.check(params.length % 2 == 0, "params length must be a multiple of 2");
+            Checks.notNull(params, "Params");
```
DV8FromTheWorld, line 712: We should make sure no keys are empty strings

src/main/java/net/dv8tion/jda/api/requests/Route.java
```diff
@@ -432,68 +635,146 @@ public boolean equals(Object o)
...
+            {
+                newQuery = new ArrayList<>(query.size() + params.length / 2);
+                newQuery.addAll(query);
+            }
 
-            for (int i = 0; i < params.length; i++)
-                newRoute.append(!hasQueryParams && i == 0 ? '?' : '&').append(params[i]).append('=').append(params[++i]);
+            // Assuming names don't need encoding
+            for (int i = 0; i < params.length; i += 2)
+                newQuery.add(params[i] + '=' + EncodingUtil.encodeUTF8(params[i + 1]));
```
DV8FromTheWorld, line 729: As a note, query strings with no value like `?use_flag` are a thing. We might want to check if the provided value for a key is `""` and not add the `=` for it.

DV8FromTheWorld: Can also cross that bridge when we come to it, but thought I'd mention it

src/main/java/net/dv8tion/jda/api/utils/SessionController.java
```diff
@@ -274,4 +302,68 @@ public int getConcurrency()
...
     }
+
+    /**
+     * Wrapper for {@link #getGlobalRatelimit()} and {@link #setGlobalRatelimit(long)}.
+     */
+    @Deprecated
+    @ForRemoval(deadline = "5.0.0")
+    @ReplaceWith("getRateLimitHandle()")
+    @SuppressWarnings("DeprecatedIsStillUsed")
+    class GlobalRateLimitAdapter implements RestRateLimiter.GlobalRateLimit
```
DV8FromTheWorld, line 313: Why is this set for deprecation and removal and setup with a ReplaceWith when it is used in the thing that is indicated to replace with? I'm confused.

DV8FromTheWorld: Is the hope here that developers implement their own GlobalRateLimit implementation to return from SessionController?

DV8FromTheWorld: I don't quite understand what this is for.

src/main/java/net/dv8tion/jda/internal/requests/Requester.java
```diff
@@ -139,29 +136,35 @@ public Long execute(Request<?> apiRequest)
...
+    public okhttp3.Response execute(WorkTask task, boolean retried, boolean handleOnRatelimit)
     {
-        Route.CompiledRoute route = apiRequest.getRoute();
-        Long retryAfter = rateLimiter.getRateLimit(route);
-        if (retryAfter != null && retryAfter > 0)
-        {
-            if (handleOnRatelimit)
-                apiRequest.handleResponse(new Response(retryAfter, Collections.emptySet()));
-            return retryAfter;
-        }
```
DV8FromTheWorld, line 156: Where did this code go? do we not need it anymore? I get that ratelimiting no longer lives in this file, so is there something that is coming before the call to `execute` in the ratelimiter?

DV8FromTheWorld: Im guessing yes because we pass out lastResponse instead of a ratelimit retry time

src/main/java/net/dv8tion/jda/internal/utils/Checks.java
```diff
@@ -38,7 +38,7 @@
 public class Checks
 {
     public static final Pattern ALPHANUMERIC_WITH_DASH = Pattern.compile("[\\w-]+", Pattern.UNICODE_CHARACTER_CLASS);
-    public static final Pattern ALPHANUMERIC = Pattern.compile("[\\w]+", Pattern.UNICODE_CHARACTER_CLASS);
+    public static final Pattern ALPHANUMERIC = Pattern.compile("\\w+", Pattern.UNICODE_CHARACTER_CLASS);
```
DV8FromTheWorld, line 41: Why the change?

src/main/java/net/dv8tion/jda/api/requests/Route.java
MinnDevelopment: Usually It's valid to do something like `?use_flag=` with no value attached as well.

src/main/java/net/dv8tion/jda/internal/utils/Checks.java
MinnDevelopment: It was redundant and gave annoying inspection warnings

src/main/java/net/dv8tion/jda/internal/requests/Requester.java
MinnDevelopment: The logic here has changed, we now simply check if `work.isDone()` in the rate-limiter to handle the 429. Essentially, whenever we complete a request, we mark the task as done and then the rate-limiter won't retry it.

Before, we used to explicitly indicate to retry using the retry-after return, which limits the capabilities and was a little confusing. Now, we simplified this to a boolean check in the rate-limiter:

```java
Response response = request.execute();
if (response != null)
    updateBucket(request.getRoute(), response);
if (!request.isDone())
    retry(request);
```

src/main/java/net/dv8tion/jda/api/utils/SessionController.java
MinnDevelopment: This is for backwards compatibility. We previously handled global rate-limits using `get/setGlobalRatelimit` on the controller, which is no longer used by JDA. The reason this adapter was introduced, is to allow existing implementations to continue working for now. However, as the annotations imply, we will remove both those getters/setters and the adapter in the full release.

Most people won't be affected by this since our default implementation is already updated.

src/main/java/net/dv8tion/jda/api/requests/Route.java
MinnDevelopment: It does the exact same but sure

src/main/java/net/dv8tion/jda/api/requests/Route.java
MinnDevelopment: This is already checked in the route constructor.

```
Checks.check(paramCount == Helpers.countMatches(route, '}'), "An argument does not have both {}'s for route: %s %s", method, route);
```

src/main/java/net/dv8tion/jda/api/exceptions/InteractionExpiredException.java
```diff
@@ -14,8 +14,18 @@
...
+ * <br>This is used for follow-up requests sent via {@link InteractionHook}, which expire after 15 minutes.
  */
-package net.dv8tion.jda.internal.requests.ratelimit;
+public class InteractionExpiredException extends RuntimeException
+{
+    public InteractionExpiredException()
+    {
+        super("The interaction has expired and can no longer be responded to!");
+    }
+}
```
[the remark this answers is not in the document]
MinnDevelopment: Likely because the majority of both files is copyright

src/main/java/net/dv8tion/jda/api/requests/Route.java
MinnDevelopment: These are specifically to get other applications by ID. I'd expect an application modifying itself would use `@me` instead of an id.

src/main/java/net/dv8tion/jda/api/requests/Route.java
MinnDevelopment: This is checked in the route constructor already, no need to check it on every compile

DV8FromTheWorld (approved):

PR #2323, opened 2022-11-05
Moved Widgets related classes into the .entities packages instead of nesting in WidgetUtil

RealYusufIsmail:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes : #2294

## Description

Moved Widget related classes instead of them being in WidgetUtil

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Widget.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Widget.java
@@ -447,4 +447,31 @@ public interface Widget extends ISnowflake
         @Nonnull
         Widget getWidget();
     }
+
+
+    /**
+     * Represents the color scheme of the widget
+     * <br>These color themes match Discord's dark and light themes
+     */
+    public enum WidgetTheme
+    {
+        LIGHT, DARK
+    }
+
+
+    /**
+     * Represents the available banner types
+     * <br>Each of these has a different appearance:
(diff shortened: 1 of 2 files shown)


Almighty-Satan:
I don't think these enums should be moved as they are only used by methods in the `WidgetUtil` class.

freya022:
> I don't think these enums should be moved as they are only used by methods in the `WidgetUtil` class.

They will be used in user code though, so it makes sense for the user to see `Widget.WidgetTheme` instead of `WidgetUtil.WidgetTheme`

Almighty-Satan:
> They will be used in user code though, so it makes sense for the user to see `Widget.WidgetTheme` instead of `WidgetUtil.WidgetTheme`

By that logic shouldn't then all methods in `WidgetUtil` also be moved to the `Widget` class?

RealYusufIsmail:
> > They will be used in user code though, so it makes sense for the user to see `Widget.WidgetTheme` instead of `WidgetUtil.WidgetTheme`
> 
> 
> 
> By that logic shouldn't then all methods in `WidgetUtil` also be moved to the `Widget` class?

Should I move everything from widget utils into widget

MinnDevelopment:
Thanks, but we think this can stay where it is.

Issue #2311, opened 2022-10-26
`application_id` support for received webhook messages?


KJP12:
### General Troubleshooting

- [X] I have checked for similar issues on the Issue-tracker.
- [X] I have updated to the [latest JDA version](https://ci.dv8tion.net/job/JDA/lastSuccessfulBuild/)
- [X] I have checked the branches or the maintainers' PRs for upcoming features.

### Feature Request

For received messages emitted by webhooks, the ability to retrieve the `application_id` for telling which bot sent the message without looking up webhooks.

This would be useful in cases where you're wanting to use another bot's API, such as [PluralKit's](https://pluralkit.me/api/), and want to limit API requests to only deletions after a webhook by the bot emits a message within the channel.

This can also be helpful for cases where you need to lookup the bot where possible without having appropriate permissions to look up the webhook directly.

It's notable that very old bots have a split on application vs. 'user' ID, so, a user lookup cannot be guaranteed. Applications can also create a webhook without having a user attached, which may also lead to another failed user lookup. This isn't really significant if all you need is the ID to compare with, but it should be noted for implementing.

### Example Use-Case

```java
class Example {
    public void onReceiveMessage(Message msg) {
        if (isModLogEnabled(msg.getGuild())) {
            msgCache.put(msg.getIdLong(), msg);
            // Assuming that Application ID is already implemented:
            if (msg.isWebhookMessage() && msg.getApplicationIdLong() == 466378653216014359L) {
                pkCache.put(m.getGuildChannel().getIdLong(), m.getIdLong());
            }
        }
    }

    public void onDeleteMessage(long id) {
        var msg = msgCache.get(id);
        if(id >= timeFilter() && pkCache.get(msg.getChannel().getIdLong()) >= timeFilter()) {
             // Query PK's API & log accordingly or not log at all.
        } else {
             // Standard logging.
        }
    }
}
```

MinnDevelopment:
This is available in [5.0.0-beta.1](https://github.com/DV8FromTheWorld/JDA/releases/tag/v5.0.0-beta.1)

PR #2335, opened 2022-11-17
Add application_id support for received messages

Almighty-Satan:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #2311 

## Description

Adds `Message#getApplicationId` and `Message#getApplicationIdLong` to easily tell which bot sent the message without the need to look up webhooks.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -484,6 +484,28 @@ public interface Message extends ISnowflake, Formattable
     boolean isWebhookMessage();
 
     /**
+     * If this message is from an application-owned {@link net.dv8tion.jda.api.entities.Webhook Webhook} or
+     * is a response to an {@link net.dv8tion.jda.api.interactions.Interaction Interaction}, this will return
+     * the application's id.
+     * 
+     * @return The application's id or {@code null} if this message was not sent by an application
+     */
+    @Nullable
+    default String getApplicationId()
+    {
+        return getApplicationIdLong() == 0 ? null : Long.toUnsignedString(getApplicationIdLong());
+    }
+
+    /**
+     * If this message is from an application-owned {@link net.dv8tion.jda.api.entities.Webhook Webhook} or
+     * is a response to an {@link net.dv8tion.jda.api.interactions.Interaction Interaction}, this will return
+     * the application's id.
+     * 
+     * @return The application's id or 0 if this message was not sent by an application
+     */
+    long getApplicationIdLong();
+
+    /**
      * Returns the {@link net.dv8tion.jda.api.entities.channel.middleman.MessageChannel MessageChannel} that this message was sent in.
      *
      * @return The MessageChannel of this Message
--- a/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
@@ -73,6 +73,7 @@ public class ReceivedMessage extends AbstractMessage
     protected final MessageChannel channel;
     protected final MessageReference messageReference;
     protected final boolean fromWebhook;
+    protected final long applicationId;
     protected final boolean pinned;
     protected final User author;
     protected final Member member;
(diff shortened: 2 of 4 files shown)


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -483,6 +483,27 @@ default boolean isFromGuild()
...
 
+    /**
+     * If this message is from an application-owned {@link net.dv8tion.jda.api.entities.Webhook Webhook} or
+     * is a response to an {@link net.dv8tion.jda.api.interactions.Interaction Interaction}, this will return
+     * the application's id.
+     * 
+     * @return The application's id or "0" if this message was not sent by an application
+     */
+    @Nonnull
+    default String getApplicationId() {
```
MinnDevelopment, line 494:
```suggestion
    default String getApplicationId()
    {
```

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -509,6 +511,11 @@ public boolean isWebhookMessage()
         return fromWebhook;
     }
 
+    @Override
+    public long getApplicationIdLong() {
```
MinnDevelopment, line 515:
```suggestion
    public long getApplicationIdLong()
    {
```

MinnDevelopment (changes requested):

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -483,6 +483,28 @@ default boolean isFromGuild()
...
+     * is a response to an {@link net.dv8tion.jda.api.interactions.Interaction Interaction}, this will return
+     * the application's id.
+     * 
+     * @return The application's id or "0" if this message was not sent by an application
+     */
+    @Nonnull
+    default String getApplicationId()
+    {
+        return Long.toUnsignedString(getApplicationIdLong());
+    }
```
DV8FromTheWorld, line 497: I feel like our pattern here is normally `id` or `null`

Almighty-Satan: This matches the behavior of `MessageChannel#getLatestMessageId`, `MessageReference#getMessageId`, `MessageReference#getChannelId` and `MessageReference#getGuildId`. So I guess those should be changed as well? Maybe in a different PR?

DV8FromTheWorld: > MessageReference#getMessageId

A message will always have an Id. The docs on this method are wrong. It being `"0"` would be a bug.

> MessageReference#getChannelId

A message will always have a channel. The docs on this method are wrong. It being `"0"` would be a bug.

> MessageReference#getGuildId

This should be returning `null` if `getGuildIdLong == 0`, so yeah we need to update it.

MinnDevelopment: I'm pretty sure we explicitly decided to use "0" for `getLatestMessageId`.

DV8FromTheWorld: Why?

DV8FromTheWorld: I just looked at the git history:
Originally this method threw an IllegalStateException when you called it if there wasn't a latestMessageId to use. We changed it to instead use `String.valueOf(id)` and later `Long.toUnsignedString(id)`.

I don't see anything in the history that indicates to me that we chose to explicitly do `"0"`

MinnDevelopment: I don't remember where or when we talked about it, but I definitely remember discussing it when we removed the exception. And I don't really see a valid reason to change it now. Is there any example of where we *do* return null for missing id?

DV8FromTheWorld: image
These at a minimum

Almighty-Satan: > > MessageReference#getMessageId
> 
> A message will always have an Id. The docs on this method are wrong. It being `"0"` would be a bug.
> 
> > MessageReference#getChannelId
> 
> A message will always have a channel. The docs on this method are wrong. It being `"0"` would be a bug.

According to the [docs](https://discord.com/developers/docs/resources/channel#message-reference-object-message-reference-structure) they are optional

MinnDevelopment: Those docs are for sending, not receiving.

DV8FromTheWorld (approved):

Issue #2358, opened 2022-12-23
MarkdownSanitizer escaping incorrectly


realSurfy:
### General Troubleshooting

- [X] I have checked for similar issues on the Issue-tracker.
- [X] I have updated to the [latest JDA version](https://ci.dv8tion.net/job/JDA/lastSuccessfulBuild/)
- [X] I have checked the branches or the maintainers' PRs for upcoming bug fixes.

### Expected Behaviour

This text is supposed to be escaped from:
```
. 
    > TEST
``` 
to:
```
. 
    \> TEST
``` 

However, this does not happen.

As shown:
![image](https://user-images.githubusercontent.com/39742246/209409059-a0e93057-8b21-4d68-9c72-ae7b732f5da4.png)

Debug:
![image](https://user-images.githubusercontent.com/39742246/209409106-f29b5abb-bfcb-4c5f-bdfe-510ed00d0d5c.png)

### Code Example for Reproduction Steps

```java
String escapedText = MarkdownSanitizer.escape(". \n    > TEST");
```

### Code for JDABuilder or DefaultShardManagerBuilder used

```java
Irrelevant
```

MinnDevelopment:
Fixed in 5.0.0-beta.10

PR #2363, opened 2022-12-27
Properly escape quotes with leading spaces

Almighty-Satan:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #2358 

## Description

Fixes an issue that caused quotes with leading spaces to not be escaped.

### DIFF
--- a/src/test/java/MarkdownTest.java
+++ b/src/test/java/MarkdownTest.java
@@ -463,5 +463,6 @@ class EscapeMarkdownAllTest
         Assertions.assertEquals("\\>\\>\\> Hello World", MarkdownSanitizer.escape("\\>\\>\\> Hello World", true));
         Assertions.assertEquals("Hello > World", MarkdownSanitizer.escape("Hello > World"));
         Assertions.assertEquals("Hello\n \\> World", MarkdownSanitizer.escape("Hello\n > World"));
+        Assertions.assertEquals("Hello\n\\> World", MarkdownSanitizer.escape("Hello\n> World"));
     }
 }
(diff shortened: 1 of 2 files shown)


realSurfy:
I've tested this, it seems like it escapes everything coming after the input string aswell, could you check?

Almighty-Satan:
> I've tested this, it seems like it escapes everything coming after the input string aswell, could you check?

I don't really understand what you mean by that. Could you please elaborate a bit?

realSurfy:
Basically, what I am doing is:

Convert the markdown from Telegram for an automatic Discord message (announcement)

I am doing so by escaping the text and then convert the Telegram entities into Discord markdown.

ORIGINAL MESSAGE: 
image

ESCAPING PRESENT DISCORD MARKDOWN ON THAT TEXT:
image

DEBUG ENTITIES:
image

Entitites get converted into discord markdown

OUTPUT STRING:
image


As you can see, in the output string everything was not being converted

Almighty-Satan:
> After some more debugging I found out that the problem is that the escape actually replaces one space char instead of just adding the \ right before the quote, as shown:
> 
> ORIGINAL: image
> 
> ESCAPED: image
> 
> This is what ends up breaking my converter, the chars basically just don't add up.

I think that should be fixed now

realSurfy:
Tested, I can confirm it does in fact work properly now

DEBUG MESSAGE AFTER ESCAPING + CONVERSION:
image

DV8FromTheWorld:
This needs tests.

src/test/java/MarkdownTest.java
```diff
@@ -459,5 +462,6 @@ public void testQuote()
         Assertions.assertEquals("\\> Hello World", MarkdownSanitizer.escape("\\> Hello World", true));
         Assertions.assertEquals("\\>\\>\\> Hello World", MarkdownSanitizer.escape("\\>\\>\\> Hello World", true));
         Assertions.assertEquals("Hello > World", MarkdownSanitizer.escape("Hello > World"));
+        Assertions.assertEquals("Hello\n \\> World", MarkdownSanitizer.escape("Hello\n > World"));
```
MinnDevelopment, line 465:
```suggestion
        Assertions.assertEquals("Hello\n \\> World", MarkdownSanitizer.escape("Hello\n > World"));
        Assertions.assertEquals("Hello\n\\> World", MarkdownSanitizer.escape("Hello\n> World"));
```

PR #2375, opened 2023-01-06
Necessary additions for role subscriptions

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This adds a lot of new constants, introduced for role subscriptions.

References:
- [x] discord/discord-api-docs#5724
- [x] discord/discord-api-docs#5766
- [x] discord/discord-api-docs#5771
- [ ] discord/discord-api-docs#5828
    We don't currently handle these flags at all


DV8FromTheWorld:
This needs to await PR merges on docs repo.

DV8FromTheWorld:
I'll review in more depth then, but from my initial scan this seemed fine

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -373,5 +373,47 @@ default String getIntegrationId()
...
+        {
+            return hasSubscriptionListing() ? Long.toUnsignedString(getSubscriptionIdLong()) : null;
+        }
+
+        /**
+         * Whether this role can be purchased by a user.
+         *
+         * @return True, if this role is purchasable
+         */
+        boolean isAvailableForPurchase();
```
DV8FromTheWorld, line 409: how is this different from `hasSubscriptionListing`?

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -373,5 +373,47 @@ default String getIntegrationId()
...
+
+        /**
+         * Whether this role is acquired through a user connection.
+         * <br>Such as twitter or reddit.
+         *
+         * @return True, if this role is acquired through a user connection
+         */
+        boolean isGuildConnections();
```
DV8FromTheWorld, line 417: This.. naming isn't great.
does it not also apply for linked roles?

src/main/java/net/dv8tion/jda/api/entities/RoleConnectionMetadata.java
```diff
@@ -0,0 +1,398 @@
...
+        return this;
+    }
+
+    @Override
+    public String toString()
+    {
+        return new EntityString(this)
+            .setType(type)
+            .addMetadata("key", key)
+            .addMetadata("name", name)
```
DV8FromTheWorld, line 279: please use `.setName`

src/main/java/net/dv8tion/jda/api/entities/RoleConnectionMetadata.java
```diff
@@ -0,0 +1,398 @@
...
+    @Override
+    public DataObject toData()
+    {
+        return DataObject.empty()
+            .put("type", type.value)
+            .put("name", name)
+            .put("key", key)
+            .put("description", description)
+            .put("name_localizations", nameLocalization)
+            .put("description_localizations", descriptionLocalization);
```
DV8FromTheWorld, line 313: These will potentially send empty maps. Not sure if that's a problem for updates

src/main/java/net/dv8tion/jda/api/entities/RoleConnectionMetadata.java
```diff
@@ -0,0 +1,398 @@
...
+     */
+    @Nonnull
+    public static RoleConnectionMetadata fromData(@Nonnull DataObject data)
+    {
+        Checks.notNull(data, "Data");
+        return new RoleConnectionMetadata(
+            MetadataType.fromValue(data.getInt("type")),
+            data.getString("name", null),
+            data.getString("key", null),
+            data.getString("description", null)
```
DV8FromTheWorld, line 338: Do we not get the localization maps back?

DV8FromTheWorld: This seems like a problem because of retrieve (GET) and update (PUT) because the localization maps will be lost.

src/main/java/net/dv8tion/jda/internal/entities/RoleImpl.java
```diff
@@ -546,7 +582,10 @@ public String toString()
             return new EntityString(this)
                     .addMetadata("bot", getBotId())
                     .addMetadata("integration", getIntegrationId())
+                    .addMetadata("subscriptionListing", getSubscriptionId())
                     .addMetadata("isBoost", isBoost())
+                    .addMetadata("isAvailableForPurchase", isAvailableForPurchase())
+                    .addMetadata("isGuildConnections", isGuildConnections())
                     .toString();
```
DV8FromTheWorld, line 589: We should include a call to `.setName(this.getName())`

src/main/java/net/dv8tion/jda/api/entities/Role.java
MinnDevelopment: I have no idea. I don't even know how to setup roles like this yet. I just implemented these tags from the docs

image

src/main/java/net/dv8tion/jda/api/entities/Role.java
MinnDevelopment: I copied the name from the docs:

image

But I'm open for suggestions for better names. Maybe `isLinkedRole`?

src/main/java/net/dv8tion/jda/api/entities/RoleConnectionMetadata.java
MinnDevelopment: I might've forgotten adding it here

src/main/java/net/dv8tion/jda/api/entities/RoleConnectionMetadata.java
MinnDevelopment: We assume updates are always replacing

src/main/java/net/dv8tion/jda/internal/entities/RoleImpl.java
MinnDevelopment: There is no name for role tags

DV8FromTheWorld: Oh, this is RoleTag, not Role. my b

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -373,5 +373,47 @@ default String getIntegrationId()
...
             return isIntegration() ? Long.toUnsignedString(getIntegrationIdLong()) : null;
         }
+
+        /**
+         * Whether this role can be acquired through a premium subscription purchase.
+         *
+         * @return True, if this is a subscription role
+         *
+         * @see    #getSubscriptionIdLong()
+         */
```
DV8FromTheWorld, line 383:
```suggestion
         * Whether this role can be acquired through a premium subscription purchase.
         * A role would also need {@link #isAvailableForPurchase()} to also be true for a user to actually be
         * able to purchase the role. 
         *
         * @return True, if this is a subscription role
         *
         * @see    #getSubscriptionIdLong()
         * @see    #isAvailableForPurchase()
         */
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -373,5 +373,47 @@ default String getIntegrationId()
...
+        boolean hasSubscriptionListing();
```
DV8FromTheWorld, line 384: This can be a default-implemented method

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -373,5 +373,47 @@ default String getIntegrationId()
...
+
+        /**
+         * The subscription listing id for this role.
+         *
+         * @return The listing id, or 0 if this role is not for a subscription listing
+         */
+        long getSubscriptionIdLong();
```
DV8FromTheWorld, line 391:
```suggestion
        /**
         * The subscription listing id for this role. If a role has a subscription id then it is a premium role that 
         * can be acquired by users via purchase.
         *
         * @return The listing id, or 0 if this role is not for a subscription listing
         *
         * @see    #isAvailableForPurchase()
         */
        long getSubscriptionIdLong();
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -373,5 +373,47 @@ default String getIntegrationId()
...
+
+        /**
+         * The subscription listing id for this role.
+         *
+         * @return The listing id, or null if this role is not for a subscription listing
+         */
```
DV8FromTheWorld, line 397:
```suggestion
        /**
         * The subscription listing id for this role. If a role has a subscription id then it is a premium role that 
         * can be acquired by users via purchase.
         *
         * @return The listing id, or null if this role is not for a subscription listing
         *
         * @see    #isAvailableForPurchase()
         */
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -373,5 +373,47 @@ default String getIntegrationId()
...
+        default String getSubscriptionId()
+        {
+            return hasSubscriptionListing() ? Long.toUnsignedString(getSubscriptionIdLong()) : null;
+        }
+
+        /**
+         * Whether this role can be purchased by a user.
+         *
+         * @return True, if this role is purchasable
+         */
```
DV8FromTheWorld, line 408:
```suggestion
        /**
         * Whether this role has been published for user purchasing. Only {@link #hasSubscriptionListing() premium roles} 
         * can be purchased. However, a premium role must be published before it can be purchased. 
         * Additionally, a premium role can be unpublished after it has been published. Doing so will make it 
         * no longer available for purchase but will not remove the role from users who have already purchased it.
         *
         * @return True, if this role is purchasable
         *
         * @see    #hasSubscriptionListing()
         */
```

src/main/java/net/dv8tion/jda/api/entities/Role.java
DV8FromTheWorld: `isLinkedRole` is probably what we want here. I'm awaiting a response from an internal contact for more info.

src/main/java/net/dv8tion/jda/api/entities/Role.java
```diff
@@ -373,5 +373,47 @@ default String getIntegrationId()
...
+        boolean isAvailableForPurchase();
+
+        /**
+         * Whether this role is acquired through a user connection.
+         * <br>Such as twitter or reddit.
```
DV8FromTheWorld, line 413: I think we should call out that this can (and does) represent roles granted via 3rd party applications too, not just connections that feel like 1st party integrations.

DV8FromTheWorld (approved):

PR #2385, opened 2023-01-22
Add some helpful functions for getting attachments

JustRed23:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds some extra functions to make getting specific attachments easier, it also includes a function to check if the message actually has attachments.

PS: I hope the documentation is good enough ;)

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -595,7 +595,6 @@ public interface Message extends ISnowflake, Formattable
 
     /**
      * An immutable list of {@link net.dv8tion.jda.api.entities.Message.Attachment Attachments} that are attached to this message, filtered by file extension.
-     * <br>Most likely this will only ever be 1 {@link net.dv8tion.jda.api.entities.Message.Attachment Attachment} at most.
      *
      * <p><b>Requires {@link net.dv8tion.jda.api.requests.GatewayIntent#MESSAGE_CONTENT GatewayIntent.MESSAGE_CONTENT}</b>
      *
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -585,7 +585,7 @@ public interface Message extends ISnowflake, Formattable
      *
      * <p><b>Requires {@link net.dv8tion.jda.api.requests.GatewayIntent#MESSAGE_CONTENT GatewayIntent.MESSAGE_CONTENT}</b>
      *
-     * @param extension The extension to filter by (ex. "png")
+     * @param extension  The extension to filter by (ex. "png")
      *
      * @return True, if this message has an attachment with the provided extension.
      *
@@ -599,7 +599,7 @@ public interface Message extends ISnowflake, Formattable
      *
      * <p><b>Requires {@link net.dv8tion.jda.api.requests.GatewayIntent#MESSAGE_CONTENT GatewayIntent.MESSAGE_CONTENT}</b>
      *
-     * @param extension The extension to filter by (ex. "png")
+     * @param extension  The extension to filter by (ex. "png")
      *
      * @return Immutable list of {@link net.dv8tion.jda.api.entities.Message.Attachment Attachments} filtered by the provided extension.
      *
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -556,15 +556,58 @@ public interface Message extends ISnowflake, Formattable
     Guild getGuild();
 
     /**
+     * Returns {@code true} if this message has any attachments.
+     *
+     * <p><b>Requires {@link net.dv8tion.jda.api.requests.GatewayIntent#MESSAGE_CONTENT GatewayIntent.MESSAGE_CONTENT}</b>
+     *
+     * @return True, if this message has attachments.
+     *
+     * @see    #getAttachments()
+     */
+    boolean hasAttachments();
+
+    /**
      * An immutable list of {@link net.dv8tion.jda.api.entities.Message.Attachment Attachments} that are attached to this message.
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -555,17 +555,60 @@ default String getApplicationId()
...
      */
     @Nonnull
     List<Attachment> getAttachments();
 
+    /**
+     * Returns {@code true} if this message has any attachments with the provided extension.
+     *
+     * <p><b>Requires {@link net.dv8tion.jda.api.requests.GatewayIntent#MESSAGE_CONTENT GatewayIntent.MESSAGE_CONTENT}</b>
+     *
+     * @param extension The extension to filter by (ex. "png")
```
freya022, line 588:
```suggestion
     * @param extension  The extension to filter by (ex. "png")
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -555,17 +555,60 @@ default String getApplicationId()
...
+     *
+     * @return True, if this message has an attachment with the provided extension.
+     *
+     * @see    #getAttachmentsByExtension(String)
+     */
+    boolean hasAttachmentsByExtension(@Nonnull String extension);
+
+    /**
+     * An immutable list of {@link net.dv8tion.jda.api.entities.Message.Attachment Attachments} that are attached to this message, filtered by file extension.
+     * <br>Most likely this will only ever be 1 {@link net.dv8tion.jda.api.entities.Message.Attachment Attachment} at most.
```
freya022, line 598: Why would there be at most 1 attachment with the same extension ?

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -555,17 +555,60 @@ default String getApplicationId()
...
+     *
+     * <p><b>Requires {@link net.dv8tion.jda.api.requests.GatewayIntent#MESSAGE_CONTENT GatewayIntent.MESSAGE_CONTENT}</b>
+     *
+     * @param extension The extension to filter by (ex. "png")
```
freya022, line 602:
```suggestion
     * @param extension  The extension to filter by (ex. "png")
```

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -460,6 +461,12 @@ public Guild getGuild()
         return getGuildChannel().getGuild();
     }
 
+    @Override
+    public boolean hasAttachments()
+    {
+        return getAttachments().size() > 0;
```
freya022, line 467:
```suggestion
        return !getAttachments().isEmpty();
```

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -468,6 +475,24 @@ public List<Attachment> getAttachments()
         return attachments;
     }
 
+    @Override
+    public boolean hasAttachmentsByExtension(@Nonnull String extension)
+    {
+        Checks.notEmpty(extension, "Extension");
+        return getAttachmentsByExtension(extension).size() > 0;
```
freya022, line 482:
```suggestion
        return !getAttachmentsByExtension(extension).isEmpty();
```

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -468,6 +475,24 @@ public List<Attachment> getAttachments()
...
+
+    @Nonnull
+    @Override
+    public List<Attachment> getAttachmentsByExtension(@Nonnull String extension)
+    {
+        Checks.notEmpty(extension, "Extension");
+        List<Attachment> results = getAttachments().stream() //No need to check intents here, we already did in getAttachments()
+                .filter(a -> Objects.equals(a.getFileExtension(), extension))
+                .collect(Collectors.toList());
+        return Collections.unmodifiableList(results);
```
freya022, line 493:
```suggestion
        return getAttachments().stream() //No need to check intents here, we already did in getAttachments()
                .filter(a -> Objects.equals(a.getFileExtension(), extension))
                .collect(Helpers.toUnmodifiableList());
```

freya022 (changes requested):

src/main/java/net/dv8tion/jda/api/entities/Message.java
JustRed23: This warning was included in getAttachments() so I thought I'll just add it to this as well

freya022: Yea I guess, but desktop users aren't limited to 1 attachment anymore, might as well remove it imo

JustRed23: Sure

MinnDevelopment:
Thank you, but we generally do not implement methods that can be replaced by trivial stream operations. We do not want to bloat the interface with a lot of unnecessary helper methods.

JustRed23:
> Thank you, but we generally do not implement methods that can be replaced by trivial stream operations. We do not want to bloat the interface with a lot of unnecessary helper methods.

What if i make an event for it? Like an AttachmentReceivedEvent

Chew:
Events are generally mapped to Discord events, and there is no such AttachmentReceivedEvent, you're better off just making a util class in your project

MinnDevelopment:
That would be a very unnecessary addition since it would only remove a single boolean check in user code. You trade a single line of code for:

- An extra function in ListenerAdapter (complexity and memory overhead)
- An extra class to load at runtime (CPU time and memory overhead)
- Cognitive complexity (Why does this event exist? Does that mean normal message received has no attachment support? When do you receive this?)

Instead, a user can simply use a guard statement:

```java
if (event.getMessage().getAttachments().isEmpty()) return;
```

PR #2388, opened 2023-02-01
Add Modal.Builder#addComponents

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This is more consistent with the message rework. We expect new layout components to exist in the future and thus make our interface prepared to deal with it.

**This is semi-breaking**: `Modal.Builder#getActionRows` used to return a mutable list to allow removing existing rows, instead it now returns an immutable list and you should use `Modal.Builder#getComponents` instead.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/modals/Modal.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/modals/Modal.java
@@ -16,12 +16,16 @@
 
 package net.dv8tion.jda.api.interactions.modals;
 
+import net.dv8tion.jda.annotations.ForRemoval;
+import net.dv8tion.jda.annotations.ReplaceWith;
 import net.dv8tion.jda.api.events.interaction.ModalInteractionEvent;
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/interactions/modals/Modal.java
```diff
@@ -295,7 +368,24 @@ public Builder addActionRow(@Nonnull Collection<? extends ItemComponent> compone
         @Nonnull
         public Builder addActionRow(@Nonnull ItemComponent... components)
         {
-            return addActionRows(ActionRow.of(components));
+            return addComponents(ActionRow.of(components));
+        }
+
+        /**
+         * Returns a immutable list of all ActionRow components
```
Xirado, line 375:
```suggestion
         * Returns an immutable list of all ActionRow components
```

Xirado (changes requested):

Xirado (approved):

PR #2390, opened 2023-02-02
Handle messages from inaccessible threads and add webhook support

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds support for sending webhook messages and handle interactions from hidden threads.

### Summary

- Added `WebhookClient.createClient(jda, url)`, which allows sending webhook messages
- Added `InteractionHook.from(jda, interactionToken)`, which allows sending followup messages for interactions
- Adjusted `Message` to be usable without a channel or guild. This is necessary for WebhookClient and InteractionHook, where the channel is potentially not reachable by the bot.
- Added `Message#hasChannel` and `Message#hasGuild` to check whether a channel or guild is provided by the message instance
- Adjusted `MessageReaction` accordingly
- Handle interactions in hidden threads (private or missing access to parent), by creating the thread using the interaction `channel` objects
- Reject all interactions from group DMs for the time being
- Remove `SystemMessage` in favor of simple checks in `ReceivedMessage`
- The `EntityBuilder` now has new methods for creating messages in webhooks:
    - `createMessageFromWebhook` creates a message instance without a channel, but potentially a guild instance
    - `createMessageBestEffort` creates a message instance based on the provided channel and guild (or null in case of not being accessible)


src/main/java/net/dv8tion/jda/api/entities/WebhookClient.java
```diff
@@ -42,8 +47,24 @@
  * @see Webhook
  * @see net.dv8tion.jda.api.interactions.InteractionHook
  */
-public interface WebhookClient<T>
+public interface WebhookClient<T> extends ISnowflake
```
Vankka, line 50: Possible to get/send/edit/delete in a specific thread for this? Possibly `setThread` or `setThreadId` on the restaction?

PR #2445, opened 2023-04-14
Add voice message read support

RedDaedalus:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds read support for voice messages: both the message flag and the waveform data. Bots cannot currently send these, but they can be received from other users. Tested against both voice and non-voice messages.

Waiting on discord/discord-api-docs#6082

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -63,6 +63,8 @@ import net.dv8tion.jda.internal.utils.IOUtil;
 import okhttp3.MultipartBody;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Range;
 
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
@@ -74,6 +76,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * Represents a Text message received from Discord.
@@ -2162,7 +2165,11 @@ public interface Message extends ISnowflake, Formattable
          * Indicates, that this message will not trigger push and desktop notifications
          * @see Message#isSuppressedNotifications
          */
-        NOTIFICATIONS_SUPPRESSED(12);
+        NOTIFICATIONS_SUPPRESSED(12),
+        /**
+         * The Message is a voice message, containing an audio attachment
+         */
+        IS_VOICE_MESSAGE(13);
 
         private final int value;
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2657,6 +2666,26 @@ public boolean isEphemeral()
             return ephemeral;
         }
 
+        /**
+         * Gets the waveform data encoded in this message. This is currently only present on
+         * {@link MessageFlag#IS_VOICE_MESSAGE voice messages}.
+         *
+         * @return A possibly-{@code null} unmodifiable list of integers representing the volume
+         *         of the audio at the given point. Volume is sampled at a rate of 10 times per
```
MinnDevelopment, line 2674: Probably best to just say `8bit 10Hz Mono PCM` and maybe provide an `AudioFormat` constant for it.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2657,6 +2665,34 @@ public boolean isEphemeral()
             return ephemeral;
         }
 
+        /**
+         * Gets the waveform data encoded in this attachment. This is currently only present on
+         * {@link MessageFlag#IS_VOICE_MESSAGE voice messages}.
+         *
+         * @return A possibly-{@code null} array of integers representing the amplitude of the
+         *         audio over time. Amplitude is sampled at a rate of 10Hz. The values in this
```
MinnDevelopment, line 2673: Apparently, this sampling rate changes on longer voice messages. So it probably has to be determined using `getDuration` and the length of the array.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2657,6 +2665,35 @@ public boolean isEphemeral()
...
+                return null;
+            return Base64.getDecoder().decode(waveform);
+        }
+
+        /**
+         * Gets the duration of this attachment. This is currently only nonzero on
+         * {@link MessageFlag#IS_VOICE_MESSAGE voice messages}.
+         *
+         * @return The duration of this attachment's audio in seconds, or {@code 0}
+         *         if not it is not a voice message.
```
MinnDevelopment, line 2690:
```suggestion
         *         if this is not a voice message.
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2657,6 +2665,35 @@ public boolean isEphemeral()
...
         }
 
+        /**
+         * Gets the waveform data encoded in this attachment. This is currently only present on
+         * {@link MessageFlag#IS_VOICE_MESSAGE voice messages}.
+         *
+         * @return A possibly-{@code null} array of integers representing the amplitude of the
+         *         audio over time. Amplitude is sampled at 10Hz, but the client will decrease
+         *         this to keep the waveform under 256 bytes. The values in this array are
+         *         <b>unsigned</b>.
```
MinnDevelopment, line 2675:
```suggestion
         *         this to keep the waveform to at most 256 bytes.
         *         The values in this array are <b>unsigned</b>.
```

MinnDevelopment (changes requested):

PR #2449, opened 2023-04-21
Add EmbedBuilder#setUrl

Xirado:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds `EmbedBuilder#setUrl`. 

Setting a url by itself without a title is not useful for most users, as the only documented behaviour is a hyperlinked title,
however it is also how you create multi-image embeds, and JDA currently doesn't allow setting the URL without also settting the title.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -261,6 +261,7 @@ public class EmbedBuilder
     /**
      * Sets the Title of the embed.
      * <br>You can provide {@code null} as url if no url should be used.
+     * <br>If you want to set a URL without a title, use {@link #setUrl(String)} instead.
      *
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
```diff
@@ -303,6 +304,32 @@ public EmbedBuilder setTitle(@Nullable String title, @Nullable String url)
         return this;
     }
 
+    /**
+     * Sets the URL of the embed.
+     * <br>The Discord client mostly only uses this property in combination with the {@link #setTitle(String) title} for a clickable Hyperlink.
```
MinnDevelopment, line 309:
```suggestion
     * <br>The Discord client mostly only uses this property in combination with the {@link #setTitle(String) title} for a clickable Hyperlink.
     *
     * <p>If multiple embeds in a message use the same URL, the Discord client will merge them into a single embed and aggregate images into a gallery view.
```

PR #2453, opened 2023-04-23
Handle retry-after issues better

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

See https://github.com/discord/discord-api-docs/issues/6109


src/main/java/net/dv8tion/jda/internal/requests/Requester.java
```diff
@@ -231,6 +234,23 @@ else if (code != 429)
             {
                 task.handleResponse(lastResponse, rays);
             }
+            else
```
MinnDevelopment, line 237: We should check content-type here

PR #2454, opened 2023-04-24
Improve handling of method discovery in AnnotatedEventManager

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Replaces #1878

This PR will incrementally update the `AnnotatedEventManager`'s registered methods when registering a listener, and throw if an annotated's method signature is invalid or an array is being passed as a listener.

Also changes internals so `JDA#addEventListener` is not called on every registered listener

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
+++ b/src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
@@ -76,6 +76,13 @@ public class AnnotatedEventManager implements IEventManager
     @Override
     public void unregister(@Nonnull Object listener)
     {
+        if (listener.getClass().isArray())
+        {
+            for (Object o : ((Object[]) listener))
+                unregister(o);
+            return;
+        }
+
         if (listeners.remove(listener))
         {
             updateMethods();
--- a/src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
+++ b/src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
@@ -59,7 +59,7 @@ public class AnnotatedEventManager implements IEventManager
     {
         if (listeners.add(listener))
         {
-            updateMethods();
+            registerListenerMethods(listener);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
```diff
@@ -57,9 +57,13 @@ public class AnnotatedEventManager implements IEventManager
     @Override
     public void register(@Nonnull Object listener)
     {
+        if (listener.getClass().isArray())
+            throw new IllegalArgumentException("Tried to register an array (" + listener.getClass().getCanonicalName() + ") as an annotated event listener." +
+                    "You might want to cast the array into Object[]");
```
MinnDevelopment, line 62: I don't really understand what this is supposed to mean. Casting here wouldn't solve anything, would it?

src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
```diff
@@ -114,32 +118,31 @@ private void updateMethods()
...
+            //Skip member methods if listener is a Class
+            if (isClass && !Modifier.isStatic(m.getModifiers()))
+                continue;
 
-                    methods.get(eventClass).get(listener).add(m);
-                }
-            }
+            final Class<?>[] parameterTypes = m.getParameterTypes();
+            if (parameterTypes.length != 1 || !GenericEvent.class.isAssignableFrom(parameterTypes[0]))
+                throw new IllegalArgumentException("Method '" + m + "' must have at most 1 parameter, which implements GenericEvent");
```
MinnDevelopment, line 140: I think its better to log this as warning instead of throwing an error

src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
MinnDevelopment: This is specifically for `addEventListeners`, not `register(object)`. The important difference here is varargs.

src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
MinnDevelopment: That's not necessarily true, sometimes people might use this for their own internal usage. For instance, you might have a custom event system making use of the same annotation. Since we never previously required this, I see no reason to introduce this limitation now.

src/main/java/net/dv8tion/jda/api/hooks/AnnotatedEventManager.java
```diff
@@ -51,15 +53,23 @@
...
 
     @Override
     public void register(@Nonnull Object listener)
     {
+        if (listener.getClass().isArray())
+        {
+            for (Object o : ((Object[]) listener))
+                register(o);
+            return;
+        }
```
MinnDevelopment, line 68: Can you also update `unregister()` to do the same?

PR #2462, opened 2023-05-04
Start migration to new username API

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Implements changes outlined in https://github.com/discord/discord-api-docs/pull/6130

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -16,6 +16,8 @@
 package net.dv8tion.jda.api.entities;
 
 
+import net.dv8tion.jda.annotations.ForRemoval;
+import net.dv8tion.jda.annotations.ReplaceWith;
 import net.dv8tion.jda.api.JDA;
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -474,8 +473,10 @@ public void updateUser(UserImpl userObj, DataObject user)
     {
         String oldName = userObj.getName();
         String newName = user.getString("username");
+        String oldGlobalName = userObj.getName();
```
My-Name-Is-Jeff, line 476: I think this causes the UserUpdateGlobalNameEvent to be repeatedly fired when the username doesn't match the display name.

```suggestion
        String oldGlobalName = userObj.getGlobalName();
```

My-Name-Is-Jeff (changes requested):

PR #2469, opened 2023-05-29
Added embed type when converting it to DataObject

OcelusPRO:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Added embed type when converting it to DataObject

(this is my first contribution to a library, don't hesitate to let me know if there is a problem, note that I also speak very poor English, I may have misunderstood certain explanations)

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/EmbedType.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/EmbedType.java
@@ -31,6 +31,15 @@ public enum EmbedType
     UNKNOWN("");
 
     private final String key;
+
+    /**
+     * The key used to identify this {@link net.dv8tion.jda.api.entities.EmbedType EmbedType}.
+     */
+    public String getKey()
+    {
+        return key;
+    }
+
     EmbedType(String key)
     {
         this.key = key;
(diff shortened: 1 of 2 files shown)


MinnDevelopment:
Could you explain why this change is needed?

OcelusPRO:
I find it rather appropriate to indicate the type of the embed when you try to export it, it is especially in relation to the fact that to recreate an embed with the entityBuilder, the type is necessary, you can certainly add it manually, but it bothers me that this part is missing when we transform the embed into a DataObject

Unless I'm wrong, in the worst case if discord removes the type, it would just be ignored by discord pending an update from JDA which would properly remove it once the discord documentation is updated, in the meantime they remains necessary even if indicated as obselette

MinnDevelopment:
Since creating embeds using the `EntityBuilder` is not an officially supported mechanism, this is not really a necessity. However, in your case you can simply add the type before storing it or before building it. Since the type is always `rich`, this would not be a problem.

We might consider an official method to create an `EmbedBuilder` instance from raw JSON data in the future. Feel free to open a feature request issue, to keep this request tracked.

I'm closing this due to low utility value and because it targets the wrong branch. In the future, please make your pull requests into the `master` branch instead.

> Unless I'm wrong, in the worst case if discord removes the type, it would just be ignored by discord pending an update from JDA which would properly remove it once the discord documentation is updated, in the meantime they remains necessary even if indicated as obselette

The worst case is Discord removing the type and returning an error, although unlikely to happen, that would be considered the worst case. The type is currently *not* necessary, which is why it's explicitly not added to avoid issues.

PR #2477, opened 2023-06-13
Fix external location value to be editable on started scheduled event

Mitmocc:
## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____

Closes Issue: NaN

## Description

Previously JDA would throw when attempting to change the location of a started scheduled event with an external location. The API allows changing the external location to another external location. What it does not allow is changing the location type after the event has already started. It also does not allow to change the channel of a started event if the location type is set to voice or stage. This commit corrects this and now allows editing of the external location value for a started scheduled event.

I also removed a redundant check, I don't recall why I put it there in the first place but it is not needed, so I assume I did it by mistake.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/managers/ScheduledEventManagerImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/managers/ScheduledEventManagerImpl.java
@@ -215,11 +215,9 @@ public class ScheduledEventManagerImpl extends ManagerBase<ScheduledEventManager
     {
         if (shouldUpdate(LOCATION))
         {
-            Checks.check(getScheduledEvent().getStatus() == ScheduledEvent.Status.SCHEDULED, "Cannot update location of non-scheduled event.");
+            Checks.check(getScheduledEvent().getStatus() == ScheduledEvent.Status.SCHEDULED || entityType == ScheduledEvent.Type.EXTERNAL, "Cannot update location of non-scheduled event.");
             if (entityType == ScheduledEvent.Type.EXTERNAL && endTime == null && getScheduledEvent().getEndTime() == null)
                 throw new IllegalStateException("Missing required parameter: End Time");
-            if (entityType == ScheduledEvent.Type.EXTERNAL)
-                Checks.check((endTime != null ? endTime : getScheduledEvent().getEndTime()).isAfter(startTime), "Cannot schedule event to end before starting!");
         }
 
         if (shouldUpdate(START_TIME))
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/managers/ScheduledEventManagerImpl.java
```diff
@@ -215,11 +215,9 @@ private void preChecks()
     {
         if (shouldUpdate(LOCATION))
         {
-            Checks.check(getScheduledEvent().getStatus() == ScheduledEvent.Status.SCHEDULED, "Cannot update location of non-scheduled event.");
+            Checks.check(getScheduledEvent().getStatus() == ScheduledEvent.Status.SCHEDULED || (entityType == ScheduledEvent.Type.EXTERNAL && getScheduledEvent().getType() == ScheduledEvent.Type.EXTERNAL), "Cannot update location type or location channel of non-scheduled event.");
             if (entityType == ScheduledEvent.Type.EXTERNAL && endTime == null && getScheduledEvent().getEndTime() == null)
                 throw new IllegalStateException("Missing required parameter: End Time");
-            if (entityType == ScheduledEvent.Type.EXTERNAL)
-                Checks.check((endTime != null ? endTime : getScheduledEvent().getEndTime()).isAfter(startTime), "Cannot schedule event to end before starting!");
```
MinnDevelopment, line 222: Is there a good reason to remove this check?

src/main/java/net/dv8tion/jda/internal/managers/ScheduledEventManagerImpl.java
```diff
@@ -215,11 +215,9 @@ private void preChecks()
     {
         if (shouldUpdate(LOCATION))
         {
-            Checks.check(getScheduledEvent().getStatus() == ScheduledEvent.Status.SCHEDULED, "Cannot update location of non-scheduled event.");
+            Checks.check(getScheduledEvent().getStatus() == ScheduledEvent.Status.SCHEDULED || (entityType == ScheduledEvent.Type.EXTERNAL && getScheduledEvent().getType() == ScheduledEvent.Type.EXTERNAL), "Cannot update location type or location channel of non-scheduled event.");
```
MinnDevelopment, line 218: Would it make more sense to just check if the type is equal to the current type?

```suggestion
            Checks.check(getScheduledEvent().getStatus() == ScheduledEvent.Status.SCHEDULED || entityType == getScheduledEvent().getType()), "Cannot update location type or location channel of non-scheduled event.");
```

Mitmocc: No, because that doesn't cover the type being Voice/Stage and the channel being changed, which is not possible. You can only edit the location value if the type is external.

MinnDevelopment: I just tested this and the client lets me change the channel for the event.

MinnDevelopment: Oh I see, you can't change it if the event is already taking place in that channel. That makes sense, then.

PR #2490, opened 2023-06-30
Fix choices having an "options" prefix when automatically localizing

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fixes automatic localization adding "options" before the choice name


MinnDevelopment:
What exactly does this change impact? Can this break existing localization files/code?

freya022:
If you made a workaround for this bug then of course it's breaking, but since the docs never mentioned that extra "options" (even in the examples) then that should be fine

PR #2494, opened 2023-07-03
Fix handling of thread update for unknown parent

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

When the parent channel is missing, we should cache the thread update.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/handle/ThreadUpdateHandler.java
+++ b/src/main/java/net/dv8tion/jda/internal/handle/ThreadUpdateHandler.java
@@ -24,6 +24,7 @@ import net.dv8tion.jda.api.utils.cache.CacheFlag;
 import net.dv8tion.jda.api.utils.data.DataArray;
 import net.dv8tion.jda.api.utils.data.DataObject;
 import net.dv8tion.jda.internal.JDAImpl;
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/handle/ThreadUpdateHandler.java
```diff
@@ -59,11 +60,26 @@ protected Long handleInternally(DataObject content)
...
+                    new ChannelUpdateArchivedEvent(
+                        api, responseNumber,
+                        thread, true, false));
+            }
+            catch (IllegalArgumentException ex)
+            {
+                if (EntityBuilder.MISSING_CHANNEL.equals(ex.getMessage()))
+                {
+                    long parentId = content.getUnsignedLong("parent_id", 0L);
+                    EventCache.LOG.debug("Received THREAD_UPDATE for a thread with uncached parent. Parent ID: {} JSON: {}", parentId, content);
```
MinnDevelopment, line 76:
```suggestion
                    EventCache.LOG.debug("Caching THREAD_UPDATE for a thread with uncached parent. Parent ID: {} JSON: {}", parentId, content);
```

PR #2495, opened 2023-07-04
Add isShutdown() method to check if ShardManager has been shut down

jandie1505:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds a method to check if a ShardManager is shut down.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManager.java
+++ b/src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManager.java
@@ -677,6 +677,6 @@ public class DefaultShardManager implements ShardManager
     @Override
     public boolean isShutdown()
     {
-        return this.shutdown.get();
+        return this.executor.isTerminated();
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/sharding/DefaultShardManager.java
```diff
@@ -673,4 +673,10 @@ protected ExecutorPair(E executor, boolean automaticShutdown)
             this.automaticShutdown = automaticShutdown;
         }
     }
+
+    @Override
+    public boolean isShutdown()
+    {
+        return this.shutdown.get();
```
CheesyGamer77, line 680: See [initial discussion in Discord](https://discord.com/channels/125227483518861312/125227483518861312/1125903549847584798). This is a logic error.

CheesyGamer77 (changes requested):

jandie1505:
I close this due to the logic error and will re-open it when it is fixed.

jandie1505:
I changed the method that it checks if the ScheduledExecutorService of the DefaultShardManager is terminated.

JDAs can still be running if isShutdown() returns true. If it would make more sense to not return true until all the JDAs are also shut down, i can add that as well.

PR #2504, opened 2023-07-22
Chage return type of method in JDA class.

kiLeo13:
* Changed the return type of "openPrivateChannelById(String)" from RestAction to CacheRestAction as it delegates to a method that returns CacheRestAction.

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

The return type of `JDA#openPrivateChannelById(long)` is `CacheRestAction`. However, the method `JDA#openPrivateChannelById(String)` returns `RestAction`, implying that the cache will not be used. This seems inconsistent, especially considering that the String method delegates to the long method.

Also changed the documentation of `JDA#openPrivateChannelById(String)` to clarify that the cache will be used.


Xirado:
Dev-facing methods should not expose internal RestAction implementations, but rather the common `RestAction` interface.

PR #2505, opened 2023-07-22
Add Member#getEffectiveNickname()

AlexCai2019:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

- Change the current Member#getEffectiveName() to Member#getEffectiveNickname().
- Add Member#getEffectiveName() which works just as same as User#getEffectiveName().
- Use ternary operator in Member#getTimeJoined() because why not.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Member.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Member.java
@@ -250,12 +250,25 @@ public interface Member extends IMentionable, IPermissionHolder, UserSnowflake
     String getNickname();
 
     /**
+     * The name visible in the UI.
+     * <br>If the {@link User#getGlobalName() global name} is {@code null}, this returns the {@link User#getName() username} instead.
+     *
+     * @see User#getEffectiveName()
+     * @return The effective display name
+     */
+    @Nonnull
+    default String getEffectiveName()
+    {
+        return getUser().getEffectiveName();
+    }
+
+    /**
      * Retrieves the Name displayed in the official Discord Client.
      *
      * @return The guild nickname of this Member or the {@link User#getEffectiveName() effective user name} if no guild nickname is present.
      */
     @Nonnull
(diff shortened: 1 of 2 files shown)


Andre601:
I also can't see the logic behind this... A Member can't have more than 1 nickname. And `getEffectiveName()` does iirc this logic:
```
Member has Nickname? -- Yes --> Return Nickname
  |
  |
  V
Member has global Displayname? -- Yes --> Return global display name
  |
  |
  V
Return username/identifier
```

MinnDevelopment:
What is the use-case for this?

AlexCai2019:
Because I think the `getEffectiveName()` method in both `User` and `Member` should return the same thing, which is the global name if present or the name, in order to prevent misuse.
On the other hand, `Member#getEffectiveNickname()` can provide the replaced function of the original `Member#getEffectiveName()`.

Andre601:
> Because I think the `getEffectiveName()` method in both `User` and `Member` should return the same thing, which is the global name if present or the name, in order to prevent misuse. On the other hand, `Member#getEffectiveNickname()` can provide the replaced function of the original `Member#getEffectiveName()`.

I think you missunderstand the logic of `getEffectiveName()` here.

Imagine it as what the bot would "see" when either looking at a User in DMs or a User on a Server.
After all, if a user has a nickname on the server do you see that and not their nickname right? That's the "effective" name they have here.

MinnDevelopment:
The point of `getEffectiveName` is to return the visible name. In the context of a user this is the globally visible name (what you see in DMs) and in the context of a member this is the guild name (visible in the member list of a guild). As such, the returned names are consistent between the two contexts.

Adding another name getter is only adding more confusion and should be avoided. Delegating getters from User into Member is a non-goal. To use the user getters, use `getUser().getXYZ()` instead.

PR #2516, opened 2023-08-02
Add initial support for media channels

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This introduces MediaChannel and the new `IPostContainer` abstraction.


src/main/java/net/dv8tion/jda/internal/handle/ChannelUpdateHandler.java
```diff
@@ -226,6 +229,44 @@ private AbstractGuildChannelImpl<?> handleChannelTypeChange(AbstractGuildChannel
...
+
+            DataObject syntheticDelete = DataObject.empty()
+                .put("t", "CHANNEL_DELETE")
+                .put("s", responseNumber)
+                .put("op", 0)
+                .put("d", DataObject.empty()
+                    .put("type", channel.getType().getId())
+                    .put("guild_id", guild.getId())
+                    .put("id", channel.getIdLong()));
+            client.getHandler("CHANNEL_DELETE").handle(responseNumber, syntheticDelete);
```
DV8FromTheWorld, line 260: I'm not sure we should do this from the perspective of it could cause bots to do things they'd normal do on `CHANNEL_DELETE` but incorrectly, like removing data for the channel even though the channel still exists

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -234,6 +242,76 @@ public void invalidate()
...
+            this.getThreadChannelsView().remove(id);
+            break;
+        }
+
+        // Remove dangling threads
+        if (channel instanceof IThreadContainer)
+        {
+            SortedSnowflakeCacheViewImpl<ThreadChannel> localView = this.getThreadChannelsView();
+            SnowflakeCacheViewImpl<ThreadChannel> globalView = api.getThreadChannelsView();
+            Predicate<ThreadChannel> predicate = thread -> channel.equals(thread.getParentChannel());
```
DV8FromTheWorld, line 291: Should this match on parentChannelId instead of parentChannel since `getParentChannel` looks at the cache and the parent has already been removed from the cache?

src/main/java/net/dv8tion/jda/internal/handle/ChannelUpdateHandler.java
```diff
@@ -216,46 +192,92 @@ private AbstractGuildChannelImpl<?> handleChannelTypeChange(AbstractGuildChannel
...
-            return textChannel;
+        long id = channel.getIdLong();
+        switch (oldType)
+        {
+        case TEXT:
+            getJDA().getTextChannelsView().remove(id);
+            guild.getTextChannelsView().remove(id);
+            break;
+        case NEWS:
+            getJDA().getNewsChannelView().remove(id);
```
DV8FromTheWorld, line 205: If there a way to dedupe some of this code between what's here and in `Guild#uncacheChannel`?

src/main/java/net/dv8tion/jda/api/entities/channel/concrete/ForumChannel.java
```diff
@@ -50,16 +41,16 @@
...
     /**
      * The maximum length of a forum topic ({@value #MAX_FORUM_TOPIC_LENGTH})
      */
-    int MAX_FORUM_TOPIC_LENGTH = 4096;
+    int MAX_FORUM_TOPIC_LENGTH = IPostContainer.MAX_POST_CONTAINER_TOPIC_LENGTH;
     /**
      * The maximum number of {@link ForumPostAction#setTags(Collection) tags} that can be applied to a forum post. ({@value #MAX_POST_TAGS})
      */
-    int MAX_POST_TAGS = 5;
+    int MAX_POST_TAGS = IPostContainer.MAX_POST_TAGS;
```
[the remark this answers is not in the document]
MinnDevelopment: I left them in case these are different per channel type in the future

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
MinnDevelopment: That doesn't matter since the channel is locally cached in the thread, and only updated if the cache has a newer non-null instance available.

PR #2519, opened 2023-08-08
fix symmetricity of CustomEmojiImpl/RichCustomEmojiImpl#equals

danthe1st:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

The contract of [`Object#equals`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)) specifies:
> It is _symmetric_: for any non-null reference values `x` and `y`, `x.equals(y)` should return `true` if and only if `y.equals(x)` returns true.

This should hold when calling `equals` on an instance of `RichCustomEmojiImpl` and `CustomEmojiImpl`.

This is because `CustomEmojiImpl` [allows any `CustomEmoji`](https://github.com/discord-jda/JDA/blob/c0582c5f397517d42bf73f156c1bf89147771aa4/src/main/java/net/dv8tion/jda/internal/entities/emoji/CustomEmojiImpl.java#L103) and checks the id while `RichCustomEmojiImpl` [expects a `RichCustomEmojiImpl`](https://github.com/discord-jda/JDA/blob/c0582c5f397517d42bf73f156c1bf89147771aa4/src/main/java/net/dv8tion/jda/internal/entities/emoji/RichCustomEmojiImpl.java#L243C8-L243C8).

```java
CustomEmoji a = new RichCustomEmojiImpl(123, someGuild);//using internal classes for demonstration
CustomEmoji b = new CustomEmojiImpl("a",123,true);
assert a.equals(b) == b.equals(a);//false
```

This PR changes `RichCustomEmojiImpl` to allow any `CustomEmoji` instances and `CustomEmoji` in order to check the emoji name just like `RichCustomEmojiImpl`.
This allows JDA users to easily compare two `CustomEmoji` instances.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/emoji/CustomEmojiImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/emoji/CustomEmojiImpl.java
@@ -90,7 +90,7 @@ public class CustomEmojiImpl implements CustomEmoji, EmojiUnion
     @Override
     public int hashCode()
     {
-        return Objects.hash(name, id, animated);
+        return Long.hashCode(id);
     }
 
     @Override
@@ -101,7 +101,7 @@ public class CustomEmojiImpl implements CustomEmoji, EmojiUnion
         if (!(obj instanceof CustomEmoji))
             return false;
         CustomEmoji other = (CustomEmoji) obj;
-        return this.id == other.getIdLong() && getName().equals(other.getName());
+        return this.id == other.getIdLong();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/entities/emoji/CustomEmojiImpl.java
```diff
@@ -101,7 +101,7 @@ public boolean equals(Object obj)
         if (!(obj instanceof CustomEmoji))
             return false;
         CustomEmoji other = (CustomEmoji) obj;
-        return other.getIdLong() == id;
+        return this.id == other.getIdLong() && getName().equals(other.getName());
```
MinnDevelopment, line 104: I also noticed that this violates the `hashCode` contract, since `a.equals(b) != (a.hashCode() == b.hashCode())`. It might be better to only compare by id, in both equals and hashCode.

danthe1st: Done

PR #2521, opened 2023-08-09
Add custom status support for bots

RedDaedalus:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____

Closes Issue: NaN

## Description

Implements support for custom status in bots. This was rolled out about an hour ago on the API side and I have tested it with my own bot.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Activity.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Activity.java
@@ -15,7 +15,6 @@
  */
 package net.dv8tion.jda.api.entities;
 
-import net.dv8tion.jda.annotations.Incubating;
 import net.dv8tion.jda.api.entities.emoji.Emoji;
 import net.dv8tion.jda.api.entities.emoji.EmojiUnion;
 import net.dv8tion.jda.internal.entities.EntityBuilder;
@@ -194,11 +193,8 @@ public interface Activity
      *         if the specified name is null, empty, blank or longer than 128 characters
      *
      * @return A valid Activity instance with the provided name with {@link net.dv8tion.jda.api.entities.Activity.ActivityType#WATCHING}
-     *
-     * @incubating This feature is not yet confirmed for the official bot API
      */
     @Nonnull
-    @Incubating
     static Activity watching(@Nonnull String name)
     {
         Checks.notBlank(name, "Name");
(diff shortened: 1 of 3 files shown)


RedDaedalus:
docs PR: https://github.com/discord/discord-api-docs/pull/6345

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -931,7 +931,7 @@ public static Activity createActivity(DataObject gameJson)
 
         if (type == Activity.ActivityType.CUSTOM_STATUS)
         {
-            if (gameJson.hasKey("state") && name.equalsIgnoreCase("Custom Status"))
+            if (gameJson.hasKey("state"))
```
MinnDevelopment, line 934: Why remove this?

src/main/java/net/dv8tion/jda/internal/managers/PresenceImpl.java
```diff
@@ -184,7 +184,15 @@ private DataObject getGameJson(Activity activity)
...
         DataObject gameObj = DataObject.empty();
-        gameObj.put("name", activity.getName());
+
+        if (activity.getType() == Activity.ActivityType.CUSTOM_STATUS)
+        {
+            gameObj.put("name", "Custom Status");
+            gameObj.put("state", activity.getName());
+        } else {
+            gameObj.put("name", activity.getName());
+        }
```
MinnDevelopment, line 194:
```suggestion
        }
        else
        {
            gameObj.put("name", activity.getName());
        }
```

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
RedDaedalus: This is an implementation detail that the client itself does not rely on and other libraries do not respect (djs allows you to set anything for name, for example.) I figured it was better for reliability to remove it.

MinnDevelopment: What does the client do if you set the name to `Custom Status` vs anything else?

src/main/java/net/dv8tion/jda/api/entities/Activity.java
```diff
@@ -230,6 +226,29 @@ static Activity competing(@Nonnull String name)
...
+     *
+     * @param  name
+     *         The not-null name of the newly created status
+     *
+     * @throws IllegalArgumentException
+     *         If the specified name is null, empty, blank or longer than 128 characters
+     *
+     * @return A valid Activity instance with the provided name with {@link net.dv8tion.jda.api.entities.Activity.ActivityType#CUSTOM_STATUS}
+     *
+     * @since  5.0.0
```
MinnDevelopment, line 241: We don't really use `@since` anymore

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
RedDaedalus: It renders it the same -- it's just ignored.

MinnDevelopment (approved):

PR #2523, opened 2023-08-10
Add Activity#withState

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Allows adding state to activities. Also makes activities with only a state and name not appear as rich.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
@@ -104,7 +104,6 @@ public class EntityBuilder
         tmp.add("flags");
         tmp.add("party");
         tmp.add("session_id");
-        tmp.add("state");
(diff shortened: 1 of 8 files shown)


src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -938,16 +937,17 @@ public static Activity createActivity(DataObject gameJson)
             }
         }
 
+        String state = gameJson.isNull("state") ? null : String.valueOf(gameJson.get("state"));
```
Xirado, line 940:
```suggestion
        String state = gameJson.isNull("state") ? null : gameJson.getString("state");
```

Xirado (changes requested):

PR #2525, opened 2023-08-20
Fix chaos in example code in the README

DWolf-19:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This simple PR fixes 2 small aspects in example code in the README:
* Curly brackets
* Beginning of sentences (comments)

Consider this an addition to #2289

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ public class ReadyListener implements EventListener
             .addEventListeners(new ReadyListener())
             .build();
 
-        // optionally block until JDA is ready
+        // Optionally block until JDA is ready
         jda.awaitReady();
     }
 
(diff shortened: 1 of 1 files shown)


MinnDevelopment:
Thank you, but all of this is intentional/unimportant.

- Braces on the same line are java convention
- Spacing or casing of comments is irrelevant and a superficial change
- None of this can be considered "chaos"

DWolf-19:
If this is a Java convention, why is this not case in all examples?

DWolf-19:
I don't agree with closure of this PR. I think in combination with #2289 it would improve readability. README doesn't look good right now. And if we agree with Java conventions, this should be the case in all examples - otherwise it looks carelessly.

Andre601:
> I don't agree with closure of this PR. I think in combination with #2289 it would improve readability. README doesn't look good right now. And if we agree with Java conventions, this should be the case in all examples - otherwise it looks carelessly.

At least for the java comments I don't see much of an improvement here given it's just capitalization of a single letter that nobody should be hung up on.
Tho, I agree on the examples being onconsistent. Some have the bracket on a new lines and others don't. It looks weird and as if multiple people with their own coding style designed the readme without checking the currently used coding style...

MinnDevelopment:
I have plans to work on the readme, but it is currently a low priority since it does the job and other more important things are going on in my life. When I find time to work on this library, it needs to be towards getting it out of beta, which means I have to prioritize what I can spend my time on. 

Ideally, I want all examples to follow the common java conventions, such as braces on the same line. If you want to make a PR specifically addressing just that, I'd be fine with merging it.

This PR, however, included both an unwanted change (moving braces on newlines) and superficial changes (casing that is irrelevant) so there is really no motivation to accept it.

The other PRs regarding grammar are mostly nitpicking, and I plan on reworking the readme anyway when the beta is completed. I keep that PR open to signal that I'm aware it's not perfect, but also not a priority at this time. For now, it is good enough to give an overview of the library and links to relevant other sources like the wiki and documentation.

PR #2527, opened 2023-08-28
Rename MessageReceivedEvent to MessageCreateEvent

DWolf-19:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Rename `MessageReceivedEvent` to `MessageCreateEvent`.

I just found this todo while reading the source code (I hope I'm not too late):
```java
//TODO-v5: Consider renaming this to MessageCreateEvent to match standard event naming (and to match Discord naming!)
```
This PR deprecate (without breaking the library interface) `MessageReceivedEvent`. Also, when accessing the `MessageReceivedEvent`, there will be a "redirected" to the new event, so that when making changes to logic, we don't need to edit same code in 2 places

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/channel/middleman/MessageChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/channel/middleman/MessageChannel.java
@@ -99,7 +99,7 @@ public interface MessageChannel extends Channel, Formattable
      * The id for the most recent message sent
      * in this current MessageChannel.
      *
-     * <p>This value is updated on each {@link net.dv8tion.jda.api.events.message.MessageReceivedEvent MessageReceivedEvent}
+     * <p>This value is updated on each {@link net.dv8tion.jda.api.events.message.MessageCreateEvent MessageCreateEvent}
      * and <u><b>the value might point to an already deleted message since the ID is not cleared when the message is deleted,
      * so calling {@link #retrieveMessageById(long)} with this id can result in an {@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE} error</b></u>
      *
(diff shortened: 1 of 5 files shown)


src/main/java/net/dv8tion/jda/api/events/message/MessageReceivedEvent.java
```diff
@@ -39,17 +40,22 @@
...
+ *             In fact, it's now a container for MessageCreateEvent.
 
  */
-//TODO-v5: Consider renaming this to MessageCreateEvent to match standard event naming (and to match Discord naming!)
+@Deprecated
+@ReplaceWith("MessageCreateEvent")
 public class MessageReceivedEvent extends GenericMessageEvent
 {
-    private final Message message;
+    private final MessageCreateEvent event;
```
duncte123, line 53: Why'd you change this? It doesn't make a whole lot of sense to me

DWolf-19: > ...so that when making changes to logic, we don't need to edit same code in 2 places.

* When adding a new method to `MessageCreateEvent`, we don't need to copy-paste the code, we just create a method of the same name in the `MessageReceivedEvent` and call the method from the new event in it.
* When editing the code, the same thing.

It seemed right to me. I think it's correct (even if it's only getters so far) to take out their logic of work to the `MessageCreateEvent`. 

And here I really don't know: if after "deprecate" the old event will not be supported in ANY way (no adding/removing logic/fixes (for backward compatibility)), then this thing is really useless.

MinnDevelopment:
This change does not seem worth it. I think we should just remove that todo comment.

DWolf-19:
I think todo should be removed in some large PR dedicated to refactoring/code cleaning (or just separate PR). Rolling back existing changes in this PR will be time consuming.

Renaming will never happen, or is it just a temporary freeze? If the latter, I think you can leave this PR for the future without deleting todo (so that others do not open duplicates). 🤔

MinnDevelopment:
> Renaming will never happen, or is it just a temporary freeze

Very likely not going to happen.

> should be removed in some large PR 

I don't see why removing a comment requires a large PR. I think it's fine to have small PRs with specific scopes.

DWolf-19:
Freezing for an indefinite time.

PR #2529, opened 2023-08-31
Allow override of headers applied by default

zaroxh:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Changed the ```applyHeaders``` method in the ```Requester``` class, so that custom headers with the same name in custom requests override the headers applied by default. This is useful if you want to override the "Authorization" header for example in a custom OAuth2 request, that needs another value for this header. The previous behaviour in this case caused there to be two "Authorization" headers in the request, which in turn lead to a HTTP 400 error being returned by the Discord API.


MinnDevelopment:
I would not recommend using JDA for OAuth requests like this, because JDA will treat all requests to be in the same rate-limit system. OAuth requests generally have different rate-limits applied per auth token, which the internal requester is not equipped to handle.

PR #2558, opened 2023-10-07
Add getting cache flags from events

freya022:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Allows getting cache flags from events, also fixes documented requirements of scheduled events, and clarifies permission overrides events

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/events/guild/override/GenericPermissionOverrideEvent.java
+++ b/src/main/java/net/dv8tion/jda/api/events/guild/override/GenericPermissionOverrideEvent.java
@@ -26,6 +26,7 @@ import net.dv8tion.jda.api.entities.channel.attribute.IPermissionContainer;
 import net.dv8tion.jda.api.entities.channel.middleman.GuildChannel;
 import net.dv8tion.jda.api.entities.channel.unions.IPermissionContainerUnion;
 import net.dv8tion.jda.api.events.guild.GenericGuildEvent;
+import net.dv8tion.jda.api.utils.cache.CacheFlag;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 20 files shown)


Andre601 (changes requested):
As per the [Structure Guide](https://jda.wiki/contributing/structure-guide/#linking):
> If the class is not already imported, use the fully qualified name when you are inking to something through JavaDoc!
> Bad: `{@link RestAction}`
> Good: `@link net.dv8tion.jda.core.requests.RestAction RestAction}`

You are creating imports that are only used for the JavaDoc which is not a good aproach here. In fact are there inconsistencies in your JavaDoc (i.e. using `{@link Guild#getPublicRole() @everyone}` but then later using the full path for the Guild class)

src/test/java/net/dv8tion/jda/CacheFlagRequirementsTest.java
```diff
@@ -0,0 +1,176 @@
...
+                    .orElse(null);
+            if (description == null)
+            {
+                LOGGER.warn("Undocumented class at {}", qualifiedEventName);
+                continue;
+            }
+
+            final List<String> links = getLinks(description);
+            final EnumSet<E> expectedFlags = mapLinks(enumType, links, enumEntryReferencePattern);
+            enumEntriesByClass.put((Class<GenericEvent>) Class.forName(qualifiedEventName), expectedFlags);
```
MinnDevelopment, line 103: Thoughts on using annotations instead of parsing javadocs? I think it would be a lot easier and clearer if we just added a `@RequiredCacheFlag(CacheFlag.EMOJI)` to the class.

freya022:
For reference, here is how https://github.com/discord-jda/JDA/pull/2558/commits/aa6788daae50067eed066933f56b0d49e2f60f87 is made: [Generate requirement annotations.patch](https://github.com/discord-jda/JDA/files/15082562/Generate.requirement.annotations.patch)

PR #2575, opened 2023-10-25
Add additional methods to get the target of an audit-log entry

Almighty-Satan:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds multiple methods to get the target of an `AuditLogEntry`. This includes methods for auto-mod rules, channels, members, scheduled events and threads.
The `integrations` field returned by Discord's `audit-logs` endpoint (https://discord.com/developers/docs/resources/audit-log) is currently being ignored because I don't think JDA has any API for integrations (please correct me if I'm wrong).

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/audit/AuditLogEntry.java
+++ b/src/main/java/net/dv8tion/jda/api/audit/AuditLogEntry.java
@@ -19,12 +19,16 @@ package net.dv8tion.jda.api.audit;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.entities.ISnowflake;
+import net.dv8tion.jda.api.entities.Member;
+import net.dv8tion.jda.api.entities.ScheduledEvent;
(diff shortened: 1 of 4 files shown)


freya022:
I think it should be the user's responsibility to check the target's type.
It would even help in the situation the target is correct, but the entity is still `null` (due to cache issues for example).

So basically, throw if the target type is incorrect, else return the `target`.

(Don't forget the dot at the end of your Javadoc descriptions btw)

Almighty-Satan:
> I think it should be the user's responsibility to check the target's type. It would even help in the situation the target is correct, but the entity is still `null` (due to cache issues for example).
>
> So basically, throw if the target type is incorrect, else return the `target`.

I thought about that, but decided against it because `AuditLogEntry#getWebhook` (which already exists) also doesn't throw. I'd be open to changing that tho.

> (Don't forget the dot at the end of your Javadoc descriptions btw)

I just copied the docs from `AuditLogEntry#getWebhook`. There are dots missing on other methods in that class too. I'll fix that later.

PR #2577, opened 2023-10-29
Add support for guild security incidents

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds support for guild security incidents.

Reference: https://github.com/discord/discord-api-docs/pull/7360


src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -3501,6 +3510,34 @@ default AuditableRestAction<Integer> prune(int days, @Nonnull Role... roles)
...
+    /**
+     * Update the current guild {@link SecurityIncidents security incidents}.
+     * <br>Security incidents are used to temporarily disable features for the purpose of moderation.
+     *
+     * <p>Possible {@link net.dv8tion.jda.api.requests.ErrorResponse ErrorResponses} caused by
+     * the returned {@link RestAction RestAction} include the following:
+     * <ul>
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#INVALID_FORM_BODY}
+     *     <br>If one of the provided timestamps is too far into the future</li>
+     * </ul>
```
Xirado, line 3522: Why can't we check for this?

MinnDevelopment: We might in the future, but since they still haven't merged docs it might be bad to check limits.

PR #2587, opened 2023-11-18
Document event data passthrough issues in extended events

freya022:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR document issues with event data passthrough when reconstructing the event from other threads


freya022:
Using the `InheritedThreadLocal` only allows the event to be passed through seamlessly in quite specific conditions, while the suggested solution should always work, so maybe I should revert that change?

PR #2604, opened 2024-01-10
Add missing retrieveMessage methods to ReactionEvents

Kaktushose:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

Adds a `retrieveMessage` method to the `MessageReactionRemoveAllEvent` and `MessageReactionRemoveEmojiEvent`. While it's just a shortcut for `getChannel().retrieveMessageById(getMessageId())` and can therefore be seen as redundant and syntatic sugar, the `GenericMessageReactionEvent` also has such a method. So this addition makes the API more consistent


duncte123:
Wouldn't it be better to move this to the `GenericMessageEvent` class instead? Seems right now that you are just repeating the same code a few times

yoyosource:
What about adding a interface like GenericMessageReactionRemoveEvent with the common code then?

PR #2622, opened 2024-02-23
Add Member#hasRoles(Role...)

AlexCai2019:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Add hasRoles(Role...) to Member.java and MemberImpl.java
hasRole(Role...) returns true if the member has all these roles.
```java
Role adminRole = guild.getRoleById(ADMIN_ROLE_ID);
if (member.hasRoles(adminRole))
    event.reply("This member is an admin!").queue();
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
@@ -38,7 +38,6 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.awt.*;
 import java.time.OffsetDateTime;
-import java.util.List;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


kazuryyx:
what in the actual ...

src/main/java/net/dv8tion/jda/internal/entities/MemberImpl.java
```diff
@@ -207,6 +206,24 @@ public List<Role> getRoles()
         return Collections.unmodifiableList(roleList);
     }
 
+    @Override
+    public boolean hasRoles(@Nonnull Role... roles)
+    {
+        switch (roles.length)
```
duncte123, line 212: Imo there is no need for this switch statement. Just do a fast return false if `this.roles.isEmpty()` and otherwise do the contains all check

MinnDevelopment:
Thank you, but we don't believe this provides much value over just using list operations.

PR #2623, opened 2024-02-26
Refactored verbose code constructs and QoL improvements

Cespito:
## Pull Request Etiquette

### Changes
- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____

Closes Issue: NaN

## Description
This pull request makes the following improvements:
- Refactor verbose code constructs
- Fix raw use of parameterized class
- Delete unnecessary import

**Details**
- Added lambda expressions and utilized library methods to reduce code verbosity
- Removed unnecessary cast expressions
- Such raw use of generic types is valid in Java, but it defeats the purpose of type parameters and may mask bugs
- Replaced C like implementations in Java classes
- Removed unused imports
- Added log in exception catch on awaitReady() function
- Simplified this construct
```java
if (currentIterator != null)
{
    if (!currentIterator.hasNext())
    {
        currentIterator.close();
        currentIterator = null;
    }
    else
    {
        if (findNext()) return true;
        currentIterator.close();
        currentIterator = null;
        }
    }
```
with
```java
if (currentIterator != null)
{
    if (currentIterator.hasNext())
    {
        if (findNext()) return true;
    }
    currentIterator.close();
    currentIterator = null;
}
```

**Benefits**
Improved code organization, maintainability, and readability.
Reduced redundancy and enhanced robustness of the code.

### DIFF
--- a/src/examples/java/MessageLoggerExample.java
+++ b/src/examples/java/MessageLoggerExample.java
@@ -26,6 +26,8 @@ import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
 import net.dv8tion.jda.api.events.message.react.MessageReactionAddEvent;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
 import net.dv8tion.jda.api.requests.GatewayIntent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import java.io.IOException;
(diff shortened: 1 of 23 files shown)


MinnDevelopment:
All these changes are superficial and not needed. Please don't open PRs that do not address concrete issues or offer meaningful improvements, especially avoid opening PRs that change many unrelated things at once.

Large unscoped changes like this are unwelcome. You could instead open specific PRs for individual issues, like replacing `size() == 0` with `isEmpty()`. It is much easier to accept or reject individual logical changes.

Please also avoid changing TweetNaclFast, that class should only be touched when absolutely necessary since it is actually a dependency that we copied rather than our own code.

Issue #2646, opened 2024-04-11
Polls


rakosi2:
### General Troubleshooting

- [X] I have checked for similar issues on the Issue-tracker.
- [X] I have updated to the [latest JDA version](https://github.com/discord-jda/JDA/releases/latest)
- [X] I have checked the branches or the maintainers' PRs for upcoming features.

### Feature Request

Allow getMessage() to see if the message contains a poll and get its contents. 
Currently, there is no way to disable polls and anyone with send message perms are able to send polls.
Discord has now enabled polls on all servers and this is causing users to bypass filters and create spam.
Messages with polls are currently MessageType.DEFAULT and have content of "".

https://support.discord.com/hc/en-us/articles/22163184112407-Polls-FAQ#h_01HSED5BPEDH4YYFVVCXQ2NKZ1

### Example Use-Case

```java
getMessage().getPoll()
```

Andre601:
Personally speaking, I feel like this would not be a reliable way at all to check for polls.

While I haven't encountered any other cases, I'm sure there can be situations where a message may not contain any content while of message type default (I would assume Webhook messages can be?).
This just wouldn't be reliable in the slightest.

What should be done instead is to request proper addition of API from Discord, so that bots have a reliable way of figuring out if a message is a poll... As an example:

- A `poll` attribute in the Message object or alternatively a `POLL` message type
- Endpoints to retrieve Poll data from a message
- Endpoints to POST own polls
- Websocket Events (or whatever it would be) for Polls (Creation, deletion, answering, ending, etc)

I agree that Polls in their current form are a mess to put it very simple, so changes are necessary here.

In the meantime, set up auto-mod rules, since Polls also are send through it, so if you have a decent auto-mod setup, polls may not get send at all.

rakosi2:
If you look at the raw data, it is similar to an embed but unlike an embed, only one poll can be in a message.

Andre601:
I've looked at the API docs repository of Discord and it seems there is actually API available, just not yet documented.
And from what I can see is it pretty much what I was talking about above, meaning:

- a `poll?` attribute in the retrieved message/interaction object that also contains info such as the question, answers, expiry date, etc.
- Gateway events for poll creation and removal

So, we better should wait for the docs to be finalized and published before making any changes to JDA here.

freya022:
[...]

PR #2649, opened 2024-04-13
Poll support

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: #2646

## Description

Adds support for message polls.

Reference: https://github.com/discord/discord-api-docs/pull/6746

### Feature Additions

- `MessagePoll`
- `MessagePollData`
- `MessagePollBuilder`
- `Message#getPoll()`
- `MessageCreateRequest#setPoll(MessagePollData)`
- `Message#replyPoll(MessagePollData)`
- `IReplyCallback#replyPoll(MessagePollData)`
- `MessageChannel#sendMessagePoll(MessagePollData)`
- `Message#endPoll()`
- `MessageChannel#endPollById(String)`
- `Message#retrievePollVoters(long)`
- `MessageChannel#retrievePollVoters(String|long, long)`
- `MessagePollAdd/RemoveEvent`

### Internal Additions

- `Checks.notLonger(Duration, Duration, TimeUnit, String)`

### Test Additions

- `<T>CheckAssertions`  
  A new way to test checks, similar to other assertion chains.
- `AbstractSocketHandlerTest`  
  Tests for socket handlers

### Example: Sending poll

```java
channel.sendMessage("Hello guys! Check my poll:")
  .setPoll(
    MessagePollData.builder("Which programming language is better?")
      .addAnswer("Java", Emoji.fromFormatted("<:java:1006323566314274856>"))
      .addAnswer("Kotlin", Emoji.fromFormatted("<:kotlin:295940257797636096>"))
      .build())
  .queue()
```


src/main/java/net/dv8tion/jda/api/utils/messages/MessagePollBuilder.java
```diff
@@ -0,0 +1,271 @@
...
+     *         If the duration is null, less than 1 hour, or longer than {@value  MessagePoll#MAX_DURATION_HOURS} hours (7 days)
+     *
+     * @return The updated builder
+     */
+    @Nonnull
+    public MessagePollBuilder setDuration(@Nonnull Duration duration)
+    {
+        Checks.notNull(duration, "Duration");
+        Checks.positive(duration.toHours(), "Duration");
+        Checks.check(duration.toHours() <= MessagePoll.MAX_DURATION_HOURS, "Poll duration may not be longer than 7 days. Provided: %d hours", duration.toHours());
```
Andre601, line 121: I feel like this text can be confusing when it speaks about text, but then gives Hours... Maybe have either the max hours mentioned, or display duration as days?

src/main/java/net/dv8tion/jda/api/utils/messages/MessagePollBuilder.java
```diff
@@ -0,0 +1,271 @@
...
+     * @throws IllegalArgumentException
+     *         If the time unit is null or the duration is not between 1 and {@value  MessagePoll#MAX_DURATION_HOURS} hours (7 days) long
+     *
+     * @return The updated builder
+     */
+    @Nonnull
+    public MessagePollBuilder setDuration(long duration, @Nonnull TimeUnit unit)
+    {
+        Checks.notNull(unit, "TimeUnit");
+        this.duration = Duration.ofHours(unit.toHours(duration));
```
Andre601, line 147: Shouldn't there be a check for max duration here to ensure it doesn't go beyond the total?

src/main/java/net/dv8tion/jda/api/utils/messages/MessagePollBuilder.java
MinnDevelopment: I don't understand whats confusing here?

src/main/java/net/dv8tion/jda/api/utils/messages/MessagePollBuilder.java
MinnDevelopment: This was supposed to call the other setter, good catch

src/main/java/net/dv8tion/jda/api/utils/messages/MessagePollBuilder.java
Andre601: I mentioned it.
It say can't be higher than 7 days, but returns your input in hours.

Scyye: Maybe "Poll duration may not be longer than 168 hours (7 days). Provided: %d hours" or something similar.
I agree with @Andre601, its a bit confusing to give a limit in a different time unit than what youre actually providing.

src/main/java/net/dv8tion/jda/api/utils/messages/MessagePollBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+     *         If the duration is null, less than 1 hour, or longer than {@value  MessagePoll#MAX_DURATION_HOURS} hours (7 days)
+     *
+     * @return The updated builder
+     */
+    @Nonnull
+    public MessagePollBuilder setDuration(@Nonnull Duration duration)
+    {
+        Checks.notNull(duration, "Duration");
+        Checks.positive(duration.toHours(), "Duration");
+        Checks.check(duration.toHours() <= MessagePoll.MAX_DURATION_HOURS, "Duration may not be longer than 168 hours (= 7 days). Provided: %d hours", duration.toHours());
```
Andre601, line 121: Wouldn't it also be better to just pass the `MAX_DURATION_HOURS` value in the error message here? That way you won't have to adjust it when duration may change.

MinnDevelopment: I doubt this will change.

Scyye: yea, but it still might be a good idea to not have it hardcoded, just in case.

is there any reason *to* have it hardcoded?

PR #2658, opened 2024-04-19
Fix skipping in thread pagination

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Attempts to fix thread pagination when using `skipTo`, also stops archived threads from being cached, since I assume they would never get removed

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/ThreadChannelPaginationActionImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/ThreadChannelPaginationActionImpl.java
@@ -14,6 +14,7 @@ import net.dv8tion.jda.api.utils.TimeUtil;
 import net.dv8tion.jda.api.utils.data.DataArray;
 import net.dv8tion.jda.api.utils.data.DataObject;
 import net.dv8tion.jda.internal.entities.EntityBuilder;
+import net.dv8tion.jda.internal.entities.GuildImpl;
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/requests/restaction/pagination/ThreadChannelPaginationActionImpl.java
```diff
@@ -59,19 +61,16 @@ protected String getPaginationLastEvaluatedKey(long lastId, ThreadChannel last)
...
-        // this should be redundant, due to calling this with PaginationAction#getLast() as last param,
-        // but let's have this here.
-        if (last == null)
+        if (last != null)
+            // If threads were retrieved
+            // OffsetDateTime#toString() is defined to be ISO8601, needs no helper method.
+            return last.getTimeArchiveInfoLastModified().toString();
+        else if (lastId != 0)
+            // If the user skipped ahead
+            return TimeUtil.getTimeCreated(lastId).toString();
```
freya022, line 70: I'm not sure if this should even be allowed, it's very easy for the user to try to skip using an ID, when in reality it uses the ID *as a iso8601 timestamp*, requiring to pass the thread archival timestamp, instead of the thread ID.

My use case was to retrieve a channel by ID by making a small window with the pagination, would be *much* better if JDA had a way to retrieve a channel by ID

PR #2659, opened 2024-04-20
Fix `CommandInteractionPayload#getCommandString` in autocomplete interactions

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

When printing a command string, in an autocomplete interaction, an `IllegalStateException` could occur when an entity is in that command, as Discord does not send resolved entities there

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/commands/CommandInteractionPayload.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/commands/CommandInteractionPayload.java
@@ -149,6 +149,26 @@ public interface CommandInteractionPayload extends Interaction
         for (OptionMapping o : getOptions())
         {
             builder.append(" ").append(o.getName()).append(": ");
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/interactions/commands/CommandInteractionPayload.java
```diff
@@ -149,6 +149,26 @@ default String getCommandString()
...
+                    continue;
+                case USER:
+                    builder.append("<@").append(o.getAsLong()).append(">");
+                    continue;
+                case ROLE:
+                    builder.append("<@&").append(o.getAsLong()).append(">");
+                    continue;
+                case MENTIONABLE:
+                    builder.append(o.getAsLong());
+                    continue;
```
MinnDevelopment, line 168: I would just only put the id, in every case.

PR #2676, opened 2024-05-05
Refactor `LocalizationMapper` to enable more checks

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [X] Other: Tests

Closes Issue: NaN

## Description

Refactors the localization mapper since we now have the full objects, instead of the raw data.

This enables the checks on options and choices, which wasn't possible before.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/interactions/command/localization/LocalizationMapper.java
+++ b/src/main/java/net/dv8tion/jda/internal/interactions/command/localization/LocalizationMapper.java
@@ -18,16 +18,12 @@ package net.dv8tion.jda.internal.interactions.command.localization;
 
 import net.dv8tion.jda.api.interactions.DiscordLocale;
 import net.dv8tion.jda.api.interactions.commands.Command;
-import net.dv8tion.jda.api.interactions.commands.OptionType;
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/internal/interactions/command/localization/LocalizationMapper.java
```diff
@@ -160,26 +155,7 @@ private void trySetTranslation(LocalizationMap localizationMap, String finalComp
...
-                {
-                    Checks.check(locale != DiscordLocale.UNKNOWN, "Localization function returned a map with an 'UNKNOWN' DiscordLocale");
-
-                    localizationMap.put(locale.getLocale(), localizedValue);
-                });
-            }
-            catch (Exception e)
-            {
-                throw new RuntimeException("An uncaught exception occurred while using a LocalizationFunction, localization key: '" + key + "'", e);
+                throw new RuntimeException("Unable to set translations from '" + localizationFunction.getClass().getName() + "' with key '" + key + "'", e);
```
MinnDevelopment, line 158: This should probably use a more concrete exception type than RuntimeException

PR #2682, opened 2024-05-25
Add Android Compatibility by removing UNICODE_CHARACTER_CLASS flag which DOES NOT run on Android

sss-ryun:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Android throws `Caused by: java.lang.IllegalArgumentException: UNICODE_CHARACTER_CLASS flag not supported` if you use JDA in Android. This prevents it from being used inside Android.

Why? Well, I want to run my Discord Bot in Android because it's power efficient and has its own UPS. I have succeeded in doing as such

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/utils/Checks.java
+++ b/src/main/java/net/dv8tion/jda/internal/utils/Checks.java
@@ -40,8 +40,8 @@ import java.util.stream.Stream;
 
 public class Checks
 {
-    public static final Pattern ALPHANUMERIC_WITH_DASH = Pattern.compile("[\\w-]+", Pattern.UNICODE_CHARACTER_CLASS);
-    public static final Pattern ALPHANUMERIC = Pattern.compile("\\w+", Pattern.UNICODE_CHARACTER_CLASS);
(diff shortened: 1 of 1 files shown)


Andre601:
Is there a general benefit for the majority of users?

Because if it only benefits you can you just make and use your fork that has these changes...

sss-ryun:
> Is there a general benefit for the majority of users?
> 
> Because if it only benefits you can you just make and use your fork that has these changes...

Yes. The majority of bot developers run small bots according to Discord, so it would be beneficial to those who want to run their bot without keeping their computers on all the time which may not be possible, and the battery of Android Phones act as UPS in case of power lost. Overall, I think this opens up a new development path for Discord Bot developers

freya022:
Android [always uses unicode character classes](https://stackoverflow.com/a/37483376), however, you cannot remove that flag from these patterns, as it would make any non-latin character be rejected, which would otherwise be valid for Discord.

And I think I'd rather have non-latin users be able to make Discord bots rather than try to please Google that chose not to support such flag

sss-ryun:
I understand. I'll try to find a middle ground when I'm not as busy. I don't want to maintain a fork

PR #2683, opened 2024-05-27
Add `JDA#listenOnce`

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds an helper method to listen for an event, once.

#### Example: Listening to a message from a channel and a user, after using a slash command:
     
```java
final Duration timeout = Duration.ofSeconds(5);
event.reply("Reply in " + TimeFormat.RELATIVE.after(timeout) + " if you can!")
       .setEphemeral(true)
       .queue();

event.getJDA().listenOnce(MessageReceivedEvent.class)
        .filter(messageEvent -> messageEvent.getChannel().getIdLong() == event.getChannel().getIdLong())
        .filter(messageEvent -> messageEvent.getAuthor().getIdLong() == event.getUser().getIdLong())
        .timeout(timeout, () -> {
            event.getHook().editOriginal("Timeout!").queue();
        })
        .submit()
        .onSuccess(messageEvent -> {
            event.getHook().editOriginal("You sent: " + messageEvent.getMessage().getContentRaw()).queue();
        });
```

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/Once.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/Once.java
@@ -3,6 +3,7 @@ package net.dv8tion.jda.api.utils;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.events.GenericEvent;
 import net.dv8tion.jda.api.hooks.EventListener;
+import net.dv8tion.jda.api.hooks.SubscribeEvent;
 import net.dv8tion.jda.api.utils.concurrent.Task;
 import net.dv8tion.jda.internal.utils.Checks;
 import net.dv8tion.jda.internal.utils.concurrent.task.GatewayTask;
@@ -89,6 +90,7 @@ public class Once<E extends GenericEvent> implements EventListener
     }
 
     @Override
+    @SubscribeEvent
     public void onEvent(@Nonnull GenericEvent event)
     {
         if (!eventType.isInstance(event))
--- a/src/main/java/net/dv8tion/jda/api/utils/Once.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/Once.java
@@ -1,0 +1,212 @@
+package net.dv8tion.jda.api.utils;
+
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.events.GenericEvent;
+import net.dv8tion.jda.api.hooks.EventListener;
+import net.dv8tion.jda.api.utils.concurrent.Task;
+import net.dv8tion.jda.internal.utils.Checks;
+import net.dv8tion.jda.internal.utils.concurrent.task.GatewayTask;
+
+import javax.annotation.CheckReturnValue;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.*;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+/**
+ * Helper class to listen to an event, once.
+ *
+ * @param <E> Type of the event listened to
+ *
+ * @see   JDA#listenOnce(Class)
+ */
+public class Once<E extends GenericEvent> implements EventListener
+{
+    private final JDA jda;
+    private final Class<E> eventType;
+    private final List<Predicate<? super E>> filters;
+    private final CompletableFuture<E> future;
+    private final GatewayTask<E> task;
+    private final ScheduledFuture<?> timeoutFuture;
+    private final Runnable timeoutCallback;
+
+    private Once(@Nonnull Once.Builder<E> builder)
+    {
+        this.jda = builder.jda;
+        this.eventType = builder.eventType;
+        this.filters = new ArrayList<>(builder.filters);
+        this.timeoutCallback = builder.timeoutCallback;
+
+        this.future = new CompletableFuture<>();
+        this.task = createTask(future);
+        this.timeoutFuture = scheduleTimeout(builder.timeout, future);
+    }
+
+    @Nonnull
+    private GatewayTask<E> createTask(@Nonnull CompletableFuture<E> future)
+    {
+        final GatewayTask<E> task = new GatewayTask<>(future, () ->
+        {
+            // On cancellation, throw cancellation exception and cancel timeout
+            jda.removeEventListener(this);
+            future.completeExceptionally(new CancellationException());
+            if (timeoutFuture != null)
+                timeoutFuture.cancel(false);
+        });
+        task.onSetTimeout(e ->
+        {
+            throw new UnsupportedOperationException("You must set the timeout on Once.Builder#timeout");
+        });
+        return task;
+    }
+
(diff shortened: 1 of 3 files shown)


duncte123:
There is an `addEventLister` method on the shardManager, maybe this helper method could be added to that as well?

freya022:
> There is an `addEventLister` method on the shardManager, maybe this helper method could be added to that as well?

This would make things more complicated, if someone has a use case where you'd need to listen once for an event across all shards, let me know

AshlyneS:
Just a quick thought, but I believe these would also be benifical. 💕
```java
<E extends GenericEvent> Task<E> listenOnce(Class<E> type, Predicate<? super E> filter);

<E extends GenericEvent> Task<E> listenOnce(Class<E> type, Predicate<? super E> filter, Duration duration, @Nullable Runnable callback);
``` 

Otherwise at least some way to `listenOnce(new SomethingCommonlyImplemented(argA, argB, etc))`

freya022:
> Just a quick thought, but I believe these would also be benifical. 💕
> 
> ```java
> <E extends GenericEvent> Task<E> listenOnce(Class<E> type, Predicate<? super E> filter);
> 
> <E extends GenericEvent> Task<E> listenOnce(Class<E> type, Predicate<? super E> filter, Duration duration, @Nullable Runnable callback);
> ```
> 
> Otherwise at least some way to
> 
> ```java
> jda.listenOnce(new SomethingCommonlyImplemented(argA, argB, etc));
> ```

After [discussing](https://ptb.discord.com/channels/125227483518861312/1236273393230286898/1257631039065423915):
- The two first overloads would just be shortcuts, except they bring more confusion on what each parameter is, this will not be added.
- The last overload might be used in interactions which require waiting for user feedback, each class would configure the event, filters and the outcome of each step. This won't be added as JDA remains a low-level enough library, this type of usage can easily be achieved using an utility class such as:
```java
public abstract class OnceFactory<E extends GenericEvent> {
    private final JDA jda;

    public OnceFactory(JDA jda) {
        this.jda = jda;
    }

    protected Class<E> getType();

    protected void configure(Once.Builder<E> builder);

    public Task<E> create() {
        final var builder = jda.listenOnce(getType());
        configure(builder);

        return builder.submit();
    }
}
```

src/main/java/net/dv8tion/jda/api/utils/Once.java
```diff
@@ -0,0 +1,214 @@
...
+    private ScheduledFuture<?> scheduleTimeout(@Nullable Duration timeout, @Nonnull CompletableFuture<E> future)
+    {
+        if (timeout == null) return null;
+
+        return jda.getGatewayPool().schedule(() ->
+        {
+            // On timeout, throw timeout exception and run timeout callback
+            jda.removeEventListener(this);
+            future.completeExceptionally(new TimeoutException());
+            if (timeoutCallback != null)
```
MinnDevelopment, line 78: Use the return value of `completeExceptionally` to check if the future was actually completed by this call. This way you prevent race-conditions where both callbacks can run. Do the same below for the success cases.

src/main/java/net/dv8tion/jda/api/utils/Once.java
```diff
@@ -0,0 +1,214 @@
...
+            {
+                try
+                {
+                    timeoutCallback.run();
+                }
+                catch (Throwable e)
+                {
+                    future.completeExceptionally(e);
```
MinnDevelopment, line 86: This never does anything, you can't complete a future twice.

src/main/java/net/dv8tion/jda/api/utils/Once.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @Override
+    @SubscribeEvent
+    public void onEvent(@Nonnull GenericEvent event)
+    {
+        if (!eventType.isInstance(event))
+            return;
+        final E casted = eventType.cast(event);
+        if (filters.stream().allMatch(p -> p.test(casted)))
```
MinnDevelopment, line 99: This needs to do exception handling, if any of the filters or callbacks throw, we should de-register the listener as well.

MinnDevelopment (changes requested):

src/main/java/net/dv8tion/jda/api/utils/Once.java
```diff
@@ -0,0 +1,268 @@
...
+        if (timeout == null) return null;
+        if (timeoutPool == null) timeoutPool = jda.getGatewayPool();
+
+        return timeoutPool.schedule(() ->
+        {
+            // On timeout, throw timeout exception and run timeout callback
+            jda.removeEventListener(this);
+            if (!future.completeExceptionally(new TimeoutException()))
+                // Return if the future was already completed
+                return;
```
MinnDevelopment, line 84:
```suggestion
            if (!future.completeExceptionally(new TimeoutException()))
                return;
```

src/main/java/net/dv8tion/jda/api/utils/Once.java
```diff
@@ -0,0 +1,268 @@
...
+            if (timeoutCallback != null)
+            {
+                try
+                {
+                    timeoutCallback.run();
+                }
+                catch (Throwable e)
+                {
+                    LOG.error("An error occurred while running the timeout callback", e);
```
MinnDevelopment, line 93: Rethrow if this is an `Error`

src/main/java/net/dv8tion/jda/api/utils/Once.java
```diff
@@ -0,0 +1,268 @@
...
+            {
+                if (timeoutFuture != null)
+                    timeoutFuture.cancel(false);
+                event.getJDA().removeEventListener(this);
+                future.complete(casted);
+            }
+        }
+        catch (Throwable e)
+        {
+            if (future.completeExceptionally(e))
```
MinnDevelopment, line 118: Rethrow errors

src/main/java/net/dv8tion/jda/api/utils/Once.java
```diff
@@ -0,0 +1,214 @@
...
+{
+    private final JDA jda;
+    private final Class<E> eventType;
+    private final List<Predicate<? super E>> filters;
+    private final CompletableFuture<E> future;
+    private final GatewayTask<E> task;
+    private final ScheduledFuture<?> timeoutFuture;
+    private final Runnable timeoutCallback;
+
+    private Once(@Nonnull Once.Builder<E> builder)
```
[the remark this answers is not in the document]
MinnDevelopment: I'd prefer doing this the standard way of passing parameters. As is, this class is very difficult to extend.

src/main/java/net/dv8tion/jda/api/utils/Once.java
```diff
@@ -0,0 +1,271 @@
+package net.dv8tion.jda.api.utils;
```
MinnDevelopment, line 1: Missing copyright

MinnDevelopment (approved):

PR #2684, opened 2024-05-28
Add abstract `createCopy` method in `SelectMenu`

Kaktushose:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds an abstract `createCopy` method in the `SelectMenu` interface. This allows to create a copy of a SelectMenu without knowing the actual type

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/components/selections/StringSelectMenu.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/components/selections/StringSelectMenu.java
@@ -100,6 +100,7 @@ public interface StringSelectMenu extends SelectMenu
      */
     @Nonnull
     @CheckReturnValue
+    @Override
     default Builder createCopy()
     {
         //noinspection ConstantConditions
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/api/interactions/components/selections/StringSelectMenu.java
```diff
@@ -100,6 +100,7 @@ default StringSelectMenu withDisabled(boolean disabled)
      */
     @Nonnull
     @CheckReturnValue
+    @Override
```
MinnDevelopment, line 103: I would move the docs to the abstract method here, otherwise that one has no docs and the implementations just duplicate them

Kaktushose: Shouldn't we keep the docs for the other methods so as to keep the links to the respective builder?

MinnDevelopment: Yes, good idea.

PR #2711, opened 2024-08-02
Fix deserialization of interaction components

WitchBoo:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

It fixes the deserialization of some interaction components. Actually, I identified two classes that are deserializing in a wrong way.

**ModalImpl**: isn't using custom_id if available. When converting to data, it saves the id as custom_id

**TextInputImpl**: serializes minimum length as min_length but actually uses min_**L**ength

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/interactions/modal/ModalImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/interactions/modal/ModalImpl.java
@@ -37,7 +37,7 @@ public class ModalImpl implements Modal
 
     public ModalImpl(DataObject object)
     {
-        this.id = object.getString("id");
+        this.id = object.getString("custom_id");
         this.title = object.getString("title");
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/interactions/modal/ModalImpl.java
```diff
@@ -37,7 +37,11 @@ public class ModalImpl implements Modal
 
     public ModalImpl(DataObject object)
     {
-        this.id = object.getString("id");
+        if (object.hasKey("custom_id")) {
+            this.id = object.getString("custom_id");
+        } else {
+            this.id = object.getString("id");
```
MinnDevelopment, line 43: I don't think `"id"` should be used at all here.

MinnDevelopment:
Thanks!

PR #2721, opened 2024-08-22
Remove extension from received audio packets

davidffa:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

With the new update (v5.1.0) that uses the new voice encryption modes, when the bot receives audio from regular users (not bots) it throws opus decoder errors in the debug logs.
The issue is that the header extension (present in user's audio packets) is not being stripped from the encoded (but decrypted) audio packet.
This PR fixes that by offseting the decrypted audio ByteBuffer by `4 * extension length in the RTP header` .


MinnDevelopment:
Please also adjust the test with these changes

PR #2728, opened 2024-09-12
Fix parsing error for AutoModResponse

StasiumDev:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Issue
While rewriting a bot from Discord.js to Discord JDA, it was discovered that `AutoModResponseImpl` always returns 0 and null for `customMessage`, `timeoutDuration` and `channel`, respectively, despite these values being set.

## Cause
Upon investigation, it was found that these values are being accessed incorrectly. They are actually nested within a sub-object called `metadata`.

## Solution
This pull request addresses the issue by correctly accessing the `custom_message`, `duration_seconds` and `channel_id` values from the `metadata` object in `AutoModResponseImpl`.

## Changes
- Updated `AutoModResponseImpl` to access `custom_message`, `duration_seconds` and `channel_id` from the `metadata` object.
- Ensured that the correct values are now being returned when these properties are accessed.

Please review and let me know if any further changes or information are needed.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/entities/automod/AutoModResponseImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/entities/automod/AutoModResponseImpl.java
@@ -66,10 +66,11 @@ public class AutoModResponseImpl implements AutoModResponse
 
     public AutoModResponseImpl(Guild guild, DataObject json)
     {
+        DataObject metadata = json.getObject("metadata");
         this.type = AutoModResponse.Type.fromKey(json.getInt("type", -1));
-        this.channel = guild.getChannelById(GuildMessageChannel.class, json.getUnsignedLong("channel_id", 0L));
-        this.customMessage = json.getString("custom_message", null);
-        this.timeoutDuration = json.getUnsignedLong("duration_seconds", 0L);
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/entities/automod/AutoModResponseImpl.java
```diff
@@ -66,10 +66,11 @@ public AutoModResponseImpl(Type type, Duration duration)
 
     public AutoModResponseImpl(Guild guild, DataObject json)
     {
+        DataObject metadata = json.getObject("metadata");
```
MinnDevelopment, line 69: This will throw an exception when `metadata` is missing, which is possible in the API. To handle this, you can use `optObject("metadata")` instead.

StasiumDev: I just pushed an Update, you reckon that would be enough? Since `getString` and `getUnsignedLong` have a defaultValue when the key doesn't exist?

MinnDevelopment (approved):

PR #2729, opened 2024-09-12
Add `Guild#retrieveMemberVoiceState`

PascalNB:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Discord recently added an API endpoint to get a user's voice state in a guild (https://discord.com/developers/docs/resources/voice).
Adds `Guild#retrieveMemberVoiceState`, `Guild#retrieveMemberVoiceStateById(long)` and `Guild#retrieveMemberVoiceStateById(String)` that send a request and create a new `GuildVoiceState` if not already cached.

Also makes changes to `EntityBuilder` to allow the creation of voice states from JSON and `Route` for the new endpoint.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2667,7 +2667,7 @@ public interface Guild extends IGuildChannelContainer<GuildChannel>, ISnowflake
     default CacheRestAction<GuildVoiceState> retrieveMemberVoiceState(UserSnowflake user)
     {
         Checks.notNull(user, "User");
-        return retrieveMemberVoiceStateById(user.getId());
+        return retrieveMemberVoiceStateById(user.getIdLong());
     }
 
     /**
--- a/src/main/java/net/dv8tion/jda/api/entities/Guild.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Guild.java
@@ -2587,13 +2587,6 @@ public interface Guild extends IGuildChannelContainer<GuildChannel>, ISnowflake
     /**
      * Load the member's voice state for the specified user.
      *
-     * <br>If the member is already loaded it will be retrieved from {@link #getMemberById(long)} and
-     * the voice state is immediately provided by {@link Member#getVoiceState()}.
-     * The cache consistency directly relies on the enabled {@link GatewayIntent GatewayIntents} as {@link GatewayIntent#GUILD_MEMBERS GatewayIntent.GUILD_MEMBERS}
-     * and {@link GatewayIntent#GUILD_VOICE_STATES GatewayIntent.GUILD_VOICE_STATES} are required to keep the cache updated with the latest information.
-     * You can use {@link CacheRestAction#useCache(boolean) useCache(false)} to always
-     * make a new request, which is the default behavior if the required intents are disabled.
-     *
      * <p>Possible {@link net.dv8tion.jda.api.exceptions.ErrorResponseException ErrorResponseExceptions} include:
      * <ul>
      *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_VOICE_STATE}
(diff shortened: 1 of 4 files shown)


src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -2584,6 +2584,91 @@ default RestAction<RichCustomEmoji> retrieveEmoji(@Nonnull CustomEmoji emoji)
     @Nonnull
     List<GuildVoiceState> getVoiceStates();
 
+    /**
+     * Load the member's voice state for the specified user.
+     * <br>If the member is already loaded it will be retrieved from {@link #getMemberById(long)} and
```
DV8FromTheWorld, line 2589:
```suggestion
     * Load the member's voice state for the specified user.
     *
     * <br>If the member is already loaded it will be retrieved from {@link #getMemberById(long)} and
```

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -2584,6 +2584,91 @@ default RestAction<RichCustomEmoji> retrieveEmoji(@Nonnull CustomEmoji emoji)
...
+     * @throws IllegalArgumentException
+     *         If provided with null
+     *
+     * @return {@link RestAction} - Type: {@link GuildVoiceState}
+     */
+    @Nonnull
+    default CacheRestAction<GuildVoiceState> retrieveMemberVoiceState(UserSnowflake user)
+    {
+        Checks.notNull(user, "User");
+        return retrieveMemberVoiceStateById(user.getId());
```
DV8FromTheWorld, line 2669:
```suggestion
        return retrieveMemberVoiceStateById(user.getIdLong());
```

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -616,13 +616,13 @@ public MemberImpl createMember(GuildImpl guild, DataObject memberJson, DataObjec
...
-                : Helpers.toTimestamp(memberJson.getString("premium_since"));
+                    ? 0
+                    : Helpers.toTimestamp(memberJson.getString("premium_since"));
             member.setBoostDate(boostTimestamp);
 
             long timeOutTimestamp = memberJson.isNull("communication_disabled_until")
-                ? 0
-                : Helpers.toTimestamp(memberJson.getString("communication_disabled_until"));
+                    ? 0
+                    : Helpers.toTimestamp(memberJson.getString("communication_disabled_until"));
```
DV8FromTheWorld, line 625: Unneeded change

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1153,6 +1153,32 @@ public List<GuildVoiceState> getVoiceStates()
...
+    @Nonnull
+    @Override
+    public CacheRestAction<GuildVoiceState> retrieveMemberVoiceStateById(long id)
+    {
+        JDAImpl jda = getJDA();
+        return new DeferredRestAction<>(jda, GuildVoiceState.class,
+                () ->
+                {
+                    MemberImpl member = (MemberImpl) getMemberById(id);
+                    return member == null ? null : member.getVoiceState();
```
DV8FromTheWorld, line 1165: If memory serves, you can have members in the cache even without the intents that would keep it up to date. Same for the voice state.
This would mean that this thing would potentially provide out of date information.

@MinnDevelopment am I remembering this correctly?

MinnDevelopment: You can't have the voice state cache enabled without enabling the intent. However, the opposite direction is problematic. When the intent is enabled and the member is removed from cache, the voice state information is also lost, even if the member is cached again later.

I would probably just never fallback on cache here.

DV8FromTheWorld: Okay, so in conclusion this should _not_ be a `CachedRestAction`

PascalNB: I changed it to a normal `RestAction`

src/main/java/net/dv8tion/jda/internal/entities/GuildImpl.java
```diff
@@ -1153,6 +1153,23 @@ public List<GuildVoiceState> getVoiceStates()
...
+    @CheckReturnValue
+    public RestAction<GuildVoiceState> retrieveMemberVoiceStateById(long id)
+    {
+        JDAImpl jda = getJDA();
+        Route.CompiledRoute route = Route.Guilds.GET_VOICE_STATE.compile(getId(), Long.toUnsignedString(id));
+        return new RestActionImpl<>(jda, route, (response, request) ->
+        {
+            EntityBuilder entityBuilder = jda.getEntityBuilder();
+            DataObject voiceStateData = response.getObject();
+            MemberImpl member = entityBuilder.createMember(this, voiceStateData.getObject("member"), voiceStateData, null);
```
MinnDevelopment, line 1167: I would not pass the voice state to `createMember` here since it will become stale and inconsistent.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -658,39 +658,47 @@ public MemberImpl createMember(GuildImpl guild, DataObject memberJson, DataObjec
...
+
+    private void updateGuildVoiceState(GuildVoiceStateImpl oldVoiceState, DataObject newVoiceStateJson, MemberImpl member)
+    {
+        Guild guild = member.getGuild();
 
-        final long channelId = voiceStateJson.getLong("channel_id");
+        final long channelId = newVoiceStateJson.getLong("channel_id");
         AudioChannel audioChannel = (AudioChannel) guild.getGuildChannelById(channelId);
         if (audioChannel != null)
             ((AudioChannelMixin<?>) audioChannel).getConnectedMembersMap().put(member.getIdLong(), member);
```
MinnDevelopment, line 682: We should avoid updating the channel connected members cache unless we actually cache voice states, otherwise this cache is incredibly unreliable.

MinnDevelopment (approved):

PR #2738, opened 2024-09-21
Support sending voice messages

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This adds support for sending voice messages.

### Example

```java
channel.sendFiles(
  FileUpload.fromData(audioFile)
    .asVoiceMessage(MediaType.parse("audio/ogg"), waveform, 10.5)
).queue();
```

The `waveform` must be a low frequency sampling of the audio file, with at most 256 samples.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/FileUpload.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/FileUpload.java
@@ -422,7 +422,10 @@ public class FileUpload implements Closeable, AttachedFile
      */
     public boolean isVoiceMessage()
     {
-        return this.mediaType.type().equals("audio");
+        return this.mediaType.type().equals("audio")
+            && this.durationSeconds > 0.0
+            && this.waveform != null
+            && this.waveform.length > 0;
     }
 
     /**
--- a/src/main/java/net/dv8tion/jda/api/utils/FileUpload.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/FileUpload.java
@@ -36,6 +36,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.OpenOption;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.Base64;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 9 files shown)


src/main/java/net/dv8tion/jda/api/utils/FileUpload.java
```diff
@@ -358,6 +363,45 @@ public FileUpload setDescription(@Nullable String description)
...
+    }
+
+    /**
+     * Whether this attachment is a valid voice message attachment.
+     *
+     * @return True, if this is a voice message attachment.
+     */
+    public boolean isVoiceMessage()
+    {
+        return this.mediaType.type().equals("audio");
```
MinnDevelopment, line 402: This should also check waveform and duration, to make sure this isn't true by mistake.

src/main/java/net/dv8tion/jda/api/utils/FileUpload.java
```diff
@@ -358,6 +363,45 @@ public FileUpload setDescription(@Nullable String description)
...
+     * @param  durationSeconds
+     *         The actual duration of the audio data in seconds.
+     *
+     * @throws IllegalArgumentException
+     *         If null is provided or the waveform is not between 1 and 256 bytes long.
+     *
+     * @return The same FileUpload instance configured as a voice message attachment
+     */
+    @Nonnull
+    public FileUpload asVoiceMessage(@Nonnull MediaType mediaType, @Nonnull byte[] waveform, double durationSeconds)
```
MinnDevelopment, line 382: Duration should be a `Duration` instead

src/main/java/net/dv8tion/jda/api/utils/FileUpload.java
```diff
@@ -358,6 +364,70 @@ public FileUpload setDescription(@Nullable String description)
         return this;
     }
 
+    /**
+     * Turns this attachment into a voice message with the provided waveform.
+     *
+     * @param  mediaType
+     *         The audio type for the attached audio file. Should be {@code audio/ogg} or similar.
+     * @param  waveform
+     *         The waveform of the audio, which is a low frequency sampling up to 256 bytes.
```
duncte123, line 373: Would it be a good idea to add a bit of sample code to get this waveform from the file?

PR #2743, opened 2024-09-30
Fix ScheduledEventManagerImpl::setStatus

raul1ro:
[documentation]: https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-object-valid-guild-scheduled-event-status-transitions
## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description
The check `status != ScheduledEvent.Status.SCHEDULED && getScheduledEvent().getStatus() != ScheduledEvent.Status.ACTIVE` is not correct.
According to Discord [documentation] you can set ACTIVE to COMPLETED, but if the event was ACTIVE, the `check` was always false and throwing error.

New checks:
 - Current status should be different than COMPLETED or CANCELED
 - If current status is SCHEDULED, the new one can be only ACTIVE or CANCELED
 - If current status is ACTIVE, the new one can be only COMPLETED

I also updated the first check which verifies the new status to not be UNKNOWN.
I added the condition that the new status to not be SCHEDULED too.

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/managers/ScheduledEventManagerImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/managers/ScheduledEventManagerImpl.java
@@ -165,10 +165,11 @@ public class ScheduledEventManagerImpl extends ManagerBase<ScheduledEventManager
     public ScheduledEventManager setStatus(@Nonnull ScheduledEvent.Status newStatus)
     {
         Checks.notNull(newStatus, "Status");
-        switch (newStatus){
+        switch (newStatus)
+        {
             case SCHEDULED:
             case UNKNOWN:
-                throw new IllegalArgumentException("Invalid new status: " + newStatus);
+                throw new IllegalArgumentException("Cannot change scheduled event status to " + newStatus);
         }
 
         //get the current status of the event. multiple-usage
@@ -180,7 +181,7 @@ public class ScheduledEventManagerImpl extends ManagerBase<ScheduledEventManager
                 //event is scheduled -> new status can be only active or cancel
                 Checks.check(
                         newStatus == ScheduledEvent.Status.ACTIVE || newStatus == ScheduledEvent.Status.CANCELED,
-                        "Cannot perform status update! A scheduled event can only be set to active or canceled status."
+                        "Cannot perform status update! A scheduled event with status SCHEDULED can only be set to ACTIVE or CANCELED status."
                 );
                 break;
 
@@ -188,7 +189,7 @@ public class ScheduledEventManagerImpl extends ManagerBase<ScheduledEventManager
                 //event is active -> new status can be only completed
                 Checks.check(
                         newStatus == ScheduledEvent.Status.COMPLETED,
-                        "Cannot perform status updated! An active event can only be set to completed status."
+                        "Cannot perform status updated! A scheduled event with status ACTIVE can only be set to COMPLETED status."
                 );
                 break;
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/managers/ScheduledEventManagerImpl.java
```diff
@@ -162,12 +162,43 @@ public ScheduledEventManager setEndTime(@Nonnull TemporalAccessor endTime)
...
+        //get the current status of the event. multiple-usage
+        ScheduledEvent.Status currentStatus = getScheduledEvent().getStatus();
+
+        switch (currentStatus)
+        {
+            case SCHEDULED:
+                //event is scheduled -> new status can be only active or cancel
+                Checks.check(
+                        newStatus == ScheduledEvent.Status.ACTIVE || newStatus == ScheduledEvent.Status.CANCELED,
+                        "Cannot perform status update! A scheduled event can only be set to active or canceled status."
```
MinnDevelopment, line 183:
```suggestion
                        "Cannot perform status update! A scheduled event with status SCHEDULED can only be set to ACTIVE or CANCELED status."
```

src/main/java/net/dv8tion/jda/internal/managers/ScheduledEventManagerImpl.java
```diff
@@ -162,12 +162,43 @@ public ScheduledEventManager setEndTime(@Nonnull TemporalAccessor endTime)
...
+                );
+                break;
+
+            case ACTIVE:
+                //event is active -> new status can be only completed
+                Checks.check(
+                        newStatus == ScheduledEvent.Status.COMPLETED,
+                        "Cannot perform status updated! An active event can only be set to completed status."
```
MinnDevelopment, line 191:
```suggestion
                        "Cannot perform status updated! A scheduled event with status ACTIVE can only be set to COMPLETED status."
```

src/main/java/net/dv8tion/jda/internal/managers/ScheduledEventManagerImpl.java
```diff
@@ -162,12 +162,43 @@ public ScheduledEventManager setEndTime(@Nonnull TemporalAccessor endTime)
...
     @Override
-    public ScheduledEventManager setStatus(@Nonnull ScheduledEvent.Status status)
+    public ScheduledEventManager setStatus(@Nonnull ScheduledEvent.Status newStatus)
     {
-        Checks.notNull(status, "Status");
-        Checks.check(status != ScheduledEvent.Status.UNKNOWN, "Cannot set the event status to an unknown status!");
-        Checks.check(status != ScheduledEvent.Status.SCHEDULED && getScheduledEvent().getStatus() != ScheduledEvent.Status.ACTIVE, "Cannot perform status update!");
-        this.status = status;
+        Checks.notNull(newStatus, "Status");
+        switch (newStatus){
```
MinnDevelopment, line 168:
```suggestion
        switch (newStatus)
        {
```

src/main/java/net/dv8tion/jda/internal/managers/ScheduledEventManagerImpl.java
```diff
@@ -162,12 +162,43 @@ public ScheduledEventManager setEndTime(@Nonnull TemporalAccessor endTime)
...
+            case SCHEDULED:
+            case UNKNOWN:
+                throw new IllegalArgumentException("Invalid new status: " + newStatus);
```
MinnDevelopment, line 171:
```suggestion
                throw new IllegalArgumentException("Cannot change scheduled event status to " + newStatus);
```

PR #2751, opened 2024-10-14
Add basic support for activities

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds basic support for activities:
- Creating/reading entry point commands (`PrimaryEntryPointCommandData`)
- Receiving/responding to entry point interactions (`PrimaryEntryPointInteractionEvent`)
- Reading the activity instance (`InteractionCallbackResponse#getActivityInstance()` from `InteractionHook#getCallbackResponse()`) created by `ILaunchActivityCallback#replyLaunchActivity()`
- Retrieving an activity instance by ID (`JDA#retrieveActivityInstanceById(String)`)

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/response/InteractionCallbackResponse.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/response/InteractionCallbackResponse.java
@@ -46,7 +46,7 @@ public interface InteractionCallbackResponse
     /**
      * Returns the {@link ActivityInstanceResource} contained in this callback response, or {@code null} if the response was not a launched activity.
      *
-     * @return ActivityInstance contained in this callback response, or null.
+     * @return {@link ActivityInstanceResource} contained in this callback response, or null.
      */
     @Nullable
     ActivityInstanceResource getActivityInstance();


src/main/java/net/dv8tion/jda/api/interactions/commands/build/CommandData.java
```diff
@@ -37,189 +40,43 @@
...
  */
-public interface CommandData extends SerializableData
+public interface CommandData
+        extends INamedCommandData, IScopedCommandData, IRestrictedCommandData, IAgeRestrictedCommandData,
+        SerializableData
 {
-    /**
-     * The maximum length the name of a command can be. ({@value})
-     */
-    int MAX_NAME_LENGTH = 32;
```
DV8FromTheWorld, line 45: Are these variables still accessible via `CommandData.MAX_NAME_LENGTH`? if not, this is technically a breaking change

src/main/java/net/dv8tion/jda/api/interactions/commands/build/Commands.java
```diff
@@ -129,11 +130,37 @@ public static CommandData context(@Nonnull Command.Type type, @Nonnull String na
         return new CommandDataImpl(type, name);
     }
 
+    /**
+     * Create an activity entry point command builder.
```
DV8FromTheWorld, line 134: EntryPoint commands are not activity only. We should not suggest as such

src/main/java/net/dv8tion/jda/api/interactions/commands/build/Commands.java
```diff
@@ -129,11 +130,37 @@ public static CommandData context(@Nonnull Command.Type type, @Nonnull String na
...
+    public static EntryPointCommandData entryPoint(@Nonnull String name, @Nonnull String description)
+    {
+        return new EntryPointCommandDataImpl(name, description);
+    }
 
     /**
      * Parses the provided serialization back into an CommandData instance.
      * <br>This is the reverse function for {@link CommandData#toData()}.
      *
+     * <p><b>Note:</b> This does not take into account {@link #entryPoint(String, String) entry point commands}.
```
DV8FromTheWorld, line 162: Does this work for ContextCommands? If it works for ContextCommands, why not for EntryPoint commands?

src/main/java/net/dv8tion/jda/api/interactions/commands/build/EntryPointCommandData.java
```diff
@@ -0,0 +1,217 @@
...
+     * Sets the handler of this entry point command.
+     * This defines the behavior when this command is used.
+     *
+     * @param  handler
+     *         The handler type for this command
+     *
+     * @return This builder instance, for chaining
+     */
+    @Nonnull
+    EntryPointCommandData setHandler(@Nonnull Handler handler);
```
DV8FromTheWorld, line 96: Similar to previous comment: Needs better docs

src/main/java/net/dv8tion/jda/api/interactions/commands/build/EntryPointCommandData.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Converts the provided {@link Command} into a {@link EntryPointCommandData} instance.
+     *
+     * @param  command
+     *         The command to convert
+     *
+     * @throws IllegalArgumentException
+     *         If null is provided or the command has illegal configuration
```
DV8FromTheWorld, line 105: call out specifically in the docs that this only works for Commands that are of PRIMARY_ENTRY_POINT type

src/main/java/net/dv8tion/jda/internal/interactions/EntryPointCommandDataImpl.java
```diff
@@ -0,0 +1,219 @@
...
+        if (localizationMapper != null) localizationMapper.localizeCommand(this, DataArray.empty());
+
+        return DataObject.empty()
+                .put("type", getType().getId())
+                .put("handler", handler.getValue())
+                .put("name", name)
+                .put("description", description)
+                .put("description_localizations", descriptionLocalizations)
+                .put("nsfw", nsfw)
+                .put("dm_permission", !guildOnly)
```
DV8FromTheWorld, line 65: dm_permission should be deprecated. We instead have `contexts`. Do not use `dm_permission`.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/EntryPointCommandData.java
```diff
@@ -0,0 +1,217 @@
...
+    {
+        Checks.notNull(object, "DataObject");
+
+        Command.Type commandType = Command.Type.fromId(object.getInt("type", 1));
+        Checks.check(commandType == Command.Type.PRIMARY_ENTRY_POINT, "Cannot convert command of type " + commandType + " to EntryPointCommandData!");
+
+        String name = object.getString("name");
+        String description = object.getString("description");
+        EntryPointCommandDataImpl command = new EntryPointCommandDataImpl(name, description);
+        command.setGuildOnly(!object.getBoolean("dm_permission", true));
```
DV8FromTheWorld, line 153: This is wrong. We should be doing things with `context`

src/main/java/net/dv8tion/jda/api/interactions/commands/build/EntryPointCommandData.java
```diff
@@ -0,0 +1,217 @@
...
+    enum Handler
+    {
+        UNKNOWN(-1),
+        /**
+         * Lets this app handle the activity start via an interaction.
+         */
+        APP_HANDLER(1),
+        /**
+         * Lets Discord handle the activity start,
+         * and sends a follow-up message without coordinating with this app.
```
DV8FromTheWorld, line 182:
```suggestion
         * Lets Discord handle the interaction automatically by launching an activity associated with this app
         * and sends a follow-up message into the chat to indicate that an activity was launched.
```

DV8FromTheWorld (commented):
Nice work. Dropped a number of comments, but many are duplicates so when a decision is reached all the code is updated.

I think we should consider moving to "PrimaryEntrypointCommand" for a measure of consistency and code-self-explanation.

src/main/java/net/dv8tion/jda/api/interactions/commands/build/attributes/IScopedCommandData.java
```diff
@@ -0,0 +1,49 @@
...
+     * Sets whether this command is only usable in a guild (Default: false).
+     * <br>This only has an effect if this command is registered globally.
+     *
+     * @param  guildOnly
+     *         Whether to restrict this command to guilds
+     *
+     * @return The builder instance, for chaining
+     */
+    @Nonnull
+    IScopedCommandData setGuildOnly(boolean guildOnly);
```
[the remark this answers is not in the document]
freya022: This is based on 5.2.0, so it wasn't merged yet

src/main/java/net/dv8tion/jda/api/interactions/commands/build/CommandData.java
freya022: Yes, all fields can be accessed from subclasses

src/main/java/net/dv8tion/jda/api/requests/restaction/interactions/LaunchActivityCallbackAction.java
```diff
@@ -0,0 +1,29 @@
...
+package net.dv8tion.jda.api.requests.restaction.interactions;
+
+import net.dv8tion.jda.api.requests.FluentRestAction;
+
+/**
+ * An {@link InteractionCallbackAction} that can be used to launch an activity.
+ *
+ * @see net.dv8tion.jda.api.interactions.callbacks.ILaunchActivityReplyCallback
+ */
+public interface LaunchActivityCallbackAction extends InteractionCallbackAction<Void>, FluentRestAction<Void, LaunchActivityCallbackAction>
```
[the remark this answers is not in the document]
freya022: With the current code, [Discord will return a 204](https://discord.com/developers/docs/interactions/receiving-and-responding#create-interaction-response). After https://github.com/discord-jda/JDA/pull/2798, it will return a 200.

The interaction callback response PR should be merged before this one, so I can add the activity-specific data to it.

freya022:
> I think we should consider moving to "PrimaryEntrypointCommand" for a measure of consistency and code-self-explanation.

Slash command properties are also merged into `Command`, so I put the entrypoint stuff in it too, having a separate `PrimaryEntrypointCommand` subtype would require the user to check if it is an instance of such, while for slash commands it isn't needed at this moment. Are you sure you want to go with it?

PR #2753, opened 2024-10-15
Add soundboard sounds

freya022:
[contributing]: https://jda.wiki/contributing/contributing/

## Pull Request Etiquette

- [X] I have checked the PRs for upcoming features/bug fixes.
- [X] I have read the [contributing guidelines][contributing].

### Changes

- [X] Internal code
- [X] Library interface (affecting end-user code) 
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Implements https://discord.com/developers/docs/change-log#soundboard-api

Also supports create/delete/update via gateway

PR #2755, opened 2024-10-16
Replace `GatewayIntent#GUILD_EMOJIS_AND_STICKERS` with `GUILD_EXPRESSIONS`

freya022:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Updates the intent to match Discord's name, will be clearer for #2753

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/requests/GatewayIntent.java
+++ b/src/main/java/net/dv8tion/jda/api/requests/GatewayIntent.java
@@ -418,10 +418,7 @@ public enum GatewayIntent
             else if (GuildBanEvent.class.isAssignableFrom(event) || GuildUnbanEvent.class.isAssignableFrom(event) || GuildAuditLogEntryCreateEvent.class.isAssignableFrom(event))
                 intents.add(GUILD_MODERATION);
             else if (GenericEmojiEvent.class.isAssignableFrom(event) || GenericGuildStickerEvent.class.isAssignableFrom(event))
-            {
-                intents.add(GUILD_EMOJIS_AND_STICKERS);
                 intents.add(GUILD_EXPRESSIONS);
-            }
             else if (GenericScheduledEventUpdateEvent.class.isAssignableFrom(event))
                 intents.add(SCHEDULED_EVENTS);
             else if (GenericGuildInviteEvent.class.isAssignableFrom(event))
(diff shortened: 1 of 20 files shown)


src/main/java/net/dv8tion/jda/api/requests/GatewayIntent.java
```diff
@@ -100,8 +101,16 @@ public enum GatewayIntent
     GUILD_MODERATION(2),
     /**
      * Custom emoji and sticker add/update/delete events.
+     *
+     * @deprecated Replaced with {@link #GUILD_EXPRESSIONS}
      */
+    @ForRemoval(deadline = "5.2.0")
```
MinnDevelopment, line 107:
```suggestion
    @ForRemoval(deadline = "5.3.0")
    @ReplaceWith("GUILD_EXPRESSIONS")
```

MinnDevelopment:
> If the user does not migrate to the new intent, and has activated cache flags that require the GUILD_EXPRESSIONS intents, the cache flags will get disabled and a warning will be logged.

I don't think it will be disabled, since the check will include both regardless. I'll consider writing a test for this.

MinnDevelopment (changes requested):

freya022:
True, as it uses the raw intents, it sees both the old and new intent enabled.

src/main/java/net/dv8tion/jda/api/requests/GatewayIntent.java
```diff
@@ -407,7 +418,10 @@ else if (GenericUserUpdateEvent.class.isAssignableFrom(event) || GenericGuildMem
             else if (GuildBanEvent.class.isAssignableFrom(event) || GuildUnbanEvent.class.isAssignableFrom(event) || GuildAuditLogEntryCreateEvent.class.isAssignableFrom(event))
                 intents.add(GUILD_MODERATION);
             else if (GenericEmojiEvent.class.isAssignableFrom(event) || GenericGuildStickerEvent.class.isAssignableFrom(event))
+            {
                 intents.add(GUILD_EMOJIS_AND_STICKERS);
+                intents.add(GUILD_EXPRESSIONS);
+            }
```
MinnDevelopment, line 424:
```suggestion
                intents.add(GUILD_EXPRESSIONS);
```

PR #2759, opened 2024-10-26
Make JDA and JDAImpl AutoClosables

mintdaniel42:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [x] Other: added a new way of shutting down the JDA by using try-with-resources statements

Closes Issue: NaN

## Description

Added a new way of shutting down the JDA by using try-with-resources statements. This makes it easier to manage the JDA lifecycle

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDA.java
+++ b/src/main/java/net/dv8tion/jda/api/JDA.java
@@ -74,7 +74,7 @@ import java.util.regex.Matcher;
  *
  * @see JDABuilder
  */
-public interface JDA extends IGuildChannelContainer<Channel>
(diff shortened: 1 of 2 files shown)


MinnDevelopment:
Thank you, but we don't want to encourage using JDA in a try-with-resources due to the cost of creating a JDA instance. It is generally recommended to keep using the same instance instead.

mintdaniel42:
I thought about use cases like main method that consist of initializing the JDA and keeping the main thread blocked or dependency injection frameworks like avaje that automatically close all beans (if they are AutoClosable). But that's okay 😄

PR #2760, opened 2024-10-27
Updating Dependencies

Lewox:
## Description

Updated tink from 1.14.1 to 1.15.0
Updated protobuf-java from 3.25.3 to 3.25.5 (transitive dependency from tink, vulnerable in the current version of tink)

Addressed CVE:
CVE-2024-7254

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: Dependency Update (Vulnerability)

Closes Issue: NaN

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -149,6 +149,9 @@ dependencies {
 
     //Audio crypto libraries
     implementation(libs.tink)
+    implementation(libs.protobuf.java) {
+        because("Overrides transitive dependency from tink to address CVE-2024-7254.")
+    }
 
     //Sets the dependencies for the examples
     configurations["examplesImplementation"].withDependencies {
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -23,8 +23,9 @@ dependencyResolutionManagement {
             library("mockito",               "org.mockito",              "mockito-core"        ).version("5.11.0")
             library("reflections",           "org.reflections",          "reflections"         ).version("0.10.2")
             library("slf4j",                 "org.slf4j",                "slf4j-api"           ).version("2.0.13")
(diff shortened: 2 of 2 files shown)


MinnDevelopment:
We don't use protobuf, so we can just wait for the transitive dependency to be updated instead.

Lewox:
Hi, thanks for checking, would you be open to upgrading `tink` if they update the transitive dependency?

Lewox:
Someone already beat me to update the dependency at tink https://github.com/tink-crypto/tink-java/pull/46 but looking at the Bazel files, it should be using `4.28.0`, which is still vulnerable.

freya022:
As Minn mentioned, JDA does not use protobuf at all, so the version of either tink, or protobuf, does not matter in the slightest.

You can verify this by running a bot with the dependency excluded, I have tried it on my own bot and audio receive works.
By looking at their documentation, I would assume that it is likely used only for [sending over the wire](https://developers.google.com/tink/wire-format).

In addition:
- JDA does not deserialize untrusted data, as all data comes from Discord
- JDA does not expose Tink or any of its transitive dependencies, as it is declared as a [`runtime` dependency](https://mvnrepository.com/artifact/net.dv8tion/JDA/5.1.2), which makes it inaccessible for JDA users. Thus, the user using Tink or Protobuf would be made aware of these issues by themselves
- The Tink PR did fix the issue, the [commit history](https://github.com/tink-crypto/tink-java/commits/7de8d49fccef6361bbfc11661324a095e99aa1cf/maven/tink-java.pom.xml) shows them going from 3.25.3 (current release) -> 4.28.0 -> 4.28.2

And to answer your earlier question, yes, I believe Minn said that we can wait for Tink's version of protobuf to get upgraded

PR #2769, opened 2024-11-09
Handle user in EntityBuilder#createApplicationEmoji as optional

Xirado:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Handle user in `EntityBuilder#createApplicationEmoji` as optional, and use self-user if it was created by `JDA#createApplicationEmoji`

### DIFF
--- a/src/main/java/net/dv8tion/jda/internal/JDAImpl.java
+++ b/src/main/java/net/dv8tion/jda/internal/JDAImpl.java
@@ -693,7 +693,7 @@ public class JDAImpl implements JDA
         return new RestActionImpl<>(this, route, body, (response, request) ->
         {
             final DataObject obj = response.getObject();
-            return entityBuilder.createApplicationEmoji(this, obj);
+            return entityBuilder.createApplicationEmoji(this, obj, true);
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/internal/JDAImpl.java
```diff
@@ -693,7 +693,7 @@ public RestAction<ApplicationEmoji> createApplicationEmoji(@Nonnull String name,
         return new RestActionImpl<>(this, route, body, (response, request) ->
         {
             final DataObject obj = response.getObject();
-            return entityBuilder.createApplicationEmoji(this, obj);
+            return entityBuilder.createApplicationEmoji(this, obj, true);
```
MinnDevelopment, line 696: Instead of this, just inject the self user into `obj`

MinnDevelopment (changes requested):

MinnDevelopment (approved):

PR #2752, opened 2024-10-14
Add premium buttons

freya022:
[contributing]: https://jda.wiki/contributing/contributing/

## Pull Request Etiquette

- [X] I have checked the PRs for upcoming features/bug fixes.
- [X] I have read the [contributing guidelines][contributing].

### Changes

- [X] Internal code
- [X] Library interface (affecting end-user code) 
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Implements changes of https://discord.com/developers/docs/change-log#premium-apps-new-premium-button-style-deep-linking-url-schemes.

Deprecated the premium reply callback, which are replaced with premium buttons.

PR #2770, opened 2024-11-10
Fix NullPointerException in ButtonImpl#toData

Xirado:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fixes a regression introduced in #2752 where the label in ButtonImpl#toData() is incorrectly checked if empty when null


MinnDevelopment:
Please:

- Change the factory method to pass in `""` instead of `null` in addition to this change
- Add a regression test for this

PR #2773, opened 2024-11-14
Prevent firing voice state update events if member is not cached

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Fixes all voice events being fired when only one property changed, for example, a member muting their microphone would fire every event that didn't correspond to the "default" voice state


carlodrift:
Yes!

PR #2778, opened 2024-12-19
Add common interface for interactions with a custom ID

Kaktushose:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: N/A

## Description

Creates a common interface, `CustomIdInteraction` for all interactions that support a user provided custom id, namely components and modals. In the [Discord API](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object), the id is always called `custom_id`, only JDA introduces a distinction between `componentId` and `modalId`. 

### Use case:
Assume an interaction framework that listens for  `GenericInteractionCreateEvent`. Using pattern matching in switch we want to determine if we need a new event handler or use an existing one. 

#### Currently this looks like this:
```java
var handler = switch (genericInteractionCreateEvent) {
    case SlashCommandInteractionEvent _, GenericContextInteractionEvent<?> _,
         CommandAutoCompleteInteractionEvent _ ->
            newHandler();
    case GenericComponentInteractionCreateEvent event when handlerExists(event.getComponentId()) ->
            getHandler(event.getComponentId());
    case ModalInteractionEvent event when handlerExists(event.getModalId()) ->
            getHandler(event.getComponentId());
    case GenericComponentInteractionCreateEvent event when shouldCreate(event.getComponentId()) ->
            newHandler();
    case ModalInteractionEvent event when shouldCreate(event.getModalId()) ->
            newHandler();
    default -> throw new UnsupportedOperationException("Unsupported jda event: %s".formatted(genericInteractionCreateEvent));
};
```

#### With common interface:
```java
var handler = switch (genericInteractionCreateEvent) {
    case SlashCommandInteractionEvent _, GenericContextInteractionEvent<?> _,
         CommandAutoCompleteInteractionEvent _ ->
            newHandler();
    case CustomIdInteraction interaction when handlerExists(interaction.getCustomId()) ->
            getHandler(interaction.getCustomId());
    case CustomIdInteraction interaction when shouldCreate(interaction.getCustomId()) ->
            newHandler();
    default -> throw new UnsupportedOperationException("Unsupported jda event: %s".formatted(genericInteractionCreateEvent));
};
```

### Pros
- Alignment with Discord API
- Useful for pattern matching

### Cons
- The current approach leaves us with two methods that do exactly the same thing, which can be confusing
- Alternatively, if we would deprecate the [`getComponentId`](https://docs.jda.wiki/net/dv8tion/jda/api/events/interaction/component/GenericComponentInteractionCreateEvent.html#getComponentId()) and [`getModalId`](https://docs.jda.wiki/net/dv8tion/jda/api/events/interaction/ModalInteractionEvent.html#getModalId()) methods we would break a lot of end user code

---
I'm up for discussions and aware that the use case might not affect some users, since JDA still uses Java 8

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/interactions/CustomIdInteraction.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/CustomIdInteraction.java
@@ -1,0 +1,29 @@
+package net.dv8tion.jda.api.interactions;
+
+import net.dv8tion.jda.api.interactions.components.ComponentInteraction;
+import net.dv8tion.jda.api.interactions.modals.ModalInteraction;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Interactions which have a custom ID, namely {@link ComponentInteraction} and {@link  ModalInteraction}.
+ *
+ * <br>This id does not have to be numerical.
+ *
+ */
+public interface CustomIdInteraction
+{
+    /**
+     * The custom ID provided to the component or modal when it was originally created.
+     * <br>This value should be used to determine what action to take in regard to this interaction.
+     *
+     * <br>This id does not have to be numerical.
+     *
+     * @return The custom ID
+     *
+     * @see ComponentInteraction#getComponentId()
+     * @see ModalInteraction#getModalId()
+     */
+    @Nonnull
+    String getCustomId();
+}
--- a/src/main/java/net/dv8tion/jda/api/interactions/components/ComponentInteraction.java
+++ b/src/main/java/net/dv8tion/jda/api/interactions/components/ComponentInteraction.java
@@ -19,6 +19,7 @@ package net.dv8tion.jda.api.interactions.components;
 import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.entities.channel.unions.GuildMessageChannelUnion;
 import net.dv8tion.jda.api.entities.channel.unions.MessageChannelUnion;
+import net.dv8tion.jda.api.interactions.CustomIdInteraction;
 import net.dv8tion.jda.api.interactions.callbacks.IMessageEditCallback;
 import net.dv8tion.jda.api.interactions.callbacks.IModalCallback;
 import net.dv8tion.jda.api.interactions.callbacks.IPremiumRequiredReplyCallback;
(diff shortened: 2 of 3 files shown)


src/main/java/net/dv8tion/jda/api/interactions/CustomIdInteraction.java
```diff
@@ -0,0 +1,29 @@
+package net.dv8tion.jda.api.interactions;
```
MinnDevelopment, line 1: Missing copyright header

src/main/java/net/dv8tion/jda/api/interactions/CustomIdInteraction.java
```diff
@@ -0,0 +1,29 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * Interactions which have a custom ID, namely {@link ComponentInteraction} and {@link  ModalInteraction}.
+ *
+ * <br>This id does not have to be numerical.
+ *
+ */
+public interface CustomIdInteraction
```
MinnDevelopment, line 14: I would make this a marker interface, we usually prefix those with an `I` (I know it's odd, but we should keep that consistent here)

```suggestion
public interface ICustomIdInteraction
```

src/main/java/net/dv8tion/jda/api/interactions/CustomIdInteraction.java
```diff
@@ -0,0 +1,29 @@
...
+import net.dv8tion.jda.api.interactions.components.ComponentInteraction;
+import net.dv8tion.jda.api.interactions.modals.ModalInteraction;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Interactions which have a custom ID, namely {@link ComponentInteraction} and {@link  ModalInteraction}.
+ *
+ * <br>This id does not have to be numerical.
+ *
```
MinnDevelopment, line 12:
```suggestion
 * Marker interface for interactions with custom IDs.
 *
 * <p>This includes modals and components.
```

src/main/java/net/dv8tion/jda/api/interactions/CustomIdInteraction.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public interface CustomIdInteraction
+{
+    /**
+     * The custom ID provided to the component or modal when it was originally created.
```
MinnDevelopment, line 17:
```suggestion
     * The relevant custom ID, that has been provided for the component or modal when it was originally created.
```

src/main/java/net/dv8tion/jda/api/interactions/CustomIdInteraction.java
```diff
@@ -0,0 +1,29 @@
...
+     * <br>This value should be used to determine what action to take in regard to this interaction.
```
MinnDevelopment, line 18: I don't think this is a good addition to the docs here. Certainly "should" is not the word I would use for this, and it also isn't really what I would describe it as. I think users can figure out what they would need this for.

```suggestion
```

src/main/java/net/dv8tion/jda/api/interactions/CustomIdInteraction.java
```diff
@@ -0,0 +1,29 @@
...
+     *
+     * <br>This id does not have to be numerical.
```
MinnDevelopment, line 20: This is just confusing, obviously a string is not numerical.

```suggestion
```

MinnDevelopment (changes requested):

PR #2782, opened 2024-12-24
Add `downloadAsFileUpload` to file proxies

freya022:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Makes re-uploading attachments easier.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/utils/AttachmentProxy.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/AttachmentProxy.java
@@ -59,6 +59,8 @@ public class AttachmentProxy extends FileProxy
      *
      * @param  url
      *         The URL to download the attachment from
+     * @param  fileName
+     *         An optional file name for {@link #downloadAsFileUpload(int, int)}
      *
      * @throws IllegalArgumentException
      *         If the provided URL is null
--- a/src/main/java/net/dv8tion/jda/api/utils/AttachmentProxy.java
+++ b/src/main/java/net/dv8tion/jda/api/utils/AttachmentProxy.java
@@ -22,6 +22,7 @@ import net.dv8tion.jda.internal.utils.IOUtil;
 
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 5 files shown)


src/main/java/net/dv8tion/jda/api/utils/AttachmentProxy.java
```diff
@@ -259,4 +259,47 @@ public CompletableFuture<Icon> downloadAsIcon(int width, int height)
...
+     *         <ul>
+     *             <li>The file name is null or blank</li>
+     *             <li>The requested width is negative or 0</li>
+     *             <li>The requested height is negative or 0</li>
+     *         </ul>
+     *
+     * @return {@link FileUpload} from this attachment.
+     */
+    @Nonnull
+    public FileUpload downloadAsFileUpload(@Nonnull String name, int width, int height)
```
MinnDevelopment, line 289: We should probably add an overload that just uses the known attachment filename

src/main/java/net/dv8tion/jda/api/utils/AttachmentProxy.java
```diff
@@ -259,4 +259,47 @@ public CompletableFuture<Icon> downloadAsIcon(int width, int height)
...
+    {
+        final String url = getUrl(width, height); // So the checks are also done outside the FileUpload
+        return FileUpload.fromStreamSupplier(name, () ->
+        {
+            try
+            {
+                // Blocking is fine on the elastic rate limit thread pool [[JDABuilder#setRateLimitElastic]]
+                return download(url).get();
```
MinnDevelopment, line 297: You can use `join()` to avoid having to try/catch

src/main/java/net/dv8tion/jda/api/utils/AttachmentProxy.java
freya022: We don't have the file name as `AttachmentProxy` is used by `MessageEmbed.Thumbnail` and `MessageEmbed.ImageInfo` and they don't have that info

freya022: I have added an overload, but it will only work for `Message.Attachment`

MinnDevelopment (approved):

PR #2796, opened 2025-01-30
Allow \ in invite links

rtm516:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This allows JDA to match both of the below instead of just the first. This brings it in line with the discord client.

```
discord.com/invite/AAAAAAAA
discord.com/invite\AAAAAAAA
```
image

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -207,7 +207,7 @@ public interface Message extends ISnowflake, Formattable
             "(?:https?://)?" +                     // Scheme
             "(?:\\w+\\.)?" +                       // Subdomain
             "discord(?:(?:app)?\\.com" +           // Discord domain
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -207,7 +207,7 @@ public interface Message extends ISnowflake, Formattable
             "(?:https?://)?" +                     // Scheme
             "(?:\\w+\\.)?" +                       // Subdomain
             "discord(?:(?:app)?\\.com" +           // Discord domain
-            "/invite|\\.gg)/(?<code>[a-z0-9-]+)" + // Path
+            "/invite|\\.gg)[/\\\\](?<code>[a-z0-9-]+)" + // Path
```
MinnDevelopment, line 210: The client doesn't allow this for `discord.gg\code` but it does for `discord.com/invite\code`

image

Your regex also applies to `.gg\code` though

PR #2806, opened 2025-02-23
Initial pass on new component v2 system

DV8FromTheWorld:
Pushing this to a branch so people can see what I'm thinking about for components v2 and so we can iterate publicly together

PR #2809, opened 2025-03-07
Add support for components V2

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This adds the new component system (aka Components V2).

**API docs:** 
1. https://discord.com/developers/docs/change-log#introducing-new-components-for-messages
2. https://discord.com/developers/docs/change-log#raised-component-limits
3. https://discord.com/developers/docs/components/overview

**Examples:** [ComponentsV2Example.java](https://github.com/freya022/JDA/blob/feat/compat-without-imports/components-v2/src/examples/java/ComponentsV2Example.java)

Make sure to give your opinion [on the JDA dev thread](https://ptb.discord.com/channels/125227483518861312/1343090348187979816).

Supersedes #2806.

## Installation (Preview)

You can use the ["Using new features"](https://jda.wiki/using-jda/using-new-features/) guide from the JDA wiki.

Alternatively, you can get the newest version for your favorite tool, by using the `/jitpack pr` command in the [JDA Discord server](https://discord.gg/0hMr4ce0tIl3SLv5). You can use the `Update PR` button to merge the latest changes if necessary (note that it may not update it if there are conflicts).

image

## Migration (Preview)
I've written an OpenRewrite recipe to update most (if not all) breaking changes.

### Requirements
- Use a release JDA version (i.e., the group ID must be `net.dv8tion`)
- Your current code must compile

### Usage
**As with any migration tool, it is highly recommended to commit/push your changes before doing anything.**

**Note:** Kotlin 2.X is not supported by OpenRewrite yet, however 1.X should work.

<details>
<summary>(Kotlin) Gradle</summary>

```kt
plugins {
    // ...
    id("org.openrewrite.rewrite") version "7.4.1"
}

repositories {
    // ...
    maven("https://jitpack.io")
}

dependencies {
    // Existing JDA dependency
    // ...

    rewrite("io.github.freya022:JDA:{{latest commit hash}}")
    rewrite("org.openrewrite.recipe:rewrite-java-dependencies:1.32.1")
}

rewrite {
    activeRecipe("net.dv8tion.MigrateComponentsV2")
}
```

#### Checking changes (dry run)
Run the `rewriteDryRun` task, this can be done by pressing CTRL twice on IntelliJ then running `gradle rewriteDryRun`.
This will generate a diff file at `build/reports/rewrite/rewrite.patch`, you can check the possible changes there.

#### Applying changes
Run the `rewriteRun` task.

</details>

<details>
<summary>Maven</summary>

1. Add the following repository:
```xml
<repository>
    <id>jitpack</id>
    <url>https://jitpack.io</url>
</repository>
```

4. Add the following plugin:
```xml
<plugin>
  <groupId>org.openrewrite.maven</groupId>
  <artifactId>rewrite-maven-plugin</artifactId>
  <version>6.6.1</version>
  <configuration>
    <activeRecipes>
      <recipe>net.dv8tion.MigrateComponentsV2</recipe>
    </activeRecipes>
  </configuration>
  <dependencies>
    <dependency>
      <groupId>org.openrewrite.recipe</groupId>
      <artifactId>rewrite-java-dependencies</artifactId>
      <version>1.32.1</version>
    </dependency>
    <dependency>
      <groupId>io.github.freya022</groupId>
      <artifactId>JDA</artifactId>
      <version>{{latest commit hash}}</version>
    </dependency>
  </dependencies>
</plugin>
```

#### Checking changes (dry run)
Run the `rewrite:dryRun` goal, this can be done by pressing CTRL twice on IntelliJ then running `mvn rewrite:dryRun`.
This will generate a diff file at `target/site/rewrite/rewrite.patch`, you can check the possible changes there.

#### Applying changes
Run the `rewrite:run` goal.

</details>

You can then set the JDA dependency, as JDA 6.0.0 is not released.

It is recommended to optimize imports (`Code | Optimize imports` in IntelliJ) after migrating.


src/main/java/net/dv8tion/jda/api/components/textdisplay/TextDisplay.java
```diff
@@ -0,0 +1,87 @@
...
+public interface TextDisplay extends Component, MessageTopLevelComponent, ContainerChildComponent, SectionContentComponent
+{
+    /**
+     * Constructs a new {@link TextDisplay} from the given content.
+     *
+     * @param  content
+     *         The content of the text display
+     *
+     * @throws IllegalArgumentException
+     *         If {@code null} is provided
```
MinnDevelopment, line 49:
```suggestion
     *         If content is {@code null}, blank, or empty
```

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -188,10 +203,20 @@ public interface Message extends ISnowflake, Formattable
     int MAX_STICKER_COUNT = 3;
 
     /**
-     * The maximum amount of {@link LayoutComponent LayoutComponents} that can be added to a message ({@value})
+     * The maximum amount of {@link MessageTopLevelComponent MessageTopLevelComponents} that can be added to a message's {@link #getComponents() root component list} when using the legacy component system.  ({@value})
      */
     int MAX_COMPONENT_COUNT = 5;
 
+    /**
+     * The maximum amount of {@link MessageTopLevelComponent MessageTopLevelComponents} that can be added to a message's {@link #getComponents() root component list} when using the {@link MessageRequest#useComponentsV2() V2 component system}.  ({@value})
```
MinnDevelopment, line 211: The [recent changelog](https://discord.com/developers/docs/change-log#raised-component-limits) said this limit was removed

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1107,12 +1157,58 @@ default MessageEditAction editMessageEmbeds(@Nonnull MessageEmbed... embeds)
...
+     *         If the message is not authored by this bot
+     * @throws IllegalArgumentException
+     *         <ul>
+     *             <li>If {@code null} is provided</li>
+     *             <li>If any of the provided components are not {@linkplain Component.Type#isMessageCompatible() compatible with messages}</li>
+     *         </ul>
+     *
+     * @return {@link MessageEditAction}
+     *
+     * @see    MessageChannel#editMessageComponentsById(long, ComponentTree)
```
MinnDevelopment, line 1202:
```suggestion
     * @see    MessageChannel#editMessageComponentsById(long, ComponentTree)
     * @see    MessageComponentTree
```

src/main/java/net/dv8tion/jda/api/entities/channel/middleman/MessageChannel.java
```diff
@@ -591,15 +595,37 @@ default MessageCreateAction sendMessageEmbeds(@Nonnull Collection<? extends Mess
      *     <br>If this message was blocked by the harmful link filter</li>
      * </ul>
      *
-     * @param  component
-     *         {@link LayoutComponent} to send
-     * @param  other
-     *         Additional {@link LayoutComponent LayoutComponents} to use (up to {@value Message#MAX_COMPONENT_COUNT})
+     * <p><b>Example: Attachment Images</b>
```
MinnDevelopment, line 598: Why does this method have an example for embeds?

src/main/java/net/dv8tion/jda/api/entities/channel/middleman/MessageChannel.java
```diff
@@ -3323,12 +3400,141 @@ default MessageEditAction editMessageComponentsById(@Nonnull String messageId, @
...
+     *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
+     *     <br>The provided {@code messageId} is unknown in this MessageChannel, either due to the id being invalid, or
+     *         the message it referred to has already been deleted. This might also be triggered for ephemeral messages.</li>
+     *
+     *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_CHANNEL UNKNOWN_CHANNEL}
+     *     <br>The request was attempted after the channel was deleted.</li>
+     * </ul>
+     *
+     * <p><b>Example</b><br>
```
MinnDevelopment, line 3434: This example doesn't use a component tree, also examples should be before error responses

src/main/java/net/dv8tion/jda/api/entities/channel/middleman/MessageChannel.java
```diff
@@ -3323,12 +3400,141 @@ default MessageEditAction editMessageComponentsById(@Nonnull String messageId, @
...
```
MinnDevelopment, line 3499: Same issue here

src/main/java/net/dv8tion/jda/api/interactions/modals/ModalInteraction.java
```diff
@@ -75,16 +76,39 @@ public interface ModalInteraction extends IReplyCallback, IMessageEditCallback
...
+     *
+     * @throws IllegalArgumentException
+     *         If the provided id is null
+     *
+     * @return ModalMapping with this numeric id, or null if not found
+     *
+     * @see    #getValues()
+     */
+    @Nullable
+    default ModalMapping getValue(int id)
```
MinnDevelopment, line 103:
```suggestion
    default ModalMapping getValueByUniqueId(int id)
```

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditData.java
```diff
@@ -357,6 +357,22 @@ public synchronized List<FileUpload> getFiles()
...
     }
 
+    /**
+     * Returns the {@link FileUpload FileUploads} that are added indirectly to this message,
+     * such as from V2 components and embeds.
+     *
+     * @return The list of additional file uploads
+     */
+    @Nonnull
+    public synchronized List<FileUpload> getAdditionalFiles()
```
MinnDevelopment, line 367: Additional seems like confusing wording here, since its an edit request. More accurate would be `getComponentFileUploads` or similar

src/main/java/net/dv8tion/jda/internal/components/actionrow/ActionRowImpl.java
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public ActionRow createCopy()
+    {
+        return new ActionRowImpl(components, uniqueId);
+    }
+
+    // TODO after removal, make this immutable starting from the constructor
```
MinnDevelopment, line 171: This is still a TODO, I'm think about just removing the modifiable behavior

src/main/java/net/dv8tion/jda/internal/components/textdisplay/TextDisplayImpl.java
```diff
@@ -0,0 +1,127 @@
...
+    public Type getType()
+    {
+        return Type.TEXT_DISPLAY;
+    }
+
+    @Nonnull
+    @Override
+    public TextDisplay withUniqueId(int uniqueId)
+    {
+        Checks.positive(uniqueId, "Unique ID");
```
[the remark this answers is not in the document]
freya022: The unique ID starts at 1, as 0 is for older components

src/main/java/net/dv8tion/jda/api/utils/AttachedFile.java
```diff
@@ -316,13 +317,34 @@ static MultipartBody.Builder createMultipartBody(@Nonnull List<? extends Attache
...
+     * @param  payloadJson
+     *         The payload data to send, null to not add a payload_json part
+     *
+     * @throws IllegalArgumentException
+     *         If the file list is null
+     *
+     * @return {@link MultipartBody.Builder}
+     */
+    @Nonnull
+    static MultipartBody.Builder createMultipartBody(@Nonnull List<? extends AttachedFile> files, @Nonnull List<? extends AttachedFile> additionalFiles, @Nullable RequestBody payloadJson)
```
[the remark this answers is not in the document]
freya022: This is an overload with "additional files" - files that are added indirectly, such as with a `FileDisplay` (and hopefully in the future, in embed images/thumbnails).

This allows users to not have to upload files separately after setting an `attachment://filename.ext` URL

src/main/java/net/dv8tion/jda/internal/components/mediagallery/MediaGalleryItemImpl.java
```diff
@@ -0,0 +1,150 @@
...
+    public boolean isSpoiler()
+    {
+        return spoiler;
+    }
+
+    @Nonnull
+    @Override
+    public DataObject toData()
+    {
+        final String fileName = Helpers.getLastPathSegment(media.getUrl());
```
MrPowerGamerBR, line 119: This causes a NullPointerException when creating a message that has a media gallery item

`jda-ktx` calls `MediaGalleryItem.fromUrl(url)`, which that calls the `MediaGalleryItemImpl(String url)` constructor, and the `media` parameter is set to null

```java
        final String mediaUrl;
        if (this.media != null) {
            final String fileName = Helpers.getLastPathSegment(media.getUrl());
            mediaUrl = "attachment://" + fileName;
        } else {
            mediaUrl = this.url;
        }
```

src/main/java/net/dv8tion/jda/api/components/attribute/IDisableable.java
```diff
@@ -0,0 +1,83 @@
...
+
+import net.dv8tion.jda.api.components.Component;
+
+import javax.annotation.CheckReturnValue;
+import javax.annotation.Nonnull;
+
+/**
+ * Component which supports being disabled, or having its children be disabled.
+ */
+public interface IDisableable extends Component
```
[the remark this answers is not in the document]
freya022: I don't think it's that beneficial, for example, `ActionComponent` must implement `IDisableable`, but its subclasses, like `Button`, will need to override the return type anyway, so it will be same as it is now.

src/main/java/net/dv8tion/jda/internal/components/actionrow/ActionRowImpl.java
freya022: It would be nicer for migration purposes to let this be mutable until the deprecated stuff gets removed.

If we suddenly make it immutable, all `updateComponent` methods on `LayoutComponent` won't be usable, they will fail silently unless removed

freya022: Unless we bump the major version, I don't think that's a good idea

Reddishye:
I have been trying it and somehow I'm having issues when sending medium/large components. I do not know whatever this is an issue on my side or is a known issue. I neither know if this should be the right place to tell this or should I make an issue on the fork itself.

freya022:
You're saying you have an issue but didn't describe what the issue is

Reddishye:
@freya022 Seemed to be an issue on my side. Was just talking and just an hour after it founding it 😭

FelixTPG:
Is it intended that a TextDisplay can't be the first argument of a `Section`?

freya022:
Yes, a `TextDisplay` cannot be an accessory.

FelixTPG:
alright, thank you

src/main/java/net/dv8tion/jda/api/components/ComponentUnion.java
```diff
@@ -0,0 +1,41 @@
...
+ * Represents a union of {@link Component Components}.
+ *
+ * @see ActionRowChildComponentUnion
+ * @see ContainerChildComponentUnion
+ * @see SectionContentComponentUnion
+ * @see SectionAccessoryComponentUnion
+ * @see MessageTopLevelComponentUnion
+ * @see ModalTopLevelComponentUnion
+ */
+public interface ComponentUnion extends Component
```
MinnDevelopment: I don't like this interface name, the naming makes me think that this is a union of all components, and thus would have all the `asFoo()` methods.

src/main/java/net/dv8tion/jda/api/components/ComponentUnion.java
```diff
@@ -0,0 +1,41 @@
...
+{
+    default boolean isUnknownComponent() {
```
MinnDevelopment: This has no docs

src/examples/resources/Prairie_ButterflyFields.webp
MinnDevelopment, line 1: I'd prefer not using WEBP for anything

src/examples/java/ComponentsV2Example.java
```diff
@@ -0,0 +1,356 @@
...
+                                .addOption("Social Space", "social_space")
+                                .addOption("Butterfly Fields", "butterfly_fields")
+                                .addOption("Prairie Village", "prairie_village")
+                                .setDefaultValues("butterfly_fields")
+                                .build()
+                )
+        );
+
+        // Notice here how we don't have to upload any files and manage ugly "attachment://" URLs?
+        // They are automatically added when they are inside a V2 component
```
MinnDevelopment: I don't like this comment, it doesn't really make sense for an example. A new user would not understand what any of this means.

src/examples/java/ComponentsV2Example.java
```diff
@@ -0,0 +1,356 @@
...
+                // Again?
+                Separator.createDivider(Separator.Spacing.SMALL),
+
+                // 🦀
+                TextDisplay.of("Powered by:"),
+                // A set of pictures to display, display in a mosaic
+                // It can also take one item, in which case it will take the most horizontal space as possible,
+                // depending on the aspect ratio.
+                MediaGallery.of(
+                        MediaGalleryItem.fromFile(getResourceAsFileUpload("/rust.gif"))
```
MinnDevelopment: Why rust? I don't get how this fits into a jda example

src/examples/java/ComponentsV2Example.java
```diff
@@ -0,0 +1,356 @@
...
+        }
+    }
+
+    private static void onComponentsV2Sample(@Nonnull SlashCommandInteractionEvent event)
+    {
+        // A simple box; looks similar to an embed but...
+        Container container = Container.of(
+                // Displays content on the left and an "accessory" on the right.
+                Section.of(
+                        // A thumbnail, it even works with WEBPs.
```
MinnDevelopment: How is webp relevant, who even uses this format in java?

src/examples/java/ComponentsV2Example.java
```diff
@@ -0,0 +1,356 @@
...
+public class ComponentsV2Example extends ListenerAdapter
+{
+    private static ApplicationEmoji backEmoji;
+
+    public static void main(String[] args) throws IOException
+    {
+        // Read our token from a "Config.txt" in the project root (where the README.md is)
+        // In Java 11+, you can replace this with "Files.readString(Path.of("Config.txt"))"
+        final String token = new String(Files.readAllBytes(Paths.get("Config.txt")), Charset.defaultCharset()).trim();
+        JDA jda = JDABuilder.createLight(token, EnumSet.noneOf(GatewayIntent.class)) // slash commands don't need any intents
```
MinnDevelopment:
```suggestion
        JDA jda = JDABuilder.createLight("YOUR_BOT_TOKEN_HERE", EnumSet.noneOf(GatewayIntent.class)) // slash commands don't need any intents
```

src/examples/java/ComponentsV2Example.java
```diff
@@ -0,0 +1,356 @@
...
+ *                     .getComponentTree()
+ *                     .replace(ComponentReplacer.byId(event.getButton(), event.getButton().asDisabled()));
+ *             event.editComponents(clickedAsDisabled).queue();
+ *         }
+ *     }
+ * </code></pre>
+ *
+ * You can also easily disable all components:
+ *
+ * <pre><code>
```
MinnDevelopment: We usually do `<pre>{@code ...}</pre>`, any reason to do this differently here?

src/examples/java/ComponentsV2Example.java
```diff
@@ -0,0 +1,356 @@
...
+import java.io.InputStream;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.EnumSet;
+import java.util.List;
+
+/**
+ * A (somewhat?) comprehensive example on how to use the new V2 components.
```
MinnDevelopment:
```suggestion
 * An example on how to use components version 2.
```

src/examples/java/ComponentsV2Example.java
```diff
@@ -0,0 +1,356 @@
...
+ *
+ * <h3>Enabling Components V2</h3>
+ * V2 components can only be used in messages which uses {@link MessageRequest#useComponentsV2()},
+ * this however has some limits, which are documented on the method.
+ *
+ * <p>Note that every component which requires that flag will be documented.
```
MinnDevelopment:
```suggestion
 * <p>Note that every component documents, whether it requires the v2 flag.
```

src/examples/java/ComponentsV2Example.java
```diff
@@ -0,0 +1,356 @@
...
+ *
+ * <h3>Component typing</h3>
+ * Components now follow a type system similar to channels, using "union" types;
```
MinnDevelopment: All of these docs read like a changelog, we can't do that for long term examples. Things like "new" and "now" will be quickly outdated.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -784,7 +825,7 @@ default Button getButtonById(@Nonnull String id)
     {
```
MinnDevelopment, line 841: The method likely needs to be renamed to `getButtonByCustomId` to avoid confusion. Although, we should probably just deprecate it entirely and not add an alternative. The better way is to just find it with a util function on a component list / iterable.

src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2577,7 +2745,31 @@ enum MessageFlag
...
+         * and allows more components in total ({@value Message#MAX_COMPONENT_COUNT_IN_COMPONENT_TREE}).
+         * <br>They also allow you to use a larger choice of components,
+         * such as any component extending {@link MessageTopLevelComponent},
+         * as long as they are {@linkplain Component.Type#isMessageCompatible() compatible}.
+         * <br>The character limit for the messages also gets changed to {@value Message#MAX_CONTENT_LENGTH_COMPONENT_V2}.
+         *
+         * <p>This, however, comes with a few drawbacks:
+         * <ul>
+         *     <li>You cannot send content, embeds, polls or stickers</li>
+         *     <li>It does not support audio files</li>
```
MinnDevelopment: No audio files at all? Like, not even a `FileDisplay`? Why?

src/main/java/net/dv8tion/jda/api/utils/messages/AbstractMessageBuilder.java
```diff
@@ -43,13 +46,18 @@
 @SuppressWarnings("unchecked")
 public abstract class AbstractMessageBuilder<T, R extends AbstractMessageBuilder<T, R>> implements MessageRequest<R>
 {
+    static boolean isDefaultUseComponentsV2 = false;
```
MinnDevelopment:
```suggestion
    protected static boolean isDefaultUseComponentsV2 = false;
```

src/main/java/net/dv8tion/jda/api/utils/messages/AbstractMessageBuilder.java
```diff
@@ -43,13 +46,18 @@
...
+    protected final List<MessageTopLevelComponentUnion> components = new ArrayList<>(Message.MAX_COMPONENT_COUNT);
     protected final StringBuilder content = new StringBuilder(Message.MAX_CONTENT_LENGTH);
     protected AllowedMentionsData mentions = new AllowedMentionsData();
     protected int messageFlags;
 
-    protected AbstractMessageBuilder() {}
+    protected AbstractMessageBuilder() {
+        if (isDefaultUseComponentsV2)
+            useComponentsV2();
+    }
```
MinnDevelopment:
```suggestion
    protected AbstractMessageBuilder()
    {
        useComponentsV2(isDefaultUseComponentsV2);
    }
```

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -229,71 +263,178 @@ default R setEmbeds(@Nonnull MessageEmbed... embeds)
...
+     * The {@link MessageTopLevelComponent MessageTopLevelComponents} that should be attached to the message.
+     * <br>You can call this method without anything to remove all components from the message.
+     *
+     * <p>The most commonly used layout is {@link ActionRow}.
+     *
+     * <p><b>Example: Set action rows</b><br>
+     * <pre>{@code
+     * channel.sendMessage("Content is still required")
+     *   .setComponents(
+     *     ActionRow.of(selectMenu) // first row
```
MinnDevelopment:
```suggestion
     *     ActionRow.of(selectMenu), // first row
```

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -229,71 +263,178 @@ default R setEmbeds(@Nonnull MessageEmbed... embeds)
...
      * <p><b>Example</b><br>
      *
      * <pre>{@code
-     * final List<ItemComponent> list = new ArrayList<>();
-     * list.add(button1);
-     * list.add(button2);
-     *
      * channel.sendMessage("Content is still required")
-     *   .setActionRow(list)
+     *   .setActionRow(button1, button2)
```
MinnDevelopment: This seems like a copy-paste error

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -100,6 +105,33 @@ static void setDefaultMentionRepliedUser(boolean mention)
...
+    static void setDefaultUseComponentsV2(boolean use)
+    {
+        AbstractMessageBuilder.isDefaultUseComponentsV2 = use;
+    }
+
+    /**
+     * Whether V2 components are used by default.
+     * <br>When enabled, {@link #useComponentsV2()} gets called for every message builder.
+     *
+     * <p>This can still be overridden by calling {@link #useComponentsV2(boolean)} manually.
```
MinnDevelopment:
```suggestion
     * <p>This can be overwritten with {@link #useComponentsV2(boolean)} on each builder instance.
```

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -100,6 +105,33 @@ static void setDefaultMentionRepliedUser(boolean mention)
         AllowedMentionsData.setDefaultMentionRepliedUser(mention);
     }
 
+    /**
+     * Sets whether V2 components will be used by default.
+     * <br>When enabled, {@link #useComponentsV2()} gets called for every message builder.
+     *
+     * <p>This can still be overridden by calling {@link #useComponentsV2(boolean)} manually.
```
MinnDevelopment:
```suggestion
     * <p>This can be overwritten with {@link #useComponentsV2(boolean)} on each builder instance.
```

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -100,6 +105,33 @@ static void setDefaultMentionRepliedUser(boolean mention)
         AllowedMentionsData.setDefaultMentionRepliedUser(mention);
     }
 
+    /**
+     * Sets whether V2 components will be used by default.
```
MinnDevelopment: This should document the default we use (false), like other setters here

src/main/java/net/dv8tion/jda/api/utils/AttachedFile.java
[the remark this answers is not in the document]
MinnDevelopment: I'd rather just make a list union instead of this new method overload, I see no reason this should exist over just doing `createMultipartBody(files + additionalFiles, payloadJson)`

src/main/java/net/dv8tion/jda/api/components/tree/ComponentTree.java
```diff
@@ -0,0 +1,229 @@
...
+     * @throws IllegalArgumentException
+     *         If {@code null} is provided
+     *
+     * @return A new tree with the new components
+     *
+     * @see ComponentReplacer
+     */
+    @Nonnull
+    @CheckReturnValue
+    ComponentTree<E> replace(ComponentReplacer replacer);
```
MinnDevelopment: I'm confused how this passed our arch unit tests, I'll have to take a look.

```suggestion
    ComponentTree<E> replace(@Nonnull ComponentReplacer replacer);
```

src/main/java/net/dv8tion/jda/api/components/tree/ComponentTree.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Represents a tree of components, in which you can find, replace or remove components recursively.
+ *
+ * <p>As with every component, component trees are immutable and will return a new instance on every mutating call.
+ *
+ * @param <E> Type of components contained by this tree
+ */
+public interface ComponentTree<E extends ComponentUnion>
```
MinnDevelopment: Why is this limited to unions?

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -229,71 +263,178 @@ default R setEmbeds(@Nonnull MessageEmbed... embeds)
...
+     *             <li>If {@code null} is provided</li>
+     *             <li>If any of the provided components are not {@linkplain Component.Type#isMessageCompatible() compatible with messages}</li>
+     *         </ul>
+     *
+     * @return The same instance for chaining
+     *
+     * @see    net.dv8tion.jda.api.components.tree.MessageComponentTree MessageComponentTree
+     */
+    @Nonnull
+    default R setComponents(@Nonnull ComponentTree<MessageTopLevelComponentUnion> tree)
```
MinnDevelopment:
```suggestion
    default R setComponents(@Nonnull ComponentTree<? extends MessageTopLevelComponentUnion> tree)
```

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateRequest.java
```diff
@@ -155,28 +162,65 @@ default R addEmbeds(@Nonnull MessageEmbed... embeds)
...
+     *             <li>If any of the provided components are not {@linkplain Component.Type#isMessageCompatible() compatible with messages}</li>
+     *         </ul>
+     *
+     * @return The same instance for chaining
+     *
+     * @see    ActionRow
+     * @see    net.dv8tion.jda.api.components.tree.MessageComponentTree MessageComponentTree
+     */
+    @Nonnull
+    default R addComponents(@Nonnull ComponentTree<MessageTopLevelComponentUnion> tree)
```
MinnDevelopment:
```suggestion
    default R addComponents(@Nonnull ComponentTree<? extends MessageTopLevelComponentUnion> tree)
```

src/main/java/net/dv8tion/jda/api/components/ActionComponent.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Returns a copy of this component with {@link #isDisabled()} set to the provided value.
+     *
+     * @param  disabled
+     *         True, if this component should be disabled
+     *
+     * @throws UnsupportedOperationException
+     *         If this component type cannot be disabled
```
MinnDevelopment, line 128: Doesn't this break `ComponentTree#withDisabled`? I believe TextInput throws this, because they can't be disabled. Right now they only work in modals, but we should probably reconsider that behavior. @DV8FromTheWorld what do you think about this?

src/main/java/net/dv8tion/jda/api/components/ComponentUnion.java
```diff
@@ -0,0 +1,41 @@
...
+        return this instanceof UnknownComponent;
```
MinnDevelopment:
```suggestion
        return this.getType() == Type.UNKNOWN;
```

src/main/java/net/dv8tion/jda/api/components/MessageTopLevelComponentUnion.java
```diff
@@ -0,0 +1,272 @@
...
+import net.dv8tion.jda.internal.components.separator.SeparatorImpl;
+import net.dv8tion.jda.internal.components.textdisplay.TextDisplayImpl;
+import net.dv8tion.jda.internal.utils.Checks;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Represents a union of {@link MessageTopLevelComponent MessageTopLevelComponents} that can be either
```
MinnDevelopment:
```suggestion
 * Represents a union of {@link MessageTopLevelComponent MessageTopLevelComponents} that can be one of
```

src/main/java/net/dv8tion/jda/api/components/MessageTopLevelComponentUnion.java
```diff
@@ -0,0 +1,272 @@
...
+    @Nonnull
+    Container asContainer();
+
+    /**
+     * Converts the provided {@link DataObject} into an {@link MessageTopLevelComponentUnion}.
+     *
+     * @param  data
+     *         The {@link DataObject} to create the component from
+     *
+     * @return An {@link MessageTopLevelComponentUnion} representing the provided data
```
MinnDevelopment: Document that this can return unknown components too

src/main/java/net/dv8tion/jda/api/components/ResolvedMedia.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * A media resolved by Discord including some metadata,
+ * typically comes from {@linkplain net.dv8tion.jda.api.entities.Message.MessageFlag#IS_COMPONENTS_V2 V2 Components}.
+ */
+public interface ResolvedMedia
+{
+    /**
+     * The URL of this media, this is always where the file originally came from.
+     * <br>This can be either {@code attachment://filename.extension} or an actual URL.
```
MinnDevelopment: When is this an `attachment://` link? Isn't this only returned from the API with unfurled URLs?

Otherwise, we **must** validate the url in `getProxy` to avoid trying to download from `attachment://` urls.

src/main/java/net/dv8tion/jda/api/components/actionrow/ActionRow.java
```diff
@@ -0,0 +1,307 @@
...
+        Checks.notNull(components, "Components");
+        return partitionOf(Arrays.asList(components));
+    }
+
+    /**
+     * How many of components of the provided type can be added to a single {@link ActionRow}.
+     *
+     * @return The maximum amount an action row can contain
+     */
+    static int getMaxAllowed(Component.Type type)
```
MinnDevelopment:
```suggestion
    static int getMaxAllowed(@Nonnull Component.Type type)
```

Please add a snapshot test for this, so we can see when this becomes outdated easily.

src/main/java/net/dv8tion/jda/api/components/actionrow/ActionRowChildComponentUnion.java
```diff
@@ -0,0 +1,170 @@
...
+     * @param  data
+     *         The {@link DataObject} to create the component from
+     *
+     * @return An {@link ActionRowChildComponentUnion} representing the provided data
+     *
+     * @throws IllegalArgumentException
+     *         If the provided data is null
+     */
+    @Nonnull
+    static ActionRowChildComponentUnion fromData(@Nonnull DataObject data)
```
MinnDevelopment: We should not have all these `fromData` methods on every union, too many places to maintain these type switches. Just the component tree is enough, in my opinion. This seems overkill.

src/main/java/net/dv8tion/jda/api/components/filedisplay/FileDisplay.java
```diff
@@ -0,0 +1,127 @@
...
+
+import javax.annotation.CheckReturnValue;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.io.InputStream;
+
+/**
+ * Component displaying a file, you can mark it as a spoiler.
+ *
+ * <p><b>Note:</b> Audio files and text files cannot be <i>previewed</i>,
```
MinnDevelopment: Sure, but they can still be shown right? Do we really need to document this?

src/main/java/net/dv8tion/jda/api/components/filedisplay/FileDisplay.java
```diff
@@ -0,0 +1,127 @@
...
+ * use {@linkplain net.dv8tion.jda.api.components.mediagallery.MediaGallery media galleries} to display images.
```
MinnDevelopment: Do file displays not display images? Whats the advantage?

src/main/java/net/dv8tion/jda/api/components/replacer/ComponentReplacer.java
```diff
@@ -0,0 +1,211 @@
...
+     * @param  newComponent
+     *         The component to replace with, {@code null} to remove the component
+     *
+     * @throws IllegalArgumentException
+     *         If {@code oldComponent} is {@code null}
+     *
+     * @return A {@link ComponentReplacer} replacing the old component with the new one
+     */
+    @Nonnull
+    static ComponentReplacer byId(@Nonnull Component oldComponent, @Nullable Component newComponent)
```
MinnDevelopment: We should use `byUniqueId` to avoid ambiguity

src/main/java/net/dv8tion/jda/api/components/replacer/ComponentReplacer.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /**
+     * Creates a {@link ComponentReplacer} which replaces a given component with another,
+     * based on their {@linkplain Component#getUniqueId() numeric ID}.
+     *
+     * @param  oldComponent
+     *         The component to replace
+     * @param  newComponent
+     *         The component to replace with, {@code null} to remove the component
```
MinnDevelopment, line 155: This doesn't work:

```java
MessageComponentTree tree = MessageComponentTree.of(
    ActionRow.of(
        Button.primary("button1", "test").withUniqueId(1),
        Button.secondary("button2", "test").withUniqueId(2))
);

tree.replace(ComponentReplacer.byId(1, (Button) null));
```

```
Button:test(id=1, custom id=button1) was replaced by an incompatible component (null), this layout only supports components of type ActionRowChildComponent
	at net.dv8tion.jda.internal.utils.Checks.check(Checks.java:77)
	at net.dv8tion.jda.internal.utils.Checks.checkComponentType(Checks.java:268)
	at net.dv8tion.jda.internal.components.utils.ComponentsUtil.doReplace(ComponentsUtil.java:83)
	at net.dv8tion.jda.internal.components.actionrow.ActionRowImpl.replace(ActionRowImpl.java:150)
	at net.dv8tion.jda.internal.components.actionrow.ActionRowImpl.replace(ActionRowImpl.java:39)
	at net.dv8tion.jda.internal.components.utils.ComponentsUtil.doReplace(ComponentsUtil.java:87)
	at net.dv8tion.jda.internal.components.tree.MessageComponentTreeImpl.replace(MessageComponentTreeImpl.java:53)
```

src/main/java/net/dv8tion/jda/api/components/IComponentUnion.java
```diff
@@ -0,0 +1,120 @@
...
+        case THUMBNAIL:
+            return new ThumbnailImpl(data);
+        case MEDIA_GALLERY:
+            return new MediaGalleryImpl(data);
+        case FILE_DISPLAY:
+            return new FileDisplayImpl(data);
+        case SEPARATOR:
+            return new SeparatorImpl(data);
+        case CONTAINER:
+            return ContainerImpl.fromData(data);
```
MinnDevelopment: Why do some of these do `fromData` and others use a constructor? I think we should be more consistent here

src/main/java/net/dv8tion/jda/api/components/IComponentUnion.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param  data
+     *         The {@link DataObject} to create the component from
+     *
+     * @return An {@link IComponentUnion} representing the provided data
+     *
+     * @throws IllegalArgumentException
+     *         If the provided data is null
+     */
+    @Nonnull
+    static IComponentUnion fromData(@Nonnull DataObject data)
```
MinnDevelopment: What's the intended use-case for this method as an end-user? You can't pass this union into anything, right?

src/main/java/net/dv8tion/jda/api/components/ResolvedMedia.java
```diff
@@ -0,0 +1,101 @@
...
+public interface ResolvedMedia
+{
+    /**
+     * The ID of the attachment represented by this resolved media,
+     * may return {@code null} if this media was created from an external link.
+     *
+     * @return The ID of the attachment, or {@code null}
+     */
+    @Nullable
+    String getAttachmentId();
```
MinnDevelopment, line 37: This is missing a `getAttachmentIdLong()`

src/main/java/net/dv8tion/jda/api/components/ResolvedMedia.java
```diff
@@ -0,0 +1,101 @@
...
+     * The height of this media, if available, or {@code 0}.
+     *
+     * <p>This may be {@code 0} if the media failed to load.
+     *
+     * @return Height of this media, or {@code 0}
+     */
+    int getHeight();
+
+    /**
+     * The <a href="https://en.wikipedia.org/wiki/Media_type" target="_blank">MIME type</a> of this media,
```
MinnDevelopment:
```suggestion
     * The <a href="https://en.wikipedia.org/wiki/Media_type" target="_blank">media type</a> of this media,
```

src/test/java/net/dv8tion/jda/test/entities/message/MessageComponentTest.java
```diff
@@ -0,0 +1,241 @@
...
+                // Edit with an entirely new message with no components in V1 mode
+                Arguments.of(messageEdit(b -> b.setReplace(true).setFiles(EXAMPLE_FILE_UPLOAD))),
+                // MAX_COMPONENT_COUNT top-level
+                Arguments.of(message(b -> b.useComponentsV2().setComponents(getMaxTopLevelV1()))),
+                // Add top-levels until it would break the tree size checks,
+                // to make sure the top-level size isn't checked, or at least equal to the max tree size
+                Arguments.of(message(b -> b.useComponentsV2().setComponents(getAbsurdTopLevelV2()))),
+                // MAX_COMPONENT_COUNT_IN_COMPONENT_TREE total
+                Arguments.of(message(b -> b.useComponentsV2().setComponents(getMaxTotal()))),
+                // Attempt to use >MAX_CONTENT_LENGTH_COMPONENT_V2
```
MinnDevelopment: Isn't this `=MAX_CONTENT_LENGTH_COMPONENT_V2`? I assume this is a copy&paste typo

src/test/java/net/dv8tion/jda/test/components/utils/ComponentsUtilTest.java
```diff
@@ -0,0 +1,89 @@
...
+                ActionRow.of(
+                        Button.primary("button1", "test").withUniqueId(1),
+                        Button.secondary("button2", "test").withUniqueId(2)
+                )
+        );
+
+        MessageComponentTree newTree = assertDoesNotThrow(() -> tree.replace(ComponentReplacer.byUniqueId(1, (Button) null)));
+        final ActionRow row = newTree.getComponents().get(0).asActionRow();
+        assertEquals(1, row.getComponents().size());
+        assertEquals(2, row.getComponents().get(0).getUniqueId());
```
MinnDevelopment:
```suggestion
        MessageComponentTree newTree = tree.replace(ComponentReplacer.byUniqueId(1, (Button) null));
        final ActionRow row = newTree.getComponents().get(0).asActionRow();
        assertThat(row.getComponents()).hasSize(1);
        assertThat(row.getComponents()).containsExactly((ActionRowChildComponentUnion) button2);
```

src/main/java/net/dv8tion/jda/internal/utils/message/MessageUtil.java
```diff
@@ -0,0 +1,56 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class MessageUtil
+{
+    @Nonnull
+    public static List<AttachedFile> getIndirectFiles(@Nonnull Collection<MessageTopLevelComponentUnion> components, boolean shouldRetain)
```
MinnDevelopment: `shouldRetain` is a confusing name for this. I would invert it as `shouldReuploadExisting` or something similar to that.

src/main/java/net/dv8tion/jda/internal/entities/ReceivedMessage.java
```diff
@@ -626,12 +630,18 @@ public List<MessageEmbed> getEmbeds()
...
+    public List<MessageTopLevelComponentUnion> getComponents()
     {
         checkIntent();
         return components;
     }
 
+    @Override
+    public boolean isUsingComponentsV2()
+    {
+        return (this.flags & MessageFlag.IS_COMPONENTS_V2.getValue()) > 0;
```
MinnDevelopment:
```suggestion
        return (this.flags & MessageFlag.IS_COMPONENTS_V2.getValue()) != 0;
```

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -1672,11 +1671,12 @@ private ReceivedMessage createMessage0(DataObject jsonObject, @Nullable MessageC
...
-        final List<StickerItem>        stickers    = map(jsonObject, "sticker_items", this::createStickerItem);
-        final List<LayoutComponent>    components  = map(jsonObject, "components",    ActionRow::fromData, obj -> obj.getInt("type", -1) == 1);
+        final List<Message.Attachment>            attachments = map(jsonObject, "attachments",   this::createMessageAttachment);
+        final List<MessageEmbed>                  embeds      = map(jsonObject, "embeds",        this::createMessageEmbed);
+        final List<MessageReaction>               reactions   = map(jsonObject, "reactions",     (obj) -> createMessageReaction(tmpChannel, channelId, id, obj));
+        final List<StickerItem>                   stickers    = map(jsonObject, "sticker_items", this::createStickerItem);
+        // Keep the unknown components so the user can read them if they want
+        final List<MessageTopLevelComponentUnion> components  = map(jsonObject, "components",    (obj) -> ComponentsUtil.deserializeTo(obj, MessageTopLevelComponentUnion.class));
```
MinnDevelopment: I think the `IComponentUnion.fromData` should be moved to a public util method similar to this `deserializeTo`. Since users likely want a way to do exactly this type of parsing.


```java
// Single component
MessageTopLevelComponentUnion component = Components.fromData(MessageTopLevelComponentUnion.class, data);
// Full tree
MessageComponentTree tree = Components.fromData(MessageComponentTree.class, data);
```

src/main/java/net/dv8tion/jda/internal/components/utils/ComponentsUtil.java
```diff
@@ -0,0 +1,178 @@
...
+    {
+        return components.stream().filter(c -> !(c instanceof ActionRow)).collect(Collectors.toList());
+    }
+
+    public static boolean hasIllegalV1Components(@Nonnull Collection<? extends Component> components)
+    {
+        return !getIllegalV1Components(components).isEmpty();
+    }
+
+    public static long getComponentTreeLength(@Nonnull Collection<? extends Component> components)
```
MinnDevelopment:
```suggestion
    public static long getComponentTreeTextContentLength(@Nonnull Collection<? extends Component> components)
```

src/main/java/net/dv8tion/jda/internal/components/tree/ModalComponentTreeImpl.java
```diff
@@ -0,0 +1,68 @@
...
+        extends AbstractComponentTree<ModalTopLevelComponentUnion>
+        implements ModalComponentTree
+{
+    private ModalComponentTreeImpl(Collection<ModalTopLevelComponentUnion> components)
+    {
+        super(components);
+    }
+
+    @Nonnull
+    public static ModalComponentTree of(@Nonnull Collection<? extends ModalTopLevelComponent> _components)
```
MinnDevelopment: Why the `_`?

src/main/java/net/dv8tion/jda/internal/components/tree/ModalComponentTreeImpl.java
```diff
@@ -0,0 +1,68 @@
...
+    {
+        // Empty trees are allowed
```
MinnDevelopment: Does an empty modal make sense?

src/main/java/net/dv8tion/jda/internal/components/tree/ModalComponentTreeImpl.java
```diff
@@ -0,0 +1,68 @@
...
+        Checks.noneNull(_components, "Components");
+
+        // Allow unknown components so [[Message#getComponentTree]] works
```
MinnDevelopment:
```suggestion
        // Allow unknown components so [[Modal#getComponentTree]] works
```

src/main/java/net/dv8tion/jda/internal/components/thumbnail/ThumbnailImpl.java
```diff
@@ -0,0 +1,203 @@
...
+    @Override
+    public ResolvedMedia getResolvedMedia()
+    {
+        return media;
+    }
+
+    @Override
+    public Stream<AttachedFile> getFiles(boolean shouldRetain)
+    {
+        if (media != null) // Retain or reupload the entire file
```
MinnDevelopment: This block seems to be duplicated in several classes, can we extract a common util method?

src/main/java/net/dv8tion/jda/internal/components/thumbnail/ThumbnailImpl.java
```diff
@@ -0,0 +1,203 @@
...
+            json.put("id", uniqueId);
+        if (description != null)
+            json.put("description", description);
+        return json;
+    }
+
+    @Override
+    public boolean equals(Object o)
+    {
+        if (!(o instanceof ThumbnailImpl)) return false;
```
MinnDevelopment, line 172: None of your equals seem to have the first part:

```java
if (o == this) return true;
```

src/main/java/net/dv8tion/jda/internal/components/section/SectionImpl.java
```diff
@@ -0,0 +1,185 @@
...
+
+    private SectionImpl(int uniqueId, Collection<SectionContentComponentUnion> components, SectionAccessoryComponentUnion accessory)
+    {
+        Checks.notEmpty(components, "Components");
+        this.uniqueId = uniqueId;
+        this.components = Helpers.copyAsUnmodifiableList(components);
+        this.accessory = accessory;
+    }
+
+    public static Section of(SectionAccessoryComponent _accessory, Collection<? extends SectionContentComponent> _components)
```
MinnDevelopment: Why the `_`?

src/main/java/net/dv8tion/jda/internal/components/actionrow/ActionRowImpl.java
```diff
@@ -0,0 +1,244 @@
...
+        return components.iterator();
+    }
+
+    @Override
+    public boolean isEmpty()
+    {
+        return false;
+    }
+
+    public boolean isValid()
```
MinnDevelopment:
```suggestion
    @Override
    public boolean isValid()
```

src/main/java/net/dv8tion/jda/internal/components/actionrow/ActionRowImpl.java
```diff
@@ -0,0 +1,244 @@
...
+
+    private ActionRowImpl(Collection<ActionRowChildComponentUnion> components)
+    {
+        this(components, -1);
+    }
+
+    private ActionRowImpl(Collection<ActionRowChildComponentUnion> components, int uniqueId)
+    {
+        Checks.notEmpty(components, "Row");
+        checkIsValid(components); // This is here so the "replace" method can't create invalid rows
```
MinnDevelopment: Preferably, this would be asserted in a test rather than a comment

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditBuilder.java
```diff
@@ -303,10 +362,42 @@ public MessageEditData build()
...
+
+    @Nonnull
+    private MessageEditData buildV2()
+    {
+        // Copy to prevent modifying data after building
+        List<AttachedFile> attachments = new ArrayList<>(this.attachments);
+        List<MessageTopLevelComponentUnion> components = new ArrayList<>(this.components);
+        AllowedMentionsData mentions = this.mentions.copy();
+
+        if ((isSet(CONTENT) && content.length() > 0) || (isSet(EMBEDS) && !embeds.isEmpty()))
```
MinnDevelopment, line 370: In `isV2Valid` you check for blank but here you check length, which one is correct?

src/main/java/net/dv8tion/jda/api/utils/messages/MessageCreateData.java
```diff
@@ -368,6 +381,37 @@ public List<FileUpload> getFiles()
...
+    private static Set<FileUpload> createAllDistinctFiles(@Nonnull Collection<FileUpload> files, @Nonnull Collection<MessageTopLevelComponentUnion> components)
+    {
+        List<AttachedFile> indirectFiles = MessageUtil.getIndirectFiles(components, false);
+        Set<FileUpload> distinctFiles = new LinkedHashSet<>(files.size() + indirectFiles.size());
+        distinctFiles.addAll(files);
+        for (AttachedFile file : indirectFiles)
+        {
+            if (!(file instanceof FileUpload))
+            {
+                LOG.warn("Tried to add an attachment that isn't a FileUpload, discarding the file. Please redirect the following message to the devs (JDA {})", JDAInfo.VERSION);
```
MinnDevelopment: This should throw an `IllegalStateException`

src/main/java/net/dv8tion/jda/api/interactions/modals/ModalTopLevelComponentUnion.java
```diff
@@ -0,0 +1,87 @@
...
+     * @throws IllegalStateException
+     *         If the component represented by this union is not actually a {@link ActionRow}.
+     *
+     * @return The component as a {@link ActionRow}
+     */
+    @Nonnull
+    ActionRow asActionRow();
+
+    @Nonnull
+    static ModalTopLevelComponentUnion fromData(@Nonnull DataObject data) {
```
MinnDevelopment: Let's remove these in favor of the previously mentioned util method

src/main/java/net/dv8tion/jda/api/interactions/components/LayoutComponent.java
```diff
@@ -16,66 +16,118 @@
...
 
 /**
  * Represents a top-level layout used for {@link ItemComponent ItemComponents} such as {@link Button Buttons}.
  *
  * <p>Components must always be contained within such a layout.
  *
  * @see ActionRow
+ *
+ * @deprecated
+ *      Will be removed in a future release, please use {@link ActionRow} instead.
```
MinnDevelopment: Let's just remove this right away, no point in keeping it around.

src/main/java/net/dv8tion/jda/api/interactions/components/ItemComponent.java
```diff
@@ -16,13 +16,22 @@
...
  * Component which can be inserted into a {@link LayoutComponent}.
  *
  * @see ActionComponent
  * @see ActionRow
  */
-public interface ItemComponent extends Component
+@ForRemoval
+@Deprecated
+@ReplaceWith("ActionRowChildComponent")
+public interface ItemComponent extends ActionRowChildComponent
```
MinnDevelopment: Remove this too

src/main/java/net/dv8tion/jda/api/components/utils/ComponentPathIterator.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.*;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * Utility class to recursively iterate on a list of components while keeping track their path.
+ */
+public class ComponentPathIterator implements Iterator<ComponentPathIterator.ComponentWithPath>
+{
+    private final Stack<Iterator<ComponentWithPath>> stack = new Stack<>();
```
MinnDevelopment: Use `ArrayDeque` instead of the obsolete stack, we don't need synchronization on an iterator.

src/main/java/net/dv8tion/jda/api/components/utils/ComponentPathIterator.java
```diff
@@ -0,0 +1,165 @@
...
+
+    protected ComponentPathIterator(String rootName, Collection<? extends Component> components) {
```
MinnDevelopment: Minor: This entire class has brackets in the wrong place

src/main/java/net/dv8tion/jda/api/components/utils/ComponentPathIterator.java
```diff
@@ -0,0 +1,165 @@
...
+    @Override
+    public ComponentWithPath next() {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+        Iterator<ComponentWithPath> iterator = stack.peek();
+        ComponentWithPath componentWithPath = iterator.next();
+        Component component = componentWithPath.component;
+
+        if (component instanceof Container) {
```
MinnDevelopment: Let's add snapshot tests for both of these iterators, so this flattening logic is not forgotten when new components are added.

src/main/java/net/dv8tion/jda/api/components/tree/ComponentTree.java
```diff
@@ -0,0 +1,281 @@
...
+     * @param  data
+     *         The {@link DataArray} to create the component tree from
+     *
+     * @return A {@link ComponentTree} representing the provided data
+     *
+     * @throws IllegalArgumentException
+     *         If the provided data is null
+     */
+    @Nonnull
+    static ComponentTree<IComponentUnion> fromData(@Nonnull DataArray data)
```
MinnDevelopment: Let's remove this in favor of the util method

src/main/java/net/dv8tion/jda/api/components/replacer/ComponentReplacer.java
```diff
@@ -0,0 +1,229 @@
...
+     * @param  tree
+     *         The {@link ComponentTree} on which to apply this replacer
+     *
+     * @throws IllegalArgumentException
+     *         If the provided tree is {@code null}
+     *
+     * @return The new component tree, with zero, one or more components replaced
+     */
+    @Nonnull
+    default <T extends Component> ComponentTree<T> applyOn(@Nonnull ComponentTree<T> tree)
```
MinnDevelopment: Let's remove this, I don't see a need to have both directions.

src/main/java/net/dv8tion/jda/api/components/actionrow/ActionRow.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    /**
+     * Check whether this row has a valid configuration.
+     *
+     * <p>This primarily checks the number of components in a single row.
+     *
+     * @return True, if this action row is valid
+     */
+    boolean isValid();
```
MinnDevelopment: Can be removed since its obsolete

src/main/java/net/dv8tion/jda/api/components/actionrow/ActionRow.java
```diff
@@ -0,0 +1,307 @@
...
+    @Override
+    @CheckReturnValue
+    ActionRow replace(@Nonnull ComponentReplacer replacer);
+
+    default boolean isDisabled()
+    {
+        return getActionComponents().stream().allMatch(ActionComponent::isDisabled);
+    }
+
+    default boolean isEnabled()
```
MinnDevelopment, line 245:
```suggestion
    @Override
    default boolean isEnabled()
```

src/main/java/net/dv8tion/jda/api/components/actionrow/ActionRow.java
```diff
@@ -0,0 +1,307 @@
...
+
+        return getComponents().stream().allMatch(Component::isModalCompatible);
+    }
+
+    @Nonnull
+    @Override
+    @CheckReturnValue
+    ActionRow replace(@Nonnull ComponentReplacer replacer);
+
+    default boolean isDisabled()
```
MinnDevelopment, line 239:
```suggestion
    @Override
    default boolean isDisabled()
```

src/main/java/net/dv8tion/jda/api/components/actionrow/ActionRow.java
```diff
@@ -0,0 +1,307 @@
...
+                .filter(Button.class::isInstance)
+                .map(Button.class::cast)
+                .collect(Helpers.toUnmodifiableList());
+    }
+
+    @Nonnull
+    @Override
+    @Deprecated
+    @ForRemoval
+    ActionRow createCopy();
```
MinnDevelopment: Remove this too

src/main/java/net/dv8tion/jda/api/components/ResolvedMedia.java
```diff
@@ -36,6 +36,19 @@ public interface ResolvedMedia
...
+     * The ID of the attachment represented by this resolved media,
+     * may return {@code null} if this media was created from an external link.
+     *
+     * @return The ID of the attachment, or {@code null}
+     */
+    @Nullable
+    default Long getAttachmentIdLong()
+    {
+        final String attachmentId = getAttachmentId();
+        return attachmentId != null ? Long.valueOf(attachmentId) : null;
```
MinnDevelopment:
```suggestion
        return attachmentId != null ? Long.parseUnsignedLong(attachmentId) : null;
```

FelixTPG:
Minn sorting out

src/main/java/net/dv8tion/jda/api/components/Components.java
```diff
@@ -0,0 +1,172 @@
...
+     * Converts the provided {@link DataArray} into a {@link ComponentTree}.
+     * <br>Note that any unsupported component will be represented as an {@link net.dv8tion.jda.api.components.UnknownComponent UnknownComponent}.
+     *
+     * @param  data
+     *         The {@link DataArray} to create the component tree from
+     *
+     * @return A {@link ComponentTree} representing the provided data
+     *
+     * @throws IllegalArgumentException
+     *         If the provided data is {@code null}, or one of the components is not of type {@link T}
```
MinnDevelopment: > one of the components is not of type {@link T}

Not compatible?

src/main/java/net/dv8tion/jda/api/interactions/modals/ModalTopLevelComponentUnion.java
```diff
@@ -0,0 +1,92 @@
...
+        switch (type) {
+            case ACTION_ROW:
+                return new ActionRowImpl(data);
+            default:
+                return new UnknownComponentImpl(data);
+        }
+    }
+
+    @Nonnull
+    static List<ModalTopLevelComponentUnion> fromData(@Nonnull DataArray data) {
```
MinnDevelopment: Let's remove these in favor of `Components.parseTree`

src/main/java/net/dv8tion/jda/api/utils/messages/MessageEditData.java
```diff
@@ -357,6 +357,33 @@ public synchronized List<FileUpload> getFiles()
...
+    }
+
+    @Nonnull
+    public static Set<AttachedFile> createAllDistinctFiles(
+            @Nullable Collection<AttachedFile> files,
+            @Nonnull Collection<MessageTopLevelComponentUnion> components
+    )
+    {
+        List<FileUpload> indirectFiles = MessageUtil.getIndirectFiles(components);
+        Set<AttachedFile> distinctFiles = new LinkedHashSet<>(files == null ? 0 : files.size() + indirectFiles.size());
```
MinnDevelopment:
```suggestion
        Set<AttachedFile> distinctFiles = new LinkedHashSet<>((files == null ? 0 : files.size()) + indirectFiles.size());
```

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -192,22 +221,24 @@ default R setEmbeds(@Nonnull MessageEmbed... embeds)
...
     @Nonnull
-    R setComponents(@Nonnull Collection<? extends LayoutComponent> components);
+    R setComponents(@Nonnull Collection<? extends MessageTopLevelComponent> components);
 
     /**
-     * The {@link LayoutComponent LayoutComponents} that should be attached to the message.
+     * The {@link MessageTopLevelComponent MessageTopLevelComponents} that should be attached to the message.
      * <br>You can call this method without anything to remove all components from the message.
      *
      * <p>The most commonly used layout is {@link ActionRow}.
```
MinnDevelopment:
```suggestion
```

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -229,104 +260,131 @@ default R setEmbeds(@Nonnull MessageEmbed... embeds)
...
     }
 
     /**
-     * Convenience method to set the components of a message to a single {@link ActionRow} of components.
-     * <br>To remove components, you should use {@link #setComponents(LayoutComponent...)} instead.
+     * The {@link MessageTopLevelComponent MessageTopLevelComponents} that should be attached to the message.
+     * <br>You can call this method without anything to remove all components from the message.
      *
-     * <p><b>Example</b><br>
+     * <p>The most commonly used layout is {@link ActionRow}.
```
MinnDevelopment:
```suggestion
```

src/main/java/net/dv8tion/jda/api/utils/messages/MessageRequest.java
```diff
@@ -229,104 +260,131 @@ default R setEmbeds(@Nonnull MessageEmbed... embeds)
...
-    default R setComponents(@Nonnull LayoutComponent... components)
+    default R setComponents(@Nonnull MessageTopLevelComponent... components)
     {
         return setComponents(Arrays.asList(components));
     }
 
     /**
-     * Convenience method to set the components of a message to a single {@link ActionRow} of components.
-     * <br>To remove components, you should use {@link #setComponents(LayoutComponent...)} instead.
+     * The {@link MessageTopLevelComponent MessageTopLevelComponents} that should be attached to the message.
```
MinnDevelopment: Did the docs get screwed up? These seem to be for a collection / varags method but are on the tree method

src/main/java/net/dv8tion/jda/internal/components/actionrow/ActionRowImpl.java
```diff
@@ -0,0 +1,217 @@
...
+        {
+            if (type != null && type != current.getType())
+            {
+                rows.add(ActionRow.of(currentRow));
+                currentRow.clear();
+            }
+
+            type = current.getType();
+
+            if (currentRow.size() == ActionRow.getMaxAllowed(type))
```
MinnDevelopment: This if branch never adds any element to the row, causing this to fail:

```java
ActionRow.partitionOf(button, button, button, button, button, button);
```

```
java.lang.IllegalArgumentException: Row may not be empty
	at net.dv8tion.jda.internal.utils.Checks.notEmpty(Checks.java:119)
	at net.dv8tion.jda.internal.components.actionrow.ActionRowImpl.<init>(ActionRowImpl.java:67)
	at net.dv8tion.jda.internal.components.actionrow.ActionRowImpl.<init>(ActionRowImpl.java:62)
	at net.dv8tion.jda.internal.components.actionrow.ActionRowImpl.of(ActionRowImpl.java:80)
	at net.dv8tion.jda.api.components.actionrow.ActionRow.of(ActionRow.java:61)
	at net.dv8tion.jda.internal.components.actionrow.ActionRowImpl.partitionOf(ActionRowImpl.java:117)
```

Let's write a unit test.

src/main/java/net/dv8tion/jda/internal/components/mediagallery/MediaGalleryItemFileUpload.java
```diff
@@ -0,0 +1,141 @@
...
+        return Stream.of(file);
+    }
+
+    @Nullable
+    @Override
+    public String getDescription()
+    {
+        // FileUpload is mutable unfortunately
+        final String fileDescription = file.getDescription();
+        if (fileDescription != null)
```
MinnDevelopment: I would expect `item.withDescription` to overwrite the file upload's description, so this should be inverted.

```java
if (description != null)
    return description;
return file.getDescription();
```

src/main/java/net/dv8tion/jda/internal/components/thumbnail/ThumbnailFileUpload.java
```diff
@@ -0,0 +1,172 @@
...
+    {
+        return null;
+    }
+
+    @Nullable
+    @Override
+    public String getDescription()
+    {
+        final String fileDescription = file.getDescription();
+        if (fileDescription != null)
```
MinnDevelopment: Same here, I expect the inverse

src/main/java/net/dv8tion/jda/internal/components/utils/ComponentsUtil.java
```diff
@@ -0,0 +1,170 @@
...
+        throw new IllegalArgumentException(Helpers.format("Cannot convert " + componentCategory + " of type %s to %s!", cleanedClassName, toUnionClass.getSimpleName()));
+    }
+
+    /**
+     * Checks the component has the target union type, and allows unknown components,
+     * throws {@link IllegalArgumentException} otherwise.
+     * <br>This should only be used for reading purposes,
+     * use {@link #membersToUnion(Collection, Class)} to verify components to be sent.
+     */
+    public static <T extends Component> T relaxedSafeUnionCast(String componentCategory, Component component, Class<T> toUnionClass)
```
MinnDevelopment:
```suggestion
    public static <T extends Component> T safeUnionCastWithUnknownType(String componentCategory, Component component, Class<T> toUnionClass)
```

src/main/java/net/dv8tion/jda/internal/components/utils/ComponentsUtil.java
```diff
@@ -0,0 +1,170 @@
...
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Retains all components extending the provided union type, keeps unknown components,
+     * throws {@link IllegalArgumentException} on invalid types.
+     * <br>This should only be used for reading purposes,
+     * use {@link #membersToUnion(Collection, Class)} to verify components to be sent.
+     */
+    public static <T extends Component> List<T> relaxedMembersToUnion(Collection<? extends Component> members, Class<T> clazz)
```
MinnDevelopment:
```suggestion
    public static <T extends Component> List<T> membersToUnionWithUnknownType(Collection<? extends Component> members, Class<T> clazz)
```

src/test/java/net/dv8tion/jda/test/compliance/ComponentConsistencyComplianceTest.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
+
+public class ComponentConsistencyComplianceTest
+{
+    final JavaClasses apiClasses = new ClassFileImporter().importPackages("net.dv8tion.jda.api");
```
MinnDevelopment: We should avoid doing this twice, since its pretty slow. It's better to share this between tests

src/test/java/net/dv8tion/jda/test/components/ContainerTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.Collections;
+
+public class ContainerTest
+{
+
+    @Test
+    void testEmptyContainerThrows()
+    {
+        Assertions.assertThatIllegalArgumentException()
```
MinnDevelopment: Minor: If you want you can extend `ChecksHelper` with a `assertCollectionChecks` to make this possible:

```java
assertCollectionChecks("Components", (list) -> new ContainerImpl(-1, list, false, null))
  .checksNotNull()
  .checksNotEmpty();
```

src/main/resources/META-INF/rewrite/rewrite.yml
```diff
@@ -0,0 +1,96 @@
...
+      artifactId: JDA
+      newVersion: 6.0.0
+
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: net.dv8tion.MigrateComponentsV2Packages
+description: Migrate to Components V2 packages.
+recipeList:
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: net.dv8tion.jda.api.interactions.components.ActionComponent
```
DV8FromTheWorld, line 26: What if people are using star imports?

src/test/java/net/dv8tion/jda/test/rewrite/MigrateComponentsV2Test.java
```diff
@@ -0,0 +1,224 @@
...
+{
+
+    @Override
+    public void defaults(RecipeSpec spec)
+    {
+        // https://docs.openrewrite.org/authoring-recipes/multiple-versions#manually-copying-jars-and-using-the-classpathfromresources-function
+        // I'd like to use their recipe-library, but it pulls a beta version of the shadow plugin, forcing us to upgrade
+        // which then breaks the buildscript for god knows what reason.
+        // I ain't putting more effort into what could have been at worst, a 5 minutes job for the end user,
+        // when it cost me my entire day.
```
MinnDevelopment: This comment can be removed

src/main/resources/META-INF/rewrite/rewrite.yml
freya022: It will prepend the new import

MinnDevelopment (approved):

PR #2810, opened 2025-03-11
Make IEventManager a FunctionalInterface

sixcorners:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Makes `IEventManager` into a `FunctionalInterface` so you can do something like `.setEventManager(System.out::println)` if you want to.


RealMangorage:
Couldn't you just have a helper?
IEventManager.of(...)?

freya022:
Y'all confused `IEventManager` with `EventListener`, which is already a functional interface.

The (quite unusual) use case you described is already possible with `EventListener`.

sixcorners:
I still think it would be a good change but ok I'll use EventListener.

PR #2814, opened 2025-03-17
remove wrong ephemeral part from javadocs

Kaktushose:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [x] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Removes the parts about ephemeral messages from the InteractionHook and Message docs that are wrong

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -1176,7 +1176,7 @@ public interface Message extends ISnowflake, Formattable
      *
      *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
      *     <br>The provided {@code messageId} is unknown in this MessageChannel, either due to the id being invalid, or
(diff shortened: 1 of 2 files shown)


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -1176,7 +1176,7 @@ default MessageEditAction editMessageComponents(@Nonnull LayoutComponent... comp
      *
      *     <li>{@link net.dv8tion.jda.api.requests.ErrorResponse#UNKNOWN_MESSAGE UNKNOWN_MESSAGE}
      *     <br>The provided {@code messageId} is unknown in this MessageChannel, either due to the id being invalid, or
-     *         the message it referred to has already been deleted. This might also be triggered for ephemeral messages.</li>
```
MinnDevelopment, line 1179: This is still true if the interaction expired I believe

Kaktushose: for all methods or just the `editMessageAttachments` methods?

MinnDevelopment: All of them

Issue #2817, opened 2025-03-25
User Installable Bot, Slash Commands stopped working in Bot DM


LordTylus:
### General Troubleshooting

- [x] I have checked for similar issues on the Issue-tracker.
- [x] I have checked for PRs that might already address this issue.

### Version of JDA

5.3.0_4d372fd

### Expected Behaviour

When using a user installable bot and I execute a Slash Command I do not want the error below, but the Command to actually be executed.

This only happens in Friend DM and Bot DM
In Group Chats and Guilds they still work fine.

Investigation shows that this code piece in JDA seems to trigger the error.

InteractionEntityBuilder.createPrivateChannel()
```
        if (recipientObj != null) {
            // We only get recipient in Bot DMs and Group DMs
            if (interactionUser.getIdLong() == recipientObj.getLong("id")) {
                channel = new PrivateChannelImpl(getJDA(), channelId, interactionUser);
            } else {
                // Friend DMs don't have recipients
                throw new IllegalArgumentException("Recipient should only be present in Bot DMs");
            }
        } else {
            // Friend DMs, no info
            channel = new DetachedPrivateChannelImpl(getJDA(), channelId);
        }
```

The error occurs since today 8pm UTC (round about, this is where it was first noticed)

Assumption is, that Discord did change something in the API, so that Friend DMs now do have a recipient. 

### Code Example for Reproduction Steps

```java
SlashCommandData slashCommandData = new CommandDataImpl(
                "test",
                "test");

        slashCommandData.setContexts(InteractionContextType.ALL);
        slashCommandData.setIntegrationTypes(IntegrationType.USER_INSTALL, IntegrationType.GUILD_INSTALL);

        jda.upsertCommand(slashCommandData).queue();
```

### Code for JDABuilder or DefaultShardManagerBuilder used

```java
JDA jda = JDABuilder
                .createDefault(botTokenConfig.getToken())
                .setAutoReconnect(true)
                .enableIntents(
                        GatewayIntent.GUILD_MESSAGES,
                        GatewayIntent.GUILD_EXPRESSIONS)
                .setEventPassthrough(true)
                .setChunkingFilter(ChunkingFilter.ALL)
                .addEventListeners(readyListener, messageListener)
                .build()
```

### Exception or Error

```yesyes
22:56:37,412 [ERROR] [JDA MainWS-ReadThread] [WebSocketClient] Got an unexpected error. Please redirect the following message to the devs:
    JDA 5.3.0_4d372fd
[...]

PR #2818, opened 2025-03-26
Fix assumptions in friend DM interactions 

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: #2817

## Description

Since Discord now sends recipients of Friend DMs, assumptions on how to detect friend DMs failed, this is now fixed.

Additionally:
- If the recipient is absent, a warn will be logged and a detached channel will be used
- Bot DMs from interactions will now target the interaction caller, this should be considered as a fix
- Detached private channels will now be able to return an `User` in some situations
- A new method `PrivateChannel#retrieveOpenPrivateChannel()` was added, this is intended for cases where the user wants the bot to *attempt to* send a DM to the friend on which the interaction was called
  - This is necessary as interactions started in friend DMs don't open a channel between the bot and the recipient

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/channel/concrete/PrivateChannel.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/channel/concrete/PrivateChannel.java
@@ -26,6 +26,15 @@ import javax.annotation.Nullable;
 /**
  * Represents the connection used for direct messaging.
  *
+ * <p>This channel may communicate with different users in interactions triggered by user-installed apps:
+ * <ul>
+ *     <li>In bot DMs, this channel will send messages to the interaction caller.</li>
+ *     <li>In friend DMs, this channel will send messages to that friend,
+ *         from the bot itself, this is different from where the interaction was executed.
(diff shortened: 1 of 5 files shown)


Reddishye:
Awesome! I was just looking for a fork or a pr with this fix (been struggling 30 minutes just to notice it was JDA5 issue)

src/main/java/net/dv8tion/jda/api/entities/channel/concrete/PrivateChannel.java
```diff
@@ -26,6 +26,15 @@
 /**
  * Represents the connection used for direct messaging.
  *
+ * <p>This channel may communicate with different users in interactions triggered by user-installed apps:
+ * <ul>
+ *     <li>In bot DMs, this channel will send messages to the interaction caller.</li>
+ *     <li>In friend DMs, this channel will send messages to that friend,
```
MinnDevelopment, line 32: More accurately, the bot just cannot send messages in these channels, it can only reply to the interaction. I think we can describe it like that to make it easier to parse.

PR #2826, opened 2025-04-11
Check for not blank strings instead of not empty

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

So users can't pass blank IDs/URLs/labels as valid ones.


MineKing9534:
Passing blank strings for button labels seems to be allowed. I tried passing ZERO_WIDTH_SPACE as button label and it works perfectly fine

freya022:
Whats the point of having a ZWSP as the label

MineKing9534:
> Whats the point of having a ZWSP as the label

I don't know a specific use case but it is possible so I don't see a reason to prevent it

PR #2833, opened 2025-04-23
No regex url check for EmbedBuilder

xDec0de:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Just a simple change to avoid using a regex Pattern when checking if a URL is valid on the EmbedBuilder class. Instead, it relies on String#regionMatches (Case insensitive) and a simple length check to have the exact same behaviour as the current pattern. I know this may be considered micro optimization, but I found it and thought it could be an easy first PR for me.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/EmbedBuilder.java
@@ -907,12 +907,29 @@ public class EmbedBuilder
         return fields;
     }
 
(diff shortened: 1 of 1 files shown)


xDec0de:
Oh, I forgot to mention that I was going to deprecate EmbedBuilder#URL_PATTERN in favor of EmbedBuilder#isUrlOrAttachment, but I decided not to and leave that decision to the maintainers.

PR #2842, opened 2025-05-01
Add `AbstractThreadCreateAction#setSlowmode`

replaceitem:

## Pull Request Etiquette

### Changes

- [x] Internal code (Only additions)
- [x] Library interface (affecting end-user code)  (Only additions)
- [x] Documentation (Only additions)
- [ ] Other: \_____

## Description

This adds the ability to set the slowmode when creating a thread channel or forum post. Just like [`ChannelAction.setSlowmode`](https://github.com/discord-jda/JDA/blob/b1f2e2873db1a95ae6da34c9d6479accadcd3fa1/src/main/java/net/dv8tion/jda/internal/requests/restaction/ChannelActionImpl.java#L215), the slowmode can be set when creating forum or thread channels. This utilizes the `rate_limit_per_user` value in the [Start Thread](https://discord.com/developers/docs/resources/channel#start-thread-without-message) endpoints.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/requests/restaction/AbstractThreadCreateAction.java
+++ b/src/main/java/net/dv8tion/jda/api/requests/restaction/AbstractThreadCreateAction.java
@@ -103,6 +103,9 @@ public interface AbstractThreadCreateAction<T, R extends AbstractThreadCreateAct
      * @throws IllegalArgumentException
      *         If the provided slowmode is negative or greater than {@value ISlowmodeChannel#MAX_SLOWMODE}
      *
+     * @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
+     *         If the bot does not have {@link Permission#MANAGE_THREADS} in the parent channel
+     *
      * @return The current action, for chaining convenience
      *
      * @see net.dv8tion.jda.api.entities.channel.attribute.ISlowmodeChannel#getSlowmode()
--- a/src/main/java/net/dv8tion/jda/api/requests/restaction/AbstractThreadCreateAction.java
+++ b/src/main/java/net/dv8tion/jda/api/requests/restaction/AbstractThreadCreateAction.java
@@ -16,9 +16,11 @@
 
 package net.dv8tion.jda.api.requests.restaction;
 
+import net.dv8tion.jda.api.Permission;
 import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.entities.channel.Channel;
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/api/requests/restaction/AbstractThreadCreateAction.java
```diff
@@ -84,4 +86,32 @@ public interface AbstractThreadCreateAction<T, R extends AbstractThreadCreateAct
...
+     *
+     * <p>A channel slowmode <b>must not</b> be negative nor greater than {@link ISlowmodeChannel#MAX_SLOWMODE}!
+     *
+     * <p>Note: Bots are unaffected by this.
+     * <br>Having {@link Permission#MESSAGE_MANAGE MESSAGE_MANAGE} or
+     * {@link Permission#MANAGE_CHANNEL MANAGE_CHANNEL} permission also
+     * grants immunity to slowmode.
+     *
+     * <p><b>Special case</b><br>
+     * {@link net.dv8tion.jda.api.entities.channel.concrete.ForumChannel ForumChannels} use this to limit how many posts a user can create.
```
MinnDevelopment, line 101: This part is irrelevant here

src/main/java/net/dv8tion/jda/internal/requests/restaction/ForumPostActionImpl.java
```diff
@@ -132,6 +135,16 @@ public ForumPostAction setAutoArchiveDuration(@Nonnull ThreadChannel.AutoArchive
         this.autoArchiveDuration = autoArchiveDuration;
         return this;
     }
+    
+    @Nonnull
+    @Override
+    public ForumPostAction setSlowmode(int slowmode)
```
MinnDevelopment, line 141: After some testing, I noticed that this requires the `MANAGE_THREADS` permission, so we need to add a permission check here (and in the other impl as well):

```java
if (!getGuild().getSelfMember().hasPermission(channel, Permission.MANAGE_THREADS))
    throw new InsufficientPermissionException(channel, Permission.MANAGE_THREADS);
```

Documented like so

```java
* @throws net.dv8tion.jda.api.exceptions.InsufficientPermissionException
*         If the bot does not have {@link Permission#MANAGE_THREADS} in the parent channel
```

replaceitem: I see, added it now, thanks.

MinnDevelopment (approved):

PR #2863, opened 2025-07-06
Add support for user primary guilds

mjezek-dev:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Adds support for [user primary guilds](https://discord.com/developers/docs/resources/user#user-object-user-primary-guild).

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/User.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/User.java
@@ -33,6 +33,7 @@ import java.awt.*;
 import java.util.Collection;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.regex.Pattern;
 
 /**
@@ -88,6 +89,8 @@ public interface User extends UserSnowflake
     String DEFAULT_AVATAR_URL = "https://cdn.discordapp.com/embed/avatars/%s.png";
     /** Template for {@link Profile#getBannerUrl()} */
     String BANNER_URL = "https://cdn.discordapp.com/banners/%s/%s.%s";
+	/** Template for {@link PrimaryGuild#getBadgeUrl()} */
+	String TAG_BADGE_URL = "https://cdn.discordapp.com/guild-tag-badges/%s/%s.png";
 
     /** Used to keep consistency between color values used in the API */
     int DEFAULT_ACCENT_COLOR_RAW = 0x1FFFFFFF; // java.awt.Color fills the MSB with FF, we just use 1F to provide better consistency
@@ -352,6 +355,14 @@ public interface User extends UserSnowflake
      * @return bitmask representation of the user's flags.
      */
     int getFlagsRaw();
+	
+	/**
+     * Returns the {@link PrimaryGuild} of this user.
+     *
+     * @return Possibly-null {@link PrimaryGuild} of this user.
+     */
+	@Nullable
+	PrimaryGuild getPrimaryGuild();
 
     /**
      * Represents the information contained in a {@link User User}'s profile.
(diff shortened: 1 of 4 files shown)


AveCard1nal:
May I also suggest adding:

```java
List<Member> Guild#getMembersByPrimaryGuild()
RestAction<List<Member>> Guild#retrieveMembersByPrimaryGuild()
```

This could be useful for cases where bots need to identify members who selected the guild as their primary. For example, to assign roles, highlight core members, or provide exclusive features.

mjezek-dev:
You could use `Guild#findMembers` with filter checking for member's primary guild to retrieve this info. Discord API doesn't directly support searching for members based on their primary guild like that is being done with usernames. Looking for cached members can be done just by iterating over list of all cached members from `Guild#getMembers` and checking for required primary guild, so I don't think it's needed to add specific method for this.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -481,6 +481,10 @@ public UserImpl createUser(DataObject user)
 
         if (newUser)
         {
+            User.PrimaryGuild primaryGuild = user.optObject("primary_guild")
+                   .map(obj -> new User.PrimaryGuild(obj.getString("identity_guild_id"), obj.getBoolean("identity_enabled"), obj.getString("tag"), obj.getString("badge")))
```
MinnDevelopment: Move this to entity builder to avoid the duplicated parsing logic

src/main/java/net/dv8tion/jda/api/entities/User.java
```diff
@@ -624,4 +636,123 @@ public static int getRaw(@Nonnull Collection<UserFlag> flags)
             return getRaw(flags.toArray(EMPTY_FLAGS));
         }
     }
+    
+    /**
+     * Represents the information about {@link User User}'s primary guild
+     */
+    class PrimaryGuild
+    {
+        private final String identityGuildId;
```
MinnDevelopment: We use `long` for this everywhere in JDA. Let's just do `private final long guildId` and make `PrimaryGuild` implement `ISnowflake` with that id.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -457,6 +457,15 @@ public GuildChannel createGuildChannel(GuildImpl guildObj, DataObject channelDat
             return null;
         }
     }
+    
+    public User.PrimaryGuild createPrimaryGuild(DataObject obj)
+    {
+        long guildId = obj.getLong("identity_guild_id", -1);
+        if (guildId == -1) // Possibly null according to docs
```
MinnDevelopment: Instead of using `-1`, which is a valid snowflake, let's just check `obj.isNull("identity_guild_id")`

FigT:
Hey! Is there anything blocking this PR? This would be useful :}

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -457,6 +457,14 @@ public GuildChannel createGuildChannel(GuildImpl guildObj, DataObject channelDat
             return null;
         }
     }
+    
+    public User.PrimaryGuild createPrimaryGuild(DataObject obj)
+    {
+        if (obj.isNull("identity_guild_id"))
+            return null;
+        
+        return new User.PrimaryGuild(obj.getLong("identity_guild_id"), obj.getBoolean("identity_enabled"), obj.getString("tag", null), obj.getString("badge", null));
```
MinnDevelopment:
```suggestion
        return new User.PrimaryGuild(obj.getUnsignedLong("identity_guild_id"), obj.getBoolean("identity_enabled"), obj.getString("tag", null), obj.getString("badge", null));
```

PR #2895, opened 2025-08-21
Add message pin permission

freya022:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR adds `Permission#PIN_MESSAGES`, updates docs and changes checks to allow `MESSAGE_MANAGE` or `PIN_MESSAGES` before Discord's backward compatibility deadline, or, `PIN_MESSAGES` after the deadline

### DIFF
--- a/src/test/java/net/dv8tion/jda/test/entities/channel/MessagePinDeadlineTest.java
+++ b/src/test/java/net/dv8tion/jda/test/entities/channel/MessagePinDeadlineTest.java
@@ -107,9 +107,6 @@ public class MessagePinDeadlineTest
                 Arguments.argumentSet("Before deadline, fail if no MESSAGE_MANAGE or PIN_MESSAGES",
                         /* timeOfCheck */ BEFORE_DEADLINE,
                         /* expectedCheckedPermissions */ EnumSet.of(Permission.MESSAGE_MANAGE, Permission.PIN_MESSAGES)),
-                Arguments.argumentSet("Before deadline, fail if no MESSAGE_MANAGE",
-                        /* timeOfCheck */ BEFORE_DEADLINE,
-                        /* expectedCheckedPermissions */ EnumSet.of(Permission.MESSAGE_MANAGE, Permission.PIN_MESSAGES)),
                 Arguments.argumentSet("After deadline, fail if no PIN_MESSAGES",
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/internal/entities/channel/mixin/middleman/GuildMessageChannelMixin.java
```diff
@@ -163,11 +166,20 @@ default void checkCanRemoveReactions()
...
+        checkPermission(Permission.PIN_MESSAGES, "You need PIN_MESSAGES to pin or unpin messages.");
     }
 
     default boolean canDeleteOtherUsersMessages()
     {
         return hasPermission(Permission.MESSAGE_MANAGE);
     }
+
+    // So we can mock it instead of JDK classes
+    default Instant currentInstant()
```
MinnDevelopment: This shouldn't be done through mocks, instead we should have a central clock that can be changed to a test clock.

`ClockProvider.getClock()`  
- By default: `Clock.systemUTC()`
- Having a setter for tests, to allow using `setClock(Clock.fixed(TEST_TIME, UTC))` with a reset before each test

Then use `clock.instant()`

src/test/java/net/dv8tion/jda/test/entities/channel/MessagePinDeadlineTest.java
```diff
@@ -0,0 +1,102 @@
...
+        channelInOrder.verify(channel, never()).hasPermission(any());
+    }
+
+    static Stream<Arguments> invalidPermissions()
+    {
+        return Stream.of(
+                Arguments.argumentSet("Before deadline, fail if no MESSAGE_MANAGE or PIN_MESSAGES",
+                        /* timeOfCheck */ BEFORE_DEADLINE,
+                        /* expectedCheckedPermissions */ EnumSet.of(Permission.MESSAGE_MANAGE, Permission.PIN_MESSAGES)),
+                Arguments.argumentSet("Before deadline, fail if no MESSAGE_MANAGE",
```
MinnDevelopment: This is identical to the previous case, no?

Issue #2904, opened 2025-08-30
Add System Channel Flags


ShadowMasterGaming:
### General Troubleshooting

- [x] I have checked for similar issues on the Issue-tracker.
- [x] I have updated to the [latest JDA version](https://github.com/discord-jda/JDA/releases/latest)
- [x] I have checked the branches or the maintainers' PRs for upcoming features.

### Feature Request

System Channel Flags from the discord docs are missing from the API.

https://discord.com/developers/docs/resources/guild#guild-object-system-channel-flags

### Example Use-Case

```java
Potential enum names for the flags.

GuildFlag.SUPPRESS_JOIN_MESSAGES
GuildFlag.SUPPRESS_BOOST_MESSAGES
GuildFlag.SUPPRESS_SETUP_TIPS
GuildFlag.SUPPRESS_WELCOME_STICKER_REPLY
GuildFlag.SUPPRESS_RENEWAL_NOTIFICATIONS
GuildFlag.SUPPRESS_PURCHASE_NOTIFICATION_REPLIES

Would need a new method to go with it; possibly:
Guild#setSystemFlags(Collection<Permission> permissions)
```

MinnDevelopment:
This has been added in 6.1.1

PR #2928, opened 2025-10-05
Implement Reading System Channel Flags

LaFriska:

## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: [#2904 Add System Channel Flags](https://github.com/discord-jda/JDA/issues/2904)

## Description

As discussed in the linked issue, JDA currently does not support reading [system channel flags](https://discord.com/developers/docs/resources/guild#guild-object-system-channel-flags) from guilds. This PR introduces system channel flags represented by a new enum type `SystemChannelFlag`. According to the Discord docs, each flag is stored as one-hot encoded bitmasks in a binary string which represents a set of flags. These bit representations are stored as `int`s in `GuildImpl`.  This is also why logic to convert between `SystemChannelFlag` and its bit-representation are necessary, and hence implemented and tested. In accordance with various guild update events, a `GuildUpdateSystemChannelFlagEvent` was also implemented and handled.

### Discord API Pull Request Link

System channel flag documentation were introduced to the Discord API [here](https://github.com/discord/discord-api-docs/pull/1262).

### Examples of Use

Below is an example subclass of `ListenerAdapter` that showcases the new feature.

```java
import net.dv8tion.jda.api.entities.channel.concrete.TextChannel;
import net.dv8tion.jda.api.entities.guild.SystemChannelFlag;
import net.dv8tion.jda.api.events.guild.update.GuildUpdateSystemChannelFlagsEvent;
import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
import net.dv8tion.jda.api.hooks.ListenerAdapter;
import org.jetbrains.annotations.NotNull;

import java.util.EnumSet;

public class ExampleEventHandler extends ListenerAdapter
{
    @Override
    public void onMessageReceived(@NotNull MessageReceivedEvent event) {
        if(event.getAuthor().isBot()) return; //Prevents infinite loop
        EnumSet<SystemChannelFlag> flags = event.getGuild().getSystemChannelFlags(); //Gets the system channel flags
        TextChannel channel = event.getChannel().asTextChannel();
        String flagList = formatFlags(flags, "System Channel Flags"); //Calls a helper method to format it as a string
        channel.sendMessage(flagList).queue();
    }

   //Called whenever the system channel flags are modified
    @Override
    public void onGuildUpdateSystemChannelFlags(@NotNull GuildUpdateSystemChannelFlagsEvent event) {
        TextChannel t = event.getGuild().getSystemChannel();
        StringBuilder sb = new StringBuilder("System channel flags changed!\n");
        if(t != null) {
            String oldFlags = formatFlags(event.getOldFlags(), "Old Flags"); //Old flags before modification
            String newFlags = formatFlags(event.getNewFlags(), "New Flags"); //New flags after modification
            sb.append(oldFlags).append("\n");
            sb.append(newFlags).append("\n");
        }
        t.sendMessage(sb.toString()).queue();
    }

    //Helper to format the flags in a nice string
    private String formatFlags(EnumSet<SystemChannelFlag> flags, String title){
        StringBuilder sb = new StringBuilder("**" + title + "**\n");
        for (SystemChannelFlag flag : flags) {
            sb.append("- ").append(flag).append("\n");
        }
        return sb.toString();
    }
}
```

### Changes

- `Guild.java`: Added new `getSystemChannelFlags` method.
- Added `SystemChannelFlag.java`.
- Added `GuildUpdateSystemChannelFlagEvent.java`.
- `ListenerAdapter.java`: added `onGuildUpdateSystemChannelFlags` method.
- `EntityBuilder.java`: Now stores system channel flags in the guild implementation.
- `GuildImpl.java`: added setter, and overrode getter from `Guild`
- `DetachedGuildImpl.java`: added override to throw `DetachedEntityException`.
- `GuildUpdateHandler.java`: added logic for handling updates to system channel flags.
- Added `SystemChannelFlagsTest.java` to test `SystemChannelFlag` logic.


src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.HashMap;
+
+/**
+ * Represents Discord's system channel flags of a guild. A system channel is a
+ * unique channel in a guild where guild notices such as welcome messages and boost events
+ * are posted. System channel flags configure suppressions of these messages and events. According
+ * to the <a href="https://discord.com/developers/docs/">Discord API docs<a/>, system channel flags
+ * are represented as one-hot encoded bitmasks, where each bit in a binary string represents whether
+ * a particular flag is enabled. This class contains methods to convert these bitmasks to an {@link EnumSet}
+ * of this class, which is the representation of system channel flags used by users of JDA.
```
freya022:
```suggestion
 * Flags which configures a {@linkplain Guild#getSystemChannel() system channel} of a {@link Guild}.
```

1. The explanation of a system channel is already in `Guild#getSystemChannel()`
2. It is technically correct right now to say system channel flags are for suppressions , but it may not be in the future, so I removed it
3. It is not necessary to explain how the flags are encoded, the user will often encounter them in a `Set<SystemChannelFlag>`
4. The conversion methods are already discoverable in your IDE and the Javadocs
5. Try to split large one liners

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+     * Suppress member join notifications.
+     * */
+    SUPPRESS_JOIN_NOTIFICATIONS(0),
+
+    /**
+     * Suppress server boost notifications.
+     * Identical to the "SUPPRESS_PREMIUM_SUBSCRIPTIONS" flag in the API
+     * docs.
+     * */
+    SUPPRESS_BOOST_NOTIFICATIONS(1),
```
freya022: Let's use the same name as the API, no need to reword it

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Suppress server setup tips.
+     * Identical to the "SUPPRESS_GUILD_REMINDER_NOTIFICATIONS" flag in the
+     * API docs.
+     * */
+    SUPPRESS_SETUP_TIPS(2),
```
freya022: Let's use the same name as the API, no need to reword it

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Hide member join sticker reply buttons.
+     * Identical to "SUPPRESS_JOIN_NOTIFICATION_REPLIES" flag in the
+     * API docs.
+     * */
+    HIDE_JOIN_REPLY_STICKERS(3),
```
freya022: Let's use the same name as the API, no need to reword it

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Suppress role subscription purchase and renewal notifications.
+     * Identical to "SUPPRESS_ROLE_SUBSCRIPTION_PURCHASE_NOTIFICATIONS" flag
+     * in the API docs.
+     * */
+    SUPPRESS_ROLE_PURCHASE_NOTIFICATIONS(4),
```
freya022: Let's use the same name as the API, no need to reword it

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Hide role subscription sticker reply buttons. Identical to
+     * "SUPPRESS_ROLE_SUBSCRIPTION_PURCHASE_NOTIFICATION_REPLIES" flag
+     * in the API docs.
+     * */
+    HIDE_ROLE_SUBSCRIPTION_STICKERS(5);
```
freya022: Let's use the same name as the API, no need to reword it

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Converts a bitmask representation of system channel flags into its
+     * {@link EnumSet}.
+     * @param bitmask the raw bitmask representing the system channel flags.
+     * @return an {@link EnumSet} of system channel flags represented by the input bitmask.
+     * @throws IllegalArgumentException if the input bitmask is invalid.
+     * */
+    @Nonnull
+    public static EnumSet<SystemChannelFlag> fromBitmask(final int bitmask)
```
freya022:
```suggestion
    public static EnumSet<SystemChannelFlag> getFlags(int raw)
```

Please refer to:
https://github.com/discord-jda/JDA/blob/e2c3b7439c46b23ca77fae804f75153d315d4b31/src/main/java/net/dv8tion/jda/api/requests/GatewayIntent.java#L265-L274

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Hide role subscription sticker reply buttons. Identical to
+     * "SUPPRESS_ROLE_SUBSCRIPTION_PURCHASE_NOTIFICATION_REPLIES" flag
+     * in the API docs.
+     * */
+    HIDE_ROLE_SUBSCRIPTION_STICKERS(5);
+
+    private final int offset;
+
+    private static final HashMap<Integer, SystemChannelFlag> OFFSET_TO_FLAG = new HashMap<>();
```
freya022: This map does not provide any benefit, a linear search will be just as good for the few values we have, and will reduce complexity when converting between raw <--> `Set<SystemChannelFlag>`

Also note that it is preferable for the type to use the corresponding interface, `Map` here

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+    {
+        for (SystemChannelFlag flag : SystemChannelFlag.values())
+        {
+            OFFSET_TO_FLAG.put(flag.offset, flag);
+        }
+    }
+
+    SystemChannelFlag(int offset)
+    {
+        this.offset = offset;
```
freya022, line 70: Introduce and expose a `rawValue`, like what `GatewayIntent` has

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+        return enumSet;
+    }
+
+    /**
+     * Converts an {@link EnumSet} of this class to its respective bitmask representing the same
+     * set of system channel flags.
+     * @param enumSet a set of system channel flags.
+     * @return an integer bitmask recognised on Discord's side.
+     * */
+    public static int toBitmask(final EnumSet<SystemChannelFlag> enumSet)
```
freya022:
```suggestion
    public static int getRaw(Collection<SystemChannelFlag> flags)
```
Staying consistent with `GatewayIntent`

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+    {
+        int result = 0;
+        for (SystemChannelFlag flag : enumSet)
+        {
+            result |= 0b1 << flag.offset;
```
freya022: Use the newly introduced `rawFlag` to rework this, see https://github.com/discord-jda/JDA/blob/e2c3b7439c46b23ca77fae804f75153d315d4b31/src/main/java/net/dv8tion/jda/api/requests/GatewayIntent.java#L287-L293

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+        return enumSet;
+    }
+
+    /**
+     * Converts an {@link EnumSet} of this class to its respective bitmask representing the same
+     * set of system channel flags.
+     * @param enumSet a set of system channel flags.
+     * @return an integer bitmask recognised on Discord's side.
+     * */
```
freya022: Please edit all of your javadocs to follow the convention https://jda.wiki/contributing/structure-guide/#example-template

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -1335,6 +1336,21 @@ default String getOwnerId()
...
+     *
+     * @throws net.dv8tion.jda.api.exceptions.DetachedEntityException
+     *         If this entity is {@link #isDetached() detached}.
+     * @see <a href="https://discord.com/developers/docs/resources/guild#guild-object-system-channel-flags">
+     *     System Channel Flags API documentation
+     *     <a/>
+     * @return the system channel flags of this guild.
+     * */
+    @Nonnull
+    EnumSet<SystemChannelFlag> getSystemChannelFlags();
```
freya022: Return and document as an **unmodifiable** `Set`, we do not want people to modify the set then complain it doesn't do anything

src/main/java/net/dv8tion/jda/api/events/guild/update/GuildUpdateSystemChannelFlagsEvent.java
```diff
@@ -0,0 +1,69 @@
...
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.entities.Guild;
+import net.dv8tion.jda.api.entities.guild.SystemChannelFlag;
+
+import javax.annotation.Nonnull;
+import java.util.EnumSet;
+
+/**
+ * Indicates that {@link Guild#getSystemChannelFlags() systemChannelFlags} have been updated. Note
+ * that these flags are stored as {@link EnumSet}s of {@link SystemChannelFlag}.
```
freya022:
```suggestion
 * Indicates that {@link Guild#getSystemChannelFlags()} have been updated.
```

src/main/java/net/dv8tion/jda/api/events/guild/update/GuildUpdateSystemChannelFlagsEvent.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * <p>This event can be used to detect changes in system channel flags and retrieve the old one.
+ *
+ * <p>Identifier: {@code system_channel_flags}
+ * */
+public class GuildUpdateSystemChannelFlagsEvent extends GenericGuildUpdateEvent<EnumSet<SystemChannelFlag>>
```
freya022: Same thing here, expose a `Set<SystemChannelFlag>` instead

src/main/java/net/dv8tion/jda/api/events/guild/update/GuildUpdateSystemChannelFlagsEvent.java
```diff
@@ -0,0 +1,69 @@
...
+        super(api, responseNumber, guild, oldFlags, newFlags, IDENTIFIER);
+    }
+
+    /**
+     * The old system channel flags for this guild.
+     *
+     * @return The old system channel flags for this guild, or null if none was set.
+     */
+    @Nonnull
+    public EnumSet<SystemChannelFlag>  getOldFlags()
```
freya022:
```suggestion
    public Set<SystemChannelFlag> getOldFlags()
```

Return and document as unmodifiable

src/main/java/net/dv8tion/jda/api/events/guild/update/GuildUpdateSystemChannelFlagsEvent.java
```diff
@@ -0,0 +1,69 @@
...
+        return getOldValue();
+    }
+
+    /**
+     * The new system channel flags for this guild.
+     *
+     * @return The new system channel flags for this guild, or null if none was set.
+     */
+    @Nonnull
+    public EnumSet<SystemChannelFlag>  getNewFlags()
```
freya022:
```suggestion
    public Set<SystemChannelFlag> getNewFlags()
```
Return and document as unmodifiable

src/main/java/net/dv8tion/jda/api/entities/Guild.java
freya022: Also introduce `int getSystemChannelFlagsRaw()`, so users can read all flags even if not in the enum yet

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,138 @@
...
+    {
+        EnumSet<SystemChannelFlag> enumSet = EnumSet.noneOf(SystemChannelFlag.class);
+        int offset = 0;
+        for (int v = bitmask; v != 0; v >>= 1)
+        {
+            int lsb = v & 0b1;
+            if(lsb == 0b1)
+            {
+                if(!OFFSET_TO_FLAG.containsKey(offset))
+                    throw new IllegalArgumentException("Input value must be a valid bitmask of Discord system channel flags.");
```
freya022: No need to throw when doing a simple conversion

freya022 (changes requested):

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
LaFriska: I will remove this map, after adapting the conversion algorithm from `GatewayIntent` it is completely redundant.

LaFriska:
Hi,

I have finished implementation of modifying system channel flags. I tested it using a Discord bot, and I was not able to detect any issues with this. I'd appreciate if a maintainer could have a look at the additional changes. 

Thanks,
Friska.

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -78,6 +79,8 @@ public interface GuildManager extends Manager<GuildManager>
     long FEATURES = 1 << 15;
     /** Used to rest the safety alerts channel field */
     long SAFETY_ALERTS_CHANNEL = 1 << 16;
+    /** Used to reset the system channel flag fields */
+    long SYSTEM_CHANNEL_FLAGS = 1 << 17;
```
freya022, line 84: You need to add it to the documented flag lists

src/main/java/net/dv8tion/jda/internal/handle/GuildUpdateHandler.java
```diff
@@ -96,6 +98,10 @@ protected Long handleInternally(DataObject content)
         TextChannel safetyAlertsChannel = content.isNull("safety_alerts_channel_id")
                 ? null : guild.getChannelById(TextChannel.class, content.getLong("safety_alerts_channel_id"));
         Set<String> features;
+
+        final int systemChannelFlagBitmask = content.getInt("system_channel_flags", 0);
+        EnumSet<SystemChannelFlag> systemChannelFlags = SystemChannelFlag.getFlags(systemChannelFlagBitmask);
```
freya022: Needs to be unmodifiable

src/main/java/net/dv8tion/jda/internal/managers/GuildManagerImpl.java
```diff
@@ -329,6 +334,40 @@ private GuildManager updateFeatures(Collection<String> changed, Consumer<String>
         return this;
     }
 
+    @Nonnull
+    @Override
+    public GuildManager setSystemChannelFlags(@Nonnull Collection<SystemChannelFlag> flags)
+    {
+        Checks.noneNull(flags, "System channel flag");
+        this.systemChannelFlags = flags;
```
freya022: Make a copy into an enum set, there is a method in Helpers for it, in case the collection is empty

freya022 (changes requested):

LaFriska:
Thank you very much for your patience, I have once again resolved the requested changes, please let me know if there are anything else that needs refinement. 👍

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +485,109 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+     * @param  flags
+     *         Collection of the new flags to use, flags not included in the collection
+     *         will be disabled.
+     *
+     * @throws IllegalArgumentException
+     *         If any of the flags in the provided collection are null
+     *
+     * @return GuildManager for chaining convenience
+     *
+     * @see SystemChannelFlag
```
freya022:
```suggestion
     * @see    SystemChannelFlag
```

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +485,109 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+    /**
+     * Enables the given {@linkplain Guild#getSystemChannelFlags() system channel flags}.
+     * <br>The list of available system channel flags, is available in the
+     * <a href="https://discord.com/developers/docs/resources/guild#guild-object-system-channel-flags" target="_blank">Official Discord API Documentation</a>.
+     *
+     * @param  flags
+     *         The flags to add
+     *
+     * @throws IllegalArgumentException
+     *         If any of the flags in the provided collection are null
```
freya022: This also happens if the collection itself is null

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +485,109 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+     * flags.add(SystemChannelFlag.SUPPRESS_JOIN_NOTIFICATIONS);
+     * guild.getManager().setSystemChannelFlags(flags).queue();
+     * }</pre>
+     *
+     * @param  flags
+     *         Collection of the new flags to use, flags not included in the collection
+     *         will be disabled.
+     *
+     * @throws IllegalArgumentException
+     *         If any of the flags in the provided collection are null
```
freya022: This also happens if the collection itself is null

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +485,109 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+    /**
+     * Enables the given {@linkplain Guild#getSystemChannelFlags() system channel flags}.
+     * <br>The list of available system channel flags, is available in the
+     * <a href="https://discord.com/developers/docs/resources/guild#guild-object-system-channel-flags" target="_blank">Official Discord API Documentation</a>.
+     *
+     * @param  flags
+     *         The flags to add
+     *
+     * @throws IllegalArgumentException
+     *         If any of the flags in the provided collection are null
```
freya022: This isn't a collection, also happens if the array itself is null

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +485,109 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+    /**
+     * Removes the given {@linkplain Guild#getSystemChannelFlags() system channel flags}.
+     * <br>The list of available system channel flags, is available in the
+     * <a href="https://discord.com/developers/docs/resources/guild#guild-object-system-channel-flags" target="_blank">Official Discord API Documentation</a>.
+     *
+     * @param  flags
+     *         The flags to remove
+     *
+     * @throws IllegalArgumentException
+     *         If any of the flags in the provided collection are null
```
freya022: This also happens if the collection itself is null

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +485,109 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
```
freya022: This isn't a collection, also happens if the array itself is null

src/main/java/net/dv8tion/jda/internal/managers/GuildManagerImpl.java
```diff
@@ -329,6 +335,40 @@ private GuildManager updateFeatures(Collection<String> changed, Consumer<String>
...
+
+    private GuildManager updateSystemChannelFlags(Collection<SystemChannelFlag> changed, Consumer<SystemChannelFlag> op)
+    {
+        Checks.noneNull(changed, "System channel flag");
+        if (this.systemChannelFlags == null)
+            this.systemChannelFlags = new HashSet<>(SystemChannelFlag.getFlags(getGuild().getSystemChannelFlagsRaw()));
+        changed.forEach(op);
+        set |= SYSTEM_CHANNEL_FLAGS;
+        return this;
+    }
```
freya022, line 371:
```suggestion
    public GuildManager enableSystemChannelFlags(@Nonnull Collection<SystemChannelFlag> flags)
    {
        return updateSystemChannelFlags(flags, Set::addAll);
    }

    @Nonnull
    @Override
    public GuildManager disableSystemChannelFlags(@Nonnull Collection<SystemChannelFlag> flags)
    {
        return updateSystemChannelFlags(flags, Set::removeAll);
    }

    private GuildManager updateSystemChannelFlags(Collection<SystemChannelFlag> flags, BiConsumer<Set<SystemChannelFlag>, Collection<SystemChannelFlag>> bulkUpdateOp)
    {
        Checks.noneNull(flags, "System channel flag");
        if (this.systemChannelFlags == null)
            this.systemChannelFlags = Helpers.copyEnumSet(SystemChannelFlag.class, getGuild().getSystemChannelFlags());
        bulkUpdateOp.accept(this.systemChannelFlags, flags);
        set |= SYSTEM_CHANNEL_FLAGS;
        return this;
    }
```

src/test/java/net/dv8tion/jda/test/entities/guild/SystemChannelFlagTest.java
freya022, line 1: Imo this has no benefit, but lets see what maintainers think about it

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +485,109 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+     *
+     * @return GuildManager for chaining convenience
+     *
+     * @see SystemChannelFlag
+     */
+    @Nonnull
+    @CheckReturnValue
+    GuildManager setSystemChannelFlags(@Nonnull Collection<SystemChannelFlag> flags);
```
freya022, line 515: Missing a vararg overload, no need to require any amount of flags

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +485,109 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+     *
+     * @return GuildManager for chaining convenience
+     */
+    @Nonnull
+    @CheckReturnValue
+    default GuildManager enableSystemChannelFlags(@Nonnull SystemChannelFlag... flags)
+    {
+        Checks.noneNull(flags, "System channel flag");
+        return enableSystemChannelFlags(Arrays.asList(flags));
+    }
```
freya022, line 580:
```suggestion
    default GuildManager enableSystemChannelFlags(@Nonnull SystemChannelFlag flag, @Nonnull SystemChannelFlag... flags)
    {
        Checks.notNull(flag, "System channel flag");
        Checks.noneNull(flags, "System channel flags");
        return enableSystemChannelFlags(EnumSet.of(flag, flags));
    }
```

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +485,109 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+     *
+     * @return GuildManager for chaining convenience
+     */
+    @Nonnull
+    @CheckReturnValue
+    default GuildManager disableSystemChannelFlags(@Nonnull SystemChannelFlag... flags)
+    {
+        Checks.noneNull(flags, "System channel flag");
+        return disableSystemChannelFlags(Arrays.asList(flags));
+    }
```
freya022, line 620:
```suggestion
    default GuildManager disableSystemChannelFlags(@Nonnull SystemChannelFlag flag, @Nonnull SystemChannelFlag... flags)
    {
        Checks.notNull(flag, "System channel flag")
        Checks.noneNull(flags, "System channel flags");
        return disableSystemChannelFlags(EnumSet.of(flag, flags));
    }
```

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -1335,6 +1336,39 @@ default String getOwnerId()
...
+    /**
+     * Returns a {@link Set} of {@link SystemChannelFlag} associated to the guild.
+     * <br>For a description of what system channel flags represent, see {@link SystemChannelFlag}.
+     *
+     * @throws net.dv8tion.jda.api.exceptions.DetachedEntityException
+     *         If this entity is {@link #isDetached() detached}.
+     *
+     * @return An <b>unmodifiable</b> set of system channel flags of this guild.
+     *
+     * @see <a href="https://discord.com/developers/docs/resources/guild#guild-object-system-channel-flags">
```
freya022: Alignment

src/main/java/net/dv8tion/jda/api/entities/Guild.java
```diff
@@ -1335,6 +1336,39 @@ default String getOwnerId()
...
+     * to the guild. This method may be preferable over {@link Guild#getSystemChannelFlags()}
+     * if new flags are introduced to the Discord API and the {@link SystemChannelFlag} enumeration
+     * cannot yet accommodate the new flags.
+     *
+     * @throws net.dv8tion.jda.api.exceptions.DetachedEntityException
+     *         If this entity is {@link #isDetached() detached}.
+     *
+     * @return An integer bitmask of system channel flags of this guild.
+     *
+     * @see <a href="https://discord.com/developers/docs/resources/guild#guild-object-system-channel-flags">
```
freya022: Alignment

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+        return enumSet;
+    }
+
+    /**
+     * Converts a {@link Collection} of this class to its respective bitmask representing the same
+     * set of system channel flags.
+     *
+     * @param flags
+     *        A set of system channel flags.
```
freya022: Align with `@return`

src/main/java/net/dv8tion/jda/api/entities/guild/SystemChannelFlag.java
```diff
@@ -0,0 +1,111 @@
...
+    {
+        this.offset = offset;
+        this.rawValue = 0b1 << offset;
+    }
+
+    /**
+     * Converts a bitmask representation of system channel flags into its {@link EnumSet}.
+     *
+     * @param raw
+     *        The raw bitmask representing the system channel flags.
```
freya022: Align with `@return`

src/main/java/net/dv8tion/jda/api/events/guild/update/GuildUpdateSystemChannelFlagsEvent.java
```diff
@@ -0,0 +1,72 @@
...
+import net.dv8tion.jda.api.entities.guild.SystemChannelFlag;
+
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * Indicates that {@link Guild#getSystemChannelFlags()} have been updated.
+ *
+ * <p>This event can be used to detect changes in system channel flags and retrieve the old one.
```
freya022:
```suggestion
 * <p>This event can be used to detect changes in system channel flags and retrieve the old ones.
```

src/main/java/net/dv8tion/jda/api/events/guild/update/GuildUpdateSystemChannelFlagsEvent.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public GuildUpdateSystemChannelFlagsEvent(@Nonnull JDA api,
+                                              long responseNumber,
+                                              @Nonnull Guild guild,
+                                              @Nonnull Set<SystemChannelFlag> oldFlags,
+                                              @Nonnull Set<SystemChannelFlag> newFlags)
+    {
+        super(api, responseNumber, guild,
+                Collections.unmodifiableSet(oldFlags),
+                Collections.unmodifiableSet(newFlags),
```
freya022: The caller already gives unmodifiable sets

freya022 (changes requested):

src/main/java/net/dv8tion/jda/internal/managers/GuildManagerImpl.java
LaFriska: This change only works if `systemChannelFlags` in `GuildManagerImpl` is a `Set`, and not a collection. Since currently the collection is only set as a `Set`, I will just convert it to a `Set` as opposed to casting the collection to a `Set` in my next commit.

However if this design choice is inconsistent or not favoured by JDA, I can change it back to a `Collection` and cast it to a set instead.

src/test/java/net/dv8tion/jda/test/entities/guild/SystemChannelFlagTest.java
LaFriska: Sounds good, I can delete it later if it is redundant.

src/test/java/net/dv8tion/jda/test/entities/guild/SystemChannelFlagTest.java
```diff
@@ -0,0 +1,97 @@
...
+                SUPPRESS_JOIN_NOTIFICATION_REPLIES,
+                SUPPRESS_ROLE_SUBSCRIPTION_PURCHASE_NOTIFICATIONS,
+                SUPPRESS_ROLE_SUBSCRIPTION_PURCHASE_NOTIFICATION_REPLIES
+        );
+    }
+
+    private void assertTo(int expectedBitmask, SystemChannelFlag... enumFlags)
+    {
+        EnumSet<SystemChannelFlag> flagSet = getEnumSet(enumFlags);
+        Assertions.assertEquals(expectedBitmask, SystemChannelFlag.getRaw(flagSet));
```
MinnDevelopment: We use assertj in all our tests, so I would prefer doing that here as well. In assertj this would look like this: `assertThat(SystemChannelFlag.getRaw(flagSet)).isEqualTo(expectedBitmask)`

src/test/java/net/dv8tion/jda/test/entities/guild/SystemChannelFlagTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.EnumSet;
+
+/**
+ * Tests the class {@link SystemChannelFlag}, especially the conversion logic
+ * between its {@link EnumSet} and bitmask representations.
+ */
+public final class SystemChannelFlagTest
+{
+    @Test
+    public void testFromBitmask()
```
MinnDevelopment: I think these tests aren't really useful, we can probably live without them since they will likely never be updated if new flags are added.

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +486,144 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+     *         The remaining system channel flags to enable
+     *
+     * @throws IllegalArgumentException
+     *         If any input flags are null, or if array itself is null
+     *
+     * @return GuildManager for chaining convenience
+     */
+    @Nonnull
+    @CheckReturnValue
+    default GuildManager enableSystemChannelFlags(@Nonnull SystemChannelFlag flag, @Nonnull SystemChannelFlag... flags)
```
MinnDevelopment: You can also set the flags to an empty list, thus we don't need the first parameter:

```suggestion
    default GuildManager enableSystemChannelFlags(@Nonnull SystemChannelFlag... flags)
```

The first parameter is only used if the list can't be empty. For instance `addFeatures` also only does `addFeatures(@Nonnull String... features)`.

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +486,144 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+     *         The remaining system channel flags to disable
+     *
+     * @throws IllegalArgumentException
+     *         If any input flags are null, or if the array itself is null
+     *
+     * @return GuildManager for chaining convenience
+     */
+    @Nonnull
+    @CheckReturnValue
+    default GuildManager disableSystemChannelFlags(@Nonnull SystemChannelFlag flag, @Nonnull SystemChannelFlag... flags)
```
MinnDevelopment: Same here:

```suggestion
    default GuildManager disableSystemChannelFlags(@Nonnull SystemChannelFlag... flags)
```

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +486,140 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+     *
+     * @return GuildManager for chaining convenience
+     */
+    @Nonnull
+    @CheckReturnValue
+    default GuildManager enableSystemChannelFlags(@Nonnull SystemChannelFlag... flags)
+    {
+        Checks.noneNull(flags, "System channel flags");
+        if (flags.length == 0) return this;
+        return enableSystemChannelFlags(EnumSet.copyOf(Arrays.asList(flags)));
```
MinnDevelopment:
```suggestion
        return enableSystemChannelFlags(Arrays.asList(flags));
```

src/main/java/net/dv8tion/jda/api/managers/GuildManager.java
```diff
@@ -480,6 +486,140 @@ default GuildManager removeFeatures(@Nonnull String... features)
...
+     *
+     * @return GuildManager for chaining convenience
+     */
+    @Nonnull
+    @CheckReturnValue
+    default GuildManager disableSystemChannelFlags(@Nonnull SystemChannelFlag... flags)
+    {
+        Checks.noneNull(flags, "System channel flags");
+        if(flags.length == 0) return this;
+        return disableSystemChannelFlags(EnumSet.copyOf(Arrays.asList(flags)));
```
MinnDevelopment:
```suggestion
        return disableSystemChannelFlags(Arrays.asList(flags));
```

MinnDevelopment (approved):

src/main/java/net/dv8tion/jda/internal/handle/GuildUpdateHandler.java
```diff
@@ -121,6 +127,15 @@ protected Long handleInternally(DataObject content)
                     guild, oldOwner,
                     oldOwnerId, ownerId));
         }
+        if (!Objects.equals(systemChannelFlags, guild.getSystemChannelFlags()))
```
MinnDevelopment: I think we should compare the raw bitmask here instead

LaFriska: I've also moved `Set<SystemChannelFlag> systemChannelFlags` construction into the if block.

PR #2942, opened 2025-10-25
Update JDK javadoc links

freya022:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR replaces the Javadoc links to the JDK to link against JDK 17
- The current JDK 8 link fails when fetching the package list: https://docs.oracle.com/javase/8/docs/api/element-list
- The nearest working package list is from JDK 10, but I believe it makes more sense to use the same version as what the CI and Gradle uses to build, which is also a much more modern version

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -275,7 +275,10 @@ val javadoc by tasks.getting(Javadoc::class) {
 
         author()
         tags("incubating:a:Incubating:")
-        links("https://docs.oracle.com/javase/8/docs/api/", "https://takahikokawasaki.github.io/nv-websocket-client/")
+        // We compile to Java 8 but JDK 8 docs don't seem to supply `element-list` anymore, failing the build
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -275,7 +275,10 @@ val javadoc by tasks.getting(Javadoc::class) {
 
         author()
         tags("incubating:a:Incubating:")
-        links("https://docs.oracle.com/javase/8/docs/api/", "https://takahikokawasaki.github.io/nv-websocket-client/")
+        // We compile to Java 8 but JDK 8 docs don't seem to supply `element-list` anymore, failing the build
```
MinnDevelopment, line 278: How is it failing? It's only a warning in the CI logs

freya022: On JDK 25 it says it is an "error", however it does not fail the build and links seem to be fine either way

PR #2943, opened 2025-10-25
Clean up build constant conditions

freya022:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [X] Other: Build script

Closes Issue: NaN

## Description

Removes conditions that became constants with the upgrade to Gradle 9, as it requires JDK 17+ to run.

The only way I know to still build using a Java <17 compiler is to use a toolchain, and JDA does not use toolchains, so I assume Gradle will only build using the JDK it runs on.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -17,11 +17,7 @@
 import com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import de.undercouch.gradle.tasks.download.Download
-import net.dv8tion.jda.tasks.VerifyBytecodeVersion
-import net.dv8tion.jda.tasks.Version
-import net.dv8tion.jda.tasks.applyAudioExclusions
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -278,24 +271,9 @@ val javadoc by tasks.getting(Javadoc::class) {
...
-
-        // Java 13 changed accessibility rules.
-        // On versions less than Java 13, we simply ignore the errors.
-        // Both of these remove "no comment" warnings.
-        if (javaVersion >= JavaVersion.VERSION_13) {
-            addBooleanOption("Xdoclint:all,-missing", true)
-        } else {
-            addBooleanOption("Xdoclint:all,-missing,-accessibility", true)
-        }
+        addBooleanOption("html5", true) // Adds search bar
```
MinnDevelopment: I think this can be removed entirely, according to `--help`:

> -html5        Generate HTML 5 output. This option is no longer required.

PR #2944, opened 2025-10-26
Remove internal class call and replace with local version

mikomikotaishi:

## Pull Request Etiquette

### Changes

- [x] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

The `net.dv8tion.jda.internal.requests.Requester` class calls a method `okhttp3.internal.http.HttpMethod::requiresRequestBody()`, which is maked `@JvmStatic` but is evidently an internal API and not exported by module `okhttp3`, causing code that uses JDA with modules to break. ~~This PR replaces that call with an internal class `HttpMethod` that supplies those same methods. It resides in the same file `Requester.java` because no other class uses `HttpMethod`, but if necessary it can be placed elsewhere.~~ This PR adds the method `net.dv8tion.jda.api.requests.Method::requiresRequestBody()` and updates the usage in `Requester.java` accordingly.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/requests/Method.java
+++ b/src/main/java/net/dv8tion/jda/api/requests/Method.java
@@ -16,9 +16,6 @@
 
 package net.dv8tion.jda.api.requests;
 
-import java.util.EnumSet;
-import java.util.Set;
-
 /**
  * Enum used to specify the HTTP method to use for a request.
  */


src/main/java/net/dv8tion/jda/internal/requests/Requester.java
```diff
@@ -47,6 +46,48 @@
 import java.util.concurrent.RejectedExecutionException;
 import java.util.function.Consumer;
 
+class HttpMethod
```
MinnDevelopment: I think we only need `requiresRequestBody`. Instead of adding this class here, please extend our existing `Method` enum:

```java
apiRequest.getRoute().getMethod().requiresRequestBody()
```

mikomikotaishi: OK, I'll just do that instead

mikomikotaishi: Made the changes as you requested, and only included the one method we need.

src/main/java/net/dv8tion/jda/api/requests/Method.java
```diff
@@ -16,6 +16,9 @@
 
 package net.dv8tion.jda.api.requests;
 
+import java.util.EnumSet;
+import java.util.Set;
```
MinnDevelopment: Are these imports used?

mikomikotaishi: Sorry, I forgot to remove those when I used an EnumSet implementation before concluding that it was pointless abstraction

mikomikotaishi:
Any other comments before we can merge this?

PR #2957, opened 2025-11-02
Update JDK to JDK 25

MinnDevelopment:

## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: Dependencies

Closes Issue: NaN

## Description

We can use JDK 25 to compile JDA. Note that we still support java 8.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -202,14 +202,6 @@ tasks.withType<DependencyUpdatesTask> {
 
 versionCatalogUpdate {
     versionSelector(VersionSelectors.STABLE)
-
-    pin {
-        libraries = listOf(
-            libs.tink,
-            libs.junit.java8,
-            libs.junit.launcher.java8,
-        )
-    }
 }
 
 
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -265,7 +293,7 @@ val javadoc by tasks.getting(Javadoc::class) {
 
         author()
         tags("incubating:a:Incubating:")
-        links("https://docs.oracle.com/javase/8/docs/api/", "https://takahikokawasaki.github.io/nv-websocket-client/")
+        links("https://docs.oracle.com/en/java/javase/$currentJavaVersion/docs/api/", "https://takahikokawasaki.github.io/nv-websocket-client/")
```
freya022, line 296: Might also want to add the `--no-fonts` option, this is new stuff that takes about 4 MB

MinnDevelopment: Given that this is just docs, I don't really see a problem with a bigger jar size for the fonts. Since we don't have our own custom fonts and want to just use the default.

PR #2966, opened 2025-12-01
Voice Channel Invite Guest Flag

austinpilz:
## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description
This PR adds the `isGuest` flag to the Invite. [Discord API Docs PR](https://github.com/discord/discord-api-docs/pull/6247/files#diff-e9cef9e469777188b4d253eff8a730d3fb64f7b915599fd45f4365dda8d98f74R45).

Applicable only to voice channel invites, the [Discord Documentation](https://discord.com/developers/docs/resources/invite#invite-object-guild-invite-flags) mentions the inclusions of a `flag` field. Discord currently only supports a single flag, the value of `1` signifying the invite makes the user a guest, and once they leave the voice channel, they are removed from the server.

The guest flag has similar overlap with the existing `temporary` attribute on Invites, however, Discord's API distinctly separates them. When a voice channel has the flag as `1`, the actual `temporary` attribute will be false.

invite

I verified this works as expected by using the local build with my bot:

invite2

This is my first JDA contribution. I chose to extract the flag into a boolean as invites (in JDA and Discord) do not support more than one flag. If you'd prefer I break the flag out into an enum, I am happy to do that, too.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Invite.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Invite.java
@@ -329,9 +329,9 @@ public interface Invite {
      * from the invite without being granted guild membership nor having visibility to other guild channels.
      * Once the user leaves the voice call, their ability to rejoin the voice call and see the guild is revoked.
      *
-     * <p>Applicable only to invites to a specific channel {@link #getChannel()}.
+     * <p>Applicable only to invites to a specific {@linkplain #getChannel() channel}.
      *
-     * @return Whether this invite is a guest invite for a voice channel or not.
+     * @return Whether this invite is a guest invite for a voice channel or not
      *
      * @see #getChannel()
      */


src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -324,6 +324,17 @@ default String getUrl() {
      */
     boolean isTemporary();
 
+    /**
+     * Whether this Invite is a guest invite for a voice channel or not.
+     *
+     * <p>Applicable only to invites to a specific channel {@link #getChannel()}.</p>
```
MinnDevelopment: Minor: you should not close paragraphs in javadocs

```suggestion
     * <p>Applicable only to invites to a specific channel {@link #getChannel()}.
```

src/main/java/net/dv8tion/jda/api/entities/Invite.java
```diff
@@ -324,6 +324,17 @@ default String getUrl() {
      */
     boolean isTemporary();
 
+    /**
+     * Whether this Invite is a guest invite for a voice channel or not.
```
MinnDevelopment: Please add a sentence explaining what "guest" means, like "Users who join through a guest invite will be removed from the guild when they disconnect from the voice channel."

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -2401,13 +2402,15 @@ public Invite createInvite(DataObject object) {
             maxUses = object.getInt("max_uses");
             uses = object.getInt("uses");
             temporary = object.getBoolean("temporary");
+            guest = object.getInt("flags", 0) == 1;
```
MinnDevelopment: This check will break once they add other flags, instead use a bitmask:

```suggestion
            guest = (object.getInt("flags", 0) & 1) == 1;
```

src/main/java/net/dv8tion/jda/internal/handle/InviteCreateHandler.java
```diff
@@ -58,6 +58,7 @@ protected Long handleInternally(DataObject content) {
 
         String code = content.getString("code");
         boolean temporary = content.getBoolean("temporary");
+        boolean guest = content.getInt("flags", 0) == 1;
```
MinnDevelopment:
```suggestion
        boolean guest = (content.getInt("flags", 0) & 1) == 1;
```

src/main/java/net/dv8tion/jda/api/entities/Invite.java
austinpilz: Thanks for the review @MinnDevelopment - I've updated the javadoc to include the requested description of what a guest invite entails.

austinpilz:
@MinnDevelopment can you take another look at this, please?

MinnDevelopment:
@austinpilz it is still on my list for the next feature release.

src/main/java/net/dv8tion/jda/internal/entities/EntityBuilder.java
```diff
@@ -2401,13 +2402,15 @@ public Invite createInvite(DataObject object) {
             maxUses = object.getInt("max_uses");
             uses = object.getInt("uses");
             temporary = object.getBoolean("temporary");
+            guest = (object.getInt("flags", 0) & 1) == 1;
```
MinnDevelopment: I think this can be moved up to the `target_type` and others. With your current code this field is only parsed on expanded invites due to the check for `hasKey("max_uses")`. This check is to differentiate between invites from the channel settings (which mods can see statistics and other insights on). In my testing, this is also set on the standard get invite endpoint used by `Invite.resolve(...)`

PR #2974, opened 2025-12-12
Fix default CV2 flag not being applied on message edits

freya022:
## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

An user with `MessageRequest.isDefaultUseComponentsV2()` returning `true` was unable to create CV2 edits by default, see https://discord.com/channels/125227483518861312/1406042603350130708/1448498020604973200.

The flag was set appropriately by `AbstractMessageBuilder` (a superclass of `MessageEditBuilder`), but then was "marked as not set" because of `MessageEditBuilder#configuredFields`'s initializer which set it back to 0.

So, this PR removes the initializer, while still keeping the `0` default, to prevent it from being overwritten.


MinnDevelopment:
An update request should by default not set flags at all, since we just keep the ones that are present on the message. The flags not being set by `deferReply()` is intended behavior by discord. I already complained about it before, but we just implement it how its intended.

PR #2982, opened 2025-12-21
Add support for bulk/streamed Zstandard transport compression

freya022:
## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

This PR implements Zstandard decompression for gateway messages using [`discord-zstd-java`](https://github.com/freya022/discord-zstd-java). As this is opt-in, JDA only includes the API for it, so, users which want to use it will have to include the implementation and configure JDA.

This is beneficial for larger bots where the CPU and GC allocations improvements will be the most noticeable, and for Discord itself. Smaller bots will only see slight benefits.

In both cases, payloads are slightly smaller mostly during startup, and quite similar during normal operation.

Maintainers note: The dependency still is 0-versioned, it will be released if no API changes are to be made before merging.

## Changes

Both `JDABuilder` and `DefaultShardManagerBuilder` had their `setMaxBufferSize`, `setCompression` and `setGatewayEncoding` deprecated. They are now configured with `setGatewayConfig` and `setGatewayConfigProvider`.

## New features

- `Compression.ZSTD` was added, this is opt-in
- Decompression using `InputStream` was added and is configurable in `GatewayConfig`
  - This is opt-in and allows even higher efficiency at the cost of no JSON being logged in case of decompression/decoding errors
  - Currently, this mode does not allow using the ETF encoding.

## Synthetic benchmarks

> [!NOTE]
> Thanks to @MrPowerGamerBR for helping me gather some data from their bot and running benchmarks on their machine!

With 10 shards starting up from scratch, giving us 364 MB of decompressed data, they ran [a few benchmarks](https://github.com/freya022/discord-zstd-java/tree/d647a0d67d00798913ee123fefc1c36d31d2e2f9/benchmarks/src/jmh/java/dev/freya02/discord/zstd) comparing zlib, zstd, as well as their bulk (what JDA currently does) and stream (letting Jackson consume an `InputStream`) variants.

They were run on their production server, with an AMD Ryzen 5 5600X and produced [this data](https://gist.github.com/freya022/0516a809d43ee1d084ed205ac4fbe56c).

If we look at the complete package, transforming the compressed data into a usable `DataObject`, we can see bulk-decompressing with Zstd cut time spent by 30% and reduced GC allocations by 35%

However, if we use streaming, Zlib gets about 10% of speed increase and removes 48% of GC allocations, while Zstd gets a 31% speed improvement and the same memory improvements.

### Other benchmarks

I have also run some numbers on my machine (an AMD Ryzen 7 3700X, with boost disabled), not much different here:

- 10 shards of [randomly generated](https://github.com/freya022/discord-zstd-java/tree/d647a0d67d00798913ee123fefc1c36d31d2e2f9/test-data-generator) data: https://gist.github.com/freya022/04d5e8cf7d44c9680ae42154808cddfd
- A bot with a single guild: https://gist.github.com/freya022/8922140965bc51a699b135ebc2f96914

## Runtime statistics

They were also kind enough to run their ~2.5K shards for a day while recording the compressed size, decompressed size and time-to-decompress.

The shards were split in 3 equally-sized sets:

- Using the current Zlib implementation
- Using Zstd with a 8 KB buffer
- Using Zstd with a 128 KB buffer

The results can be seen there: https://gist.github.com/freya022/7b35aa412a4f125ca1b139b71360ab45

## Links

- How Discord Reduced Websocket Traffic by 40%: https://discord.com/blog/how-discord-reduced-websocket-traffic-by-40-percent
- Docs: https://discord.com/developers/docs/events/gateway#zstdstream

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/requests/GatewayConfig.java
+++ b/src/main/java/net/dv8tion/jda/api/requests/GatewayConfig.java
@@ -23,14 +23,14 @@ import net.dv8tion.jda.api.GatewayEncoding;
 import net.dv8tion.jda.api.utils.Compression;
 import net.dv8tion.jda.internal.requests.GatewayConfigImpl;
 import net.dv8tion.jda.internal.requests.gateway.decoder.Decoder;
-import net.dv8tion.jda.internal.requests.gateway.messages.GatewayBulkMessageReader;
+import net.dv8tion.jda.internal.requests.gateway.messages.GatewayBufferMessageReader;
 import net.dv8tion.jda.internal.requests.gateway.messages.GatewayMessageReader;
 import net.dv8tion.jda.internal.requests.gateway.messages.GatewayStreamMessageReader;
 import net.dv8tion.jda.internal.utils.Checks;
 import net.dv8tion.jda.internal.utils.compress.disabled.DisabledDecompressor;
-import net.dv8tion.jda.internal.utils.compress.zlib.ZlibBulkDecompressor;
+import net.dv8tion.jda.internal.utils.compress.zlib.ZlibBufferDecompressor;
 import net.dv8tion.jda.internal.utils.compress.zlib.ZlibStreamDecompressor;
-import net.dv8tion.jda.internal.utils.compress.zstd.ZstdBulkDecompressorAdapter;
+import net.dv8tion.jda.internal.utils.compress.zstd.ZstdBufferDecompressorAdapter;
 import net.dv8tion.jda.internal.utils.compress.zstd.ZstdStreamDecompressorAdapter;
 
 import java.util.function.Function;
@@ -59,7 +59,7 @@ public interface GatewayConfig {
      */
     class Builder {
         private Function<Decoder, GatewayMessageReader> messageReaderFunction =
-                (decoder) -> new GatewayBulkMessageReader(decoder, new ZlibBulkDecompressor(2048));
+                (decoder) -> new GatewayBufferMessageReader(decoder, new ZlibBufferDecompressor(2048));
         private boolean isStreamDecompression = false;
         private GatewayEncoding encoding = GatewayEncoding.JSON;
 
@@ -71,7 +71,8 @@ public interface GatewayConfig {
          */
         @Nonnull
         public Builder disableCompression() {
-            this.messageReaderFunction = (decoder) -> new GatewayBulkMessageReader(decoder, DisabledDecompressor.INSTANCE);
+            this.messageReaderFunction =
+                    (decoder) -> new GatewayBufferMessageReader(decoder, DisabledDecompressor.INSTANCE);
             this.isStreamDecompression = false;
 
             return this;
@@ -94,17 +95,16 @@ public interface GatewayConfig {
          * @param  maxBufferSize
          *         The size of the intermediary buffer that has to be reached before it has to be shrunk down
          *
-         *
          * @throws IllegalArgumentException
          *         If the provided buffer size is negative
          *
          * @return This builder for chaining convenience
          */
         @Nonnull
-        public Builder useBulkZlibDecompression(int maxBufferSize) {
+        public Builder useBufferZlibDecompression(int maxBufferSize) {
             Checks.notNegative(maxBufferSize, "Max buffer size");
             this.messageReaderFunction =
-                    (decoder) -> new GatewayBulkMessageReader(decoder, new ZlibBulkDecompressor(maxBufferSize));
+                    (decoder) -> new GatewayBufferMessageReader(decoder, new ZlibBufferDecompressor(maxBufferSize));
             this.isStreamDecompression = false;
 
             return this;
@@ -138,11 +138,11 @@ public interface GatewayConfig {
          * @return This builder for chaining convenience
          */
         @Nonnull
-        public Builder useBulkZstdDecompression(int bufferSizeHint) {
+        public Builder useBufferZstdDecompression(int bufferSizeHint) {
             DiscordZstdDecompressorFactory underlyingFactory =
                     DiscordZstdProvider.get().createDecompressorFactory(bufferSizeHint);
-            this.messageReaderFunction = (decoder) ->
-                    new GatewayBulkMessageReader(decoder, new ZstdBulkDecompressorAdapter(underlyingFactory.create()));
+            this.messageReaderFunction = (decoder) -> new GatewayBufferMessageReader(
+                    decoder, new ZstdBufferDecompressorAdapter(underlyingFactory.create()));
             this.isStreamDecompression = false;
             return this;
         }
@@ -165,7 +165,7 @@ public interface GatewayConfig {
          * @return This builder for chaining convenience
          */
         @Nonnull
-        public Builder useStreamingDecompression(@Nonnull Compression compression) {
+        public Builder useStreamDecompression(@Nonnull Compression compression) {
             Checks.notNull(compression, "Compression");
             switch (compression) {
                 case NONE:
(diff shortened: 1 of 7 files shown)


build.gradle.kts
```diff
@@ -133,6 +133,9 @@ dependencies {
     api(libs.websocket.client)
     api(libs.okhttp)
 
+    // Zstd decompression API
+    api(libs.discord.zstd.java.api)
```
MinnDevelopment, line 137: I don't want a hard dependency on this library, rather just an interface we provide and users can add this implementation if they want to. We do the same for lavalink's hooks and jda-nas.

src/main/java/net/dv8tion/jda/api/utils/Compression.java
```diff
@@ -21,14 +21,24 @@
...
-    ZLIB("zlib-stream");
+    ZLIB("zlib-stream"),
+    /**
+     * Use Zstandard transport compression,
+     * this requires you to include a dependency,
+     * see <a href="https://github.com/freya022/discord-zstd-java" target="_blank">discord-zstd-java</a>.
+     *
+     * <p>This uses way less resources than Zlib, however this typically is noticeable for large bots.
+     */
+    ZSTD("zstd-stream"),
```
MinnDevelopment, line 40: I think we can just deprecate this enum entirely and exclusively rely on the implementation. Our decompressor hook interface can have a `getQueryParameter()` or similar that JDA just uses.

src/main/java/net/dv8tion/jda/internal/requests/gateway/decoder/JsonDecoder.java
```diff
@@ -0,0 +1,55 @@
...
+            // Print the string that could not be parsed and re-throw the exception
+            LOG.error("Failed to parse json: {}", jsonString);
+            throw e;
+        }
+    }
+
+    @Nonnull
+    @Override
+    public DataObject decode(@Nonnull InputStream stream) {
+        // No exception handling as we can't get read the stream twice to get the parsed string
```
MinnDevelopment, line 52: Yes you can, with `BufferedInputStream`

src/main/java/net/dv8tion/jda/internal/requests/gateway/messages/GatewayMessageReader.java
```diff
@@ -0,0 +1,42 @@
...
+import net.dv8tion.jda.internal.utils.compress.DecompressionException;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Bulk + JSON = OK
+ * Bulk + ETF = OK
+ * Streaming + JSON = OK
+ * Streaming + ETF = NO (adapt ExTermDecoder to use input streams)
```
MinnDevelopment: Are these the final docs?

src/main/java/net/dv8tion/jda/api/exceptions/DecompressionException.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.internal.utils.compress;
+
+public class DecompressionException extends Exception {
```
MinnDevelopment, line 22: Move this to the exception package

src/main/java/net/dv8tion/jda/internal/utils/compress/noop/NullDecompressor.java
```diff
@@ -0,0 +1,55 @@
...
+import net.dv8tion.jda.api.utils.Compression;
+import net.dv8tion.jda.internal.utils.compress.BulkDecompressor;
+import net.dv8tion.jda.internal.utils.compress.StreamDecompressor;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+
+import javax.annotation.Nullable;
+
+public class NullDecompressor implements BulkDecompressor, StreamDecompressor {
```
MinnDevelopment:
```suggestion
public class DisabledDecompressor implements BulkDecompressor, StreamDecompressor {
```

src/main/java/net/dv8tion/jda/api/requests/GatewayConfig.java
```diff
@@ -0,0 +1,219 @@
...
+         * @throws IllegalArgumentException
+         *         If the {@linkplain #useEncoding(GatewayEncoding) encoding} is set to {@link GatewayEncoding#ETF ETF}
+         *
+         * @return This builder for chaining convenience
+         */
+        @Nonnull
+        public Builder useStreamingDecompression(@Nonnull Compression compression) {
+            Checks.notNull(compression, "Compression");
+            Checks.check(
+                    encoding != GatewayEncoding.ETF, "Cannot use streaming decompression with ETF payload encoding");
```
MinnDevelopment: These checks should be in `build()`

src/main/java/net/dv8tion/jda/api/requests/GatewayConfig.java
```diff
@@ -0,0 +1,219 @@
...
+        return new Builder();
+    }
+
+    /**
+     * A builder of {@link GatewayConfig}.
+     */
+    class Builder {
+        private Function<Decoder, GatewayMessageReader> messageReaderFunction =
+                (decoder) -> new GatewayBulkMessageReader(decoder, new ZlibBulkDecompressor(2048));
+        private boolean isBulkDecompression = false;
```
MinnDevelopment: This can be determined by the interface instead. Also, why call it "bulk"? The discord terminology is "Payload Compression" and "Transport Compression".

I expect we can have two interfaces:

```java
interface PayloadGatewayDecompressor extends GatewayDecompressor {
  default CompressionStrategy getStrategy() {
    return CompressionStrategy.PAYLOAD;
  }
  
  String getCompressionType();
  byte[] decompress(byte[] data);
}
```

```java
interface TransportGatewayDecompressor extends GatewayDecompressor {
  default CompressionStrategy getStrategy() {
    return CompressionStrategy.TRANSPORT;
  }

  String getCompressionType();
  byte[] decompress(byte[] data);
}
```

src/main/java/net/dv8tion/jda/internal/requests/gateway/messages/GatewayMessageReader.java
freya022: Forgot to remove, I had tried adapting `ExTermDecoder` to use input streams but it was way slower even with buffering

src/main/java/net/dv8tion/jda/internal/requests/gateway/decoder/JsonDecoder.java
freya022: Sure with `mark` + `reset` but that would defeat the performance improvements from using an `InputStream`, which is the point of it

src/main/java/net/dv8tion/jda/api/requests/GatewayConfig.java
freya022: It's called bulk because of how JDA goes from a compressed `byte[]` to a decompressed `byte[]`, while the "streaming" decompressor is `byte[]` -> `InputStream`.

Additionally JDA does not support *payload* compression and I have no plan to support it at all.

MinnDevelopment: I see, that naming convention is confusing. I'd go with "Buffer" and "Stream" which is more intuitive

Issue #2984, opened 2025-12-23
Possibility to send invalid modal (with disabled components) via JDA to discord api


fabirednik:
### General Troubleshooting

- [x] I have checked for similar issues on the Issue-tracker.
- [x] I have checked for PRs that might already address this issue.

### Version of JDA

6.2.0

### Expected Behaviour

JDA should remove setDisabled() function for StringSelectMenu (or just throw invalid argument error before sending this invalid modal to discord api) when used as component in Modal like in TextInput

### Code Example for Reproduction Steps

```java
event.replyModal(
        Modal.create("id", "title")
                .addComponents(
                        Label.of("label", StringSelectMenu.create("id")
                                .addOption("label", "id")
                                .setDisabled(true)
                                .build()
                        )
                )
                .build()
        )
        .queue();
```

### Code for JDABuilder or DefaultShardManagerBuilder used

```java
-
```

### Exception or Error

```yesyes
2025-12-23 21:55:00.610 [ForkJoinPool.commonPool-worker-1] ERROR t.r.f.d.m.m.m.l.s.EditFieldStringSelected - 50035: Invalid Form Body
data.components[0].component.disabled
	- MODAL_COMPONENT_DISABLED: Modals do not support disabled components
net.dv8tion.jda.api.exceptions.ErrorResponseException: 50035: Invalid Form Body
data.components[0].component.disabled
	- MODAL_COMPONENT_DISABLED: Modals do not support disabled components
	at net.dv8tion.jda.api.exceptions.ErrorResponseException.create(ErrorResponseException.java:195)
	at net.dv8tion.jda.api.requests.Request.createErrorResponseException(Request.java:156)
	at net.dv8tion.jda.internal.requests.RestActionImpl.handleResponse(RestActionImpl.java:297)
	at net.dv8tion.jda.internal.requests.restaction.interactions.InteractionCallbackImpl.handleResponse(InteractionCallbackImpl.java:145)
	at net.dv8tion.jda.api.requests.Request.handleResponse(Request.java:290)
	at net.dv8tion.jda.internal.requests.Requester$WorkTask.handleResponse(Requester.java:413)
	at net.dv8tion.jda.internal.requests.Requester$WorkTask.access$300(Requester.java:360)
	at net.dv8tion.jda.internal.requests.Requester.execute(Requester.java:241)
	at net.dv8tion.jda.internal.requests.Requester.execute(Requester.java:156)
[...]

PR #2987, opened 2025-12-24
Added Component Verification in Modal Builder

Crain-32:
## Pull Request Etiquette

### Changes

- [x] Internal code

Closes Issue: 2984

## Description

Adds a check for "isDisabled" on Modal Creation. Disabled Components within Modals will error out, as noted in #2984.
Primarily internal, but it's worth noting that the Modal's Builder Interface has "changed", since it's now more strict than in previous releases.


src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
```diff
@@ -0,0 +1,96 @@
...
+            Component.Type.STRING_SELECT, StringSelectMenu.class,
+            Component.Type.USER_SELECT, EntitySelectMenu.class,
+            Component.Type.ROLE_SELECT, EntitySelectMenu.class,
+            Component.Type.MENTIONABLE_SELECT, EntitySelectMenu.class,
+            Component.Type.CHANNEL_SELECT, EntitySelectMenu.class);
+
+    @MethodSource("modalComponents")
+    @ParameterizedTest
+    public void testDisabledComponents(Component component) {
+        Assertions.assertInstanceOf(IDisableable.class, component);
```
MinnDevelopment: We use assertj in our tests:

```java
assertThat(component).isInstanceOf(IDisableable.class).isInstanceOf(LabelChildComponent.class)
```

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Assertions.assertInstanceOf(LabelChildComponent.class, component);
+        LabelChildComponent mutatedComponent = ((LabelChildComponent) ((IDisableable) component).asDisabled());
+        Assertions.assertThrows(IllegalArgumentException.class, () -> {
```
MinnDevelopment:
```java
assertThatIllegalArgumentException().isThrownBy(() -> { ... })
```

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Assertions.assertDoesNotThrow(() -> {
+            Modal.create("id", "title")
+                    .addComponents(Label.of("label", (LabelChildComponent) component))
+                    .build();
+        });
+    }
+
+    @Test
+    public void testOneInvalidComponentInList() {
+        var components = new ArrayList<>(testComponents().toList());
```
MinnDevelopment:
```suggestion
        List<LabelChildComponent> components = new ArrayList<>(testComponents().toList());
```

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
```diff
@@ -0,0 +1,96 @@
...
+        var mutatedComponentIndex = ThreadLocalRandom.current().nextInt(components.size());
```
MinnDevelopment:
```suggestion
        int mutatedComponentIndex = ThreadLocalRandom.current().nextInt(components.size());
```

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Assertions.assertInstanceOf(IDisableable.class, components.get(mutatedComponentIndex));
+        components.set(mutatedComponentIndex, ((IDisableable) components.get(mutatedComponentIndex)).asDisabled());
+
+        Assertions.assertThrows(IllegalArgumentException.class, () -> {
```
MinnDevelopment: If you extend `AbstractSnapshotTest`, you can snapshot the exception output:

```java
assertThatIllegalArgumentException()
    .isThrownBy(() -> {
        ...
    })
    .satisfies(exception -> assertWithSnapshot(exception.toString(), null));
```

Feel free to add `assertWithSnapshot(String)` overload to avoid the `null` parameter. To update snapshots you can either run the test with intellij's junit test runner or using the gradle task `./gradlew updateTestSnapshots`

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
```diff
@@ -0,0 +1,96 @@
...
+                    .addComponents(Label.of("label", (LabelChildComponent) component))
+                    .build();
+        });
+    }
+
+    @Test
+    public void testOneInvalidComponentInList() {
+        var components = new ArrayList<>(testComponents().toList());
+        var mutatedComponentIndex = ThreadLocalRandom.current().nextInt(components.size());
+        Assertions.assertInstanceOf(IDisableable.class, components.get(mutatedComponentIndex));
```
MinnDevelopment: To avoid potentially creating a flaky test later, this should just assert that the entire list is the expected type:

```java
assertThat(components).allSatisfy(component -> assertThat(component).isInstanceOf(IDisableable.class));
```

This will give you a very helpful error message if one of the components in the test data is not the right type:

```java
java.lang.AssertionError: 
Expecting all elements of:
  [SelectMenu[ROLE_SELECT](id=-1, custom id=role-menu, placeholder=null),
    SelectMenu[MENTIONABLE_SELECT](id=-1, custom id=mentionable-menu, placeholder=null),
    SelectMenu[USER_SELECT](id=-1, custom id=user-menu, placeholder=null),
    SelectMenu[CHANNEL_SELECT](id=-1, custom id=channel-menu, placeholder=null),
    SelectMenu[STRING_SELECT](id=-1, custom id=select-menu, placeholder=null),
    AttachmentUpload(custom_id=attachment-upload, required=true)]
to satisfy given requirements, but these elements did not:

AttachmentUpload(custom_id=attachment-upload, required=true)
error: java.lang.AssertionError: 
Expecting actual:
  AttachmentUpload(custom_id=attachment-upload, required=true)
to be an instance of:
  net.dv8tion.jda.api.components.attribute.IDisableable
but was instance of:
  net.dv8tion.jda.internal.components.attachmentupload.AttachmentUploadImpl
```

src/main/java/net/dv8tion/jda/api/modals/Modal.java
```diff
@@ -327,13 +330,29 @@ public String getId() {
...
+        /**
+         * Verifies Disabled Component state for a Modal,
+         * based on the <a href="https://discord.com/developers/docs/components/reference">Component Reference</a>
+         *
+         * @param component Expected to be a {@link ModalTopLevelComponentUnion} Component
+         *
+         * @return false iff the Component is a {@link Label}, and the Child Component of the Label is Disabled.
+         */
+        private static boolean componentIsEnabled(Component component) {
+            if (component instanceof Label && ((Label) component).getChild() instanceof IDisableable) {
```
MinnDevelopment: To improve the exception message, you should validate the child directly instead of going through the label:

```java
return !(component instanceof IDisableable) || ((IDisableable) component).isEnabled();
```

This will change the path in the exception to resolve the actual component that violates the constraint.

Before:

```
java.lang.IllegalArgumentException: Components cannot be disabled in Modals
- root.components[2]<LABEL>
```

After:

```
java.lang.IllegalArgumentException: Components cannot be disabled in Modals
- root.components[2]<LABEL>.component<CHANNEL_SELECT>
```

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
MinnDevelopment: I would use a consistent index here to make the exception message stable, which allows us to snapshot it.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Assertions.assertThrows(IllegalArgumentException.class, () -> {
+            Modal.create("id", "title")
+                    .addComponents(Label.of("label", mutatedComponent))
+                    .build();
+        });
+    }
+
+    @MethodSource("modalComponents")
+    @ParameterizedTest
+    public void testEnabledComponents(Component component) {
```
MinnDevelopment: You can directly put `LabelChildComponent component` as the test parameter, since JUnit will already check the type for you. This removes the need for the instanceof check.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
```diff
@@ -0,0 +1,96 @@
...
+                            .toList())
+                    .build();
+        });
+    }
+
+    static Stream<Arguments> modalComponents() {
+        return testComponents().map(Arguments::arguments);
+    }
+
+    static Stream<Component> testComponents() {
```
MinnDevelopment:
```suggestion
    static Stream<LabelChildComponent> testComponents() {
```

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
```diff
@@ -0,0 +1,96 @@
...
+    static Map<Component.Type, Class<? extends LabelChildComponent>> LABEL_CHILDREN = Map.of(
+            Component.Type.STRING_SELECT, StringSelectMenu.class,
+            Component.Type.USER_SELECT, EntitySelectMenu.class,
+            Component.Type.ROLE_SELECT, EntitySelectMenu.class,
+            Component.Type.MENTIONABLE_SELECT, EntitySelectMenu.class,
+            Component.Type.CHANNEL_SELECT, EntitySelectMenu.class);
+
+    @MethodSource("modalComponents")
+    @ParameterizedTest
+    public void testDisabledComponents(Component component) {
```
MinnDevelopment:
```suggestion
    public void testDisabledComponents(LabelChildComponent component) {
```

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
Crain-32: The random index is on purpose. I'd rather test a random index, or test all the indexes. This felt like the best approach considering comprehensive and overall speed.

src/main/java/net/dv8tion/jda/api/modals/Modal.java
Crain-32: Does the label delegate "isEnabled" to the child? I just double checked across the LabelImpl, and didn't see it delegating, but I could be missing something in the class hierarchy.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
Crain-32: This lets the test input mirror the signature of `Modal#componentIsEnabled`, and I consider the `LabelChildComponent` check part of the test itself, so I'd rather have it fail in the test body than in the input. 

Ideally I would have tested `componentIsEnabled` directly (And had it be package-private), but since the Test package path is `net.dv8tion.jda.test` I'd have to put the Test in an awkward package from the rest of the test.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
Crain-32: See comment about Enabled Component. I'm marking this one as "resolved", but I'll mirror whatever we end up deciding on from the other conversation.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
Crain-32: I'll have the type mirror the input type of the tests (just for ease), and so the enabled Component test conversation resolution will impact this. Rather have the test method source type mirror the input type itself.

I won't mark this as "resolved" however, in case you want this decoupled from the test method inputs.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
Crain-32: I'll play around with this. `assertThat(components).allMatch(IDisableable.class::isInstance)` looks cleaner on a first pass, but if it doesn't give as good of an error message than I'll swap to this.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
Crain-32: I'll adjust all the assertions to AssertJ in that case.

Going to wait to push an update until we've gotten a path forward on the other aspects. Marking this as "resolved" since it's a straight forward project standard.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
Crain-32: Since the exception message is a static string I won't do it for this test. But that is good to note.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
MinnDevelopment: Then I would prefer doing all indexes and snapshotting the error message.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
MinnDevelopment: It is not a static string, it is determined by the builder input. I think the snapshot is quite useful information to have as a test output.

src/main/java/net/dv8tion/jda/api/modals/Modal.java
MinnDevelopment: What do you mean by that? The `Checks.checkComponents` will recursively iterate all components and also end up at the child components of the label, since *those* are `IDisableable` you get the output as shown in my example

Crain-32: `ComponentPathIterator`, is what I had missed. Didn't realize it was doing a recursive stream against the components.

Will adjust.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
MinnDevelopment: The reason I prefer changing it, is because it makes the test more explicit, since you're actually testing `LabelChildComponent` and not just `Component`. The test additionally wraps all its input in a `Label`, which you can only know if you actually read the test code. Alternatively, you could rename the test to `testEnabledLabelChildComponent(Component component)` but that ends up achieving the same thing as just changing the parameter type, with the downside of still having to explicitly add an assertion.

I think it's good to test the modal's validation through the modal builder, instead of only testing the internal check method.

src/test/java/net/dv8tion/jda/test/modals/ModalTest.java
MinnDevelopment: By using `allMatch` you lose the actual check in the exception, since you reduce it to just a "true/false" instead of an "is instance of"

MinnDevelopment (approved):

PR #3054, opened 2026-03-24
Improve and complement regex patterns

Vampire:
## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [x] Other: Regex patterns

### Description

A few improvements to the regex patterns.
- Added possessive quantifiers where possible to improve fail-fast behavior and reduce needless backtracking
- Named groups everywhere
- Improved groups documentation
- Ensure mentions are not escaped and thus effectively no mentions
- Documented the regex parts using comment mode for easier understanding
- Added language marker for IDEs where the patterns are used as regex constructor parameters
- Support `@me` as channel ID in message jump url pattern for non-guild channels
- Add channel jump url pattern

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/entities/Message.java
+++ b/src/main/java/net/dv8tion/jda/api/entities/Message.java
@@ -218,19 +218,39 @@ public interface Message extends ISnowflake, Formattable {
     /**
      * Pattern used to find instant invites in strings.
      *
-     * <p>The only named group is at index 1 with the name {@code "code"}.
+     * <p><b>Groups</b><br>
+     * <table>
+     *   <caption style="display: none">JavaDoc is stupid, this is not a required tag</caption>
+     *   <tr>
+     *     <th>Index</th>
+     *     <th>Name</th>
+     *     <th>Description</th>
+     *   </tr>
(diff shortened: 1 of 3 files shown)


src/main/java/net/dv8tion/jda/api/entities/Message.java
```diff
@@ -2520,38 +2544,241 @@ default ReactionPaginationAction retrieveReactionUsers(@Nonnull Emoji emoji) {
...
+         *
+         * You can use the names with {@link java.util.regex.Matcher#group(String) Matcher.group(String)}
+         * and the index with {@link java.util.regex.Matcher#group(int) Matcher.group(int)}.
          */
-        USER("<@!?(\\d+)>", "users"),
+        // language=regexp
+        USER(
+                "(?x)                          # enable comment mode \n"
+                        + "(?<!                # negative lookbehind (do not have uneven amount of backslashes before) \n"
+                        + "    (?<!\\\\)       # negative lookbehind (do not have one backslash before) \n"
```
MinnDevelopment, line 2575: This is unwanted and a breaking change, we also want to match escaped ones.

PR #3070, opened 2026-03-29
Add JDABuilder#setHttpClientBuilderPreprocessor

Vampire:
## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

## Description

This PR adds the ability to add a consumer to the jda builder that is used to preprocess the http client builder before `build()` is called on it.
This can be used to do own settings, but on top of the defaults that JDA uses.

If for example #3069 gets rejected, this PR would be very useful to just add the OkHttp logging interceptor, without needing to redefine the JDA built-in configuration for the builder, or alternatively using the internal `IOUtil` to get the JDA defaults and enhance them.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/JDABuilder.java
+++ b/src/main/java/net/dv8tion/jda/api/JDABuilder.java
@@ -45,6 +45,7 @@ import okhttp3.OkHttpClient;
 
 import java.util.*;
 import java.util.concurrent.*;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import javax.annotation.CheckReturnValue;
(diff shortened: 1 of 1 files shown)


src/main/java/net/dv8tion/jda/api/JDABuilder.java
```diff
@@ -796,14 +798,39 @@ public JDABuilder setToken(@Nullable String token) {
...
+     * @return The JDABuilder instance. Useful for chaining.
+     *
+     * @see    #setHttpClient
+     * @see    #setHttpClientBuilder
+     */
+    @Nonnull
+    public JDABuilder setHttpClientBuilderPreprocessor(@Nullable Consumer<OkHttpClient.Builder> builderPreprocessor) {
+        this.httpClientBuilderPreprocessor = builderPreprocessor;
+        return this;
+    }
```
MinnDevelopment, line 820: How about this:

```java
public JDABuilder configureHttpClient(@Nonnull Consumer<? super OkHttpClient.Builder> configuration) {
  Checks.notNull(configuration, "Configuration");

  if (this.httpClientBuilder == null) {
    this.httpClientBuilder = IOUtil.newHttpClientBuilder();
  }

  configuration.apply(this.httpClientBuilder);
  return this;
}
```

PR #3116, opened 2026-07-03
Add partial API for Discord's messages resource

freya022:
## Pull Request Etiquette

### Changes

- [X] Internal code
- [X] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

As there is an increasing demand to send/edit messages from potentially archived threads (and thus absent from the thread cache), this PR adds an API that lets users interact with part of Discord's messages resource, without requiring a channel instance.

The new API (`ApiEndpoints`) is marked as experimental (and requires an opt-in for Kotlin users), it may change in the future as it can differ from a future, proper REST-only API.

### DIFF
--- a/src/main/java/net/dv8tion/jda/api/endpoints/ApiEndpoints.java
+++ b/src/main/java/net/dv8tion/jda/api/endpoints/ApiEndpoints.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2015 Austin Keener, Michael Ritter, Florian Spieß, and the JDA contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.dv8tion.jda.api.endpoints;
+
+import net.dv8tion.jda.annotations.ExperimentalRestApi;
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.internal.endpoints.MessageApiImpl;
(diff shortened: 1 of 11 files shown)


src/main/java/net/dv8tion/jda/internal/requests/restaction/AbstractMessageCreateActionImpl.java
```diff
@@ -0,0 +1,243 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.BooleanSupplier;
+import java.util.stream.Collectors;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+abstract class AbstractMessageCreateActionImpl extends RestActionImpl<Message>
```
freya022, line 47: Not sure if making separate actions was a good call, there are only a few differences, but I figured it would make it cleaner.

src/main/java/net/dv8tion/jda/internal/requests/restaction/AbstractMessageCreateActionImpl.java
```diff
@@ -0,0 +1,243 @@
...
+    @Override
+    public MessageCreateAction setMessageReference(@Nullable String messageId) {
+        if (messageId == null) {
+            this.messageReference = null;
+            return this;
+        }
+
+        Checks.isSnowflake(messageId);
+        this.messageReference =
+                new MessageReferenceData(MessageReference.MessageReferenceType.DEFAULT, null, channelId, messageId);
```
freya022, line 160: This no longer sets the guild ID (which we don't have in the REST API), I found out only the channel ID is required.

src/main/java/net/dv8tion/jda/api/endpoints/ApiEndpoints.java
```diff
@@ -0,0 +1,76 @@
...
+import javax.annotation.Nonnull;
+
+/**
+ * Entrypoint for REST-only APIs.
+ *
+ * <h2>Stability note</h2>
+ * The API and ABI of this interface and all objects returned by it, are <b>subject to changes without warning</b>.
+ */
+@ExperimentalRestApi
+public interface ApiEndpoints {
```
freya022, line 33: Maybe `DiscordEndpoints`

PR #3123, opened 2026-07-07
Fix incremental compilation support

freya022:
## Pull Request Etiquette

### Changes

- [ ] Internal code
- [ ] Documentation
- [X] Other: Build logic

Closes Issue: NaN

## Description

This PR fixes multiple build issues:
1. Full rebuilds were triggered by any change, this was especially annoying for feature PRs, where a single documentation change can result in 20+ seconds builds
    * Logs found that the compile task warned about inputs having non-source roots being passed
    * Instead of filtering for a `JDAInfo.java`, it is now generated, and unchanged source set roots are now passed
2. Gradle fingerprinted the API model generator's output before it could be filtered (by deleting excluded files), resulting in non-existent files being passed to the compile task, thus failing
    * This was fixed by using an intermediary folder, bridging the output of the API model generator to the input of the filter task, then passing the filter's output to the compiler
3. Tasks had to have their inputs manually set to accommodate the filtered sources
    * As `JDAInfo` is now generated, sources only consists of unchanged source sets
    * The generated sources are now all added to the `sourceSets`, removing the need to explicitly set `dependsOn`, all tasks responsible for generating those sources are automatically called
4. Some tasks had directory paths as their input, requiring a `dependsOn`
    * Fixed by passing the task directly, which takes all outputs and depends on the task


buildSrc/src/main/kotlin/net/dv8tion/jda/gradle/Version.kt
```diff
@@ -16,12 +16,14 @@
...
 
+import java.io.Serializable
+
 data class Version(
         val major: String,
         val minor: String,
         val revision: String,
         val classifier: String? = null
-) {
+) : Serializable {
```
MinnDevelopment, line 26: Add a serialVersionUID for this

PR #3137, opened 2026-08-25
docs: fix typo inheritence -> inheritance

vaibhav8a:
One-word typo fix in `src/main/java/net/dv8tion/jda/internal/entities/channel/mixin/README.md:3`: "Code reuse through composition instead of inheritence" → "inheritance".


MinnDevelopment:
Hi, I don't accept low effort spam PRs from obvious bots. This kind of behavior only adds unnecessary review effort for me that I do not appreciate. I've fixed the typo myself, so other bots don't pick it up as well.

PR #3147, opened 2026-09-14
Add CompiledRoute getter in RestActionImpl

HeyBanditoz:
## Pull Request Etiquette

### Changes

- [X] Internal code
- [ ] Documentation
- [ ] Other: \_____ 

Closes Issue: NaN

## Description

Exposes the Route.CompiledRoute a RestActionImpl uses via public getter.


freya022:
Why?

HeyBanditoz:
Route.CompiledRoute is already exposed in [HttpRequestEvent#getRoute](https://github.com/DV8FromTheWorld/JDA/blob/c392b04bbc10b39e3b907dd299ef48b248ba3850/src/main/java/net/dv8tion/jda/api/events/http/HttpRequestEvent.java#L131-L134), we can also expose it from inside a constructed `RestActionImpl` for example use: anyone building observability on top of `RestAction`s

Also, everything inside Route.CompiledRoute is immutable so I don't see any risk with exposing it from an API perspective

freya022:
Is `HttpRequestEvent#getRoute` not enough? Can't quite see how you'd have a `RestAction` easily be kept around for observability. Also, if it were to be exposed, it would be on `RestAction`, not `RestActionImpl`.

MinnDevelopment:
A lot of restactions use `finalizeRoute` to change the route for individual requests, for instance this is heavily utilized in PaginationAction. I believe you probably want to instead use either okhttp interceptor or add a rate-limiter proxy for observability. As a note for observability, if you use routes for metrics, you might want to use the uncompiled routes which don't have the path parameters filled in.

